PR #6, opened 2019-09-30
Add Nix support and run Travis with nix

gila:
These two commits do the following:

1. Update the cargo dependencies that we are using
2. Introduce build support for Nix and change Travis to use nix as the base image 

### A little more about Nix

Our current way of building system is constructing a layer cake out of userland components of a distro which adds, to begin with, a lot of undeeded binaries and libraries in the base image.

Typically, we create a clean image, and `apt-get` our way until we are able to run the MayaStor or, have a special "dev" container that hosts the build and we copy over the artifacts to "release" container.

One of the problems with this approach, next to the bloat of things we don't need is that it is almost impossible to keep the image in sync, updated and deterministic. With Nix, we are able to create functional expressions that result in an immutable "derivation" or build result. This result can be easily placed into a *single layer* container, VMs and/or purposely built customized OS ISO images for distribution later on and we can always rebuild the exact same contents and also target other platforms for example ARM.

Nix runs on different operating systems and distro's -- it is opt-in _not_ all in. This means that contributors can simply do:

```bash
cd mayastor
nix-shell
# have fun
```
On their distro of choice. As we do not have an actual release yet, there is no Maystor expression (and thus no deterministic build). Prior to releasing an updated image, one has to be created and we need to wait for a version of a stable Rust compiler or pin to a nightly.


nix/README.md
```diff
@@ -0,0 +1,38 @@
+## Overview
+
+These are a collection of packages we need, or packages where we 
+want to control the exact version(s) of.
+
+The packages are imported through the `nix-shell` automatically. If you
+run NixOS, read the following section.
+
+## Adding adding the overlay(s)
```
jkryl: typo: double "adding"

nix/README.md
```diff
@@ -0,0 +1,38 @@
...
+mkdir -p ~/.config/nixpkgs/overlays
+ln -s $(pwd)/nix/mayastor-overlay.nix ~/.config/nixpkgs/overlays/rust-overlay.nix
+```
+
+Now the package is integrated natively:
+
+```
+nix-env -qaP libiscsi
+nixos.libiscsi  libiscsi-1.19.0
+gila@gixos ~/MayaStor>
```
jkryl: nit: the first line is missing a prompt and the last line has it though there should not be any (probably).

nix/README.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+Like wise for rust nightly:
+
+```
+$ git clone https://github.com/mozilla/nixpkgs-mozilla.git
+$ ln -s $(pwd)/nixpkgs-mozilla/rust-overlay.nix ~/.config/nixpkgs/overlays/rust-overlay.nix
+```
+
+With the overlay in place you can start nix-shell a within the project root.
```
jkryl: typo: superfluous 'a' before within.

nix/README.md
```diff
@@ -0,0 +1,38 @@
...
+These are a collection of packages we need, or packages where we 
+want to control the exact version(s) of.
+
+The packages are imported through the `nix-shell` automatically. If you
+run NixOS, read the following section.
+
+## Adding adding the overlay(s)
+```
+mkdir -p ~/.config/nixpkgs/overlays
+ln -s $(pwd)/nix/mayastor-overlay.nix ~/.config/nixpkgs/overlays/rust-overlay.nix
```
jkryl: the link name should be mayastor-overlay.nix instead of rust-overlay.nix?

nix/README.md
gila: Thanks! Fixed 👍

Issue #1, opened 2019-09-16
Reserve space for metadata


gila:
The nexus driver should, implicitly, reserve disk space for metadata. The metadata is used during the rebuild process.

gila:
Implemented by #7

PR #7, opened 2019-10-07
Nexus: Implement GPT labeling support

gila:
This change will make the nexus look for GPT labels. The table
entries will be used to differentiate between data and meta
data partitions.

If no GPT labels exist, they will be created implicitly. This implements #1


mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -498,8 +373,106 @@ impl Nexus {
...
+            ent_guid: GptGuid::new_random(),
+            ent_start: entries[0].ent_end + 1,
+            ent_end: hdr.lba_end,
+            ent_attr: 0,
+            ent_name: GptName {
+                name: "MayaData".into(),
+            },
+        };
+
+        // entries.resize(128, GptEntry::default());
```
jkryl: is this comment left alone as a remark for future task or is it stale?

gila: Should be removed, thanks!

PR #27, opened 2019-10-25
nix: Initial package for Mayastor

gila:
This also bumps some dependencies. Simply run nix-env -iA mayastor and you are all set once you have added the overlay (to a method of your choosing).

There are some things to consider though:

 - the package right now is purely there to create containers from
 - once we actually release we should ensure we change `src` to an actual commit


gila:
merged with 94ca54b

PR #36, opened 2019-10-29
Rebase buildenv container on nixos and use nix shell for dependencies

jkryl:


doc/build.md
```diff
@@ -53,30 +53,18 @@ DISTRIB_DESCRIPTION="Ubuntu 18.04.3 LTS"
...
-```
-
-Now you can run, but lets ensure we have recent nightly first:
-
-```bash
-rustup update nightly
-rustup default nightly
+cd MayaStor
+nix-shell
+git submodule update --init --recursive
```
gila: recursive is not needed here? Just spdk-sys (and not its subrepo's)

jkryl: Somehow it sneaked through again. I will fix it - thx 👍

PR #41, opened 2019-11-01
mayastor: update to v19.10

gila:
This updated our package to v19.1. There are, of course come changes too non public headers but overall its not all that bad.  The structure  of the repo has changes somewhat but this is transparent to our sys crate. 

One of the reasons I personally want this merged so quickly is because of the work done in reset handling.

### DIFF
--- a/mayastor/src/nvmf_target.rs
+++ b/mayastor/src/nvmf_target.rs
@@ -30,6 +30,7 @@ use spdk_sys::{
     spdk_nvmf_subsystem_set_sn,
     spdk_nvmf_subsystem_start,
     spdk_nvmf_subsystem_stop,
+    spdk_nvmf_target_opts,
     spdk_nvmf_tgt,
     spdk_nvmf_tgt_accept,
     spdk_nvmf_tgt_add_transport,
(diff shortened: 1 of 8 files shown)


mayastor/src/nvmf_target.rs
```diff
@@ -245,6 +246,37 @@ impl Iterator for SubsystemIter {
...
+
+impl TargetOpts {
+    fn new(name: &str, max_subsystems: u32) -> Self {
+        let mut opts = spdk_nvmf_target_opts::default();
+
+        unsafe {
+            std::ptr::copy_nonoverlapping(
+                name.as_ptr() as *const _ as *mut libc::c_void,
+                &mut opts.name[0] as *const _ as *mut libc::c_void,
+                256,
```
jkryl: just a nit. using spdk_sys::NVMF_TGT_NAME_MAX_LENGTH would be safer here.

gila: I've added a new issue for this because for some odd reason its not there, I think its due to the white labelling but, things used should be automatically white listed as well.

Issue #49, opened 2019-11-06
UDS listner does not handle existing sockets properly


gila:
When starting mayastor-agent, with a preexisting CSI socket, we fail the bind.

```rust
let mut tcp = tokio::net::UnixListener::bind(std::path::Path::new(path))?.incoming();
```

*note this code is in the fork of tonic*

The above line creates a new listener if and only if  `path` does not exist. We should bind to the socket even when it exists or, clean up any stale/existing sockets prior to starting the server.

gila:
fixed with https://github.com/openebs/MayaStor/commit/85e3815fe1c5f39015de1edb7e5d82eb8b4990a1

Issue #46, opened 2019-11-04
Do not expose CSI grpc services on TCP port


jkryl:
CSI methods are called from k8s sidecars using unix domain socket file. Exposing CSI services on TCP port is useless and not safe without implementing authentication. It is believed that it is required for running the mayastor tests. Remove the CSI services from TCP port and make necessary changes to the test suite (if needed).

jkryl:
This has been fixed on behalf of https://github.com/openebs/MayaStor/pull/52. Closing.

PR #52, opened 2019-11-06
stale socket file

jkryl:
Two issues which are somewhat related because they required similar changes in the test suite:

UDS listner does not handle existing sockets properly #49
Do not expose CSI grpc services on TCP port #46

PS: This will also fix an issue we have been seeing in ci/cd pipeline with weird socket file in top level dir of the repo.


gila:
This fixes #49

PR #72, opened 2019-12-10
Want a rocket when mayastor starts

gila:
Rework the logging such that all logging passes through rust.
This commit also bumps some of the dependencies.


mayastor/src/lib.rs
```diff
@@ -58,6 +65,69 @@ pub extern "C" fn cps_init() {
...
+        }
+
+        let lvl = match level {
+            -1 => return,
+            0 => Level::Error,
+            1 => Level::Warn,
+            2 => Level::Info,
+            3 => Level::Debug,
+            4 => Level::Trace,
+            _ => Level::Error,
```
jkryl: IMHO it would have been better to match levels by SPDK enum names rather than numbers (i.e. spdk_sys::SPDK_LOG_WARN instead of 1).

PR #74, opened 2019-12-12
nexus: avoid submitting IO we know that will fail

gila:
Add test cases for creating filesystems on a nexus device. Today, I had very strange failures, that looked like, to be related to UNMAP but that turned out to be incorrect. 

There appear to be some issues with NVMe and mkfs, but I can't reproduce the error anymore. In any case, it is likely a good idea to test the creation of filesystems.

If we submit IO to a child that does not support UNMAP, we should get back two failures (in case of two replicas) before we fail the parent IO. I've added trace!() statements to observe that in the future. However,  if we know we don't support it, we might as well safe the two IO's to begin with. As such, we verify that that all child devices support UNMAP before passing it down. If this turns out to be white hot in production we can make it a variable within the nexus itself.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -54,7 +54,11 @@ build-mayastor:
   artifacts:
     expire_in: 1 day
     paths:
-      - target/debug/
+        - target/debug/mayastor
+        - target/debug/mayastor-agent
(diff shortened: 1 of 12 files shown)


gila:
merged with 5244b1a

PR #76, opened 2019-12-19
Convert all error types in nexus code to snafu errors.

jkryl:
Besides the main group of nexus errors, there are a couple of other
error groups for child bdev, rebuild, labels, nbd, uri parsing, event
and poller modules.

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_share.rs
+++ b/mayastor/src/bdev/nexus/nexus_share.rs
@@ -1,16 +1,23 @@
-use std::ffi::CString;
-
-use futures::channel::oneshot;
-
-use spdk_sys::create_crypto_disk;
-
 use crate::{
     bdev::{
         bdev_lookup_by_name,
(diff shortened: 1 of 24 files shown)


mayastor/src/bdev/nexus/nexus_share.rs
```diff
@@ -33,49 +42,38 @@ impl Nexus {
             let name = format!("crypto-{}", self.name);
 
             // constant
-            let flavour = CString::new(CRYPTO_FLAVOUR)?;
+            let flavour = CString::new(CRYPTO_FLAVOUR).unwrap();
```
gila: This will crash the system, can we go back to "?" here?

jkryl: We discussed this offline. The summary: For certain basic failures like memory allocation failures or invalid strings that come from trusted sources, we can (or should) assert that the things are as we expect them as it makes the error handling easier. 👍

PR #81, opened 2019-12-30
Test utility for reading/writing to/from nexus child - replica

jkryl:
* Refactor the code around bdev descriptor object to feel more rusty
  and to avoid code duplication with spdk_bdev_open().
* Rust wrapper object for DMA buffer lives in its own file now.
* Remove dma_zmalloc() in favour of dma_malloc() - both methods
  returned zero initialized memory.
* Fix bug when return code from mayastor_stop wasn't properly
  propagated to mayastor_start caller.
* Fix default log expression which filtered out all messages not coming
  from the mayastor lib (i.e. messages from the binaries using the lib).
* Remove "deprecated" tag from mayastor_start() as it's preferred
  alternative to mayastor env lib for simple test & utility programs.

Help message for the new tool:
```
Test initiator for nexus replica
Connect, read or write a block to a nexus replica using its URI

USAGE:
    initiator [OPTIONS] <URI> [SUBCOMMAND]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -o, --offset <NUMBER>    Offset of IO operation on the replica in bytes (default 0)

ARGS:
    <URI>    URI of the replica to connect to

SUBCOMMANDS:
    connect    Connect to and disconnect from the replica
    help       Prints this message or the help of the given subcommand(s)
    read       Read bytes from the replica
    write      Write bytes to the replica
```

Example usage (all three major subcommands):
```
  initiator nvmf://127.0.0.1:8420/nqn.2019-05.io.openebs:disk2 connect
  initiator nvmf://127.0.0.1:8420/nqn.2019-05.io.openebs:disk2 read /tmp/out
  initiator nvmf://127.0.0.1:8420/nqn.2019-05.io.openebs:disk2 write /tmp/out
```

TODO:
 * Make starting nvmf and iscsi target in mayastor lib optional. We don't
   need those in initiator utility. We need to squeeze mem requirements
   and the targets conflict with a mayastor instance running on the same
   host (workaround is MY_POD_IP=<extern-IP>).
 * Use the new utility in JS unit tests to test the IO on replica.

### DIFF
--- a/mayastor/src/descriptor.rs
+++ b/mayastor/src/descriptor.rs
@@ -13,7 +13,7 @@
 //! use mayastor::descriptor::Descriptor;
 //! // open a descriptor to the bdev, in readonly or read/write
 //! let bdev = Descriptor::open("my_bdev", true).unwrap();
-//! let mut buf = bdev.dma_zmalloc(4096).unwrap();
+//! let mut buf = bdev.dma_malloc(4096).unwrap();
 //! buf.fill(0xff);
 //! bdev.write_at(0, &buf).await.unwrap();
 //!
@@ -28,7 +28,6 @@
 use std::{
     ffi::c_void,
     ops::{Deref, DerefMut},
-    slice::{from_raw_parts, from_raw_parts_mut},
 };
 
 use futures::channel::oneshot;
@@ -44,22 +43,23 @@ use spdk_sys::{
     spdk_bdev_open,
     spdk_bdev_read,
     spdk_bdev_write,
-    spdk_dma_free,
-    spdk_dma_zmalloc,
     spdk_get_io_channel,
     spdk_io_channel,
     spdk_put_io_channel,
 };
 
 use crate::{
-    bdev::{bdev_lookup_by_name, Bdev},
+    bdev::Bdev,
+    dma::{DmaBuf, DmaError},
     executor::{cb_arg, errno_result_from_i32},
 };
 
 #[derive(Debug, Snafu)]
-pub enum DmaError {
-    #[snafu(display("Failed to allocate DMA buffer"))]
-    Alloc {},
+pub enum DescError {
+    #[snafu(display("Failed to open bdev"))]
+    OpenBdev { source: Errno },
+    #[snafu(display("Failed to obtain IO channel for bdev"))]
+    GetIoChannel {},
     #[snafu(display("Invalid IO offset {}", offset))]
     InvalidOffset { offset: u64 },
     #[snafu(display(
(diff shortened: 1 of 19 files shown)


mayastor/src/lib.rs
```diff
@@ -296,25 +301,13 @@ where
 /// Cleanly exit from the program.
 /// NOTE: cannot be called from a future -> double borrow of executor.
 pub fn mayastor_stop(rc: i32) -> i32 {
-    if rc != 0 {
```
gila: This won't work.. the global thread the app framework uses is not set. Its not accessible outside of the app framework either. 

This will cause the shutdown code to crash or trip an assert or something nasty like that. Maybe code that uses the app framework should use a different shutdown callback? (which more or less is what at is was but but simply named differently?)

jkryl: yea, I found out exactly what you described :-) I believe it's fixed now, but let's wait for test results. The main idea of the fix is to avoid mixing shutdown path of new env lib and original rust wrappers in lib.rs. mayastor_stop should be used only for progs started by mayastor_start and mayastor_env_stop only for mayastor::environment::env::MayastorEnvironment::new 🤓

mayastor/src/descriptor.rs
```diff
@@ -88,103 +88,52 @@ pub enum DmaError {
...
 
-/// DmaBuf that is allocated from the memory pool
+/// Block Device Descriptor
 #[derive(Debug)]
-pub struct DmaBuf {
-    /// a raw pointer to the buffer
-    pub buf: *mut c_void,
-    /// the length of the allocated buffer
-    len: usize,
+pub struct Descriptor {
```
gila: @jkryl, Thanks for clean this up, in retrospect, the initially chosen naming here chose very poorly.

The initial idea was to use the descriptor -- that is analogous to a file descriptor, and it would provide all the methods to work with it. 

However, as we are moving towards multiple cores, there is an inherent problem with this approach that I'm hitting while implementing the new per segment rebuild implementation.

To sum that up, the idea here is that we divide the block device into segments, and have the nexus dispatch segments to whatever cores are available.

Because of that, that's where _this_ descriptor model brakes as per core we need an io channel. So it works only for whatever core we happen to initialise on. 

So we need to decouple the descriptor and the io_channel. We can then also get rid of the Rc around it. 

BTW, I'm not expecting you to do that as part of this PR or anything like that. I'm implementing two new types around these primitives.  I'll go back and update this part to make use of that once we agree on the code that implements it.

jkryl: thanks for explaining 👍

PR #84, opened 2020-01-07
moac accepts a new parameter from storage class - # of replicas

jkryl:
Storage class parameters section may contain a new "repl" key with
string value which is supposed to be a number for now. moac will
create as many replicas of nexus as specified by this count (one local
running on the same node as nexus and the rest remote using nvmf
protocol). Note that this does not impact the IO protocol for
nexus device itself which still uses NBD.

This is just the first part of the support for multiple replicas.
To finalize the feature we need to come up with custom resource(s)
representing the replica/nexus configuration in k8s's etcd. Relying
just on CSI methods is not sufficient.


gila:
Great demo @jkryl

PR #88, opened 2020-01-08
Utility for monitoring mayastor IOPS using influxdb and grafana.

jkryl:
This is the grafana work I did today for the demo. The code was taken from the old mayastor repo and it was polished to catch up with the changes in mayastor. One change though is that all the source code for the monitoring is kept aside from the rest of the mayastor code in `monitor` subdirectory (source code, docker file, k8s deployment file, configuration files, ...). It is not part of the product. Though it is generally useful when we want to present the mayastor to others.

### DIFF
--- a/monitor/Dockerfile
+++ b/monitor/Dockerfile
@@ -1,0 +1,9 @@
+FROM node:10
+
+RUN mkdir /monitor
+WORKDIR /monitor
+COPY package*.json ./
+RUN npm install
+COPY *.js /monitor/
+
+ENTRYPOINT ["/monitor/index.js"]
--- a/monitor/README.md
+++ b/monitor/README.md
@@ -1,0 +1,51 @@
+# Monitoring extension for MayaStor
+
+It consists of monitor script written in NodeJS, which continuously
+retrieves stats data from moac REST API server and saves them to
+influxDB. From there the data can be taken and displayed by Grafana.
+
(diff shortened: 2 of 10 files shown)


richardelling:
I'll not comment on the code itself, other than to say a quick scan didn't show any obvious problems.

However, going forward, it is easier to write or configure an agent for an aggregator, like telegraf, and have that handle forwarding the metrics to influxdb, prometheus, or whatever. If the metrics themselves are easily grokked from an API call, then it could just be a simple matter of configuring 
the inputs.httpjson agent for telegraf.

gila:
@richardelling , Thanks for the suggestions, telegraf seems very useful for collection not just CSI volume metrics, will keep that on the list to look into for sure!

jkryl:
@richardelling : thanks for your comment! That looks like exactly what we need.

I will close this PR and eventually submit a new one with telegraf. I will do changes to our HTTP server to make the format of json data telegraf-friendly and create deployment yaml for telegraf. IMHO that's what mayastor should provide. Users are then free to choose what database and what graphing tool they want to use. That said it would be nice if we had something basic & simple in the repo for displaying the stats. Not sure if influxdb + grafana is the best choice. It's open for comments.

richardelling:
@jkryl there are several options to implement metrics, perhaps we can open a discussion on slack (NB, I don't see you in the contributor's channel) or email... issues don't work well for these discussions.

PR #89, opened 2020-01-13
mayastor: switch to upstream tonic-rs

gila:
Tonic has recently added support for UDS so we do not need to make use of our
fork anymore.


jsonrpc/src/test.rs
```diff
@@ -52,10 +54,9 @@ where
     };
 
     // create tokio futures runtime
```
jkryl: I think this comment should be removed too

gila: thx Jan fixed

PR #96, opened 2020-01-14
Disable "fortify" causing annoying warning when in nix shell.

jkryl:
Even better would be to disable fortify conditionally for
libspdk and mayastor package when doing debug build. But that
would need a different solution - this variable works only in
nix shell.


gila:
Thanks @jkryl

PR #105, opened 2020-02-19
fix NBD and reactor shutdown

gila:
NBD was hanging when we shutdown as we did not wait for any outstanding IO completed. These IO's where not submitted by us rather by systemd-udev.

Secondly, when we use `block_on` our reactor state is not set to running which implies we need to account for handling this case in the shutdown future. This typically affects testing and not normal mode of operations.

### DIFF
--- a/mayastor/src/core/reactor.rs
+++ b/mayastor/src/core/reactor.rs
@@ -1,13 +1,7 @@
-//!
-//! The reactor is the main loop that will run on each core that is available to
-//! use given by the the coremask argument.
-//!The reactor is the main loop that runs on each core that is available to use
-//! given by the core mask argument.
-//!
-//! The first thing that needs to happen is to initialize DPDK, this among others
-//! provides us with lockless queues which we use to send messages between the
-//! cores.  Typically these messages contain simple function pointers and
-//! argument pointers.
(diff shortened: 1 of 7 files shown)


mayastor/src/core/reactor.rs
```diff
@@ -54,11 +48,11 @@ use spdk_sys::{
 
 use crate::core::{Cores, Mthread};
 
-pub(crate) const INIT: usize = 1 << 1;
-pub(crate) const RUNNING: usize = 1 << 2;
-pub(crate) const SHUTDOWN: usize = 1 << 3;
-pub(crate) const SUSPEND: usize = 1 << 4;
-pub(crate) const DEVELOPER_DELAY: usize = 1 << 5;
+pub(crate) const INIT: usize = 1;
```
tiagolobocastro: I'm afraid I'm not not familiar enough with all this code yet so I'll just make a nitpick suggestion: 1 << 0 ? :)

gila: I actually had that at first but Clippy suggested to change that to 1 (as its a nop) When Clippy tells me to do something, I usually do it =) 

BTW, we run Clippy as part of the CI so its suggestions, actually become failures. Thanks!

PR #106, opened 2020-02-19
CAS-21 Add ShareProtocol field to PublishNexusRequest

blaisedias:


blaisedias:
Untested at the moment, cause unable to get the tests to run successfully with no source modifications.

PR #111, opened 2020-02-26
docs: Add a troubleshooting guide to help new developers get started with our tests

tjoshum:


jkryl:
I personally prefer to keep the minimal requirements really minimal. if huge pages are not leaked or if multiple instances of mayastor are not run, then 512 should be enough. Though we could recommend users to allocate more (1024) for a more smooth user experience (in cases when HPs are leaked) :-)  Interestingly I haven't seen leaked hugepages when running mayastor in k8s. I think the leaks are related to unclean shutdowns of mayastor that are less likely in k8s env compared to test environments where all kinds of crazy things happen.

tjoshum:
I've focused this doc more on a smooth developer experience, rather than minimal requirements so I'll keep the large number here, but not change the minimal requirements in other docs.
I'll close this PR and open one from a branch to trigger the GitLab CI.

tjoshum:
Replaced by #112

PR #121, opened 2020-03-06
nix: Use mayastor-overlay to build liburing

jonathan-teh:
Upstream nix-unstable moved from liburing 0.2 to 0.4pre, which introduced libdevdir in the configure script. This was not set as part of the build so libspdk_fat.so ended up being linked to liburing.so.1 without a full path, breaking the build of Mayastor.

Override upstream's nix expression by setting libdevdir to libdir, and also moving up to the current 0.5 release.


jonathan-teh:
This fixes #117 for users where nix had installed liburing-0.4pre514 instead of 0.2 when CI built it.

PR #124, opened 2020-03-10
add terraform helpers to construct k8s clusters

gila:
This is an initial pass to providing automated setup of a k8s cluster. We can extend this as we need to for other providers. The purpose of this is to get something "fast but working" and not delve into the more sophisticated systems like KOPS and/or Kubespray. 

Some notes:

 - it might be useful to build an ISO to avoid pkg installing which consumes the most time 
- installing mayastor on this cluster automatically is not done yet the intent is to use terraform for this as well
- the setup of the OS, ie huge pages and kubelet config is purposely not setup. the reason for this is that we must provide an easy way to do this with a system that is not configured correctly yet
- we always grab the latest release we can pin the exact release in the YAML files if we want

### DIFF
--- a/terraform/README.adoc
+++ b/terraform/README.adoc
@@ -1,0 +1,114 @@
+:source-highlighter: highlightjs
+:toc:
+
+== Quick start
+
+. Ensure you have libvirtd running and configured properly
+. Edit `variables.tf` in particular make sure you the public key is correct
+. run `terraform init`
+. run `terraform plan`
+. run `terraform apply`
+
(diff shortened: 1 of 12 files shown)


gila:
Thanks, @jkryl I want to add one more thing before I merge it I think.  One of the missing pieces (depending on what you exactly want of course) is to suck in the the `admin.conf` from the created cluster to the local system. This then allows you to developer code on your own box and of course allows you to use ku-be-cuttle.

PR #125, opened 2020-03-10
nexus: add ability to share through iSCSI

tjoshum:
Make it possible to specify the nexus front-end protocol for use with CSI and add support for  iSCSI. 

See task CAS-54.
The set of supported interfaces is planned to be nbd, iSCSI and nvmf.
Of these, nbd is implemented. This change adds iSCSI, Nvmf is still to be implemented.
A publish nexus request now takes an additional parameter to specify which protocol to use. Specifying "iscsi" now results in the creation of an iSCSI target to act as the front-end for the specified nexus. 
To separate the domain of the front-end iscsi target from that of the existing iscsi targets used for the back-end replicas, two portal groups are now created, which differ in the choice of TCP port used.

The choice of ports for the two portal groups are likely to change as a result of CAS-97.

(Changes by @chriswldenyer and @blaisedias to allow a nexus to be shared via iSCSI.)

Changes have been made resulting from the first code reviews.
Iscsi nexus tests have been added to test_nexus.js, as requested and as required by CAS-65.
The publish nexus request now returns the URI of the created iSCSI target.
Repeated publish requests and now handled correctly, a repeated request with a different protocol returns an error.

### DIFF
--- a/cli/src/convert.rs
+++ b/cli/src/convert.rs
@@ -1,4 +1,5 @@
 use byte_unit::Byte;
+use rpc::mayastor::ShareProtocolNexus;
 
 /// converts a human string into a blocklen
 #[allow(dead_code)]
@@ -21,5 +22,14 @@ pub(crate) fn parse_size(src: &str) -> Result<u64, String> {
         Ok(val.get_bytes() as u64)
     } else {
         Err(src.to_string())
+    }
+}
+
+pub(crate) fn parse_proto(src: &str) -> Result<ShareProtocolNexus, &str> {
+    match src.to_lowercase().trim() {
+        "nbd" => Ok(ShareProtocolNexus::NexusNbd),
+        "nvmf" => Ok(ShareProtocolNexus::NexusNvmf),
+        "iscsi" => Ok(ShareProtocolNexus::NexusIscsi),
+        _ => Err("Protocol needs be either NVMf, iSCSI or NBD"),
     }
 }
--- a/mayastor-test/test_csi.js
+++ b/mayastor-test/test_csi.js
@@ -24,6 +24,7 @@ const protoLoader = require('@grpc/proto-loader');
 // possible only with grpc-uds.
 const grpc = require('grpc-uds');
 const common = require('./test_common');
+const mayastorProto = require('./mayastor_proto');
 // Without requiring wtf module the ts hangs at the end. It seems that it is
 // waiting for sudo'd mayastor progress which has already exited!?
 const wtfnode = require('wtfnode');
@@ -197,7 +198,12 @@ describe('csi', function() {
               let uuid = BASE_UUID + n;
               common.dumbCommand(
                 'publish_nexus',
-                { uuid: uuid, key: '' },
+                {
+                  uuid: uuid,
+                  key: '',
+                  share: mayastorProto.getConstants().ShareProtocolNexus
+                    .NEXUS_NBD,
+                },
                 next
               );
             },
--- a/mayastor-test/test_nexus.js
+++ b/mayastor-test/test_nexus.js
@@ -15,6 +15,7 @@ const { exec } = require('child_process');
 const { createClient } = require('grpc-kit');
 const grpc = require('grpc');
 const common = require('./test_common');
+const mayastorProto = require('./mayastor_proto');
 // just some UUID used for nexus ID
 const UUID = 'dbe4d7eb-118a-4d15-b789-a18d9af6ff21';
 const UUID2 = 'dbe4d7eb-118a-4d15-b789-a18d9af6ff22';
@@ -27,6 +28,7 @@ const uringFile = '/tmp/uring-backend';
 const diskSize = 64 * 1024 * 1024;
 // external IP address detected by common lib
 const externIp = common.getMyIp();
+const mayastorProtoConstants = mayastorProto.getConstants();
 
 // Instead of using mayastor grpc methods to create replicas we use a config
 // file to create them. Advantage is that we don't depend on bugs in replica
@@ -49,7 +51,7 @@ const configNexus = `
   ImmediateData Yes
   ErrorRecoveryLevel 0
   # Reduce mem requirements for iSCSI
-  MaxSessions 1
+  MaxSessions 2
   MaxConnectionsPerSession 1
 
 [PortalGroup1]
@@ -154,6 +156,7 @@ var doUring = (function() {
 describe('nexus', function() {
   var client;
   var nbd_device;
+  var iscsi_uri;
 
   const unpublish = args => {
     return new Promise((resolve, reject) => {
@@ -399,11 +402,17 @@ describe('nexus', function() {
   });
 
   it('should publish the nexus using nbd', done => {
-    client.PublishNexus({ uuid: UUID }, (err, res) => {
-      assert(res.device_path);
-      nbd_device = res.device_path;
-      done();
-    });
+    client.PublishNexus(
+      {
+        uuid: UUID,
+        share: mayastorProtoConstants.ShareProtocolNexus.NEXUS_NBD,
+      },
+      (err, res) => {
+        assert(res.device_path);
+        nbd_device = res.device_path;
+        done();
+      }
+    );
   });
 
   it('should un-publish the nexus device', done => {
@@ -414,11 +423,18 @@ describe('nexus', function() {
   });
 
   it('should re-publish the nexus using NBD, and a crypto key', done => {
-    client.PublishNexus({ uuid: UUID, key: '0123456789123456' }, (err, res) => {
-      assert(res.device_path);
-      nbd_device = res.device_path;
-      done();
-    });
+    client.PublishNexus(
+      {
+        uuid: UUID,
+        share: mayastorProtoConstants.ShareProtocolNexus.NEXUS_NBD,
+        key: '0123456789123456',
+      },
+      (err, res) => {
+        assert(res.device_path);
+        nbd_device = res.device_path;
+        done();
+      }
+    );
   });
 
   it('should be able to write to the NBD device', async () => {
@@ -439,6 +455,115 @@ describe('nexus', function() {
 
     buffer.forEach(e => {
       assert(e === 122);
+    });
+  });
+
+  it('should un-publish the NBD nexus device', done => {
+    client.unpublishNexus({ uuid: UUID }, (err, res) => {
+      if (err) done(err);
+      done();
+    });
+  });
+
+  it('should publish the nexus using iscsi', done => {
+    client.PublishNexus(
+      {
+        uuid: UUID,
+        share: mayastorProtoConstants.ShareProtocolNexus.NEXUS_ISCSI,
+      },
+      (err, res) => {
+        assert(res.device_path);
+        done();
+      }
+    );
+  });
+
+  it('should un-publish the iscsi nexus device', done => {
+    client.unpublishNexus({ uuid: UUID }, (err, res) => {
+      if (err) done(err);
+      done();
+    });
+  });
+
+  it('should publish the nexus using iscsi', done => {
+    client.PublishNexus(
+      {
+        uuid: UUID,
+        share: mayastorProtoConstants.ShareProtocolNexus.NEXUS_ISCSI,
+      },
+      (err, res) => {
+        assert(res.device_path);
+        done();
+      }
+    );
+  });
+
+  it('should fail another publish request using a different protocol', done => {
+    client.PublishNexus(
+      {
+        uuid: UUID,
+        share: mayastorProtoConstants.ShareProtocolNexus.NEXUS_NBD,
+      },
+      (err, res) => {
+        if (!err) return done(new Error('Expected error'));
+        assert.equal(err.code, grpc.status.INVALID_ARGUMENT);
+        done();
+      }
+    );
+  });
+
+  it('should succeed another publish request using the existing protocol', done => {
+    client.PublishNexus(
+      {
+        uuid: UUID,
+        share: mayastorProtoConstants.ShareProtocolNexus.NEXUS_ISCSI,
+      },
+      (err, res) => {
+        if (err) done(err);
+        assert(res.device_path);
+        done();
+      }
+    );
+  });
+
+  it('should un-publish the iscsi nexus device', done => {
+    client.unpublishNexus({ uuid: UUID }, (err, res) => {
+      if (err) done(err);
+      done();
+    });
+  });
+
+  it('should succeed another un-publish request', done => {
+    client.unpublishNexus({ uuid: UUID }, (err, res) => {
+      if (err) done(err);
+      done();
+    });
+  });
+
+  it('should re-publish the nexus using iSCSI and a crypto-key', done => {
+    client.PublishNexus(
+      {
+        uuid: UUID,
+        share: mayastorProtoConstants.ShareProtocolNexus.NEXUS_ISCSI,
+        key: '0123456789123456',
+      },
+      (err, res) => {
+        assert(res.device_path);
+        iscsi_uri = res.device_path;
+        done();
+      }
+    );
+  });
+
+  it('should send io to the iscsi nexus device', done => {
+    let uri = iscsi_uri + '/0';
+    // runs the perf test for 1 second
+    exec('iscsi-perf -t 1 ' + uri, (err, stdout, stderr) => {
+      if (err) {
+        done(stderr);
+      } else {
+        done();
+      }
     });
   });
 
@@ -495,10 +620,25 @@ describe('nexus', function() {
     });
   });
 
-  it('should create, publish, un-publish and finally destroy the same nexus', async () => {
+  it('should create, publish, un-publish and finally destroy the same NBD nexus', async () => {
     for (let i = 0; i < 10; i++) {
       await createNexus(createArgs);
-      await publish({ uuid: UUID });
+      await publish({
+        uuid: UUID,
+        share: mayastorProtoConstants.ShareProtocolNexus.NEXUS_NBD,
+      });
+      await unpublish({ uuid: UUID });
+      await destroyNexus({ uuid: UUID });
+    }
+  });
+
+  it('should create, publish, un-publish and finally destroy the same iSCSI nexus', async () => {
+    for (let i = 0; i < 10; i++) {
+      await createNexus(createArgs);
+      await publish({
+        uuid: UUID,
+        share: mayastorProtoConstants.ShareProtocolNexus.NEXUS_ISCSI,
+      });
       await unpublish({ uuid: UUID });
       await destroyNexus({ uuid: UUID });
     }
@@ -512,10 +652,24 @@ describe('nexus', function() {
     });
   });
 
-  it('should create, publish, and destroy but without un-publishing the same nexus', async () => {
+  it('should create, publish, and destroy but without un-publishing the same nexus, with NBD protocol', async () => {
     for (let i = 0; i < 10; i++) {
       await createNexus(createArgs);
-      await publish({ uuid: UUID });
+      await publish({
+        uuid: UUID,
+        share: mayastorProtoConstants.ShareProtocolNexus.NEXUS_NBD,
+      });
+      await destroyNexus({ uuid: UUID });
+    }
+  });
+
+  it('should create, publish, and destroy but without un-publishing the same nexus, with iSCSI protocol', async () => {
+    for (let i = 0; i < 10; i++) {
+      await createNexus(createArgs);
+      await publish({
+        uuid: UUID,
+        share: mayastorProtoConstants.ShareProtocolNexus.NEXUS_ISCSI,
+      });
       await destroyNexus({ uuid: UUID });
     }
   });
--- a/mayastor/src/bdev/nexus/nexus_bdev.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev.rs
@@ -5,6 +5,7 @@
 //! application needs synchronous mirroring may be required.
 
 use std::{
+    fmt,
     fmt::{Display, Formatter},
     os::raw::c_void,
 };
@@ -39,8 +40,9 @@ use crate::{
             nexus_channel::{DREvent, NexusChannel, NexusChannelInner},
             nexus_child::{ChildError, ChildState, NexusChild},
             nexus_io::{io_status, Bio},
+            nexus_iscsi::{NexusIscsiError, NexusIscsiTarget},
             nexus_label::LabelError,
-            nexus_nbd::{Disk, NbdError},
+            nexus_nbd::{NbdDisk, NbdError},
         },
     },
     core::{Bdev, DmaBuf, DmaError},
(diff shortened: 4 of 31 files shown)


mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -186,10 +198,14 @@ pub struct Nexus {
...
     /// nbd device which the nexus is exposed through
-    pub(crate) nbd_disk: Option<Disk>,
+    pub(crate) nbd_disk: Option<NbdDisk>,
     /// the handle to be used when sharing the nexus, this allows for the bdev
     /// to be shared with vbdevs on top
     pub(crate) share_handle: Option<String>,
+    /// frontend share protocol used when the nexus is published
+    pub share_protocol: Option<ShareProtocolNexus>,
+    /// iscsi target which the nexus is exposed through
+    pub(crate) iscsi_target: Option<NexusIscsiTarget>,
```
jkryl: Currently, we have three optional members of the struct related to sharing and there are combinations that don't make a sense. For example, both, nbd_disk and iscsi_target, being Some(). Or share_protocol being None and nbd_disk or iscsi_target Some. How can we simplify this and eliminate inconsistent combinations at the struct definition level? 🤔 

A possible solution could be a new enum Disk (or Target) having three variants: Nbd, Iscsi and Nvmf. These variatns would hold a value NbdDisk, NexusIscsiTarget and later NexusNvmfTarget. That way we can easily enforce that only one target/disk is set for the nexus at any given time. And we would not need share_protocol at all as it is contained in the enum.

mayastor-test/test_csi.js
```diff
@@ -24,9 +24,11 @@ const protoLoader = require('@grpc/proto-loader');
 // possible only with grpc-uds.
 const grpc = require('grpc-uds');
 const common = require('./test_common');
+const mayastorProto = require('./mayastor_proto');
```
gila: nit; you could just do a require('./mayastor_proto).getConstants() here and get rid of line 31

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -66,8 +71,10 @@ pub enum Error {
     AlreadyShared { name: String },
     #[snafu(display("The nexus {} has not been shared", name))]
     NotShared { name: String },
-    #[snafu(display("Failed to share nexus {}", name))]
-    ShareNexus { source: NbdError, name: String },
+    #[snafu(display("Failed to share NBD nexus {}", name))]
```
gila: nit; "would failed to share nexus over NBD" not be more descriptive. One could read this as a nexus called NBD failed to share.

mayastor/src/bdev/nexus/nexus_bdev.rs
gila: I think that indeed we can make this generic over the ShareProtocol.

mayastor/src/target/iscsi.rs
```diff
@@ -82,101 +88,70 @@ thread_local! {
...
         spdk_iscsi_tgt_node_construct(
-            idx,
-            c_iqn.as_ptr(),
-            ptr::null(),
-            &mut group_idx as *mut _,
-            &mut group_idx as *mut _,
-            1, // portal and initiator group list length
-            &mut spdk_bdev_get_name(bdev.as_ptr()),
-            &mut lun_id as *mut _,
+            idx,                             // target_index
```
gila: nice thanks for adding that 👍

rpc/proto/mayastor.proto
```diff
@@ -47,13 +47,22 @@ message ListPoolsReply {
...
   ISCSI = 2;  // iSCSI
 }
 
+// Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type,
+// not children of it.
+// So cannot use NBD, NVMF, and ISCSI as symbols for ShareProtocolNexus
+enum ShareProtocolNexus {
+  NBD_FE = 0;    // local
+  NVMF_FE = 1;   // NVMe over Fabrics (TCP)
+  ISCSI_FE = 2;  // iSCSI
```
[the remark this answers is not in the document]
gila: I actually think we don't need the ShareProtocolNexus enum at all. Telling a replica to share, or a nexus is subject to the same protocols, so why would we need a new enum here? 🤔

mayastor/src/bdev/nexus/nexus_rpc.rs
```diff
@@ -13,6 +13,7 @@ use rpc::mayastor::{
     PublishNexusRequest,
     RemoveChildNexusRequest,
     UnpublishNexusRequest,
+    ShareProtocolNexus,
```
gila: I've commented on the proto file -- but to be sure I'll add it here as well. The enum variants are the same for `ShareProtocol` as well as `ShareProtocolNexus` Why cant we use the same enum?

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -186,10 +198,14 @@ pub struct Nexus {
     /// the offset in num blocks where the data partition starts
     pub data_ent_offset: u64,
     /// nbd device which the nexus is exposed through
-    pub(crate) nbd_disk: Option<Disk>,
+    pub(crate) nbd_disk: Option<NbdDisk>,
     /// the handle to be used when sharing the nexus, this allows for the bdev
     /// to be shared with vbdevs on top
     pub(crate) share_handle: Option<String>,
+    /// frontend share protocol used when the nexus is published
+    pub share_protocol: Option<ShareProtocolNexus>,
```
gila: If the nexus is not shared, I assume this share_protocol would have the variant, so we really don't need to put this in an option, it should always have a value, right?

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -264,8 +280,10 @@ impl Nexus {
             dr_complete_notify: None,
             data_ent_offset: 0,
             nbd_disk: None,
+            iscsi_target: None,
             share_handle: None,
             size,
+            share_protocol: None,
```
gila: This could be `ShareProtocolNexus::None` and we can get rid of the `Option<> `

mayastor/src/bdev/nexus/nexus_rpc.rs
```diff
@@ -130,8 +132,13 @@ pub(crate) fn register_rpc_methods() {
             let key: Option<String> =
                 if args.key == "" { None } else { Some(args.key) };
 
+            let share_proto = ShareProtocolNexus::from_i32(args.share);
+            if share_proto.is_none() {
```
gila: IMHO, share_proto itself should never be None IMHO. it should always be one of the variants where::None is one of them.

mayastor/src/bdev/nexus/nexus_rpc.rs
```diff
@@ -130,8 +132,13 @@ pub(crate) fn register_rpc_methods() {
...
+                return Err(Error::InvalidShareProtocol {sp_value: args.share as i32});
+            }
+
             let nexus = nexus_lookup(&args.uuid)?;
-            nexus.share(key).await.map(|device_path| PublishNexusReply {
+            nexus.share(share_proto.unwrap(), key).await.map(|device_path| PublishNexusReply {
```
gila: If we remove the Option we can remove the unwrap()

mayastor/src/bdev/nexus/nexus_share.rs
```diff
@@ -12,33 +12,39 @@ use crate::{
...
     pub async fn share(
         &mut self,
+        share_protocol: ShareProtocolNexus,
         key: Option<String>,
     ) -> Result<String, Error> {
-        if self.nbd_disk.is_some() {
+
+        assert_eq!(self.share_handle, None);
+
+        if self.share_protocol.is_some() {
```
gila: This is tricky! First off, if we ditch the Option we can write this as:

```rust
if self.share_protocol != ShareProtocol::None {
// we should already be shared -- as CSI is idempotent chances are we get called for some odd 
// reason. Validate indeed -- that we are shared by walking the target. If so,  simply return Ok()
// if we are not shared but the variant says we should be, try to correct the state if the variants are the 
// same. If they are not, return an error making sure we return the right error.
}
```

mayastor/src/bdev/nexus/nexus_share.rs
```diff
@@ -69,52 +75,97 @@ impl Nexus {
...
+                        name: self.name.clone(),
+                    })?;
             }
-            None => Err(Error::NotShared {
-                name: self.name.clone(),
-            }),
+        } else {
+            warn!("Missing bdev for a shared device");
         }
+        self.share_protocol = None;
```
gila: ShareProtocol::None.

mayastor/src/target/iscsi.rs
```diff
@@ -82,101 +88,70 @@ thread_local! {
...
 pub fn init(address: &str) -> Result<()> {
-    let portal_host = CString::new(address.to_owned()).unwrap();
-    let portal_port = CString::new(ISCSI_PORT.to_string()).unwrap();
-    let initiator_host = CString::new("ANY").unwrap();
-    let initiator_netmask = CString::new("ANY").unwrap();
 
-    let pg = unsafe { spdk_iscsi_portal_grp_create(0) };
-    if pg.is_null() {
-        return Err(Error::CreatePortalGroup {});
+    if let Err(e) = create_portal_group(address, ISCSI_PORT_BE, ISCSI_PORTAL_GROUP_BE) {
```
gila: you can simply write:

```rust
create_portal_group(...)?;
```

If there is an error, it will return it.

mayastor/src/target/iscsi.rs
```diff
@@ -82,101 +88,70 @@ thread_local! {
...
-/// Export given bdev over iscsi. That involves creating iscsi target and
-/// adding the bdev as LUN to it.
-pub fn share(uuid: &str, bdev: &Bdev) -> Result<()> {
-    let iqn = target_name(uuid);
+fn share_as_iscsi_target(bdev_name: &str, bdev: &Bdev, pg_idx: c_int, ig_idx: c_int ) -> Result<String ,Error>{
+
+    let iqn = target_name(bdev_name);
     let c_iqn = CString::new(iqn.clone()).unwrap();
-    let mut group_idx: c_int = 0;
+    let mut portal_group_idx = pg_idx;
```
gila: Why copy these values?

chriswldenyer: Agreed, I'll fix it.

mayastor/src/target/iscsi.rs
chriswldenyer: Ok, will change as suggested.

mayastor/src/bdev/nexus/nexus_bdev.rs
chriswldenyer: Ok, will fix.

mayastor/src/bdev/nexus/nexus_rpc.rs
```diff
@@ -130,8 +132,13 @@ pub(crate) fn register_rpc_methods() {
             let key: Option<String> =
                 if args.key == "" { None } else { Some(args.key) };
 
+            let share_proto = ShareProtocolNexus::from_i32(args.share);
+            if share_proto.is_none() {
+                return Err(Error::InvalidShareProtocol {sp_value: args.share as i32});
+            }
```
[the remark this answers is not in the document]
blaisedias: Fixed. https://github.com/openebs/Mayastor/pull/125/commits/74322e937b211cca53728d59bf35938af0e6c4ed

rpc/proto/mayastor.proto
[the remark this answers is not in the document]
blaisedias: If we want to use None and Nbd then the enum will not be an exact set for either replica or nexus. This means the code will be peppered with handle Nbd correctly for replicas and handle None correctly for nexus. IMHO more error prone.

rpc/proto/mayastor.proto
```diff
@@ -47,13 +47,22 @@ message ListPoolsReply {
   repeated Pool pools = 1;  // list of the pools
 }
 
-// Protocol for remote storage access which exposes a replica.
+// Protocol for remote storage access which exposes a replica and frontend.
```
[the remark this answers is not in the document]
blaisedias: Will do.

rpc/proto/mayastor.proto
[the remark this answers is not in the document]
blaisedias: Address at least jkryls request, https://github.com/openebs/Mayastor/pull/125/commits/b7341b5c775ab69c18188420ea8efff6a07dbe3f

mayastor-test/test_nexus.js
```diff
@@ -343,8 +345,27 @@ describe('nexus', function() {
...
+          assert.equal(err.code, grpc.status.INVALID_ARGUMENT);
+          done();
+      });
+  });
+*/
+
   it('should publish the nexus using nbd', done => {
-    client.PublishNexus({ uuid: UUID }, (err, res) => {
+    // TODO: repeat this test for iSCSI and Nvmf
+    client.PublishNexus({ uuid: UUID, share: mayastorProtoConstants.ShareProtocolNexus.NBD_FE }, (err, res) => {
```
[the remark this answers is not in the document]
blaisedias: Changed https://github.com/openebs/Mayastor/pull/125/commits/b7341b5c775ab69c18188420ea8efff6a07dbe3f

rpc/proto/mayastor.proto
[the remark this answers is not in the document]
blaisedias: Fixed https://github.com/openebs/Mayastor/pull/125/commits/b7341b5c775ab69c18188420ea8efff6a07dbe3f

cli/src/convert.rs
```diff
@@ -23,3 +24,12 @@ pub(crate) fn parse_size(src: &str) -> Result<u64, String> {
         Err(src.to_string())
     }
 }
+
+pub (crate) fn parse_proto(src: &str) -> Result<ShareProtocolNexus, &str> {
+    match src.to_lowercase().trim() {
+        "nbd" => Ok(ShareProtocolNexus::NbdFe),
+        "nvmf" => Ok(ShareProtocolNexus::NvmfFe),
+        "iscsi" => Ok(ShareProtocolNexus::IscsiFe),
+        _ => Err("Protocol needs be either NVMf, ISCI or NBD"),
```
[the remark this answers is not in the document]
blaisedias: https://github.com/openebs/Mayastor/pull/125/commits/da6fc6490194d70bd6e5387c2666384bc4351f10

mayastor-test/test_csi.js
blaisedias: Fixe, https://github.com/openebs/Mayastor/pull/125/commits/85459d5655e148ae4eab771a3685a17e3d2089ab

mayastor/src/bdev/nexus/nexus_rpc.rs
blaisedias: Fixed, https://github.com/openebs/Mayastor/pull/125/commits/74322e937b211cca53728d59bf35938af0e6c4ed

blaisedias: Fixed by https://github.com/openebs/Mayastor/pull/125/commits/74322e937b211cca53728d59bf35938af0e6c4ed

mayastor/src/bdev/nexus/nexus_share.rs
```diff
@@ -69,52 +75,97 @@ impl Nexus {
         debug!("creating share handle for {}", name);
         // The share handle is the actual bdev that is shared through the
         // various protocols.
-        let disk = Disk::create(&name).await.context(ShareNexus {
-            name: self.name.clone(),
-        })?;
-        let device_path = disk.get_path();
+
+        let return_val = match share_protocol {
```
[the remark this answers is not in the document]
chriswldenyer: Fixed in commit https://github.com/openebs/Mayastor/commit/8fe37c12e6ad29c1c6c7d94da22f9d67a47a026b

mayastor/src/bdev/nexus/nexus_bdev.rs
chriswldenyer: Fixed in commit https://github.com/openebs/Mayastor/commit/8fe37c12e6ad29c1c6c7d94da22f9d67a47a026b

mayastor/src/target/iscsi.rs
chriswldenyer: Fixed in commit https://github.com/openebs/Mayastor/commit/8fe37c12e6ad29c1c6c7d94da22f9d67a47a026b

mayastor/src/target/iscsi.rs
chriswldenyer: Fixed in commit https://github.com/openebs/Mayastor/commit/8fe37c12e6ad29c1c6c7d94da22f9d67a47a026b

mayastor/src/bdev/nexus/nexus_bdev.rs
chriswldenyer: Will do as suggested.

chriswldenyer: Resolved in https://github.com/openebs/Mayastor/pull/125/commits/d2abbaa1d04637148b05dac687b3b713b529794b

mayastor/src/bdev/nexus/nexus_share.rs
chriswldenyer: Resolved in https://github.com/openebs/Mayastor/pull/125/commits/d2abbaa1d04637148b05dac687b3b713b529794b

mayastor/src/bdev/nexus/nexus_share.rs
chriswldenyer: self.share_protocol no longer exists

blaisedias:
> 
> How about the CLI using grpc? It uses proto file as well. Do we need to make any changes there?
mctl has been updated and client.rs only handles pool and replicas so no change required there.

chriswldenyer:
Code review changes have been addressed. 
Iscsi publish nexus test-cases have been added to test_nexus.js to match those done for NBD.
In addition, the publish nexus request now returns the URI of the nexus, not just the IQN.

chriswldenyer:
Re-open to trigger CI/CD

mayastor/src/bdev/nexus/nexus_iscsi.rs
```diff
@@ -0,0 +1,72 @@
...
+            Ok(()) => (),
+            Err(e) =>  error!("Failed to destroy iscsi frontend target, error {}", e),
+        }
+    }
+
+    fn get_iqn(&self) -> String {
+        target_name(&self.bdev_name)
+    }
+
+    pub fn get_uri(&self) ->String {
```
gila: nit; 

Why cant we call create_uri(Side::Nexus, &self.bdev_name)?

More rustic would be to call the function "as_uri()" or "into_uri()" into typically consumes the variable though.

mayastor/src/core/mod.rs
```diff
@@ -78,3 +78,9 @@ pub enum CoreError {
         len: usize,
     },
 }
+
+// Which kind of target interface to use for a bdev
```
gila: This does not belong here. We should move it to ../target

mayastor-test/test_csi.js
```diff
@@ -197,7 +198,13 @@ describe('csi', function() {
               let uuid = BASE_UUID + n;
               common.dumbCommand(
                 'publish_nexus',
-                { uuid: uuid, key: '' },
+                {
+                  uuid: uuid,
+                  key: '',
+                  // TODO: repeat this test for iSCSI and Nvmf
```
gila: I see a lot of TODO's lets remove the TODO's and add issues for this such that we can actually do them. Otherwise they get burried in the code and we will forget about them

mayastor-test/test_nexus.js
```diff
@@ -343,30 +346,65 @@ describe('nexus', function() {
     });
   });
 
+  /*  FIXME:
```
gila: This looks like something we should either file an issue for or fix before integration?

mayastor-test/test_nexus.js
```diff
@@ -343,30 +346,65 @@ describe('nexus', function() {
...
+  });
+*/
+
   it('should publish the nexus using nbd', done => {
-    client.PublishNexus({ uuid: UUID }, (err, res) => {
-      assert(res.device_path);
-      nbd_device = res.device_path;
-      done();
-    });
+    // TODO: repeat this test for Nvmf
```
gila: Todo...

mayastor-test/test_nexus.js
```diff
@@ -343,30 +346,65 @@ describe('nexus', function() {
...
     });
   });
 
   it('should re-publish the nexus using NBD, and a crypto key', done => {
-    client.PublishNexus({ uuid: UUID, key: '0123456789123456' }, (err, res) => {
-      assert(res.device_path);
-      nbd_device = res.device_path;
-      done();
-    });
+    // TODO: repeat this test for Nvmf
```
gila: Todo...

mayastor-test/test_nexus.js
```diff
@@ -343,30 +346,65 @@ describe('nexus', function() {
...
+      (err, res) => {
+        assert(res.device_path);
+        nbd_device = res.device_path;
+        done();
+      }
+    );
   });
 
   it('should be able to write to the NBD device', async () => {
+    // TODO: repeat this test for Nvmf
```
gila: Todo....

mayastor-test/test_nexus.js
```diff
@@ -387,6 +426,116 @@ describe('nexus', function() {
     });
   });
 
+  it('should un-publish the NBD nexus device', done => {
+    // TODO: repeat this test for Nvmf
+    client.unpublishNexus({ uuid: UUID }, (err, res) => {
```
gila: todo...

mayastor-test/test_nexus.js
```diff
@@ -440,10 +589,26 @@ describe('nexus', function() {
     });
   });
 
-  it('should create, publish, un-publish and finally destroy the same nexus', async () => {
+  it('should create, publish, un-publish and finally destroy the same NBD nexus', async () => {
     for (let i = 0; i < 10; i++) {
       await createNexus(createArgs);
-      await publish({ uuid: UUID });
+      // TODO: repeat this test for Nvmf
```
gila: Todo...

deploy/pvc.yaml
```diff
@@ -7,5 +7,5 @@ spec:
    - ReadWriteOnce
   resources:
     requests:
-      storage: 64Mi
+      storage: 1Gi
```
gila: Why do we need to create such a big volume? this implies we MUST have 1Gi everywhere. If 64Mi is not suffucient, would 128Mi do?

mayastor/src/bdev/nexus/nexus_iscsi.rs
chriswldenyer: > Why cant we call create_uri(Side::Nexus, &self.bdev_name)?
True - that's the only place the function is called now so it could go. I'll rename it to as_uri().

mayastor-test/test_csi.js
chriswldenyer: There is already a Jira (CAS-26) for Nvmf integration testing. I'll remove the TODO comments.

mayastor-test/test_nexus.js
chriswldenyer: There is already a Jira (CAS-26) for Nvmf integration testing. I'll remove the TODO comments.

mayastor-test/test_nexus.js
chriswldenyer: Will remove comment

mayastor-test/test_nexus.js
chriswldenyer: Will remove comment

mayastor-test/test_nexus.js
chriswldenyer: Will remove comment

mayastor-test/test_nexus.js
chriswldenyer: Will remove comment

mayastor/src/core/mod.rs
chriswldenyer: Will move to target.

deploy/pvc.yaml
blaisedias: This change was made so that we could use the scripts in the deploy directory to do what is described in quick.md leading to less confusion for people trying out Mayastor.
I will revert, or change quick.md.

blaisedias: Reverted.

mayastor-test/test_csi.js
blaisedias: Fixed by https://github.com/openebs/Mayastor/pull/125/commits/94d78ef5c45d51693a4f85c8f134259bf02ca7c3

mayastor-test/test_nexus.js
chriswldenyer: Commented-out test removed. JIRA raised: https://mayadata.atlassian.net/browse/CAS-161

PR #127, opened 2020-03-17
Implement initial rebuild task with a few limitations

tiagolobocastro:
(CAS-94) Initial rebuild task imp with a few constraints:
- Single Mayastor Reactor
- Rebuild acts on 1 Segment at time
- Rebuilds from only 1 child to only 1 child
- Assume no host IO during the rebuild
- Assume no IO failures during the rebuild

When a child is added it remains in the faulted state until it is rebuilt from another healthy child (the nexus also remains degraded)
the rebuild is not automatically triggered, but rather by using the RPC start_rebuild, this is try and isolate this initial effort from the rest of mayastor.
Once the rebuild completes the newly added child can then start servicing nexus IOs and the nexus state also moves back to online.

coded by @paulyoong  & @tiagolobocastro


mayastor/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -138,10 +141,17 @@ impl Nexus {
...
+
                 self.children.push(child);
                 self.child_count += 1;
-                // TODO -- rsync labels
-                Ok(self.set_state(NexusState::Degraded))
+                self.set_state(NexusState::Degraded);
+
+                if let Err(e) = self.sync_labels().await {
+                    error!("Failed to sync labels {:?}", e);
+                    // todo: how to signal this?
```
gila: Can we create an issue for this? IMHO its rather important and not something we should shoehorn in later.

mayastor/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -158,6 +168,138 @@ impl Nexus {
         }
     }
 
+    pub async fn start_rebuild(
+        &mut self,
+        destination: &str,
+    ) -> Result<(), Error> {
+        trace!("{}: start rebuild request for {}", self.name, destination);
+
+        let source = match self
```
gila: I see this pattern in several places, perhaps its an idea to create a function for this? Something like child_lookup()?

mayastor/src/rebuild.rs
```diff
@@ -1,284 +1,224 @@
...
-    Suspended,
+    Completed,
 }
 
-/// struct that holds the state of a copy task. This struct
-/// is used during rebuild.
 #[derive(Debug)]
 pub struct RebuildTask {
+    nexus_name: String,
+    source: String,
```
gila: If we instead of storing Strings and BdevHandles, simply store the Descriptor, we can get the name and handle *from* the descriptor.

iirc. Descriptor as a method .get_bdev() which in turn has .get_name(). The name though is not really important to us outside of logging or things like that. 

I think we discussed this yesterday, so we can leave it for now, but please do file an issue for this.

mayastor/src/rebuild.rs
```diff
@@ -1,284 +1,224 @@
...
 
-    /// function called when the rebuild has completed. We record something in
-    /// the logs that provides some information about the time in seconds
-    /// and average throughput mbs.
-    fn rebuild_completed(&mut self) {
-        let elapsed = self.start_time.unwrap().elapsed().unwrap();
-        let mb = (self.source.get_bdev().block_len() as u64
-            * self.source.get_bdev().num_blocks())
-            >> 20;
+pub trait RebuildActions {
```
gila: As mentioned yesterday: https://docs.rs/async-trait/0.1.24/async_trait/

mayastor/src/rebuild.rs
```diff
@@ -1,284 +1,224 @@
...
-    /// Ok(true). When a new IO has been successfully dispatched in returns
-    /// Ok(false)
-    ///
-    /// When memory allocation fails, it shall return an error no attempts will
-    /// be made to restart a build automatically, ideally we want this to be
-    /// done from the control plane and not internally, but we will implement
-    /// some form of implicit retries.
-    fn next_segment(&mut self) -> Result<RebuildState, Error> {
-        let num_blocks = self.num_blocks();
+            self.copy_buffer = self
```
gila: We already allocated a buffer when we constructed the rebuild tasks, why allocate a new one?

mayastor/src/rebuild.rs
```diff
@@ -1,284 +1,224 @@
...
-    }
-
-    /// progress function that prints to the log; this will be removed in the
-    /// future and will be exposed via an API call
-    extern "C" fn progress(_ctx: *mut c_void) -> i32 {
-        unimplemented!()
+    fn send_complete(&self) {
+        self.stats();
+        let complete = self.complete;
+        complete(self.nexus_name.clone(), self.destination.clone());
```
gila: nit; you can change this to (self.complete)(..args)

mayastor/src/rebuild.rs
```diff
@@ -1,284 +1,224 @@
...
-                self.start_progress_poller();
-            }
-        }
+        None
+    }
+
+    // todo: ideally we'd want the nexus out of here but sadly rust does not yet
+    // support async trait's
+    // the course of action might just be not using traits
+    fn start(&mut self) {
```
gila: https://docs.rs/async-trait/0.1.24/async_trait/

mayastor/tests/nexus_rebuild.rs
```diff
@@ -0,0 +1,144 @@
...
+    std::thread::spawn(move || {
+        s.send(compare_nexus_device(&nexus_device, DISKNAME2, false))
+    });
+    reactor_poll!(r);
+
+    // add the second child
+    nexus.add_child(BDEVNAME2).await.unwrap();
+
+    // kick's off the rebuild (NOWAIT)
+    nexus.start_rebuild(BDEVNAME2).await.unwrap();
```
gila: We should have start_rebuild return a Receiver<T> and then we can await it and get rid of the thread sleep hack.

mayastor/tests/nexus_rebuild.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+async fn create_nexus() {
+    let ch = vec![BDEVNAME1.to_string()];
+    nexus_create(NEXUS_NAME, NEXUS_SIZE, None, &ch)
+        .await
+        .unwrap();
+}
+
+pub fn dd_urandom(device: &str) -> String {
```
gila: the helper functions should be moved into test/common.rs

mayastor/src/bdev/nexus/nexus_bdev_children.rs
tiagolobocastro: @cjones1024 this is the one

mayastor/src/rebuild.rs
tiagolobocastro: It's done temporarily as the BDevHandle seem to use the full len of the DmaBuff when issue IO, and there seems to be no way of setting the len to a smaller size.
So we add new read/write fn or enhance the DmaBuff to expose a "virtual" len? hmm

mayastor/tests/nexus_rebuild.rs
tiagolobocastro: will be coming shortly :)

mayastor/tests/nexus_rebuild.rs
tiagolobocastro: yeah I wondered about that, but didn't want to pollute it without approval, I'll do so now :)

PR #132, opened 2020-03-18
cargo fmt changes are not being detected by the linter

tiagolobocastro:
the changes are only applied during test and do not fail the PR so unformatted code can be checked in master...

currently the spdk-sys submodule has ill formatted code... we avoid running the format check against submodules as that should be handled in the respective repos

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -25,7 +25,9 @@ lint:
     - nix-shell --run 'prettier --version'
     - nix-shell --run 'prettier --config .prettierrc --check csi/moac/*.js mayastor-test/*.js'
     - nix-shell --run 'jshint --config .jshintrc csi/moac/*.js mayastor-test/*.js'
-    - nix-shell --run 'cargo fmt --all'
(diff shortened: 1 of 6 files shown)


.gitlab-ci.yml
```diff
@@ -25,7 +25,9 @@ lint:
     - nix-shell --run 'prettier --version'
     - nix-shell --run 'prettier --config .prettierrc --check csi/moac/*.js mayastor-test/*.js'
     - nix-shell --run 'jshint --config .jshintrc csi/moac/*.js mayastor-test/*.js'
-    - nix-shell --run 'cargo fmt --all'
+    # We don't want to perform this check for the submodules as it should be done in their respective repos
+    - echo 'set -e; for d in */Cargo.toml; do dir=`dirname "$d"`; if [ ! -f "$dir"/.git ]; then echo "cargo fmt -p "$dir" -- --check"; cargo fmt -p "$dir" -- --check; fi; done' > fmt.sh
```
jonathan-teh: Prefer  `$(...)` to backticks but otherwise looks fine.

PR #135, opened 2020-03-20
Implementation of gRPC in Mayastor using Tokio

yannis218:
Mayastor now implements its own gRPC interface, so that calls no longer have to go via the agent.
Tests can now be pointed at the new interface by setting the environment variable TEST_PORT.


mayastor/src/pool.rs
```diff
@@ -355,7 +405,55 @@ impl Iterator for PoolsIter {
     }
 }
 
-fn list_pools() -> Vec<jsondata::Pool> {
+pub async fn create_pool(args: rpc::mayastor::CreatePoolRequest)
```
gila: Some functions are pub some of them are pub (crate) let us try to use pub (crate) everywhere where possible.

shell.nix
```diff
@@ -29,6 +29,7 @@ mkShell {
     nodejs-10_x
     nvme-cli
     pre-commit
+    python2
```
gila: Why do we need python2?

gila:
Also, please run the pre-commit hooks, as the PR fails cargo and I think fmt as well.

mayastor/src/grpc.rs
```diff
@@ -0,0 +1,286 @@
...
+        -> std::result::Result<(), ()> {
+    info!("gRPC server configured at address '{}:{}'", addr, port);
+    let saddr = format!("{}:{}", addr, port).parse().unwrap();
+    let svc = Server::builder()
+        .add_service(MayastorServer::new(MayastorGrpc {}))
+        .serve(saddr);
+
+    match svc.await {
+        Ok(_) => Ok(()),
+        Err(_) => Err(()),
```
gila: At the every least we should print the error to stdout here?

mayastor/src/core/env.rs
```diff
@@ -199,6 +213,9 @@ impl Default for MayastorEnvironment {
     fn default() -> Self {
         Self {
             config: None,
+            enable_grpc: false,
```
jkryl: This cannot be optional. grpc server must run as control plane depends on it. Or is it something you plan to change once mayastor grpc code is removed from ms agent?

mayastor/src/core/env.rs
```diff
@@ -199,6 +213,9 @@ impl Default for MayastorEnvironment {
...
+            grpc_addr: "None".into(),
```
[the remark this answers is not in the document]
jkryl: The default should be really the POD IP address taken from environment variable MY_POD_IP. And 127.0.0.1 can be used if MY_POD_IP is not set and cmd line argument not specified.

mayastor/src/core/env.rs
```diff
@@ -199,6 +213,9 @@ impl Default for MayastorEnvironment {
...
+            grpc_port: "None".into(),
```
jkryl: Why don't we pick a new port number here? Just like mayastor-agent has a default port used for grpc, the mayastor should have a default port number different from mayastor-agent, to make it simple to run both concurrently until mayastor grpc server is removed from mayastor-agent.

mayastor/src/core/env.rs
gila: @jkryl this actually more complicated than just that. For example, the nvmf initiator should not start the gRPC server when we are running this during the tests. Does that make sense to you?

jkryl: AFAIK our initiator utility for testing iscsi and nvmf targets does not use core::env. mayastor seems to be the only one using this code. Though if it makes more sense to leave this optional and hardcode enable_grpc parameter in mayastor code that creates the env, then that is also possible.

csi/src/client.rs
```diff
@@ -27,6 +43,68 @@ fn parse_share_protocol(pcol: Option<&str>) -> Result<i32, Status> {
...
+    client.destroy_nexus(request).await?;
+
+    Ok(())
+}
+
+async fn create_nexus(
+    mut client: MayaClient,
+    matches: &ArgMatches<'_>,
+) -> Result<(), Status> {
+    let size = parse_size(matches.value_of("size").unwrap()).unwrap();
```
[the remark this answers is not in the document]
yannis218: Done.

mayastor/src/core/env.rs
```diff
@@ -632,14 +651,34 @@ impl MayastorEnvironment {
...
+            .unwrap();
+
+        let local = task::LocalSet::new();
+        rt.block_on(async {
+            local
+                .run_until(async {
+                    let master = Reactors::current();
+                    master.send_future(async { f() });
+                    if grpc {
+                        let _out = tokio::try_join!(grpc_server_init(&grpc_addr, &grpc_port), master);
```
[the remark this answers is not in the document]
yannis218: The try_join returns Result<(), ()> so there is no info to print out.

mayastor/src/pool.rs
yannis218: OK, I have tried to make it more consistent.

shell.nix
yannis218: Eh... because it is used?  I suspect that it has been working for people because it was already there.  I did most of this work on my VM at work, which runs Ubuntu Server, which does not come with python2 by default, and the tests did not work.  I do not now remember what exactly failed, but it worked again when I added this.  On my VM at home, however, I run Lubuntu, which has python2 by default, so when I took it out it all still worked.  I suggest we keep it here as an explicit entry, because as I said it does appear to be a requisite.

mayastor-test/test_nexus.js
```diff
@@ -118,7 +118,7 @@ function createGrpcClient(service) {
         oneofs: true,
       },
     },
-    common.endpoint
```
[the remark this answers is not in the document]
yannis218: Added extra runs of test_nexus and test_replica with the port redirected to the new one.

mayastor/src/replica.rs
```diff
@@ -137,6 +147,23 @@ impl RpcErrorCode for Error {
...
+        Error::PoolNotFound { .. } => Self::not_found(e.to_string()),
+        Error::ReplicaExists { .. } => Self::already_exists(e.to_string()),
+        Error::InvalidParams { .. } => Self::invalid_argument(e.to_string()),
+        Error::CreateLvol { .. } => Self::invalid_argument(e.to_string()),
+        Error::DestroyLvol { .. } => Self::internal(e.to_string()),
+        Error::ReplicaShared { .. } => Self::internal(e.to_string()),
+        Error::ShareNvmf { source:_ } => Self::internal(e.to_string()),//source.into(),
+        Error::ShareIscsi { source:_ } => Self::internal(e.to_string()),//source.into(),
+        Error::UnshareNvmf { source:_ } => Self::internal(e.to_string()),//source.into(),
+        Error::UnshareIscsi { source:_ } => Self::internal(e.to_string()),//source.into(),
```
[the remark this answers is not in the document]
yannis218: Commented out stuff that never got cleaned up.  Cleaned up now.

mayastor-test/test_replica.js
```diff
@@ -85,7 +85,7 @@ function createGrpcClient(service) {
 describe('replica', function() {
   var client;
 
-  this.timeout(10000); // for network tests we need long timeouts
+  this.timeout(50000); // for network tests we need long timeouts
```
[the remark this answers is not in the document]
yannis218: On my heavily loaded system it was timing out, until I did this.  I am happy to not push this.

shell.nix
yannis218: Taken out.

mayastor/src/grpc.rs
yannis218: OK, done.

yannis218: Done.

shell.nix
yannis218: Removed.

mayastor/src/core/env.rs
yannis218: The initiator utility *does* use core::env

mayastor/src/grpc.rs
yannis218: Done.

mayastor/src/core/env.rs
jkryl: Indeed, initially, it did not, but it was changed 2 months ago. So I agree that the current approach makes sense then 👍

mayastor/src/core/env.rs
yannis218: Yep, that's exactly what happens.  The "None" string here gets overwritten by the defaults from the structopt, which as can be seen elsewhere use a new port, 10125.  The invalid value is here as a sort of guard - if the overwriting is missed out, something will choke further down.

mayastor-test/test_replica.js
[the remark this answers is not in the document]
yannis218: I am reverting this.

mayastor/src/core/env.rs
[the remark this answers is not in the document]
yannis218: Structopt claims you can write ``default_value`` without the ``="whatever" `` in the declaration, and then it will pick up the value from the Default impl.  In practice, I could not get this to work, and had to put the default in the structopt declaration.  However, the Default impl still needed a value.  That is why the "127.0.0.1" ended up in two places.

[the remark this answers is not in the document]
jkryl: Just an idea, but perhaps we could avoid enable_grpc and strange "None" default value by having an optional member in ms env that is a tuple `(address, port)`. We could call it `grpc_endpoint` and it would be simply set to `None` if grpc is not used. So host part would be `grpc_endpoint.0` and port `grpc_endpoint.1`. It also eliminates the case when enable_grpc=true but grpc_addr remains "None" because it was not set.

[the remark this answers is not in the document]
yannis218: No, no.  The value is always set and never remains "None".  The "None" is overwritten by the value from structopt unconditionally.  As for merging the enable flag with the presence or not of the socket settings, I think that will confuse the logic - the flag is set by the client code that uses env.  The socket settings are set by the end user.

yannis218:
> You seem to have some odd commits in there which we do not want in the master branch, can you look into that?

What commits are you referring to?

mayastor/src/core/env.rs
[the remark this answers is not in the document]
jkryl: I will try to explain better what I mean:

in struct MayastorEnvironment we would have:
```
+       grpc_endpoint: (String, String),
```
In impl Default for MayastorEnvironment:
```
+       grpc_endpoint: None,
```
and in MayastorEnvironment::new:
```
   Self {
        ...
+      grpc_endpoint: (args.addr, args.port),
        ..Default::default()
   }
```
I find this as a more natural way how to express what we want in rust. First, we don't need a special placeholder string "None" if rust already has a special value to denote empty values (Optional). Second, I don't need three independent variables in mayastor env where only a few combinations of them make sense - so less error-prone I would say. Not sure how that seems to you guys?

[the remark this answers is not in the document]
yannis218: "None" is not a valid value - it is not used to signify a not-specified address, unlike an Option which is None.  Also, in your scheme, how would the client code enable or disable the grpc?

PR #136, opened 2020-03-20
mayastor-test: Test uring bdevs in test_nexus

jonathan-teh:
Only test uring bdevs if the filesystem and kernel support it.
Move the functions testing for io_uring support from the core unit test to bdev::uring_util. Add a CLI tool that calls these functions given a path so that it can be called from test_nexus.

### DIFF
--- a/mayastor-test/test_nexus.js
+++ b/mayastor-test/test_nexus.js
@@ -125,10 +125,10 @@ function createGrpcClient(service) {
   );
 }
 
-var doUring = (function () {
+var doUring = (function() {
   var executed = false;
   var supportsUring = false;
-  return function () {
+  return function() {
     if (!executed) {
       executed = true;
       const { exec } = require('child_process');
@@ -140,7 +140,7 @@ var doUring = (function () {
         'uring-support'
       );
       const CMD = URING_SUPPORT_CMD + ' ' + uringFile;
-      exec(CMD, (error) => {
+      exec(CMD, error => {
         if (error) {
           return;
         }
@@ -148,7 +148,7 @@ var doUring = (function () {
       });
     }
     return supportsUring;
-  }
+  };
 })();
 
 describe('nexus', function() {
@@ -263,10 +263,8 @@ describe('nexus', function() {
           fs.unlink(aioFile, err => next());
         },
         next => {
-          if (doUring())
-            fs.unlink(uringFile, err => next());
-          else
-            next();
+          if (doUring()) fs.unlink(uringFile, err => next());
+          else next();
         },
       ],
       err => {
@@ -289,8 +287,7 @@ describe('nexus', function() {
         `nvmf://127.0.0.1:8420/nqn.2019-05.io.openebs:disk2`,
       ],
     };
-    if (doUring())
-      args.children.push(`uring:///${uringFile}?blk_size=4096`);
+    if (doUring()) args.children.push(`uring:///${uringFile}?blk_size=4096`);
 
     client.CreateNexus(args, done);
   });
@@ -321,7 +318,10 @@ describe('nexus', function() {
       );
       assert.equal(nexus.children[3].state, 'open');
       if (doUring()) {
-        assert.equal(nexus.children[4].uri, `uring:///${uringFile}?blk_size=4096`);
+        assert.equal(
+          nexus.children[4].uri,
+          `uring:///${uringFile}?blk_size=4096`
+        );
         assert.equal(nexus.children[4].state, 'open');
       }
       done();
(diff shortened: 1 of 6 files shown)


mayastor/src/bdev/uring_util.rs
```diff
@@ -0,0 +1,87 @@
+use std::{
+    fs::{File, OpenOptions},
+    io::{BufRead, BufReader, ErrorKind},
+    os::unix::fs::OpenOptionsExt,
+};
+
+pub fn fs_supports_direct_io(path: &str) -> bool {
+    // SPDK uring bdev uses IORING_SETUP_IOPOLL which requires O_DIRECT
```
gila: nit; what I read is that the FS or Blk dev needs to support polling, not direct IO.

jonathan-teh: According to io_uring_setup(2):
```
       IORING_SETUP_IOPOLL
              Perform  busy-waiting  for an I/O completion, as opposed to get‐
              ting notifications via an asynchronous IRQ (Interrupt  Request).
              The  file  system (if any) and block device must support polling
              in order for this to work.  Busy-waiting provides lower latency,
              but  may  consume  more CPU resources than interrupt driven I/O.
              Currently, this feature is usable  only  on  a  file  descriptor
              opened using the O_DIRECT flag.  When a read or write is submit‐
              ted to a polled context, the application must poll  for  comple‐
              tions  on the CQ ring by calling io_uring_enter(2).  It is ille‐
              gal to mix and match polled and non-polled I/O  on  an  io_uring
              instance.
```
I guess that means both, polling and IO_DIRECT.

mayastor/src/bdev/uring_util.rs
```diff
@@ -0,0 +1,89 @@
...
+        Ok(_f) => true,
+        Err(e) => {
+            assert_eq!(e.kind(), ErrorKind::InvalidInput);
+            println!("Skipping uring bdev, open: {:?}", e);
+            false
+        }
+    }
+}
+
+fn get_mount_filesystem(path: &str) -> Option<String> {
```
jkryl: just a remark: we have a bunch of mount related code in csi/src/mount.rs that can serve as an inspiration. But if this code works ok, then I guess it's fine. Sharing the code between ms and agent would be too difficult.

mayastor/src/bdev/uring_util.rs
```diff
@@ -0,0 +1,89 @@
...
+    let mut path = std::path::Path::new(path);
+    loop {
+        let f = match File::open("/etc/mtab") {
+            Ok(f) => f,
+            Err(e) => {
+                eprintln!("open: {}", e);
```
jkryl: I'm not sure about these println's and eprintln's in this file. Currently, this library module is used only by a small command -line utility used in our tests to determine if io_uring tests should or should not be run. From that POV that seems to be ok. However, eventually (correct me if I'm wrong) this code is going to be used from mayastor itself to determine if aio or uring should be used for a storage pool. In that case, println's must be removed or replaced by proper log function calls. What is the plan here?

mayastor-test/test_nexus.js
```diff
@@ -187,6 +216,17 @@ describe('nexus', function() {
...
           fs.truncate(aioFile, diskSize, next);
         },
+        next => {
+          fs.writeFile(uringFile, '', next);
+        },
+        next => {
+          fs.truncate(uringFile, diskSize, next);
+        },
+        next => {
+          if (doUring())
```
jkryl: just a general note, it is possible to skip tests in mocha. In that case, the test is highlighted in the output making it easy for anyone to see that the test exists but wasn't run for some reason. As it is written now, it does not give a clue that some functionality of mayastor was not tested. On the other hand, duplicating current tests just for the purpose of testing io_uring would be probably even worse :-)

jkryl: Replying to myself... :-) Would we prefer to fail the tests instead? I mean if we ever make a mistake and misconfigure our ci/cd pipeline to run on a system that does not support io_uring, I would rather see it to fail (with error message explaining why) than to not notice that something is not tested.

mayastor/src/bdev/uring_util.rs
jonathan-teh: Yes, when it is eventually called from the pool code then it should switch to proper logging.

mayastor-test/test_nexus.js
jonathan-teh: I went through this with @gila and there were discussions on whether we could set up a loop device and either use it directly or format it as XFS and use a file there. The problem comes with the clean up stage and, from past experience, failing to clean up properly then leads to intermittent failures later. At the moment, it skips it on CI (where /tmp is overlay fs as it's inside the container) and on Ubuntu (which defaults to ext4), but will run elsewhere like CentOS/RHEL (defaults to XFS) or @gila's machine.

PR #141, opened 2020-03-24
First steps for a library that provides NVMf initiator support

AntonioCarlini:
First steps for a library that provides NVMf initiator support

CAS-113 The ability to connect to an NVMe remote target 
CAS-115 Discover NVME disks offered by a remote target


nvme-library/Cargo.lock
```diff
@@ -0,0 +1,349 @@
+# This file is automatically @generated by Cargo.
```
gila: This file should not be checked in, we use a single cargo.lock for the whole of mayastor.

gila:
@AntonioCarlini I cant see the changes in the library you made like this, what if I commit the original library and you put your changes on top of that? That way I can also prevent the cargo.lock issues you seem to be having.

nvme-library/Cargo.lock
AntonioCarlini: OK. I'll undo that

AntonioCarlini:
I did change the way connect worked slightly at one stage because it didn't seem to work on my test system. However, that was because of the way /dev/nvme-fabrics works (I think). So I reverted those. The only changes to src/*.rs now should be whatever formatting gubbins (if any) rustfmt threw in.

If you prefer, I can scrap the PR, you insert your sources and then I can issue a new PR?

PR #143, opened 2020-03-25
rebuild: Add stop rebuild support

paulyoong:
Support also included for getting rebuild state.
Unit tests added for the rebuild process.

### DIFF
--- a/mayastor-test/test_rebuild.js
+++ b/mayastor-test/test_rebuild.js
@@ -1,0 +1,379 @@
+// Unit tests for rebuild tasks
+
+'use strict';
+
+const { createClient } = require('grpc-kit');
+const async = require('async');
+const fs = require('fs');
+const common = require('./test_common');
+const path = require('path');
+const assert = require('chai').assert;
+
+// backend file for aio bdev
+const child1 = '/tmp/child1';
+const child2 = '/tmp/child2';
+// 64MB is the size of nexus and replicas
+const diskSize = 64 * 1024 * 1024;
+// nexus UUID
+const UUID = 'dbe4d7eb-118a-4d15-b789-a18d9af6ff21';
+// external IP address detected by common lib
+const externIp = common.getMyIp();
+
+// Instead of using mayastor grpc methods to create replicas we use a config
+// file to create them. Advantage is that we don't depend on bugs in replica
+// code (the nexus tests are more independent). Disadvantage is that we don't
+// test the nexus with implementation of replicas which are used in the
+// production.
+const configNexus = `
+[Malloc]
+  NumberOfLuns 2
+  LunSizeInMB  64
+  BlockSize    4096
+
+[iSCSI]
+  NodeBase "iqn.2019-05.io.openebs"
+  # Socket I/O timeout sec. (0 is infinite)
+  Timeout 30
+  DiscoveryAuthMethod None
+  DefaultTime2Wait 2
+  DefaultTime2Retain 60
+  ImmediateData Yes
+  ErrorRecoveryLevel 0
+  # Reduce mem requirements for iSCSI
+  MaxSessions 1
+  MaxConnectionsPerSession 1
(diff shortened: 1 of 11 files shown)


mayastor-test/test_rebuild.js
```diff
@@ -0,0 +1,379 @@
...
+  };
+
+  const rebuildState = args => {
+    return new Promise((resolve, reject) => {
+      client.getRebuildState(args, (err, data) => {
+        if (err) return reject(err);
+        resolve(data);
+      });
+    });
+  };
```
jkryl: please, have a look at https://github.com/openebs/Mayastor/blob/master/csi/moac/grpc_client.js#L66. I think grpc_promise and its `promisifyAll()` can help you here.

mayastor-test/test_rebuild.js
```diff
@@ -0,0 +1,379 @@
...
+        },
+        next => {
+          createNexus(nexusArgs)
+            .then(() => {
+              next();
+            })
+            .catch(err => {
+              assert(err);
+            })
+            .catch(done);
```
jkryl: IMO the three lines above this one are not needed. The last catch calls done callback with error argument.

mayastor-test/test_rebuild.js
```diff
@@ -0,0 +1,379 @@
...
+    });
+  });
+
+  describe('stopping rebuild', function() {
+    beforeEach(async () => {
+      await addChild(rebuildArgs);
+      await startRebuild(rebuildArgs);
+      await stopRebuild(rebuildArgs);
+      // TODO: Check for rebuild stop rather than sleeping
+      await sleep(1000); // Give time for the rebuild to stop
```
jkryl: Should not the stopRebuild return only after the rebuild was stopped? Just like startRebuild returns once rebuild has started. That would eliminate the need for the sleep at this line. Though I'm not sure if it isn't a bad idea from the implementation perspective.

rpc/proto/mayastor.proto
```diff
@@ -205,6 +207,11 @@ message StartRebuildRequest {
   string uri = 2;   // uri of the child to be rebuilt
 }
 
+message StopRebuildRequest {
+  string uuid = 1;  // uuid of the nexus
+  string uri = 2;   // uri of the child to be rebuilt
```
jkryl: copy-paste error in comment for the uri (and the same for the uri in rebuild-state-request)

mayastor-test/test_rebuild.js
```diff
@@ -0,0 +1,379 @@
...
+      let nexus = res.nexus_list[0];
+      assert.equal(nexus.uuid, UUID);
+
+      assert.equal(nexus.rebuilds, expected);
+    });
+  }
+
+  function sleep(ms) {
+    return new Promise(resolve => setTimeout(resolve, ms));
+  }
```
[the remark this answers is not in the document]
paulyoong: Done - now using sleep-promise

mayastor-test/test_rebuild.js
paulyoong: Done - modified the test code to use grpc_promise

mayastor-test/test_rebuild.js
paulyoong: Done - removed unnecessary lines

mayastor-test/test_rebuild.js
paulyoong: At the moment the stopRebuild is an asynchronous call which sets a "stop" flag. It doesn't actually wait for the stop to complete. This is because, in the future, there may be a number of segments rebuilding and we cannot stop the process until these complete. Therefore we didn't want to block the rpc call during this time.

rpc/proto/mayastor.proto
paulyoong: Amended comments.

PR #145, opened 2020-03-25
CAS-145: refactor nexus disk label generation 

cjones1024:
Extend NexusLabel struct to include protective MBR and secondary GPT header,
so that it contains all componets required to generate a complete disk label.

Refactor nexus_label and related packages so that all data related to a disk
label is written to each disk in sequence before moving on to the next disk.


cjones1024:
closing temporarily to trigger CI

cjones1024:
closing temporarily to trigger CI

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -252,65 +259,112 @@ impl NexusChild {
...
-        let primary = u64::from(block_size);
-        let secondary = bdev.num_blocks() - 1;
-
         let mut buf = desc
             .dma_malloc(block_size as usize)
             .context(LabelAlloc {})?;
 
-        self.read_at(primary, &mut buf)
+        self.read_at(0, &mut buf).await.context(LabelRead {})?;
+        let mbr = match Pmbr::from_slice(&buf.as_slice()[440 .. 512]) {
```
gila: This makes an invalid MBR an error, I don't think that is needed. The PMBR and the GPT label do not depend on one other AFAIK, right?

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -252,65 +259,112 @@ impl NexusChild {
...
             .await
             .context(LabelRead {})?;
+        let primary = match GPTHeader::from_slice(buf.as_slice()) {
+            Ok(header) => header,
+            Err(_) => {
+                warn!(
+                    "{}: {}: The primary GPT header is invalid!",
+                    self.parent, self.name
+                );
+                return Err(ChildError::LabelInvalid {});
```
gila: If the primary is invalid, why don't we try the second label anymore, as we did before this change?

gila:
I can't comment on binary files but why did you change them?

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -252,65 +259,112 @@ impl NexusChild {
...
+                    "{}: {}: The secondary GPT header is invalid!",
+                    self.parent, self.name
+                );
+                return Err(ChildError::LabelInvalid {});
+            }
         };
 
+        if primary.guid != secondary.guid {
+            warn!("{}: {}: The primary and secondary GPT headers are inconsistent: GUIDs differ!", self.parent, self.name);
+            return Err(ChildError::LabelInvalid {});
```
tiagolobocastro: How do other tools handle primary vs secondary in case of a mismatch? Eg do they just defer to the secondary as the current source of truth?

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -319,12 +373,41 @@ impl NexusChild {
...
-        Ok(nl)
+    /// write a label to this child
+    pub async fn write_label(
+        &mut self,
+        label: &NexusLabel,
+        data: &LabelData,
+        block_size: u64,
+    ) -> Result<(), ChildIoError> {
+        // Protective MBR
+        self.write_at(0 as u64, &data.mbr).await?;
```
tiagolobocastro: btw I think you can just use '0u64'

mayastor/src/bdev/nexus/nexus_child.rs
cjones1024: We should check the MBR as well before reconstructing the primary GPT from the secondary.
I had intended to address this fully in CAS-146, but will revert to previous behaviour.

mayastor/src/bdev/nexus/nexus_child.rs
cjones1024: To be fair, this is something that should not ever happen. The spec requires that the secondary GPT must be updated whenever the primary changes, so that they are always in sync. If they disagree then (assuming the primary is valid) the secondary should be updated accordingly.
However my line of thinking was that, as the nexus requires identical labels accross all child disks, and the existing code will update all labels on all disks anyway whenever it finds at least one that is different, we can mark this as a failure and then the nexus will update everything and all will be well.
Granted this is not as efficient as it could be, but I was hoping to address that as part of CAS-146.

mayastor/src/bdev/nexus/nexus_child.rs
cjones1024: There is a dependency between the MBR and the GPT. Acording to the spec, you should not attempt to reconstruct a primary GPT from the secondary unless it is consistent with the MBR. This is to avoid the following scenario:
A third party reformats the disk as MBR rather than GPT. This will probably result in the primary GPT label being overwritten more or less immediately, but more than likely leave the secondary GPT label situated at the end of the disk completeley intact - even though it is now invalid. If we reconstruct the primary GPT from this secondary without first checking for the existence of a protective partition in the MBR, then we will not detect that we are potentitally overwriting data.

cjones1024:
> 
> 
> I can't comment on binary files but why did you change them?

The original binary file contained the MBR, primary GPT header and the partition table from the beginning of the disk. I have now added a second binary file containing the secondary partition table and GPT header from the end of the disk. In order to ensure absolute consistency, the new binary files were copied directly from an actual disk of the correct size created (independently of mayastor) expressly for this purpose. I did recreate the exact label GUIDs in order to minimise any necessary changes. The differences (other than the checksums) from the original are:
- the partition size in the MBR now correctly reflects the true size of the disk
- the LBA of the last usable block in the GPT now correctly reflects the true size of the disk 
- the LBA of the secondary GPT header is now correct

mayastor/src/bdev/nexus/nexus_child.rs
gila: Hi, @cjones1024 It been a while since I've read the spec,  Based on your explanation what it would amount to is that if the first 512 bytes are toast, we simply discard the whole partitioning scheme. This seems rather harsh. 

Also, we know that we only care about GPT and so we don't do anything with the PMBR whenever its good or bad. All it would say in our case is that it contains one partition that spans the whole disk which is lie -- and we lie about it because the MBR code knows nothing about GPT and we mostly use it to shut up boot code anyway. 

The only reason I added it the first place was because gptdisk said that the PMBR was invalid/missing but that was all it really did: noise. 

Anyways, if the spec says its invalid, let us follow the spec then, but please add a comment for future reference, thanks.

gila:
> > I can't comment on binary files but why did you change them?
> 
> The original binary file contained the MBR, primary GPT header and the partition table from the beginning of the disk. I have now added a second binary file containing the secondary partition table and GPT header from the end of the disk. In order to ensure absolute consistency, the new binary files were copied directly from an actual disk of the correct size created (independently of mayastor) expressly for this purpose. I did recreate the exact label GUIDs in order to minimise any necessary changes. The differences (other than the checksums) from the original are:
> 
> * the partition size in the MBR now correctly reflects the true size of the disk
> * the LBA of the last usable block in the GPT now correctly reflects the true size of the disk
> * the LBA of the secondary GPT header is now correct

Cool thanks!

mayastor/src/bdev/nexus/nexus_child.rs
cjones1024: Maybe a compromise is to just issue a warning but not consider this to be a fatal error?

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -252,79 +259,221 @@ impl NexusChild {
...
+                        (bdev.num_blocks() - 1) * u64::from(block_size),
+                        &mut buf,
+                    )
+                    .await
+                {
+                    Ok(secondary) => {
+                        header.push(secondary);
+                        self.check_consistency(&primary, &secondary)?;
+                    }
+                    Err(_) => {
```
tiagolobocastro, line 305: should we capture the error and log it as part of the warn!?

tiagolobocastro, line 305: should we capture the error and log it as part of the warn!?

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -252,79 +259,221 @@ impl NexusChild {
...
+                        warn!(
+                            "{}: {}: The secondary GPT header is invalid!",
+                            self.parent, self.name
+                        );
+                        // Recreate secondary from primary.
+                        header.push(primary.to_backup());
+                    }
+                }
+            }
+            Err(_) => {
```
tiagolobocastro, line 317: same as above (plus a few more below)

mayastor/src/bdev/nexus/nexus_label.rs
```diff
@@ -126,113 +164,110 @@ impl Nexus {
...
-        serialize_into(&mut writer, &pmbr).context(SerializeError {})?;
-
-        for child in &mut self.children {
-            child.write_at(0, &buf).await.context(WriteError {})?;
+        serialize_into(&mut writer, &label.mbr).context(SerializeError {})?;
+
+        // Primary GPT header
+        writer
+            .seek(SeekFrom::Start(label.primary.lba_self * block_size))
+            .unwrap();
```
tiagolobocastro, line 210: should these unwraps be handled?

mayastor/src/bdev/nexus/nexus_child.rs
cjones1024: done

mayastor/src/bdev/nexus/nexus_child.rs
cjones1024: done

mayastor/src/bdev/nexus/nexus_label.rs
cjones1024: This is how the code was originally so I did not change it. In this case we are writing (and seeking) to an in-memory buffer of known size, and so I guess it is not unreasonable to assume that they won't ever fail (if they do you have got bigger problems anyway).

cjones1024:
> 
> 
> I didn't look at the existing test in detail, but it might be interesting to add some more test cases to test the various error cases (could be done at a later PR)

Agreed, should at more tests, but can do that as a different Jira

gila:
bors try

mayastor/src/bdev/nexus/nexus_child.rs
cjones1024: Done.
We now recreate the primary from the secondary if possible. We also do the converse as well - if the secondary is not valid then we recreate it from the primary.

mayastor/src/bdev/nexus/nexus_label.rs
```diff
@@ -567,8 +642,8 @@ pub struct GptName {
 ///
 /// The struct should have a 440 byte code section here as well, this is
 /// omitted to make serialisation a bit easier.
-#[derive(Serialize, Deserialize)]
-struct Pmbr {
+#[derive(Copy, Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub struct Pmbr {
```
gila, line 646: Why do these things need to become pub?

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -253,79 +260,229 @@ impl NexusChild {
...
-            Err(_) => return Err(ChildError::LabelInvalid {}),
-        };
-
-        // determine number of blocks we need to read from the partition table
-        let num_blocks =
-            ((label.entry_size * label.num_entries) / block_size) + 1;
+        if mbr.entries[0].num_sectors != 0xffff_ffff
+            && mbr.entries[0].num_sectors as u64 != header[0].lba_alt
+        {
+            warn!("{}: {}: The protective MBR disk size does not match the GPT disk size!", self.parent, self.name);
```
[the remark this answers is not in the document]
cjones1024: The num_sectors field is a u32. The 2TB max is only possible with a 4K sector/block size
If the size of the disk in blocks does not fit into 32 bits, we are required to set the value to 0xffff_ffff. (See line 122)

mayastor/src/bdev/nexus/nexus_label.rs
```diff
@@ -126,113 +164,110 @@ impl Nexus {
...
-        pmbr.entries[0].ent_type = 0xee;
-        pmbr.entries[0].num_sectors = if size < u32::max_value().into() {
-            size as u32
-        } else {
-            u32::max_value()
-        };
-
-        pmbr.signature = [0x55, 0xaa];
-
+    fn get_primary_data(
```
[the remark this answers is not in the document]
cjones1024: This creates a raw buffer containing the data that needs to be written to each disk to update the "label". It is created once and then written to each disk. will add a descriptive comment.

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -253,79 +260,229 @@ impl NexusChild {
...
+        }
+        Ok(())
+    }
 
-        let nl = NexusLabel {
-            primary: label,
-            partitions: parts,
-        };
+    /// write both (primary and secondary) labels to this child
+    pub async fn write_labels(
```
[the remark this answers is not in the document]
cjones1024: There is only going to be ever either one or two. I thought that allowing an arbitrary slice was unnecessary. Having write_labels() in addition to write_label() makes the caller's intention clearer.
But can dispense with it if you feel its unnecessary.

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -253,79 +260,229 @@ impl NexusChild {
...
+            primary: header[0],
+            partitions: entries,
+            secondary: header[1],
+        })
+    }
 
-        self.read_at(label.lba_table * u64::from(block_size), &mut buf)
-            .await
-            .context(PartitionTableRead {})?;
+    fn get_size(size: u32, block_size: u32) -> u32 {
```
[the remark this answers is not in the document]
cjones1024: Will do

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -7,7 +7,13 @@ use snafu::{ResultExt, Snafu};
 use spdk_sys::{spdk_bdev_module_release_bdev, spdk_io_channel};
```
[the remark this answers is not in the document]
cjones1024: will look into this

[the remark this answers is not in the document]
cjones1024: All "helper" functions have now been moved to nexus_label.rs

mayastor/src/bdev/nexus/nexus_child.rs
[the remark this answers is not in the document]
cjones1024: function has been renamed to get_aligned_size()

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -253,79 +260,229 @@ impl NexusChild {
...
+        }
+    }
+
+    async fn read_header(
+        &self,
+        offset: u64,
+        buf: &mut DmaBuf,
+    ) -> Result<GPTHeader, ChildError> {
+        self.read_at(offset, buf).await.context(LabelRead {})?;
+        match GPTHeader::from_slice(buf.as_slice()) {
```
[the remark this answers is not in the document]
cjones1024: This has been fixed as part of the refactoring to move things to nexus_label.rs

mayastor/src/bdev/nexus/nexus_child.rs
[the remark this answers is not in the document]
cjones1024: Fter moving things to nexus_label.rs have dispensed with these functions entirely

mayastor/src/bdev/nexus/nexus_label.rs
cjones1024: Pmbr needs to be pub as it is now part of the NexusLabel struct so that it can be read and checked for consistency with the GPT header. However all fields that don't *need* to be public at present have been made private.

mayastor/src/bdev/nexus/nexus_child.rs
cjones1024: no longer applicable

PR #147, opened 2020-03-26
Rename pool state enums to avoid conflicts with other state enums

jkryl:
This is a preparation commit for converting string states of nexus
and nexus child to enum states (in grpc proto file). Other changes:

* "faulty" was changed to more correct "faulted"
* improved mapping between enum names and integer values in JS code


jkryl:
For reference I'm including the planned changes to the proto file that will eventually follow this commit:
```
diff --git a/rpc/proto/mayastor.proto b/rpc/proto/mayastor.proto
index cb06a67..f47de76 100644
--- a/rpc/proto/mayastor.proto
+++ b/rpc/proto/mayastor.proto
@@ -140,17 +140,31 @@ message CreateNexusRequest {
   repeated string children = 3; // uris to the targets we connect to
 }

+// State of the nexus child.
+enum ChildState {
+  CHILD_ONLINE = 0;   // healthy and contains the latest bits
+  CHILD_REBUILD = 1;  // rebuild of the child is in progress
+  CHILD_FAULTED = 2;  // unrecoverable error
+}
+
 // represents a child device part of a nexus
 message Child {
   string uri = 1;   // uri of the child device
-  string state = 2; // TODO: enum
+  ChildState state = 2; // state of the child
+}
+
+// State of the nexus (terminology inspired by ZFS).
+enum NexusState {
+  NEXUS_ONLINE = 0;   // healthy and working
+  NEXUS_DEGRADED = 1; // not healthy but is able to serve IO (i.e. rebuild is in progress)
+  NEXUS_FAULTED = 2;   // broken and unable to serve IO
 }

 // represents a nexus device
 message Nexus {
   string uuid = 1;             // name of the nexus
   uint64 size = 2;             // size of the volume in bytes
-  string state = 3;            // current state of the nexus (TODO: enum)
+  NexusState state = 3;        // current state of the nexus
   repeated Child children = 4; // array of children
   // Path to device file for the volume (missing if not published).
   // Missing property and empty string are treated the same.
```

jkryl:
Blaise, thanks for reviewing the changes 👍

PR #148, opened 2020-03-27
CAS-176 disconnect from an NVMe remote target

AntonioCarlini:
Add a disconnect function that finds any nvme device that exactly matches the specified name and disconnects all the ones it finds.


AntonioCarlini:
OK, here's a re-write ... there might be a merge or two in there but I'll squash everything when this gets pushed.

AntonioCarlini:
Not sure that I rebased as clearly as I'd thought ... but I'm pretty sure I've pushed the re-written code!

tjoshum:
Looks like the code is called from anywhere. Is it worth adding some unit tests or something?

AntonioCarlini:
I do have a test program which calls the library and runs on a real client VM against a real server VM.

I could add a call to disconnect in the test code, although it won't do anything (like the call to connect).

I'll do that as soon as my VM is configured to build (rather than just talk to the office machine).

nvmeadm/src/nvmf_discovery.rs
```diff
@@ -403,3 +405,23 @@ impl DiscoveryLogEntry {
...
+    let _subsys: Result<Vec<Subsystem>, Error> = NvmeSubsystems::new()?
+        .filter_map(Result::ok)
+        .filter(|e| e.nqn == nqn)
+        .map(|e| {
+            disconnected_devices += 1;
+            e.disconnect()?;
+            Ok(e)
+        })
+        .collect();
+    Ok(disconnected_devices)
```
jonathan-teh, line 426: I think you can return subsys.unwrap().len() instead of maintaining an additional counter (and the return type has to change to usize).

AntonioCarlini:
Update and added a test. Closed/Opened to kick off a build.

PR #150, opened 2020-03-30
Facilitate specification of the nexus protocol using CSI.

blaisedias:
This change does not implement sharing of the nexus,
but lays the foundation for that to happen.
Supported nexus protocols are
        nbd
        nvmf
        iscsi

### DIFF
--- a/csi/moac/test/csi_test.js
+++ b/csi/moac/test/csi_test.js
@@ -214,6 +214,31 @@ module.exports = function() {
         }
       });
 
+      it('should create a volume and return parameters in volume context', async () => {
+        createVolumeStub.resolves(returnedVolume);
+        var parameters = { protocol: 'nbd', repl: 3, blah: 'again' };
+        let result = await client.createVolume().sendMessage({
+          name: 'pvc-' + UUID,
+          capacityRange: {
+            requiredBytes: 10,
+            limitBytes: 20,
+          },
+          volumeCapabilities: [
+            {
+              accessMode: { mode: 'SINGLE_NODE_WRITER' },
+              block: {},
+            },
+          ],
+          parameters: parameters,
+        });
+        // volume context is a of type map<string><string>
+        var expected = {};
+        for (const key in parameters) {
+          expected[key] = parameters[key].toString();
+        }
+        sinon.assert.match(result.volume.volumeContext, expected);
+      });
+
       it('should fail if topology requirement other than hostname', async () => {
         createVolumeStub.resolves(returnedVolume);
         await shouldFailWith(GrpcCode.INVALID_ARGUMENT, () =>
@@ -233,6 +258,7 @@ module.exports = function() {
               requisite: [{ segments: { rack: 'some-rack-info' } }],
               preferred: [],
             },
+            parameters: { protocol: 'nbd' },
           })
         );
       });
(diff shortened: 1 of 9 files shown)


csi/moac/test/csi_test.js
```diff
@@ -392,7 +424,7 @@ module.exports = function() {
               block: {},
             },
           ],
-          parameters: { repl: '3' },
+          parameters: { repl: '3', protocol: 'nexus_nbd' },
```
jkryl: it does not matter too much as the createVolume method just passes the string from input to output value, though it might be better to pass a correct value for the parameter: `nbd`. The same comment applies to line 453.

csi/moac/test/csi_test.js
```diff
@@ -392,7 +424,7 @@ module.exports = function() {
...
         });
         sinon.assert.calledWith(createVolumeStub, UUID, {
```
jkryl: can we add a check for return value of createVolume CSI method now when it actually matters? It should return `protocol = nbd` in the output params.

csi/moac/test/nexus_test.js
```diff
@@ -184,10 +184,61 @@ module.exports = function() {
       callStub.reset();
     });
 
-    it('should publish the nexus', async () => {
+    it('should not publish the nexus with whatever protocol', async () => {
+      callStub.resolves({ devicePath: '/dev/whatever0' });
+      callStub.rejects(
+        new GrpcError(GrpcCode.INVALID_ARGUMENT, 'Test failure')
```
jkryl: Can we change the error code to something else (i.e. NOT_FOUND or whatever) so that we can make the distinction between the case when grpc call fails and when the argument check fails more apparent?

csi/moac/test/csi_test.js
blaisedias: I think this is done in the test case at line 217 "It should create a volume and return parameters in volume context"

blaisedias:
> ... Also we should test the case of calling csi publish volume with missing protocol parameter. What should be the result in that case? Do we want to check this case and return an error to client or do we want to supply some kind of default value to call nexus publish with? Returning an error seems safer to me (we know that something is not right as opposed to the bug being undiscovered and making harm later when it's more difficult to find out why it happened).

It should fail and it would fail in the e2e case but not with the mock testing. I've fixed the csi code so that it is explicitly tests for the presence and not the value, and if absent throws an INVALID_ARGUMENT exception - in keeping with other bits of code in controllerPublishVolume.

blaisedias:
>I'm missing a change to publish volume CSI method tests. I think it should get parameters from the createVolume with the protocol set to a value and we should check that the nexus publish method was called with the right argument then.

Unless I am mistake, I think this is checked for by the assert I added to the test case 
`      it('should publish volume', async () => {`
line 616 in the diffs
`        sinon.assert.calledWith(publishStub, 'nbd');`

Since we are not testing within a CSI framework, a call to createVolume will not result in a call to controllerPublishVolume (which invokes volume.Publish which invokes nexus.Publish).

csi/moac/test/csi_test.js
blaisedias: Done. https://github.com/openebs/Mayastor/pull/150/commits/2c509c7bc5230fda308031b496a234ad631b66ad

PR #151, opened 2020-03-30
Allow nexus configuration from files

gila:
To make life a bit easier, in particular during testing, we want to ability to load or bootstrap mayastor with a given, specific config file.

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_bdev_children.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev_children.rs
@@ -68,9 +68,9 @@ impl Nexus {
             .for_each(drop);
     }
 
-    /// register a single child to nexus, only allowed during the nexus init
-    /// phase
-    pub async fn register_child(
+    /// Create and register a single child to nexus, only allowed during the
+    /// nexus init phase
+    pub async fn create_and_register(
         &mut self,
         uri: &str,
     ) -> Result<(), BdevCreateDestroy> {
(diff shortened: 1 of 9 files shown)


mayastor/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -440,16 +440,16 @@ impl Nexus {
...
     pub fn examine_child(&mut self, name: &str) -> bool {
-        for mut c in &mut self.children {
-            if c.name == name && c.state == ChildState::Init {
+        self.children
+            .iter_mut()
+            .filter(|c| c.state == ChildState::Init && c.name == name)
+            .any(|c| {
                 if let Some(bdev) = Bdev::lookup_by_name(name) {
-                    debug!("{}: Adding child {}", self.name, name);
                     c.bdev = Some(bdev);
```
chriswldenyer: Was removing the debug message intentional?

gila: I considered the message to be redundant. We have a message that says that a child came online in nexus_module.rs:126. If the nexus is able to open, it will show in the logs, and if it can't a message that its incomplete is added. This is only visible, however, when we debug messages are enabled.

mayastor/src/bdev/nexus/nexus_module.rs
```diff
@@ -145,6 +151,41 @@ impl NexusModule {
...
+            let uris = nexus
+                .children
+                .iter()
+                .map(|c| c.name.clone())
+                .collect::<Vec<String>>();
+
+            let json = json!({
+                "method": "create_nexus",
+                "params": {
+                    "uuid" : nexus.name.as_str()[6 ..],
```
chriswldenyer: Maybe should explain the 6 chars here "nexus-" ?. Could it ever change? 
I thought "crypto-" can get prefixed onto the name as well?

mayastor/src/bdev/nexus/nexus_bdev_children.rs
chriswldenyer: Ok,

chriswldenyer:
I left two comments, one resolved.
I don't know where/how dump_info_json() and config_json() get called as I don't understand the architecture well enough, but otherwise it looks reasonable AFAIUI.

mayastor/src/bdev/nexus/nexus_module.rs
gila: Good question.  But the one level "up" so the crypto instance is stacked ontop of it.

PR #160, opened 2020-04-03
CI: enable github actions

gila:


jkryl:
nice Jeffry! There are some issues that you have probably already noticed:

* all github action tests run twice because you run them for each PR and each push.
* checksum of mayastor dependencies does not match (see gitlab result). It seems this PR did not break it, but regardless must be fixed.
* according to what I see in actions you always checkout master branch - does it really test PR?

gila:
This just mostly allows us to run Clippy actions that do not allow for specifying your own command(s). I personally think though that's a limitation of the actual plugins. 

However, I've yet to see how much this will bring us (the auto-comments added to the PR)  from the history of the action, it does not seem to very active it the moment.

Also, as far as I can tell, a rustfmt plugin like this does not exist, so I'm not sure if we want to use these actions or use different ones scattered around the net. Maybe we should use the ones from rust directly. They are, perhaps less fancy (well the Clippy one at least and also the only one AFAIK). 

Also, we should also add the logic we want to have run in a bash script (or whatever) and not it the YAML file as they are specific to the CI vendor in use. 

All in all, IMHO we can merge this for now, as it does not hurt anything, but I'm working on more fundamental changes that won't require nix-shelling at all.

gila:
> nice Jeffry! There are some issues that you have probably already noticed:
> 
> * all github action tests run twice because you run them for each PR and each push.

Yeah not sure how to filter that out more accurately, it is possible, but -- as you fork you might want to run things "as you push". 

> * checksum of mayastor dependencies does not match (see gitlab result). It seems this PR did not break it, but regardless must be fixed.

I'm not sure what's going on there, or how we even got it merged in the first place if that's the case but yeah we should fix that for sure. 
> * according to what I see in actions you always checkout master branch - does it really test PR?

Yes. the @master refers to the master branch of the checkout action itself, not our the code :)

PR #162, opened 2020-04-06
rebuild: run n copy tasks at a time (async tasks)

tiagolobocastro:
a management async tasks schedules all copy tasks, awaits them and schedules enough tasks until the bdev is fully rebuilt
(still using a single reactor this time)

### DIFF
--- a/mayastor/src/rebuild.rs
+++ b/mayastor/src/rebuild.rs
@@ -5,6 +5,8 @@ use snafu::{ResultExt, Snafu};
 use spdk_sys::spdk_get_thread;
 use std::{cell::UnsafeCell, collections::HashMap, fmt};
 
+use futures::{channel::mpsc, StreamExt};
+
 pub struct RebuildInstances {
     inner: UnsafeCell<HashMap<String, RebuildJob>>,
 }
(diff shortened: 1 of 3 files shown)


mayastor/src/rebuild.rs
```diff
@@ -60,7 +62,24 @@ impl fmt::Display for RebuildState {
     }
 }
 
-#[derive(Debug)]
+struct TaskResult {
```
gila, line 65: please add comments to each of the fields as we do (mostly). Using 3 slashes creates a doc comment i.e

```rust
/// this struct is used for ... 
struct  foo {
/// the blk (??) being rebuilt
blk: u64
/// the  id that is used for...??
id: 64
}
```

mayastor/src/rebuild.rs
```diff
@@ -60,7 +62,24 @@ impl fmt::Display for RebuildState {
...
+    blk: u64,
+    id: u64,
+    error: Option<RebuildError>,
+}
+
+const SEGMENT_TASKS: u64 = 4;
+const SEGMENT_SIZE: u64 = 10 * 1024; // 10KiB
+
+struct RebuildTasks {
+    buffers: Vec<DmaBuf>,
```
gila, line 75: Likewise, please try to comment all structs and members, and functions.. it really pays of to do this consistently. you can in fact even tell the compiler to help you by setting that you do not allow missing docs. 

iirc its `#[deny(missing_docs)]`

mayastor/src/rebuild.rs
```diff
@@ -5,6 +5,8 @@ use snafu::{ResultExt, Snafu};
 use spdk_sys::spdk_get_thread;
 use std::{cell::UnsafeCell, collections::HashMap, fmt};
 
+use futures::{channel::mpsc, StreamExt};
```
gila, line 8: Please consider using the crossbeam channels for this they are considered to be far more efficient at various aspects.

mayastor/src/rebuild.rs
tiagolobocastro: yep I should have and will do indeed. I'll add that attribute, that'll be quite handy :+1:

mayastor/src/rebuild.rs
tiagolobocastro: I actually started with the crossbeam which we currently use for something already but it didn't seem to be using futures so ended up looking up async-std and finally this future:channel:mpsc. I'll have another look at the crossbeam ones.

tiagolobocastro:
> Some minor nits, can be addressed in follow up PRs as well. +1

Thanks, I'll address those as a follow PR

PR #165, opened 2020-04-06
Pin

gila:


nix/pkgs/libspdk/default.nix
```diff
@@ -42,70 +40,84 @@ stdenv.mkDerivation rec {
...
   ] ++ stdenv.lib.optionals enableDebug [ cunit lcov ];
 
-  CONFIGURE_OPTS = ''
-    ${enableFeature enableDebug "debug"}
-    --without-isal --with-iscsi-initiator
-    --with-internal-vhost-lib --disable-tests --with-dpdk-machine=native
-    --with-crypto
-    --with-uring
-  '';
+  #${enableFeature enableDebug "unit-tests"}
```
jkryl, line 43: do we need this comment?

nix/mayastor-overlay.nix
```diff
@@ -1,29 +1,20 @@
...
+  mayastorImage = (super.callPackage ./pkgs/mayastor { }).mayastorImage;
+  mayastorCSIImage = (super.callPackage ./pkgs/mayastor { }).mayastorCSIImage;
+  ms-buildenv = super.callPackage ./pkgs/ms-buildenv { };
+  mkContainerEnv = super.callPackage ./lib/mkContainerEnv.nix { };
   node-moac = (import ./../csi/moac { pkgs = super; }).package;
   node-moacImage = (import ./../csi/moac { pkgs = super; }).buildImage;
+
+  # the upstream package in nix adds phantomjs we really dont want that
+  # as that pulls in QTWeb, and as a consequence, X libraries
+  jshint = super.nodePackages.jshint.override (o: rec { buildInputs = [ ]; });
```
jkryl, line 19: nice catch!

nix/pkgs/ms-buildenv/default.nix
```diff
@@ -0,0 +1,201 @@
...
+        "PAGER=less"
+        "PATH=/nix/var/nix/profiles/default/bin"
+
+        # in order for tools that download stuff to work we need certs
+        "SSL_CERT_FILE=/nix/var/nix/profiles/default/etc/ssl/certs/ca-bundle.crt"
+        "GIT_SSL_CAINFO=/nix/var/nix/profiles/default/etc/ssl/certs/ca-bundle.crt"
+        "RUST_BACKTRACE=1"
+      ];
+      Labels = {
+        # https://github.com/microscaling/microscaling/blob/55a2d7b91ce7513e07f8b1fd91bbed8df59aed5a/Dockerfile#L22-L33
```
jamie-0, line 192: whats this, why is it a comment?

nix/pkgs/ms-buildenv/default.nix
```diff
@@ -0,0 +1,201 @@
...
+        "org.label-schema.vcs-ref" = "master";
+        "org.label-schema.vcs-url" = "https://github.com/gila/mayastor";
```
jkryl, line 194: I followed the link and after reading the docker file there it's not more clear to me what these definitions are good for. And I think certainly there should not be a URL to your private mayastor repo. Maybe we can just remove those definitions?

nix/pkgs/ms-buildenv/default.nix
```diff
@@ -0,0 +1,201 @@
...
+, libiscsi
+, mkContainerEnv
+, libaio
+, binutils-unwrapped
+, nodejs
+, protobuf
+, llvmPackages
+, gzip
+, iproute
+, # vscode
```
jkryl, line 52: I think this comment should be removed. It is unlikely that anyone would like to run vscode editor inside the container. Instead it's just going to be used for remote executions.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - uses: actions-rs/clippy-check@v1.0.5
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          args: --all --all-targets -- -D warnings
+  nix-build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: cachix/install-nix-action@v7
+      - run: nix-build -A mayastorImage
```
jamie-0, line 37: Aren't we building this on line 10-18 above?

nix/pkgs/libspdk/default.nix
gila: For the upcoming releases the have split out some of the tests in 'test' and 'unit-test' that's not in our tree yet, I did try it out so I left it here as comment so that I don't forget to add it back later, I will make the comment more explicit 👍

.github/workflows/build.yaml
gila: This builds a docker image, the lines above build the code in 'debug' mode. So they are strong related and we can use this approach above as well to create "debug images" but - we we can not run the checks this way. (unless we hack the actions themselves).

nix/pkgs/ms-buildenv/default.nix
gila: To provide context as to why those labels are there but -- i'll remove them for now.

nix/pkgs/ms-buildenv/default.nix
gila: I'll move the references now but -- AFAIK there are two "modes" and if I where to use this stuff, i would run vscode *in* the container, and not SSH (i think...)

PR #174, opened 2020-04-08
nvmf: target instantiation is redundant

gila:
A target is always created. Depending on config file we should
only create a transport and add it to the existing target.

### DIFF
--- a/mayastor/src/target/nvmf.rs
+++ b/mayastor/src/target/nvmf.rs
@@ -30,21 +30,17 @@ use spdk_sys::{
     spdk_nvmf_subsystem_add_ns,
     spdk_nvmf_subsystem_create,
     spdk_nvmf_subsystem_destroy,
-    spdk_nvmf_subsystem_get_first,
-    spdk_nvmf_subsystem_get_next,
     spdk_nvmf_subsystem_get_nqn,
     spdk_nvmf_subsystem_set_allow_any_host,
     spdk_nvmf_subsystem_set_mn,
     spdk_nvmf_subsystem_set_sn,
     spdk_nvmf_subsystem_start,
     spdk_nvmf_subsystem_stop,
-    spdk_nvmf_target_opts,
     spdk_nvmf_tgt,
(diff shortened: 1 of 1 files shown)


jkryl:
I don't understand why do we remove something that works and gives us flexibility to do things like having a target per replica/volume in the future. Even SPDK folks don't seem to be happy with the default target and want to make it more flexible in future reworking jsonrpc methods around this. Building on assumption of the default target in SPDK can bring us problems in the future when it is changed. Staying independent is more reasonable from this point of view.

gila:
There are several reasons for that:

- we create two targets one implicit that always starts (which we use now)
and one we create ourselves. Both of them allocate a mempool where one Is
entirely unused, which is just a waste of huge pages.

- the one we created has no means of setting any options like retries,
timeout and the likes we really need to be able to control that and right
now there is no way of doing that due to the way the code bootstraps.

- also in the same domain we can not preload the systems with any bdevs
which makes testing anything outside of BDD tests rather hard.

- the current code does not handle any error conditions and simply creates
a new transport even if it is there on top of the target it also -- already
created. These changes handle that case now (by looking up the transport)
and simply return if the transport exists.

I don't understand why do we remove something that works and gives us
> flexibility to do things like having a target per replica/volume in the
> future.
>

I can not think of any situation where we want would want that, do you have
an example of how that would look like? Are you not confusing this with
listeners? AFAIK all exports should go over the same target regardless if
they are for the nexus or replica. The selection of what goes where is
dictated by the address and port, not by the target.

Even SPDK folks don't seem to be happy with the default target and want to
> make it more flexible in future reworking jsonrpc methods around this.
> Building on assumption of the default target in SPDK can bring us problems
> in the future when it is changed. Staying independent is more reasonable
> from this point of view
>
So do I -- but there current implementation makes it less flexible and all
it does really, is remove one thing: a target. We can always add it back if
they ever get around to it which I don't think will be very soon. I would
think it's more likely we stop linking in the .a’s that handle all of this
 implicit  creation and implement something that matches what we need.

PR #176, opened 2020-04-09
bors: update configuration

gila:
- we require 2 reviewers
 - we do not want merge commits
 - clippy in a PR must always pass before we stage it

Not 100% sure yet if this is is final but I think it is.
!!!

Below here things are ignored CAS-1


gila:
bors try

PR #177, opened 2020-04-09
nvmf: target instantiation is redundant

gila:
A target is always created. Depending on config file we should
only create a transport and add it to the existing target.

Ideally this implicit creation is not there at all but its not trivial
to remove that now without going over a proper design phase that
reimplements configuration.


gila:
We need a much better way, closing this to implement subsystems and dependencies

PR #178, opened 2020-04-13
build changes

gila:
This little easter egg pops up some changes to the build system, the changes themselves are not all that involving except for that fact that this actually moved us to GitHub for 100%. 

Most noticeable changes are:

- The buildenv has been updated to add the pkgs needed during tests (i.e mkfs)
-  jshint and prettier have been added but -- are not enabled for now (commented out) they work, however, @jkryl has some changes to the linter and formatter inflight and so I did not want to change more JS code than needed to avoid to the churn
- I've disabled the cargo lock validation for the dependencies. I'm not sure if we need it at all (as it also hashes the dependencies) 
- enable (at long last) cachix to cache libspdk and libiscsi (although only libspdk is built all dependencies are pushed)
- the whole build is now debug enabled by default, this includes the images for releases we then have to disable debug
- change/add workflows for:
  - on PR run linters for  Rust and JS code
  - on bors +r -- run build.yaml (which is the same cycle as gitlab) bors will not run unless PR stage has succeeded
  - when a PR changes any .nix file, we run nix.yaml

@jamie-0 it is all yours now! 🎁


gila:
bors +r

PR #184, opened 2020-04-15
ci: minor fixes

gila:
I've ran this several times on my fork and I hope that
with the two cpu args it will be a bit more stable.


gila:
> how many cpus have we got on github?

2

tiagolobocastro:
> > how many cpus have we got on github?
> 
> 2

so we're already on the max then :(

PR #185, opened 2020-04-16
build: ensure precommit hooks pass

tjoshum:
The toolchain is now pinned to a known version (since #165)
so shouldn't change without a further commit. This commit brings the
style of all files in the repository up to date with the current
tooling. I've done this in order to avoid changes in future commits
which do not relate to the purpose of those commits.

I've also made the javascript linting around parans less strict for now,
to avoid spurious changes the next time we upgrade prettier.


blaisedias:
Looks okay to me.

PR #186, opened 2020-04-16
build: fix prettier linting and make hooks read-only

tjoshum:
Correct a spelling mistake in .prettierrc, and commit the changes
now necessary to have the precommit hooks pass against all of our
files.

I've also altered the precommit hooks so that they no longer alter
code in the repository when they run. This should minimise the effort
required to repair any damage from teething problems with these tools.


tjoshum:
Changes in .pre-commit-config.yaml and .prettierrc are mine, everything else was done by the linters.

jonathan-teh:
Fixes #185.

PR #189, opened 2020-04-16
CAS-16: New grpc interface for Mayastor

yannis218:
Mayastor now implements its own gRPC interface,
so that calls no longer have to go via the agent.

### DIFF
--- a/csi/src/client.rs
+++ b/csi/src/client.rs
@@ -7,13 +7,29 @@
 #[macro_use]
 extern crate clap;
 
+use byte_unit::Byte;
 use bytesize::ByteSize;
 use clap::{App, AppSettings, Arg, ArgMatches, SubCommand};
-
+use rpc::mayastor::{
+    CreateNexusRequest,
+    DestroyNexusRequest,
+    Null,
+    PublishNexusRequest,
+};
 use tonic::{transport::Channel, Code, Request, Status};
 
 use rpc::service::mayastor_client::MayastorClient;
 
+type MayaClient = MayastorClient<Channel>;
+
+/// parses a human string into bytes accounts for MiB and MB
+pub(crate) fn parse_size(src: &str) -> Result<u64, String> {
+    if let Ok(val) = Byte::from_str(src) {
+        Ok(val.get_bytes() as u64)
+    } else {
+        Err(src.to_string())
+    }
+}
 fn parse_share_protocol(pcol: Option<&str>) -> Result<i32, Status> {
     match pcol {
         None => Ok(rpc::mayastor::ShareProtocolReplica::ReplicaNone as i32),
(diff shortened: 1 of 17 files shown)


mayastor-test/test_csi.js
```diff
@@ -111,7 +112,7 @@ function getFsType(mp) {
 }
 
 describe('csi', function() {
-  this.timeout(10000); // for network tests we need long timeouts
+  this.timeout(60000); // for network tests we need long timeouts
```
jkryl, line 115: 60s is a bit too much. Why did you change this?

mayastor-test/test_rebuild.js
```diff
@@ -141,6 +141,8 @@ describe('rebuild tests', function() {
     DESTINATION_CHILD: 2,
   };
 
+  this.timeout(60000); // for network tests we need long timeouts
+
```
jkryl, line 145: the same question here

mayastor-test/test_replica.js
```diff
@@ -60,6 +60,7 @@ function destroyTestDisk(done) {
 }
 
 function createGrpcClient(service) {
+  console.log('Starting client at ', endpoint);
```
jkryl, line 63: I suppose this was used for debugging and can be removed now

csi/src/server.rs
```diff
@@ -197,6 +197,7 @@ async fn main() -> Result<(), String> {
     builder.init();
 
     let saddr = format!("{}:{}", addr, port).parse().unwrap();
+    println!("Agent starting service on {}", saddr);
```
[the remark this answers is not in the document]
yannis218: Why not?

mayastor/src/core/env.rs
```diff
@@ -110,6 +119,8 @@ pub struct MayastorCliArgs {
 impl Default for MayastorCliArgs {
     fn default() -> Self {
         Self {
+            addr: "None".into(),
```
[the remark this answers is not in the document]
yannis218: It is not an Option; it is a String, for reasons already discussed.  The specific value does not matter, so if you do not like "None", fine, but I do not think that "" would be a good idea, again as already discussed.

mayastor/src/core/env.rs
```diff
@@ -169,6 +180,9 @@ type Result<T, E = EnvError> = std::result::Result<T, E>;
 #[derive(Debug, Clone)]
 pub struct MayastorEnvironment {
     pub config: Option<String>,
+    pub enable_grpc: bool,
+    grpc_addr: String,
```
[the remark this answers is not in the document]
yannis218: Already an option - you were not looking at the latest diff.

mayastor-test/test_cli.js
```diff
@@ -54,6 +54,8 @@ function runMockServer(rules) {
 }
 
 describe('cli', function() {
+  this.slow(300);
+
```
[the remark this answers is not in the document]
yannis218: Not necessary at all - it only configures the colouring in the console output.  Too many of the tests were being tagged with red for "being slow" (at least on my loaded system) that it became difficult to notice actual test failures.

mayastor-test/test_csi.js
yannis218: I am happy to try smaller values, but 10sec on my system fails the test with timeout.  Given that this is basically to guard against tests that hang, what is the harm in using 1min?  If you envisage a test case that has to take less than a particular amount of time to be considered a pass, then that can surely be asserted explicitly by that test.  This is better that having an entire test suite that relies on a performant system.

mayastor-test/test_rebuild.js
yannis218: See above.

mayastor-test/test_replica.js
yannis218: I took out a lot of my debugging prints but not this, as I thought it was not unreasonable for the test to announce the socket for the interface the test suite was about to test.

csi/src/client.rs
```diff
@@ -378,6 +458,63 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                                .index(1),
+                        )
+                        .arg(
+                            Arg::with_name("key")
+                                .help("crypto key to use")
+                                .required(false)
+                                .index(2),
+                        )
+                )
+        )
```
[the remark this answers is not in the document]
yannis218: I do not know - these were Jeffry's changes that I "inherited".

mayastor-test/test_cli.js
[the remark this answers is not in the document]
yannis218: I am removing this.

mayastor-test/test_csi.js
yannis218: I tried reverting this on my new setup (running on the host, no VM) and the tests passed!
(I should have tried doing this in the first place, when I got rid of the VM - it was obvious that everything was running faster)  so I am reverting this.

mayastor-test/test_replica.js
yannis218: Particularly as some of the tests now run twice, with different sockets.

jkryl: In general there should be no output from mocha test suite except the green check marks (or red if it fails). That said sometimes the code we are testing logs messages to stderr - nothing we can do about it. And perhaps there are other few exceptions like printing log output of mayastor process or csi agent process (as there is no way to get this output otherwise). But log messages like "Starting something ... ", "Doing this or that" should be avoided as they clutter the output more than necessary. In the long term we won't run the same tests against two endpoints, but there will be just one endpoint. Then how useful is the information that I'm running the tests against some particular endpoint. But we can remove that later when the obsolete code in csi-agent goes away 👍

csi/src/client.rs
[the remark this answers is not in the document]
jkryl: I seems you inherited the code with its problems too :-) And the problem of this code is that it does not seem to be tested. I'm fine if you just create a ticket addressing the missing tests and tackle it in the next sprint 👍  Or remove this code and push it along with the tests as it should have been done. Whatever you prefer.

csi/src/server.rs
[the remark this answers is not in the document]
jkryl: why not to use log functions? The same comment for the other println you added below.

mayastor-test/test_replica.js
yannis218: I can see the reason you are describing, but when I run the tests I get thousands of lines of output, as it dumps out the std{out,err} of the processes it was running.  If all you got was the grey text with the ticks, I would have hesitated to add this kind of thing.  So OK, I'll remove it again when I get rid of the agent's gRPC and the tests go back to a single socket.

csi/src/server.rs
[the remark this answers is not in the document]
yannis218: Oh, as in change the println! to info!

PR #190, opened 2020-04-16
rebuild: clean up rebuild job on child removal

tiagolobocastro:
rebuild: clean up rebuild job on child removal

when a child is being removed we check for any existing jobs that have:
1. child as dest
in this case, we simply stop the job and wait for it to finish
2. child as src
same as above but we also need to start a new rebuild job to rebuild the
dest child.

Atm the wait infra is not here so we have to rely on the child destroy
to block until all the bdevs are closed.
This will be addressed once cas-194 is resolved.

CAS-126

-------------------------------

rebuild: add missing_docs attribute

various bits of the rebuild have now been documented
any missing doc will now be a compiler warning

CAS-201

---------------------------------

move replica&rebuild into new dir


tiagolobocastro:
@gila the 3 commits are isolated, I hope this is ok otherwise I can split this up

gila:
> @gila the 3 commits are isolated, I hope this is ok otherwise I can split this up

Thanks @tiagolobocastro  LGTM

PR #191, opened 2020-04-16
pool: Return base bdev type in list pools

jonathan-teh:
Return a URI in the disks field where the URI scheme contains the
base bdev type.

Add a unit test to test creation of pools that use uring bdevs,
though skip the test if the kernel lacks support, which requires
changes to uring-support to only test for kernel support if the
pathname argument is not present.

### DIFF
--- a/mayastor/src/pool.rs
+++ b/mayastor/src/pool.rs
@@ -644,7 +644,11 @@ pub(crate) fn list_pools() -> Vec<jsondata::Pool> {
     for pool in PoolsIter::new() {
         pools.push(jsondata::Pool {
             name: pool.get_name().to_owned(),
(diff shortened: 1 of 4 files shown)


mayastor/src/pool.rs
```diff
@@ -644,7 +644,11 @@ pub(crate) fn list_pools() -> Vec<jsondata::Pool> {
     for pool in PoolsIter::new() {
         pools.push(jsondata::Pool {
             name: pool.get_name().to_owned(),
-            disks: vec![pool.get_base_bdev().name()],
+            disks: vec![
+                pool.get_base_bdev().driver()
```
gila, line 648: Oh  nice, I've not even thought of that 👍

PR #198, opened 2020-04-22
Run semistandard with --fix on JS files

jkryl:
The remaining 300 errors that cannot be auto-corrected will be fixed
incrementally as we touch the files in future commits.

NOTE: the semistandard github action test is not going to pass as I modified files with
errors that cannot be autocorrected - that's expected.


jkryl:
@yannis218 : yes

PR #199, opened 2020-04-22
Add github action for testing moac

jkryl:
Resolves: CAS-220

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,6 +1,9 @@
-name: "Testing"
+name: "Mayastor tests"
 on:
   pull_request:
+    paths-ignore:
+      - 'csi/moac/**'
+      - 'doc/**'
 jobs:
   Build:
     name: Build and run tests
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -1,6 +1,9 @@
-name: "Testing"
+name: "Mayastor tests"
```
gila, line 1: I think we also need to update the -- what's the name again? The guards for merge block/unblock. It's not called that but I think you know what I mean😅

jkryl: There we refer to job names that have not changed. So no change is required there 👍

.github/workflows/moac.yaml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  Build:
+    name: Build and run moac tests
+    runs-on: ubuntu-latest
+    container:
+      image: docker.io/mayadata/ms-buildenv:nix
+    steps:
+      - uses: actions/checkout@v2
+      # npm prepare is normally done by npm install but not if run as a root
+      - run: cd csi/moac && npm install && npm run-script prepare
```
jamie-0, line 16: we don't currently run `npm run-script prepare` in the build yaml. Why is it needed here? Should it be also in build.yaml?

jkryl: I assume you mean gitlab.yaml. In build.yaml we didn't build moac at all (only the docker image  using nix). "run-script prepare" is needed because when npm install is run as root (which is the case here), it does not run installation scripts and we have an installation script that makes a copy of grpc proto files in moac's dir. In gitlab.yaml we use a different trick and that is "npm install --unsafe-perm" that runs the scripts even if run as root. So two different tricks to overcome the npm quirk. One cannot say that one is better than the other.

PR #201, opened 2020-04-23
build: add sudo to build image

gila:
We really don't need sudo, however, some of the GH actions require it.
The reason that some of them need/assume that, is that they can be run
on the host directly or within a container.


gila:
> :(
> out of curiosity what does actually need it?
> bonus: if we have sudo now, maybe we don't need to rm the cargo config file?

cachix -- when installed in a container, blindly called sudo, so its mostly just to avoid issues because of that. 

As for the cargo file, we can. OTOH that changes is more work then it solves.

PR #203, opened 2020-04-24
Rebuild: ensure all state transitions are correct and...

tiagolobocastro:
...now we can add the hability to wait for a rebuild state and termination

Commit breakdown:

rebuild: split rebuild module

splits the rebuild module into impl and api
no functional changes
this should help contain the next set of coming changes

-----------------------------------------------------

rebuild: split the rebuild state logic

Fixes the rebuild state by introducing and current and a pending state
the pending state is set by the nexus whenever it wants to perform
an operation on the rebuild. The pending state is reconciled when the
job's runner adjusts to the pending state, eg running -> paused

The states are now fully validated by a state machine. No catch all
"_ => ()", was used so the compiler should complain if a new state is
added.

If a pending state is "pending" then a new operation is rejected until
the previous completes.
Internal operations are allowed to override the pending state

Started to make the trait a first class citizen

bonus: fully implement the child src removal by waiting on the rebuild
job to complete before creating a new job to replace the removed one.
This means we now enable the commented out test
Also fix issues pointed out by semi-standard

CAS-194

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_bdev_rebuild.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev_rebuild.rs
@@ -1,4 +1,4 @@
-use crossbeam::channel::Receiver;
+use futures::channel::oneshot::Receiver;
 use rpc::mayastor::RebuildStateReply;
 use snafu::ResultExt;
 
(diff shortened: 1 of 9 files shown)


mayastor/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -18,13 +18,13 @@ use crate::{
...
     core::Reactors,
-    rebuild::{RebuildJob, RebuildOperations, RebuildState},
+    rebuild::{ClientOperations, RebuildJob, RebuildState},
 };
 
 impl Nexus {
     /// Starts a rebuild job in the background
     pub async fn start_rebuild_rpc(&mut self, name: &str) -> Result<(), Error> {
-        self.start_rebuild(name).await?;
+        let _ = self.start_rebuild(name).await?;
```
gila, line 27: you sure you want to swallow whatever this returns?

tiagolobocastro: I think so because we don't want to return the rust channel on the rpc call

PR #205, opened 2020-04-28
mayastor: add config subsystem framework

gila:
With this new subsystem we get the ability to load our own configuration
file. Additionally, it also allows us to apply the configuration file
that makes more sense for our particular use case (i.e import pools)

Currently, we have support for setting global options for the targets
that normally would be applied using ini files. To not break all
test cases, we keep support for loading .ini files but this change does
remove the support for native JSON loading.

The new config file can be saved on request by invoking a new RPC
method and the config will be saved to the path that was passed in during
startup.

This certainly not final, but more like a foundation on which we can
extent any config options we want to add in the future. Moreover this
file will also allow us to recover from any crashes quickly without
depending exclusively on k8s to detect that we are down.

This implements CAS-207


mayastor/src/subsys/config.rs
```diff
@@ -0,0 +1,362 @@
...
+                }
+
+                let uuid = my_bdev.uuid_as_string();
+                assert_eq!(bdev.uuid.as_ref(), Some(&uuid));
+
+                if !self.implicit_share_base {
+                    continue;
+                }
+
+                // share the bdev over nvmf
```
chriswldenyer, line 267: Not supporting iscsi replicas via yaml, because we'll be nvmf-only long-term? If so, maybe that should be in the pull-request comments.

gila: @chriswldenyer absence of iSCSI does not imply anything in a particular right. As stated in the PR, this is an initial pass at a framework that we can extent easily and better suits our needs. (we can debate if it achieved that of course...)

The reason I did nvmf here is because we can already do iSCSI easily (with the older ini file) but for nvmf this was not yet possible in a way that would allow us to set various settings easily.

chriswldenyer: Ok

.github/workflows/build.yaml
```diff
@@ -8,6 +8,7 @@ jobs:
   Build:
     name: Build and run tests
     runs-on: self-hosted
+    timeout-minutes: 20
```
tiagolobocastro, line 11: this should help with the hangs :+1:

mayastor/src/subsys/opts.rs
```diff
@@ -0,0 +1,458 @@
...
+            max_srq_depth: 0, // RDMA
+            no_srq: false,    // RDMA
+            sock_priority: 0,
+        }
+    }
+}
+
+/// we cannot add derives for YAML to these structs directly, so we need to
+/// copy them. The upside though, is that if the FFI structures change, we will
+/// now about it during compile time.
```
tiagolobocastro, line 155: now -> know

mayastor/src/subsys/opts.rs
```diff
@@ -0,0 +1,458 @@
...
+pub struct NvmeBdevOpts {
+    /// action take on timeout
+    action_on_timeout: u32,
+    /// timeout for each command
+    timeout_us: u64,
+    /// retry count
+    retry_count: u32,
+    /// TODO
+    arbitration_burst: u32,
+    /// max number of low priority cmds a controller my launch at one time
```
tiagolobocastro, line 188: my -> may?

mayastor/src/subsys/opts.rs
```diff
@@ -0,0 +1,458 @@
...
+    /// todo
+    allow_duplicate_isid: bool,
+}
+
+impl Default for IscsiTgtOpts {
+    fn default() -> Self {
+        Self {
+            authfile: "".to_string(),
+            nodebase: "iqn.2019-05.io.openebs".to_string(),
+            timeout: 5,
```
tiagolobocastro, line 379: I wonder, if all children have the same timeout, would we want to make the frontend timeout the (child timeout x number of child io retries) + slack?

mayastor/src/core/env.rs
```diff
@@ -580,12 +603,37 @@ impl MayastorEnvironment {
...
+        // load the config and apply it before any subsystems have started.
+        // there is currently no run time check that enforces this.
+        let cfg = if let Some(yaml) = &self.mayastor_config {
+            info!("loading yo-mama configfile");
+            Config::get_or_init(|| {
+                if let Ok(cfg) = Config::read(yaml) {
+                    cfg
+                } else {
+                    // if the configuration is invalid exit early
+                    std::process::exit(-1);
```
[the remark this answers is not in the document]
gila: Yes, but a valid crash loop I would think.

mayastor/src/replica.rs
```diff
@@ -449,15 +449,26 @@ impl ReplicaIter {
     }
 }
 
+// XXX: shut this not simply be a bdev iterator with .filter()?
```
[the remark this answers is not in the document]
gila: No, it is more like a question for future readers

mayastor/src/subsys/opts.rs
gila: The IO's are  sent out parallel, so I don't think we want to do that.

mayastor/src/subsys/opts.rs
gila: thx fixed :+1:

mayastor/src/subsys/opts.rs
gila: fixed :+1:

mayastor/src/core/env.rs
```diff
@@ -580,12 +603,37 @@ impl MayastorEnvironment {
         Self::target_init().unwrap();
     }
 
+    fn load_yaml_config(&self) {
+        // load the config and apply it before any subsystems have started.
+        // there is currently no run time check that enforces this.
+        let cfg = if let Some(yaml) = &self.mayastor_config {
+            info!("loading yo-mama configfile");
```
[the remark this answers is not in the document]
gila: hehehe =) thx, no, let me remove that :dancers:

mayastor/src/subsys/opts.rs
tiagolobocastro: not the retry IOs right? so for, if the frontend timeout is 1s, the child timeout is 1s with 5 retries, if even the first retry would not be useful because the frontend has already timeout? So the frontend IO timeout should allow for the retries plus some slack

gila: That assumes we do implicit retries without the initiator asking us to do so, is that what you mean?  We can talk about it some in the future what is best, but I think we dont want to do to many things on behalf of  the initiator here... :thinking:

tiagolobocastro: ah I thought that was the case now as I saw the retries being set to 5 on NvmeBdevOpts...
yes this is not the place to do it, I was just wondering :)

PR #208, opened 2020-04-30
Enhancements to terraform scripts and documentation update

jkryl:


terraform/mod/lxd/main.tf
```diff
@@ -38,7 +38,7 @@ resource "lxd_container" "c8s" {
 
   config = {
     "boot.autostart"       = true
-    "raw.lxc"              = "lxc.mount.auto = proc:rw cgroup:rw sys:rw\nlxc.mount.entry = /lib/modules lib/modules none bind,ro 0 0\nlxc.mount.entry = /boot boot none bind.ro 0 0\nlxc.apparmor.profile = unconfined\nlxc.cgroup.devices.allow = a\nlxc.cap.drop="
+    "raw.lxc"              = "lxc.mount.auto = proc:rw cgroup:rw sys:rw\nlxc.apparmor.profile = unconfined\nlxc.cgroup.devices.allow = a\nlxc.cap.drop="
```
jonathan-teh, line 41: Without these mounts, I had to add `SystemVerification` to kubeadm's `--ignore-preflight-errors` in `mod/k8s/master.sh`.

terraform/README.adoc
```diff
@@ -71,7 +86,9 @@ users.users.gila = {
 === Setting up LXD on Nixos
 
 Using LXD is faster and requires fewer resources. Also, you can do many things
-with lxd that you typically can't do quickly or easily with VMs.
+with lxd that you typically can't do quickly or easily with VMs. Make sure
+that your system is using *unstable channel* for nixpkgs (at least LXD v4 is
```
jonathan-teh, line 90: I was able to use LXD v3, on Debian 10 because that's the latest version that can be installed due to its older version of snap.

terraform/README.adoc
```diff
@@ -34,6 +47,9 @@ new nvmf target or what have you.
 So we need a form of infra management but also - configuration management.
 These scripts are intended to do just that.
 
+Currently the scripts don't support anything else than KVM and LXD
+deployments.
```
jonathan-teh, line 51: I think this is clearer as "...scripts only support KVM and LXD deployments".

terraform/README.adoc
jkryl: For NixOS the requirement is LXD v4. I suppose it's not related to LXD itself, but rather to improvements done to packaging scripts in the nix. This section is about setting up LXD on NixOS, but if you agree I can add a note to the following section that is about LXD on other distros that a v3 is sufficient. If you confirm that you agree I will make the change. Thanks

terraform/README.adoc
jkryl: agree. thanks!

terraform/mod/lxd/main.tf
jkryl: that's interesting as I was able to run k8s without any changes even without these filesystems (and @gila too). Normally I would suspect NixOs vs Ubuntu difference but in this case it happens in the container that is the same for all of us. So I can't explain why it fails for you. I have seen error messages about missing kernel modules etc. but they are all accompanied by a follow-up text "please ignore if you run k8s in a container". Are you sure that your kubelet failed because of the missing directories? Perhaps the root cause was a different problem.

terraform/README.adoc
jonathan-teh: Yes, that's fine.

terraform/mod/lxd/main.tf
yannis218: On my workspace I removed the /boot, which allowed it to proceed to the next step, at which point it failed due to missing `configs` kernel module.  I then added `SystemVerification` to the ignore list.  I would like to see some definitive solution to these difficulties, lest we all end up with a variety of local patches.

terraform/README.adoc
```diff
@@ -82,45 +99,92 @@ overlay
...
+lxc exec first -- /bin/bash -c 'wget http://google.com/'
+----
 
-Now comes a somewhat tricky part. We need to install the terraform lxd provider,
-but it does not exist in nixpkgs. I'll create a PR to NixOS shortly.
+=== Setting up LXD with terraform on other distros than NixOS
 
-On other distributions, manually install the lxd provider from
+Read the above section and ignore NixOs specific configuration and
+workarounds. When it comes to installing terraform with LXD provider,
```
yannis218, line 184: Could the earlier steps be copied and modified rather than referring the reader those steps but requiring them to modify on-the-fly?  Then the thinking needed can be done once, rather than every time.

terraform/main.tf
```diff
@@ -14,14 +15,15 @@ module "lxd" {
   source   = "./mod/lxd"
   ssh_user = var.ssh_user
   ssh_key  = var.ssh_key
+  num_nodes = var.num_nodes
```
yannis218, line 18: Yes!  I did think it was missing!

terraform/mod/lxd/main.tf
jonathan-teh: I started with `SystemVerification` but wasn't sure what other checks it disabled, hence the mount entries. It's looking like it's easier to just add that and move on.

terraform/README.adoc
```diff
@@ -82,45 +99,92 @@ overlay
...
-    enable = true;
-    zfsSupport = false;
-  };
 
- users.users.gila = {
-    isNormalUser = true;
-    extraGroups = [ "wheel" ... "lxd" "lxc" ];
-  };
+After that run `lxd init` and configure it to suit k8s cluster and your
+needs. In particular do not use btrfs or ZFS as a storage pool. Docker's
```
jamie-0, line 131: This needs to mentioned in the quick start. We say setup lxd correctly but don't say what "correct" is. btrfs is the default choice when running lxd init for the first time

terraform/mod/lxd/main.tf
jonathan-teh: Without the mount entries, this is error I get:
`module.k8s.null_resource.k8s[0] (remote-exec): 	[ERROR SystemVerification]: failed to parse kernel config: unable to load kernel module: "configs", output: "modprobe: ERROR: ../libkmod/libkmod.c:586 kmod_search_moddep() could not open moddep file '/lib/modules/5.4.0-0.bpo.4-amd64/modules.dep.bin'\nmodprobe: FATAL: Module configs not found in directory /lib/modules/5.4.0-0.bpo.4-amd64\n", err: exit status 1`

terraform/README.adoc
jkryl: I will add that 👍

terraform/mod/lxd/main.tf
jonathan-teh: The SystemVerification check starts here: https://github.com/kubernetes/kubernetes/blob/release-1.18/cmd/kubeadm/app/preflight/checks.go#L519 and includes kernel config validation: https://github.com/kubernetes/system-validators/blob/v1.1.2/validators/kernel_validator.go#L179

gila: Yeah, I would agree, just remove the SystemVerification, otherwise, we have to keep this file generic for all possible permutations and that's really not what we want i would say.

terraform/variables.tf
```diff
@@ -1,13 +1,19 @@
...
   type        = string
   description = "Where the images will be stored"
-  default     = "/code/ubuntu-pool"
+  default     = "/home/user/images"
+}
+
+variable "private_key_path" {
+  type        = string
+  description = "SSH private key path"
+  default     = "/home/user/.ssh/k8s_rsa"
```
gila, line 10: I'd remove the "default" here @jkryl --then terraform wil prompt you instead of going off with a known to be broken default.

terraform/variables.tf
```diff
@@ -1,13 +1,19 @@
...
 }
 
 variable "ssh_key" {
   type        = string
   description = "SSH pub key to use"
-  default     = "<contents of ~/.ssh/id_rsa.pub>"
+  default     = "ssh-rsa ..."
```
gila, line 16: Same here - just remove it. It will prompt the user to fill in the value

terraform/variables.tf
```diff
@@ -17,7 +23,7 @@ variable "ssh_key" {
 variable "ssh_user" {
   type        = string
   description = "The user that should be created and who has sudo power"
-  default     = "gila"
+  default     = "user"
```
gila, line 26: as the user "gila" come standard on all distro's it worked in the past ;) maybe remove this too.

terraform/README.adoc
jkryl: right. I'm reworking this section and splitting it into three sections: common instructions, nixos specific, other distros specific.

terraform/variables.tf
jkryl: Oh, and the default is wrong. I left the path from my env 🤦  Thanks! will fix

terraform/variables.tf
jkryl: ok 👍

terraform/variables.tf
tiagolobocastro: just an idea for all defaults, what about leaving the reasonably sensible defaults but commented out? so we'd know what to edit should anyone want to "hardcode" them?

jkryl: good idea 👍

terraform/variables.tf
jkryl: Either comment out the default or pick something more standard that is commonly used as a default. Could be "jack", "demo", ... but whatever you prefer guys

terraform/mod/lxd/main.tf
jkryl: I have added SystemVerification flag as suggested by Jonathan. Thanks!

jkryl:
I have addressed all review comments. Please check if everything is according to your liking and if it, please resolve comments raised by you and approve the PR. That way I will be sure I haven't missed anything. Thanks for all the comments!

PR #212, opened 2020-05-04
CAS-225. Added handling for reset IO in nexus function table.

chriswldenyer:
CAS-225.
Adds functionality in the nexus to propagate reset IOs to the children.

This includes rust-based testing of reset functionality in a
new unit test plus additional support for the reset IO in the core handle
code, (thanks to @gila ).


chriswldenyer:
Closing to squash the commits

PR #215, opened 2020-05-06
CAS-174 Create initial framework for persisting metadata to MayaMeta partition

cjones1024:
Initial design of basic framework for persisting metadata to the first disk partition reserved for this purpose. Note that the second commit is a CI test,  but can usefully serve as an example of how the API is to be used.


mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -60,6 +65,27 @@ pub enum ChildError {
...
+    MetaDataIndexRead { source: ChildIoError },
+    #[snafu(display("Metadata is invalid"))]
+    MetaDataIndexInvalid { source: MetaDataError },
+
+    #[snafu(display("Failed to allocate buffer for metadata"))]
+    MetaDataObjectAlloc { source: DmaError },
+    #[snafu(display("Failed to read metadata from child"))]
+    MetaDataObjectRead { source: ChildIoError },
+    #[snafu(display("Metadata is invalid"))]
+    MetaDataObjectInvalid { source: MetaDataError },
```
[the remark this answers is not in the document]
cjones1024: Actually, the problem here is that the display messages do not distinguish the different errors. That was an oversight. What I would prefer to do is to have distinct display messages for each Error type.

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,620 @@
...
+// creates a circular definition, resulting in the compiler error:
+//   "cycle detected when processing ChildError"
+// This is resolved by providing an explicit conversion.
+impl std::convert::From<ChildError> for MetaDataError {
+    fn from(error: ChildError) -> MetaDataError {
+        MetaDataError::NexusChildError {
+            error: format!("{}", error),
+        }
+    }
+}
```
jkryl, line 620: Cycles like this usually indicate that there is something wrong with the architecture. It would be good to have a look again if the code is split optimally between the modules. For example there are NexusChild methods implemented in metadata module using metadata error type. I think that's the first confusing thing. I would expect NexusChild methods being defined in nexus_child module along with the nexus-child struct and using nexus-child-error instead of metadata-error. But I don't have a good understanding of the whole feature. Perhaps we can take the discussion offline to clarify things.

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -256,31 +282,25 @@ impl NexusChild {
         self.state == ChildState::Open || self.state == ChildState::Faulted
     }
 
-    /// read and validate this child's label
-    pub async fn probe_label(&self) -> Result<NexusLabel, ChildError> {
+    pub fn get_dev(&self) -> Result<(&Bdev, &BdevHandle), ChildError> {
```
gila, line 285: lets not make this pub :+)

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -403,6 +423,115 @@ impl NexusChild {
         })
     }
 
+    pub async fn probe_index(
```
gila, line 426: Please document what the function does, and make it a good habit, please. And should this be pub?

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -403,6 +423,115 @@ impl NexusChild {
...
+            let mut buf = desc
+                .dma_malloc((blocks * block_size) as usize)
+                .context(MetaDataObjectAlloc {})?;
+            self.read_at(
+                (metadata.header.self_lba + entry.data_start) * block_size,
+                &mut buf,
+            )
+            .await
+            .context(MetaDataObjectRead {})?;
+            list.push(
```
gila, line 526: what happens this allocated memory? This memory is precious, and if you need to keep it around, we should copy it out, and free the DMA-able memory, makes sense?

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,620 @@
...
+
+    #[snafu(display("{}", error))]
+    NexusChildError { error: String },
+}
+
+#[derive(Debug, Deserialize, PartialEq, Default, Serialize, Copy, Clone)]
+pub struct MetaDataHeader {
+    /// Signature identifying this as a MetaDataHeader object
+    pub signature: [u8; 8],
+    /// Size of this header in bytes
```
gila, line 82: Looks nice with the comments, doesn't it? :+1: Please also add a /// above the struct itself :)

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,620 @@
...
+                    * MetaDataHeader::INDEX_ENTRY_SIZE,
+                block_size,
+            );
+
+        MetaDataHeader {
+            signature: [0x4d, 0x61, 0x79, 0x61, 0x44, 0x61, 0x74, 0x61],
+            header_size: MetaDataHeader::METADATA_HEADER_SIZE,
+            self_checksum: 0,
+            generation: 0,
+            self_lba: partition.ent_start + 1, /* skip the first block of the
```
gila, line 158: Lets make sure we keep the terminoligy right -- we are dealing with the metadata itself, dont think that that deals with partitions :)

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,620 @@
...
+pub struct MetaDataIndexEntry {
+    pub revision: u64,
+    pub timestamp: u64,
+    pub data_checksum: u32,
+    pub data_start: u64,
+    pub data_end: u64,
+}
+
+impl MetaDataIndexEntry {
+    ///
```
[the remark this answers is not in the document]
gila: :+1:

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,620 @@
...
+            });
+        }
+
+        let (bdev, _desc) = self.get_dev()?;
+        let block_size = bdev.block_len() as u64;
+
+        let timestamp = now
+            .duration_since(UNIX_EPOCH)
+            .context(TimeStampError {})?
+            .as_secs();
```
gila, line 315: I think it would be a good idea to have a higher resolution here (i.e usec or nsec?)

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,620 @@
...
+        }
+
+        Err(MetaDataError::MissingPartition {})
+    }
+
+    // Retrieve header + index from "MetaData" partition.
+    pub async fn get_metadata(&self) -> Result<NexusMetaData, MetaDataError> {
+        if let Some(partition) = self.probe_label().await?.partitions.get(0) {
+            if partition.ent_type
+                == GptGuid::from_str("27663382-e5e6-11e9-81b4-ca5ca5ca5ca5")
```
gila, line 525: Let us define this somewhere once as a const

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,620 @@
...
+
+    // Create a new header + index on "MetaData" partition.
+    pub async fn create_metadata(
+        &mut self,
+    ) -> Result<NexusMetaData, MetaDataError> {
+        let (bdev, _desc) = self.get_dev()?;
+
+        if let Some(partition) = self.probe_label().await?.partitions.get(0) {
+            if partition.ent_type
+                == GptGuid::from_str("27663382-e5e6-11e9-81b4-ca5ca5ca5ca5")
```
gila, line 505: Lets define these somewhere once as a const

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,620 @@
+use std::{
```
gila, line 1: Add the top of this file, please add a big theory statement where you explain (roughly) how it works and how it can be used. Top file comments like these should be documented using `//!` as a maker.  as an example you see reactor.rs 

(again, all of this renders into nice docs)

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,620 @@
...
+    /// CRC-32 checksum of the index table
+    pub index_checksum: u32,
+    /// Offset of first usable data block relative to self_lba
+    pub data_start: u64,
+    /// Offset of last usable data block relative to self_lba
+    pub data_end: u64,
+}
+
+impl MetaDataHeader {
+    pub const METADATA_HEADER_SIZE: u32 = 72;
```
gila, line 107: This maps to 136 bytes? .. orso? what about the others i.e 512-136? are they reserved? should we pad the struct to 512 and mark it reserved? I think thats wise to do...

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,620 @@
...
+            return Err(MetaDataError::IndexChecksum {});
+        }
+        Ok(index)
+    }
+
+    /// Construct and validate an empty index
+    pub fn empty_index(
+        header: &MetaDataHeader,
+    ) -> Result<Vec<MetaDataIndexEntry>, MetaDataError> {
+        if header.index_checksum != 0 {
```
gila, line 238: Not sure what this is supposed to do... :thinking:

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,620 @@
...
+        let checksum = crc32::checksum_ieee(buf.as_slice());
+        if checksum != entry.data_checksum {
+            return Err(MetaDataError::ObjectChecksum {});
+        }
+        Ok(NexusConfig::from_slice(buf.as_slice())
+            .context(DeserializeError {})?)
+    }
+}
+
+impl NexusChild {
```
gila, line 258: I think this belongs here.

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,620 @@
...
+        }
+
+        self.write_at(metadata.header.self_lba * block_size, &buf)
+            .await
+            .context(IndexWrite {})?;
+
+        Ok(())
+    }
+
+    pub async fn write_config_object(
```
gila, line 296: these are specific to labels and such, lets not add that to the chilld

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,620 @@
...
+                entry.data_end = start + blocks;
+            }
+            start = entry.data_end + 1;
+        }
+
+        Ok(())
+    }
+
+    // Update checksums and write out MetaData header + index to disk.
+    pub async fn sync_metadata(
```
gila, line 487: This syncs the header + index, not all the metadata right? problably we want to be explicitit about that?

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,620 @@
...
+                && partition.ent_name.name == "MayaMeta"
+            {
+                return Ok(self.probe_index(partition.ent_start).await?);
+            }
+        }
+
+        Err(MetaDataError::MissingPartition {})
+    }
+
+    // Retrieve specified config object from "MetaData" partition.
```
gila, line 536: I assume get the object at the at index N? the comment is not really clear ifso can we rename n to index?

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,620 @@
...
+    pub async fn get_config_object(
+        &self,
+        metadata: &NexusMetaData,
+        n: u32,
+    ) -> Result<Option<NexusConfig>, MetaDataError> {
+        if n < metadata.header.used_entries {
```
gila, line 542: we should also verify that n > max index entries (28?)

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,620 @@
...
+        Ok(None)
+    }
+
+    // Retrieve latest config object from "MetaData" partition.
+    pub async fn get_latest_config_object(
+        &self,
+        metadata: &NexusMetaData,
+    ) -> Result<Option<NexusConfig>, MetaDataError> {
+        if metadata.header.used_entries > 0 {
+            let n = metadata.header.used_entries - 1;
```
gila, line 554: here you use -1 but for the above you dont...

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,620 @@
...
+        metadata: &mut NexusMetaData,
+        n: u32,
+    ) -> Result<(), MetaDataError> {
+        if metadata.index.len() != metadata.header.used_entries as usize {
+            return Err(MetaDataError::IndexInconsistent {});
+        }
+
+        if n < metadata.header.used_entries {
+            metadata.index.remove(n as usize);
+            metadata.header.used_entries -= 1;
```
gila, line 572: here again, there is get_object() get_latest, delete, n -1 or n?

mayastor/src/bdev/nexus/nexus_state.rs
```diff
@@ -0,0 +1,36 @@
+use serde::{Deserialize, Serialize};
```
gila, line 1: nexus_state to me means something radically different then Config versions.

gila: Let us not start out with different versions. For testing, if you want to test those, then thats great and by all means but lets not start with deadcode from day zero by introductuuinng versions straight away.

mayastor/src/bdev/nexus/nexus_state.rs
```diff
@@ -0,0 +1,36 @@
...
+use std::collections::HashMap;
+
+#[derive(Debug, Deserialize, PartialEq, Default, Serialize, Clone)]
+pub struct NexusConfigVersion1 {
+    pub name: String,
+    pub labels: Vec<String>,
```
gila, line 7: whats "labels" ?

gila: from the tests it seems that labels should be "key?

cjones1024: The NexusConfigVersion<N> structs are really just placeholders for whatever we finally decide to define as our "data structure". The fields are just made up nonsense. The point was to demonstrate that we could handle different Versions of the struct" by wrapping them in an enum. You are also right in that we should have just a "version1" for now and try not to add newer versions too often. But the mechanism for handling multiple versions should be there right from the start.

PR #218, opened 2020-05-06
CAS-177 Redirect to new gRPC & remove old interface

yannis218:


csi/src/mount.rs
```diff
@@ -111,58 +111,97 @@ pub fn mount_opts_compare(m1: &[String], m2: &[String], ro: bool) -> bool {
...
+
+    match probe_defaults(fsname, &mi) {
+        Ok(opts) => {
+            cleanup_mount(&mi).unwrap_or_else(|e| {
+                warn!("{}", e);
+            });
+            Ok(opts)
+        }
+        Err(err) => Err(format!(
+            "Filesystem {} will not be available: {}",
```
jkryl, line 150: it seems that the same message will be printed twice (in probe_filesystems you have the same). I would say that `let opts = probe_defaults(fsname, &mi)?;` would work just fine.

csi/src/mount.rs
```diff
@@ -111,58 +111,97 @@ pub fn mount_opts_compare(m1: &[String], m2: &[String], ro: bool) -> bool {
...
-    }
 
-    Ok(filesystems)
+    if code == 0 {
+        return Ok(());
+    }
+    Err(format!(
+        "Failed to cleanup default mount options files: {}",
+        stderr
+    ))
```
jkryl, line 204: thanks for doing this! It will make the patch for lxd cluster much easier than before.

csi/src/mount.rs
yannis218: Oh, yes - didn't notice that...  fixing.

PR #219, opened 2020-05-07
save container images to github as artifacts

jamie-0:
keep the docker images as artefacts for each master build


jamie-0:
> Upload to where?

github actions as build artefacts, as we don't want every master build being published to a container registry

jamie-0:
added moac but I still need to stop the images being built for every pull request

jamie-0:
> Why are we building moac on every build now rather than only when the files change? What are we trying to do exactly, there is not much information in the jira... (eg when to upload the images, every PR, every build, release branch?)

At the moment we just want the images to be uploaded on master, i have now addressed that issue with an if statement in the image.yaml file.

tiagolobocastro:
> > Why are we building moac on every build now rather than only when the files change? What are we trying to do exactly, there is not much information in the jira... (eg when to upload the images, every PR, every build, release branch?)
> 
> At the moment we just want the images to be uploaded on master, i have now addressed that issue with an if statement in the image.yaml file.

Can we not keep the old moac yaml and avoid building moac and the image everytime? Takes extra 12 minutes to do it which isn't massive, but less would be better :)

jamie-0:
> > > Why are we building moac on every build now rather than only when the files change? What are we trying to do exactly, there is not much information in the jira... (eg when to upload the images, every PR, every build, release branch?)
> > 
> > 
> > At the moment we just want the images to be uploaded on master, i have now addressed that issue with an if statement in the image.yaml file.
> 
> Can we not keep the old moac yaml and avoid building moac and the image everytime? Takes extra 12 minutes to do it which isn't massive, but less would be better :)

It's done in parallel to the other jobs  :slightly_smiling_face:

PR #221, opened 2020-05-11
CSI iSCSI updates for quick.md

blaisedias:
Resolves CAS-248

### DIFF
--- a/doc/quick.md
+++ b/doc/quick.md
@@ -14,8 +14,9 @@ This has steps have been tested on:
 
  * Only works on Intel Core i7 processor architecture (march=corei7) and higher
  * MayaStor Daemonset (MDS) requires privileged mode
- * MDS requires 2MB hugepages
+ * MDS requires 2MB huge pages
  * For testing, MDS requires the Network Block Device driver (NBD) and the XFS filesystem kernel module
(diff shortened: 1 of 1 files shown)


blaisedias:
> Hi Blaise,
> 
> I think in general, this is fine for the time being, until such time as we've had some time to try applying the instructions as provided. However I'm going to suggest changes to the first section:
> 
>     1. The first line requires a grammatical correction
> 
>     2. Since this is a new build I think we should remove all configurations listed under the "has been tested on" section which we haven't actually yet tested on, not simply rely on these still being correct.
> 
> 
> Thanks,
> Glenn

doc/quick.md
```diff
@@ -1,21 +1,16 @@
...
-    * You will need to add following directory mapping to `services_kubelet->extra_binds` in your `cluster.yml`:
-     `/opt/rke/var/lib/kubelet/plugins:/var/lib/kubelet/plugins`. Otherwise the CSI socket paths won't match and the CSI
-     driver registration process will fail.
 
 ### Requirements
 
  * Only works on Intel Core i7 processor architecture (march=corei7) and higher
  * MayaStor Daemonset (MDS) requires privileged mode
- * MDS requires 2MB hugepages
+ * MDS requires 2MB huge pages
```
gila, line 11: We require a huge page *size* of 2MB  and require at least 512 of those. We will bump that in the future but technically, 1GB is enough right now.

PR #225, opened 2020-05-14
reactor: enable full speed mode

gila:
When we moved to the future polling of the reactor, the ability to run
without a developer delay was omitted. When we do release builds we want
to run at full speed.

Also this change adds a release option to mayastor/default.nix

### DIFF
--- a/nix/pkgs/mayastor/default.nix
+++ b/nix/pkgs/mayastor/default.nix
@@ -1,4 +1,5 @@
 { stdenv
+, release ? false
 , e2fsprogs
 , libaio
 , libiscsi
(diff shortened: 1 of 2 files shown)


nix/pkgs/mayastor/default.nix
```diff
@@ -29,95 +30,96 @@ let
...
+      protobuf
+      utillinux
+      xfsprogs
+      utillinux.dev
+    ];
 
-      doCheck = false;
-      meta = { platforms = stdenv.lib.platforms.linux; };
-    };
+    buildType = if release then "debug" else "release";
```
jkryl, line 89: looks like the condition is inverted.

gila: Yes Jan, thanks, @blaisedias also mentioned this to mee this morning. After considering for a while, I think I want to redo this file in a bit more,  but later. 

I think its cleaner to have the docker images themselves split out into separate files. The processes would remain the same, but the logic is elsewhere. Mayastor/default.nix would then only have one artefact instead of multiple things.

PR #228, opened 2020-05-14
Improved test for changes related to iscsi feature

jkryl:
Minor change to adjust the test closer to reality and avoid useless parsing of JSON in the code.


jkryl:
tested in k8s cluster (creating iscsi vol and running a fio on it) and works fine. merging ..

PR #229, opened 2020-05-14
We lost a lot of useful log messages after we flipped mayastor from j…

jkryl:
…sonrpc to grpc

The messages were part of CSI agent, but now when the implementation was moved to mayastor, they were lost.

Note that some messages might be superfluous as similar information is already logged by the internal routines used by grpc methods, BUT it's certainly better to have a few duplicated messages than having none when troubleshooting user's problem (especially so when we plan to release new images now). We can find the right balance later by removing some of them.

### DIFF
--- a/mayastor/src/grpc.rs
+++ b/mayastor/src/grpc.rs
@@ -28,8 +28,8 @@ use crate::{
 #[derive(Debug)]
 struct UnixStream(tokio::net::UnixStream);
 
-/// Lookup a nexus by its uuid. Return error if uuid is invalid or nexus
-/// not found.
+/// Lookup a nexus by its uuid prepending "nexus-" prefix. Return error if
+/// uuid is invalid or nexus not found.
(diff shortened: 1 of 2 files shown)


mayastor/src/grpc.rs
```diff
@@ -61,245 +83,324 @@ impl Mayastor for MayastorGrpc {
...
-        Ok(Response::new(ListPoolsReply {
-            pools,
-        }))
+        let reply = ListPoolsReply {
+            pools: pool::list_pools()
+                .iter()
+                .map(|pool| Pool {
+                    name: pool.name.clone(),
+                    disks: pool.disks.clone(),
+                    state: match pool.state.as_str() {
```
gila, line 134: The whole idea of getting away from jsonrpc for our own methods is to get rid of these mappings, to make sure we do that I've filed an issue for this, thanks for the itermediate solution :+1:

PR #230, opened 2020-05-14
Automatically rebuild when adding a new child

tiagolobocastro:
When a  child is added the control plane expects mayastor to rebuild
the child. Added a new parameter to add_child which allows for this
whilst also allowing for a finer grained control over rebuild.

Adjusted all the required tests for this including semistandard fixes

Added a test to verify that the new parameter works as expected

Resolves CAS-149

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_bdev.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev.rs
@@ -165,11 +165,11 @@ pub enum Error {
     },
     #[snafu(display(
         "Failed to execute rebuild operation on job {} of nexus {}",
-        child,
+        job,
         name,
     ))]
     RebuildOperationError {
-        child: String,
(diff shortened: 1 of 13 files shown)


mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -163,8 +163,16 @@ pub enum Error {
         child: String,
         name: String,
     },
-    #[snafu(display(""))]
-    RebuildOperationError { source: RebuildError },
+    #[snafu(display(
+        "Failed to execute rebuild operation on job {} of nexus {}",
```
gila, line 167: Is this message correct? job == child?

tiagolobocastro: good catch thanks :+1:

PR #232, opened 2020-05-14
Change so that mount does not use loopback

blaisedias:
Mounting iscsi targets with the std rust library,
results in creation of loopback devices and mounting
of the volume on the loopback devices.
Also unmounts were not destroying the loopback devices.
Workaround by resolving the absolute path of the iscsi
target to avoid the issue.

### DIFF
--- a/csi/src/node/iscsiutil.rs
+++ b/csi/src/node/iscsiutil.rs
@@ -32,7 +32,7 @@ fn wait_for_path_to_exist(devpath: &str, max_retries: i32) -> bool {
     device_path.exists()
 }
 
-fn realpath(path: String) -> String {
+fn iscsi_realpath(path: String) -> String {
     match std::fs::read_link(path.as_str()) {
         Ok(linkpath) => {
             // For iscsi the root path is /dev/disk/by-path
@@ -70,7 +70,7 @@ fn attach_disk(
     // method has succeeded.
     if wait_for_path_to_exist(device_path.as_str(), 1) {
         trace!("path already exists!");
-        return Ok(realpath(device_path));
+        return Ok(iscsi_realpath(device_path));
     }
 
     let args_discoverydb_new = [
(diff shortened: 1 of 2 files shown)


csi/src/node/iscsiutil.rs
```diff
@@ -32,6 +32,19 @@ pub fn wait_for_path_to_exist(devpath: &str, max_retries: i32) -> bool {
     device_path.exists()
 }
 
+fn realpath(path: String) -> String {
```
jonathan-teh, line 35: Nitpick but I think this function should be renamed because it's not exactly the same as realpath(3), itself provided by fs::canonicalize, but rather an internal iSCSI-specific variant.

csi/src/node/iscsiutil.rs
```diff
@@ -214,7 +232,7 @@ pub fn iscsi_detach_disk(device_path: &str) -> Result<(), String> {
     }
 }
 
-pub fn iscsi_find(uuid: &str) -> Option<String> {
+pub fn iscsi_find_(uuid: &str) -> Option<String> {
```
jkryl, line 235: I think it does not need to be public. Can we get some more meaningful name for it? In particular a name that would make clear the difference between iscsi_find and iscsi_find_. Comments for the functions (/// ...) would not make harm :-)

csi/src/node/iscsiutil.rs
blaisedias: Fixed by https://github.com/openebs/Mayastor/pull/232/commits/974b40c85561e4771b75746145bf986f01355458

csi/src/node/iscsiutil.rs
blaisedias: Fixed by https://github.com/openebs/Mayastor/pull/232/commits/eb83c5027105004374de9d3eca8b798a7b8fd232

blaisedias:
@jkryl we could avoid using the sys-mount crate, but that would be a larger change than I would like to do at this point.

jkryl:
agree. But let's create a ticket so that we don't forget to do that in the future 👍

PR #233, opened 2020-05-14
Reformat and update the Quickstart Guide

GlennBullingham:
Rework the structure of the Quickstart guide:

1. Prioritise Storage Node configuration ahead of deploying mayastor on the cluster
2. Provide a definition of what a Storage Node is, within the context of mayastor
3. Update tested platforms with AKS information
4. Update example commands given for checking and setting 2MB huge page support.
        (sudo tee now used as this works on AKS)
5. Update example commands given for creating storage classes, to clearly identify the transport
       (ie. nbd vs iscsi)

Also, remove reference to NVME_TCP module in "Requirements" section as this isn't supported by the CSI driver in this release

### DIFF
--- a/doc/quick.md
+++ b/doc/quick.md
@@ -7,7 +7,7 @@ This quickstart guide has been tested against the following platforms and config
 - Microsoft Azure Kubernetes Service (AKS)
     - k8s version 1.16.7
     - Ubuntu 16.04.6 LTS (GNU/Linux 4.15.0-1077-azure x86_64)
-    - not tested with scale-sets
+    - Not tested with scale-sets
 
     > AKS is not currently recommended for use with Mayastor in production owing to the need, at this time, to make configuration changes to worker nodes which are not readily compatible with upgrades or autoscaling behavior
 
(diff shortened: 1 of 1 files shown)


GlennBullingham:
@blaisedias We came to consensus on (ie. Jeffry told us) the correct capitalisation for Mayastor quite some time ago.  If there are any camel case instances now then they have been left in by myself in error, and I'll correct them.

doc/quick.md
```diff
@@ -62,49 +52,90 @@ These steps have been tested on:
...
+
+    This change can be made persistent across reboots by adding the required value to `/etc/sysctl.conf`, thus:
+    ```
+    vm.nr_hugepages = 512
+    ```
+
+    > If you modify the huge page configuration of a node, you *must* restart the kubelet or reboot the node
+    
+
+2.  Load the NBD and XFS kernel modules (if not already loaded).
```
gila, line 70: This step is not mandatory actually. Only if you want to use nbd, nbd should be loaded and -- XFS is optional but i recommond using XFS. Should make this step optinal?

GlennBullingham: If loading XFS is optional but recommended, can I ask what the basis of that recommendation is?  What's the potential benefit of loading it and does it apply only in specific circumstances?

PR #234, opened 2020-05-15
nexus: address child sizing mismatches

tiagolobocastro:
When a child fails to be added due to a different block size or invalid
size the returned error does not use the name form the bdev which is
now destroyed.

When adding a new child the size of the child has to be larger than the
minimum required.

The nexus bdev now takes into account the requested size when it was
created rather than just looking at its children sizes.
At the moment new children need to be as large as the smallest child,
this is due to the partition being done in relation to the smallest
child. We might want to consider doing it in relation to the nexus size
+ metadata.

Resolves CAS-247


tiagolobocastro:
Pushed another bug fix pointed by @cjones1024, thank you :+1:

gila:
@cjones1024 as you've reviewed it can you then also approve it if you think it is ready to go in?

PR #235, opened 2020-05-15
CAS-174 Create initial framework for persisting metadata to MayaMeta partition

cjones1024:
Initial design of basic framework for persisting metadata to the first disk partition reserved for this purpose.

This is a new PR that obsoletes #215, and changes have been made to address review comments in the former. In doing so the code has been reorganised somewhat. In particular, in order to avoid circular references in the snafu error definitions, the probe_xxx() functions have been moved from nexus_child.rs to nexus_label.rs and nexus_metadata.rs respectively.


mayastor/src/bdev/nexus/nexus_label.rs
```diff
@@ -462,7 +496,9 @@ impl GPTHeader {
             deserialize_from(&mut reader).context(DeserializeError {})?;
 
         if gpt.header_size != 92 {
-            return Err(LabelError::GptHeaderSize {});
+            return Err(LabelError::GptHeaderSize {
+                size: 92,
```
gila, line 500: would we want to say here that the expected size (92) is not the gpt.header_size?

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,828 @@
...
+//!    });
+//!    let now = SystemTime::now();
+//!    let child = &mut nexus.children[0];
+//!    let mut metadata = child.create_metadata().await?;
+//!    child.append_config_object(&mut metadata, &config, &now).await?;
+//!
+//! Sample code to retrieve the latest config object from an existing index:
+//!
+//!    let metadata = child.get_metadata().await?;
+//!    let config = child.get_latest_config_object(&metadata).await?;
```
gila, line 30: :+1:

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,828 @@
...
+    /// MetaDataHeader)
+    pub data_start: u64,
+    /// Location of last data block (relative to self_lba in the
+    /// MetaDataHeader)
+    pub data_end: u64,
+}
+
+impl MetaDataIndexEntry {
+    /// Convert a slice into an index array
+    pub fn from_slice(
```
gila, line 229: tip, these days rust has this: https://doc.rust-lang.org/std/convert/trait.TryFrom.html

The difference here is that the commonly known Trait from<T> for U -- can not fail, as it simply returns U. TryFrom, returns a Result<U, E> which means that it can try but is allowed to fail. 

so you *COULD* impl TryFrom for this but in the near future somewhere :)

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,828 @@
...
+
+#[derive(Debug, PartialEq, Default, Clone)]
+pub struct NexusMetaData {
+    pub header: MetaDataHeader,
+    pub index: Vec<MetaDataIndexEntry>,
+}
+
+impl NexusMetaData {
+    /// Construct a MetaDataHeader from raw data
+    fn read_header(buf: &DmaBuf) -> Result<MetaDataHeader, MetaDataError> {
```
gila, line 261: nit; when I read the function name I assume it "reads" from disk, but its already read from disk as its int he DmaBuf...

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -0,0 +1,828 @@
...
+        MetaDataHeader::from_slice(buf.as_slice())
+    }
+
+    /// Construct index array from raw data
+    fn read_index(
```
gila, line 266: ... same here, its not wrong fo sure, but all these functions convert "from A to B" -- with a Result<> it has TryFrom written all over it :)

mayastor/tests/nexus_metadata.rs
```diff
@@ -0,0 +1,127 @@
...
+
+const DISKNAME: &str = "/tmp/disk1.img";
+const BDEVNAME: &str = "aio:///tmp/disk1.img?blk_size=512";
+
+pub mod common;
+
+#[test]
+fn get_metadata() {
+    common::mayastor_test_init();
+    let output = Command::new("truncate")
```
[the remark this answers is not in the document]
cjones1024: will leave for now as really should do this for the other tests as well

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -33,40 +25,21 @@ pub enum ChildError {
...
-    #[snafu(display("Failed to read partition table from child"))]
-    PartitionTableRead { source: ChildIoError },
-    #[snafu(display("Partition table is invalid"))]
-    PartitionTableInvalid { source: LabelError },
 }
 
 #[derive(Debug, Snafu)]
 pub enum ChildIoError {
-    #[snafu(display("Error writing to {}", name))]
+    #[snafu(display("Error writing to {}: {}", name, source))]
```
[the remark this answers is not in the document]
cjones1024: Actually you do need the source here explicitly if you want its message to be included. And rust will definitely complain if the number of "{}" specifiers does not match the number of arguments. What is true is that you do not need to set the "source" field explicitly when you are creating a "context" for this error as snafu will automatically do that for you.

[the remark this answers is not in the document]
tiagolobocastro: Actually you don't strictly need to... what I meant was not adding the {} as well as the source.

The created error will indeed have the source information as the snafu ext does this for you. But extra information might not included in the default Display trait, you'd have to extend it. You need however to unwind the error chain when printing errors which we do on rpc/grpc code... yes it's a bit messy what happens when there are sources in the display message (not just because of your code) is that some rpc errors end up with duplicated messages, eg:
Error: RpcError { code: InternalError, msg: "Failed to write label of nexus nexus-be8d0ecf-ed31-497e-a554-56c47d401b4c: Write error: **Invalid descriptor for child bdev aio:///dev/ram0: Invalid descriptor for child bdev aio:///dev/ram0"** }

So we might need to remove the sources from all messages, or add them to all.  @gila  any thoughts?

PR #241, opened 2020-05-20
CAS-214: Add support to lock/unlock LBA ranges

paulyoong:
- Provide the ability to lock/unlock LBA ranges when rebuilding a segment.
- Temporarily change the Docker build image that is used so that we can use the SPDK changes to support the locking/unlocking.
- Additional sleeps have been introduced in various test cases to allow time for the rebuild operations to take effect. The locking/unlocking seems to have affected the test timings somewhat.


.github/workflows/build.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: self-hosted
     timeout-minutes: 20
     container:
-      image: docker.io/mayadata/ms-buildenv:nix
+      image: docker.io/teamrebuild/ms-buildenv:spdk-changes
```
gila, line 15: You are creating a lot of work for yourselves this way. I’m fine with it, just saying that merging all of this back is not going to be trivial.

mayastor/src/core/descriptor.rs
```diff
@@ -109,3 +166,41 @@ impl Debug for Descriptor {
         )
     }
 }
+
+extern "C" fn spdk_range_cb(
```
gila, line 170: I think we can use the cb handlers for this in the ffihelper.rs ?

mayastor/src/replicas/rebuild/rebuild_impl.rs
```diff
@@ -205,6 +213,17 @@ impl RebuildJob {
             &mut self.tasks.buffers[id as usize]
         };
 
+        let len = copy_buffer.len() as u64 / self.block_size;
+        let mut ctx = RangeContext::new(blk, len, self.nexus_channel.clone());
+
+        self.nexus_descriptor
+            .lock_lba_range(&mut ctx)
```
gila, line 220: If we need to hold on to the context where are going to put it?

mayastor/src/core/descriptor.rs
```diff
@@ -109,3 +166,41 @@ impl Debug for Descriptor {
...
+        }
+    }
+}
+
+/// Store the context for calls to lock_lba_range and unlock_lba_range.
+/// Corresponding lock/unlock calls require the same context to be used.
+pub struct RangeContext {
+    pub offset: u64,
+    pub len: u64,
+    io_channel: Arc<IoChannel>,
```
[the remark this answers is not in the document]
paulyoong: Everything that is saved in the context is there to ensure that the parameters passed to the bdev_lock_lba_range function match those passed to the bdev_unlock_lba_range. 

The unlock expects all the parameters to match the corresponding lock; if not it returns an error.

.github/workflows/build.yaml
paulyoong: This is meant to be temporary until we update the SPDK version. It was the best way we could think of to get the build to pass.

mayastor/src/core/descriptor.rs
paulyoong: The problem with the cb handlers was that they create a new oneshot sender. As mentioned above, the unlock expects the same parameters as the lock, so this would cause the unlock to return an error.

mayastor/src/replicas/rebuild/rebuild_impl.rs
paulyoong: The ctx is only required for the duration of the copy. Once the unlock is called on line 241 we no longer require it.

mayastor/src/core/descriptor.rs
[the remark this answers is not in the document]
paulyoong: I used Arc<> here because multiple rebuild jobs can be running concurrently and using the same channel. This was primarily to prevent a channel having to be created each time a rebuild job runs.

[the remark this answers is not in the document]
gila: I think what it asserted is that the pointer address is still the same, and it is not deep comparing bit for bit that what there during lock is there for unlock? 

If you hold onto the RangeContext, the fact that you hold onto it as you need it later implies that it is not dropped. The Arc<> keeps a reference count (atomic) to the channel, such that when nobody is using it its get dropped automatically (i.e garbage collected). 

Also, Arc<> implies that the channel is used by multiple different cores, we should never share channels across cores.

mayastor/src/core/descriptor.rs
gila: I think here I might indeed not fully get it but:

the callback here has a fixed address with the code, it by definition always lives on the same address. So it seems to be the case that the void *ctx needs to be point to the same thing, what is _in_ the thing is not something that it cares about, right?

mayastor/src/replicas/rebuild/rebuild_impl.rs
gila: Right, but where we do we store for the duration of the lock?

PR #242, opened 2020-05-21
CAS-284 import device traversal code

gila:
This code is based on older device inventory code. We dont need the
inventory bits, but this code allows us to lookup a device based in the
storage URI's such that we can find devices based on the UUIDs.


devinfo/src/block_device.rs
```diff
@@ -0,0 +1,114 @@
...
+
+        let uuid = Uuid::parse_str(nq[1]).map_err(|e| NqnInvalid {
+            value: format!("the UUID is invalid {}", e.to_string()),
+        })?;
+
+        match value.scheme() {
+            "iscsi" => Ok(BlkDev::Scsi(uuid)),
+            "nvmf" => Ok(BlkDev::Nvmf(uuid)),
+            _ => Err(DevInfoError::NotSupported {
+                value: "".to_string(),
```
paulyoong, line 68: Should we return the scheme that isn't supported here?

devinfo/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+mod block_device;
+use snafu::Snafu;
+
+#[derive(Debug, Snafu)]
+pub enum DevInfoError {
+    #[snafu(display("Device {} not found", path))]
+    NotFound { path: String },
+    #[snafu(display("Failed to parse value {}", value))]
+    ParseError { value: String },
+    #[snafu(display(" Name qualifier invalid {} ", value))]
```
paulyoong, line 13: Erroneous space before "Name"

gila: :+1:

devinfo/src/block_device.rs
gila: Thanks! fixed :+1:

PR #243, opened 2020-05-21
mayastor-test: Test Nexus child going offline

jonathan-teh:
Create a new replica timeout test, based on the existing reset test,
that performs I/O, brings down one Nexus child via a SIGTERM followed
by performing a read, which must pass, then bringing down the
remaining Nexus child followed by another read, which must fail.

This requires Bio::assess to mark the IO as failed if the child IO
failed otherwise it can return a read as success with the buffer
untouched.

Also fix an off-by-one bug in the existing io_test.

### DIFF
--- a/mayastor/tests/common/ms_exec.rs
+++ b/mayastor/tests/common/ms_exec.rs
@@ -172,14 +172,19 @@ impl MayastorProcess {
     }
 
     /// terminate the mayastor process and wait for it to die
-    pub fn sig_term(&self) {
+    pub fn sig_term(&mut self) {
+        if self.child == 0 {
+            return;
+        }
+        let child = self.child;
+        self.child = 0;
         Command::new("kill")
-            .args(&["-s", "SIGTERM", &format!("{}", self.child)])
+            .args(&["-s", "SIGTERM", &format!("{}", child)])
             .spawn()
             .unwrap();
 
         // blocks until PID is gone, signals are racy by themselves however
-        waitpid(Pid::from_raw(self.child as i32), None).unwrap();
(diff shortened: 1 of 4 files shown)


mayastor/tests/common/ms_exec.rs
```diff
@@ -172,14 +172,19 @@ impl MayastorProcess {
     }
 
     /// terminate the mayastor process and wait for it to die
-    pub fn sig_term(&self) {
+    pub fn sig_term(&mut self) {
+        if self.child == 0 {
```
chriswldenyer, line 176: These changes weren't mentioned in the description - it looks like it avoids attempting to kill the same process twice?

mayastor/tests/replica_timeout.rs
```diff
@@ -0,0 +1,155 @@
...
+        }
+        Err(_e) => {
+            assert_eq!(ok, false);
+            return;
+        }
+    };
+
+    let slice = buf.as_slice();
+
+    for &it in slice.iter().take(511) {
```
chriswldenyer, line 151: Should this be 512   :-)

mayastor/tests/common/ms_exec.rs
jonathan-teh: Yes, when it shuts down normally it calls sig_term again.

chriswldenyer: Ah, right.

mayastor/tests/replica_timeout.rs
jonathan-teh: You're right. I originally had a C-style for loop from 0-511 but clippy suggested this instead.

PR #244, opened 2020-05-22
Reworked tabular output

yannis218:
Added more tests

### DIFF
--- a/csi/src/client.rs
+++ b/csi/src/client.rs
@@ -9,6 +9,7 @@ extern crate clap;
 
 use byte_unit::Byte;
 use clap::{App, AppSettings, Arg, ArgMatches, SubCommand};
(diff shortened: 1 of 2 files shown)


yannis218:
> The code looks great! Just wondering what shall we do when it comes to pretty-printing nexus
> children. Currently they don't seem to be printed by nexus list command. There can be many of
> them. Should we have a separate command for them (just for printing the children - one line per
> each child) or should we embed them to nexus list output (i.e. one line per child but indented at
> the beginning).

The code did use to print out the children, one per line, after the nexus details, but when I did the tabular output with columns, etc, I thought it would look odd as the children info would not fit the columns.  We could have an option to add them at the end of each line, making the line overlong, and a command for listing the children of a specific nexus it a nice table.  What do you think?

PR #246, opened 2020-05-22
rebuild: more testing bug fixes (CAS-246)

tiagolobocastro:
added a few more testcases and that led to fixing a few bugs

rebuild with childs of different sizes (prior that would fail) including
using children with sizes < metadata
using non multiple of the segment size for completeness
the lookup of rebuild jobs and concurrent rebuild jobs which revealed a
bug
rebuild operations idempotence (currently on the client api) though we
may want to restrict it to RPC specifically

bonus: when the rebuild tests fail (local or CI) the cargo test itself
gets stuck which is caused by panics on a different thread (?) so moved
the failures to the reactor thread and also use different nexus and bdev
names on each test - this produces a cleaner failure log

rebuild code coverage reported by grcov is now ~84% (if you trust it)

todo: cas-287 will test the IO failures

Resolves: CAS-246

### DIFF
--- a/mayastor/tests/nexus_rebuild.rs
+++ b/mayastor/tests/nexus_rebuild.rs
@@ -477,21 +477,24 @@ fn rebuild_operations() {
 }
 
 #[test]
-fn rebuild_concurrently() {
-    test_ini("rebuild_concurrently");
+// rebuilds N children at the same time
+// creates the nexus with 1 healthy and then adds N children which
+// have to be rebuilt - this means we have N active rebuilds jobs
+fn rebuild_multiple() {
+    test_ini("rebuild_multiple");
 
-    let concurrent_rebuilds = 4;
+    let active_rebuilds = 4;
     Reactor::block_on(async move {
         nexus_create(1).await;
         let nexus = nexus_lookup(nexus_name()).unwrap();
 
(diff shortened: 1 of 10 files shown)


mayastor/tests/nexus_rebuild.rs
```diff
@@ -226,15 +244,273 @@ async fn nexus_test_child(child: u64) {
...
+        // already stopped
+        nexus.stop_rebuild(&get_dev(1)).await.unwrap();
+
+        nexus_lookup(nexus_name()).unwrap().destroy().await.unwrap();
+    });
+
+    test_fini();
+}
+
+#[test]
```
gila, line 473: can you add a comment to spell out what it means to rebuild concurent? it not obvious from looking at the code what is happening here.

tiagolobocastro: yep will do.
(it basically adds N children to the nexus causing N rebuild jobs to be created; it then removes the children and tries to add them back again to make sure new rebuild jobs get created)

tiagolobocastro: I've updated the fn name (concurrently was perhaps an unfortunate choice) and comments plus the linter fixes

gila:
LGTM

PR #251, opened 2020-05-26
Update CSI sidecar containers to latest versions

jkryl:


blaisedias:
FWIW I tested with fio on my local k8s cluster.

PR #254, opened 2020-05-27
CAS-214: Add support to lock/unlock LBA ranges

paulyoong:
- Provide the ability to lock/unlock LBA ranges when rebuilding a segment.

- Temporarily change the Docker build image that is used so that we can use the SPDK changes to support the locking/unlocking.

- Additional sleeps have been introduced in various test cases to allow time for the rebuild operations to take effect. The locking/unlocking seems to have affected the test timings somewhat.

### DIFF
--- a/mayastor/tests/lock_lba_range.rs
+++ b/mayastor/tests/lock_lba_range.rs
@@ -1,0 +1,225 @@
+#[macro_use]
+extern crate log;
+
+pub mod common;
+
+use crossbeam::channel::unbounded;
+use mayastor::{
+    bdev::{nexus_create, nexus_lookup},
+    core::{
+        Bdev,
+        DmaBuf,
+        MayastorCliArgs,
+        MayastorEnvironment,
+        RangeContext,
+        Reactor,
+        Reactors,
+    },
+};
+use std::sync::{Arc, Mutex};
+
+const NEXUS_NAME: &str = "lba_range_nexus";
+const NEXUS_SIZE: u64 = 10 * 1024 * 1024;
+const NUM_NEXUS_CHILDREN: u64 = 2;
+
+fn test_ini() {
+    test_init!();
+    for i in 0 .. NUM_NEXUS_CHILDREN {
+        common::delete_file(&[get_disk(i)]);
+        common::truncate_file_bytes(&get_disk(i), NEXUS_SIZE);
+    }
+
+    Reactor::block_on(async {
+        create_nexus().await;
+    });
+}
+
+fn test_fini() {
+    for i in 0 .. NUM_NEXUS_CHILDREN {
+        common::delete_file(&[get_disk(i)]);
+    }
+
+    Reactor::block_on(async {
+        let nexus = nexus_lookup(NEXUS_NAME).unwrap();
+        nexus.destroy().await.unwrap();
+    });
+}
+
+fn get_disk(number: u64) -> String {
+    format!("/tmp/disk{}.img", number)
+}
+
+fn get_dev(number: u64) -> String {
+    format!("aio://{}?blk_size=512", get_disk(number))
+}
+
+async fn create_nexus() {
+    let mut ch = Vec::new();
+    for i in 0 .. NUM_NEXUS_CHILDREN {
+        ch.push(get_dev(i));
+    }
+
+    nexus_create(NEXUS_NAME, NEXUS_SIZE, None, &ch)
+        .await
+        .unwrap();
+}
+
+fn get_shareable_ctx(offset: u64, len: u64) -> Arc<Mutex<RangeContext>> {
+    let nexus = Bdev::open_by_name(NEXUS_NAME, true).unwrap();
+    Arc::new(Mutex::new(RangeContext::new(offset, len, &nexus)))
+}
+
+async fn lock_range(ctx: &mut RangeContext) {
+    let mut nexus = Bdev::open_by_name(NEXUS_NAME, true).unwrap();
+    let _ = nexus.lock_lba_range(ctx).await;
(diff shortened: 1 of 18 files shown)


mayastor/tests/lock_lba_range.rs
```diff
@@ -0,0 +1,269 @@
...
+
+#[test]
+// Test acquiring and releasing a lock.
+fn lock_unlock() {
+    test_ini();
+    Reactor::block_on(async {
+        let mut nexus = Bdev::open_by_name(NEXUS_NAME, true).unwrap();
+        let mut ctx = RangeContext::new(1, 5);
+        let ch = nexus.get_channel().unwrap();
+        let _ = nexus.lock_lba_range(&mut ctx, &ch).await;
```
gila, line 95: You get back here (and ingore) a result<(), Error> -- what should it be? Ok() or Err() we should test for whatever is needed here?

mayastor/tests/lock_lba_range.rs
```diff
@@ -0,0 +1,269 @@
...
+#[test]
+// Test that an error is received when the lock/unlock contexts don't match
+fn lock_unlock_different_context() {
+    test_ini();
+    Reactor::block_on(async {
+        let mut nexus = Bdev::open_by_name(NEXUS_NAME, true).unwrap();
+
+        let mut ctx = RangeContext::new(1, 5);
+        let ch = nexus.get_channel().unwrap();
+        let _ = nexus.lock_lba_range(&mut ctx, &ch).await;
```
gila, line 110: Same here, if this MUST succeed, you should unwrap()

mayastor/tests/lock_lba_range.rs
```diff
@@ -0,0 +1,269 @@
...
+        let ctx_clone = Arc::clone(&ctx);
+        let ch = get_shareable_nexus_channel();
+        let ch_clone = Arc::clone(&ch);
+        reactor.send_future(async move {
+            lock_range(
+                &mut ctx_clone.lock().unwrap(),
+                &ch_clone.lock().unwrap(),
+            )
+            .await;
+            let _ = s.send(());
```
gila, line 141: if you dont care for the result, its best to just unwrap :)

mayastor/tests/lock_lba_range.rs
```diff
@@ -0,0 +1,269 @@
...
+                &mut ctx1_clone.lock().unwrap(),
+                &ch_clone1.lock().unwrap(),
+            )
+            .await;
+            trace!("Second lock succeeded");
+            let _ = lock_sender.send(());
+        });
+        reactor_poll!(100);
+
+        // First lock is held, second lock shouldn't succeed
```
gila, line 162: Thank you for the comments -- really helpful to understand what you are trying to test :+1:

mayastor/tests/lock_lba_range.rs
```diff
@@ -0,0 +1,269 @@
...
+        let ch = get_shareable_nexus_channel();
+        let ch_clone = Arc::clone(&ch);
+
+        reactor.send_future(async move {
+            lock_range(
+                &mut ctx_clone.lock().unwrap(),
+                &ch_clone.lock().unwrap(),
+            )
+            .await;
+            trace!("Lock succeeded");
```
gila, line 223: if you unwrap() the lock_range, the trace is not needed :)

mayastor/tests/lock_lba_range.rs
```diff
@@ -0,0 +1,269 @@
...
+    {
+        let reactor = Reactors::current();
+
+        // First lock
+        let (s, r) = unbounded::<()>();
+        let ctx = get_shareable_ctx(1, 10);
+        let ctx_clone = Arc::clone(&ctx);
+        let ch = get_shareable_nexus_channel();
+        let ch_clone = Arc::clone(&ch);
+        reactor.send_future(async move {
```
gila, line 135: I'm not sure why you need the Arc<T> here, you are not creating a thread? Also, along those lines, it looks like you can simply use block_on() here as in the previous test? 

Typically, using reactor_poll!() is only needed when you issue IO and you don't know how many futures there it will spawn (or at least hard to determine).

If you run exactly one future, and when that future is completed, you are done, block_on() is your friend. 

reactor_poll!() for example is more appropriate when you format a filesystem, you don't know how many times poll() you have to call, so what we typically do here is: "poll the reactor until your receive a message" or "poll the reactor N times". 

In the first case, (wait for a message) you typically send something over a channel when format completes.

paulyoong: You're right, Arc isn't need here - I'll remove it.

I initially tried using block_on() but it didn't seem to work for this test case. I got into a deadlock scenario. The second lock blocks waiting for the first lock to be released. The first lock can't be released because the reactor is blocked waiting on the second lock.

mayastor/tests/lock_lba_range.rs
paulyoong: Done

mayastor/tests/lock_lba_range.rs
paulyoong: Done

mayastor/tests/lock_lba_range.rs
paulyoong: Done

mayastor/tests/lock_lba_range.rs
paulyoong: :+1:

mayastor/tests/lock_lba_range.rs
paulyoong: Done

PR #259, opened 2020-06-01
nexus: Close child when destroying all Nexus children

jonathan-teh:
Make Nexus::destroy_children close a child first before destroying it,
making its behaviour match the code comment.

Addresses this panic when creating a Nexus with multiple children
where the 2nd or later child is inaccessible:

thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Init`,
 right: `Closed`', mayastor/src/bdev/nexus/nexus_child.rs:218:9

Fixes: 63bd078 ("nexus: add more test cases for adding children")

Resolves: CAS-297


jonathan-teh:
Thanks for the approvals. Given that it's only clippy that's failing with code that wasn't even touched here due to the use of nightly clippy in CI, I'm going to merge this.

PR #260, opened 2020-06-01
Add CI tests for CSI iSCSI

blaisedias:
Restructure the CSI test javascript files so
that we can repeat the test suite with different
parameters for different protocols.
Update workflow to run CSI nbd and iscsi as well.
Run iscsiadm as chroot on CI/CD
Remove ISCSIADM environment setting from mayastor-daemonset
yaml.
Using the environment variable it is still supported,
but we fallback to trying to use in order
 - /bin/mayastor-iscsiadm
 - iscsiadm

Precise iscsiadm login to target

Streamline iscsi disk attach
	- Remove rescan
	- use "discovery" instead of "discoverydb"
	- check output of discovery command before iscsi login.

Set the csi iscsi test timeout to 120000
resolves: CAS-235


nvmeadm/src/nvmf_discovery.rs
```diff
@@ -171,7 +171,7 @@ impl Discovery {
         let _ret = unsafe {
             convert_ioctl_res!(nix_ioctl(
                 f.as_raw_fd(),
-                u64::from(NVME_ADMIN_CMD_IOCLT).try_into().unwrap(),
+                u64::from(NVME_ADMIN_CMD_IOCLT),
```
cjones1024, line 174: Since we're here ... I'm sure Jeffry won't mind if we change this to the proper spelling: NVME_ADMIN_CMD_IOCTL

mayastor-test/test_common.js
```diff
@@ -358,11 +358,7 @@ function dumbCommand (method, args, done) {
       JSON.stringify(args) +
       "'",
     (err, stdout, stderr) => {
-      if (err) {
-        done(new Error(stderr));
-      } else {
-        done();
-      }
+      done(err, stdout);
```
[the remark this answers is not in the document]
blaisedias: The motivation for that change is so that we can use the output to retrieve the actual published URIs from the nexuses. That is what we store in published URIs

nvmeadm/src/nvmf_discovery.rs
blaisedias: Done

.github/workflows/build.yaml
```diff
@@ -50,23 +50,27 @@ jobs:
...
-      - run: echo 8192 | tee /sys/kernel/mm/hugepages/hugepages-2048kB/nr_hugepages
-      - run: ( cd mayastor-test && npm install)
-      - run: ( cd mayastor-test && ./node_modules/mocha/bin/mocha test_cli.js )
-      - run: ( cd mayastor-test && ./node_modules/mocha/bin/mocha test_replica.js )
-      - run: ( cd mayastor-test && ./node_modules/mocha/bin/mocha test_csi.js )
-      - run: ( cd mayastor-test && ./node_modules/mocha/bin/mocha test_nexus.js )
-      - run: ( cd mayastor-test && ./node_modules/mocha/bin/mocha test_rebuild.js )
+      - run: ( sudo modprobe nbd )
+      - run: ( sudo modprobe xfs )
+      - run: ( lsmod | grep -e xfs -e nbd )
```
tiagolobocastro, line 61: Guess this and 63 was just for debugging? Should probably be removed

.github/workflows/build.yaml
```diff
@@ -50,23 +50,27 @@ jobs:
...
+      - run: ( echo 2048 | sudo tee /sys/kernel/mm/hugepages/hugepages-2048kB/nr_hugepages )
```
tiagolobocastro, line 62: I imagine this change is because we're now running on the cloud? If so a comment would be good because if we switch back to the self hosted we'd be overriding the 8192 from the other tests (which can run in parallel)

mayastor-test/test_common.js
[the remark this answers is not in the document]
jkryl: yes, I got why the stdout is there, but I don't get why you throw stderr away in case of an error.

.github/workflows/build.yaml
blaisedias: Done. https://github.com/openebs/Mayastor/pull/260/commits/18ee8732816be85c37efcd7d39bb8303bd0dc415

.github/workflows/build.yaml
blaisedias: Fixed. https://github.com/openebs/Mayastor/pull/260/commits/18ee8732816be85c37efcd7d39bb8303bd0dc415

blaisedias:
@jkry as requested, see https://github.com/openebs/Mayastor/pull/260/commits/447b35cd6bb024c519fb7ddf219472fc9b3458bf

PR #261, opened 2020-06-02
CAS-144 Allow default nexus and replica iSCSI target and nvmf ports to be overridden.

cjones1024:
The preferred method for having multiple iSCSI target portals is to use different IP addresses, together with the well known port (3260). As this is not straightforward to do for kubernetes pods (which have a single IP address) we need to use other non-standard ports. In this case we need to allow the port values to be configurable.
resolves CAS-144

### DIFF
--- a/mayastor/src/target/iscsi.rs
+++ b/mayastor/src/target/iscsi.rs
@@ -70,10 +70,7 @@ impl RpcErrorCode for Error {
 
 type Result<T, E = Error> = std::result::Result<T, E>;
 
-/// Default iSCSI target port numbers
-pub const ISCSI_PORT_NEXUS: u16 = 3260;
-pub const ISCSI_PORT_REPLICA: u16 = 3262;
-
+/// Portal Group Tags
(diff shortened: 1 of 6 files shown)


mayastor/src/target/iscsi.rs
```diff
@@ -69,9 +70,9 @@ impl RpcErrorCode for Error {
 
 type Result<T, E = Error> = std::result::Result<T, E>;
 
-/// iscsi target port number
-const ISCSI_PORT_NEXUS: u16 = 3260;
-const ISCSI_PORT_REPLICA: u16 = 3262;
+/// Default iSCSI target port numbers
```
gila, line 73: Now that the port numbers are defined by the config subsystem, we should be able to get rid of this consts, right? That is to say, if you add them to the opts.rs (iirc)

cjones1024: Have moved these constants (and also those for the nvmf ports) to opts.rs

PR #263, opened 2020-06-03
Remove unused vec from filesystem struct

blaisedias:


csi/src/mount.rs
```diff
@@ -14,7 +14,6 @@ pub struct MountInfo {
 #[derive(Clone, Debug)]
 pub struct Fs {
```
gila, line 15: What about just putting strings in the supported list?

blaisedias: Fixed by https://github.com/openebs/Mayastor/pull/263/commits/e24901e2e1dd4d5691e57b450068d981089f93a9

PR #264, opened 2020-06-03
CAS-53: Initial NVMEoF nexus implementation

tjoshum:
A version of the NVMEoF nexus support work which is slimmed down as much as possible. Limited to nexus using the same target as the replicas and without discovery support.

### DIFF
--- a/mayastor-test/test_nexus.js
+++ b/mayastor-test/test_nexus.js
@@ -29,6 +29,12 @@ const uringFile = '/tmp/uring-backend';
 const diskSize = 64 * 1024 * 1024;
 // external IP address detected by common lib
 const externIp = common.getMyIp();
+// port at which iscsi replicas are available
+const iscsiReplicaPort = '3261';
+
+// NVMEoF frontends don't play nicely with iSCSI backend at the time of writing,
+// so temporarily disable these tests.
+const doIscsiReplica = false;
 
 // Instead of using mayastor grpc methods to create replicas we use a config
 // file to create them. Advantage is that we don't depend on bugs in replica
@@ -55,7 +61,7 @@ const configNexus = `
   MaxConnectionsPerSession 1
 
 [PortalGroup1]
-  Portal GR1 0.0.0.0:3261
+  Portal GR1 0.0.0.0:${iscsiReplicaPort}
 
 [InitiatorGroup1]
   InitiatorName Any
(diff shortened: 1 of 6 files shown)


mayastor/src/bdev/nexus/nexus_nvmf.rs
```diff
@@ -0,0 +1,72 @@
...
+    target,
+    target::nvmf::{share, unshare},
+};
+
+#[derive(Debug, Snafu)]
+pub enum NexusNvmfError {
+    #[snafu(display("Bdev not found {}", dev))]
+    BdevNotFound { dev: String },
+    #[snafu(display(
+        "Failed to create iscsi target for bdev uuid {}, error {}",
```
chriswldenyer, line 18: should be "nvmf target" ?

mayastor/src/subsys/opts.rs
```diff
@@ -43,7 +43,7 @@ impl Default for NexusOpts {
     fn default() -> Self {
         Self {
             replica_port: 8420,
-            nvmf_nexus_port: 4421,
+            nvmf_nexus_port: 8420, // Same target for now.
```
jkryl, line 46: don't we have the conflict then? If replica_port and nvmf_nexus_port are the same then we won't be able to create a nexus on a node with a replica, if my understanding is correct.

mayastor-test/test_nexus.js
```diff
@@ -320,16 +429,20 @@ describe('nexus', function () {
...
-      assert.equal(nexus.children[2].state, 'CHILD_ONLINE');
-      assert.equal(
-        nexus.children[3].uri,
         'nvmf://127.0.0.1:8420/nqn.2019-05.io.openebs:disk2'
       );
-      assert.equal(nexus.children[3].state, 'CHILD_ONLINE');
+      assert.equal(nexus.children[2].state, 'CHILD_ONLINE');
+      if (doIscsiReplica) {
+        assert.equal(
+          nexus.children[2].uri,
```
chriswldenyer, line 440: Shouldn't the index be 3? doIscsiReplica will also affect the index used by uring below, I think.

mayastor-test/test_nexus.js
```diff
@@ -29,6 +29,12 @@ const uringFile = '/tmp/uring-backend';
 const diskSize = 64 * 1024 * 1024;
 // external IP address detected by common lib
 const externIp = common.getMyIp();
+// port at which iscsi replicas are available
+const iscsiReplicaPort = '3261';
+
+// NVMEoF frontends don't play nicely with iSCSI backend at the time of writing,
+// so temporarily disable these tests.
+const doIscsiReplica = false;
```
jkryl, line 37: we just discussed with @gila  yesterday that we need to move iscsi target to a different process (to the process that contains nvmf target). So it looks that by doing that we could solve this problem as well. No need to change anything in this PR, just a note for myself and Jeffry. 👍

tjoshum:
> Although not merged yet, can you have a quick look at PR#261 as one of us will need to resolve the conflicts that will arise.

The conflicts don't look horrendous, so just merge your changes when they're approved and I can mop up.

mayastor/src/subsys/opts.rs
tjoshum: I hit problems with the separate targets for nexus and replicas, and apparently the way they work also changes in the new SPDK, so we agreed we'll consider using separate targets for nexus and replicas to be 'blocked' on the SPDK update, but with this PR we'll unblock the CSI for nvmf work so that can get progress in the mean time,

mayastor-test/test_nexus.js
tjoshum: Yup, pushing a fix now.

mayastor/src/bdev/nexus/nexus_nvmf.rs
tjoshum: Yup, pushing a fix now.

tjoshum:
As discussed, I'll let @gila merge this into his changes, then onto develop.

tjoshum:
Oops, didn't mean to add a merge commit. Rebased again now.

PR #266, opened 2020-06-05
CAS-241. Functionality to store and query a record of recent errored IOs

chriswldenyer:
CAS-241. Functionality to store and query a record of recent errored IOs in each nexus child and a means of enabling the feature and setting the buffer size.
Test added which adds an error bdev layer between the nexus and one of its
replicas and allows unit testing of the feature. Also included unit testing
of the error store independenly of the bdev.

Addressed some code-review comments, ops and errors are stored
instead of flags, also made the ordering of parameters more consistent.
Use monolithic time instead of absolute time in error store.
Using message-passing to the management core.

Addressed some code-review comments, ops and errors are stored
instead of flags, also made the ordering of parameters more consistent.
Use monolithic time instead of absolute time in error store.
Using message-passing to the management core.
The age parameter in the query is now an Option, if absent all ages allowed.

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_child_error_store.rs
+++ b/mayastor/src/bdev/nexus/nexus_child_error_store.rs
@@ -1,0 +1,306 @@
+use spdk_sys::{spdk_bdev, spdk_bdev_io_type};
+use std::{
+    fmt::{Debug, Display},
+    time::{Duration, Instant},
+};
+
+use crate::subsys::Config;
+
+use crate::bdev::nexus::nexus_io::{io_status, io_type};
+
+use crate::core::{Cores, Reactors};
+
+use crate::bdev::nexus::{
+    nexus_bdev,
+    nexus_bdev::{
+        nexus_lookup,
+        Error::{ChildMissing, ChildMissingErrStore},
+        Nexus,
+    },
+};
+
+use serde::export::{fmt::Error, Formatter};
+
+#[derive(Copy, Clone)]
+pub struct NexusChildErrorRecord {
+    io_offset: u64,
+    io_num_blocks: u64,
+    timestamp: Instant,
+    io_error: i32,
+    io_op: spdk_bdev_io_type,
+}
+
+impl Default for NexusChildErrorRecord {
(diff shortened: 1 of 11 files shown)


mayastor/src/bdev/nexus/nexus_child_error_store.rs
```diff
@@ -0,0 +1,306 @@
...
+}
+
+impl Default for NexusChildErrorRecord {
+    fn default() -> Self {
+        Self {
+            io_op: 0,
+            io_error: 0,
+            io_offset: 0,
+            io_num_blocks: 0,
+            timestamp: Instant::now(), // for lack of another suitable default
```
paulyoong, line 40: FYI - Because you are using default values for the u64 and i32 variables you could just use their default values. So I think you could do something like
Self {timestamp: Instant::now(), ..Default::default()};

mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -65,7 +65,7 @@ pub mod io_type {
     //    pub const IO_NUM_TYPES: u32 = 14;
 }
 
-/// the status of an IO
+/// the status of an IO - note: values copied from spdk bdev_module.h
 pub mod io_status {
     //pub const NOMEM: i32 = -4;
     //pub const SCSI_ERROR: i32 = -3;
```
paulyoong, line 71: Should these commented out lines be removed?

chriswldenyer: I think the idea is to un-comment them as the code that uses them gets implemented (they are from spdk). For example the reset IO recently. We may need to use the other error types at a later point when we need to support them.

paulyoong: Ok :+1:

mayastor/src/bdev/nexus/nexus_child_error_store.rs
chriswldenyer: Oh OK, didn't know about that. Thanks!

PR #267, opened 2020-06-05
[WIP] 

gila:
I've tried my best to reduce any changes as part of the upgrade as much as possible. Follow up changes will happen later that address the todo and the target creation.


.github/workflows/build.yaml
```diff
@@ -11,21 +11,22 @@ jobs:
     runs-on: self-hosted
     timeout-minutes: 20
     container:
-      image: docker.io/mayadata/ms-buildenv:latest
+      image: docker.io/mayadata/ms-buildenv:gila
```
tiagolobocastro, line 14: guess you'll move this to latest before you merge?

jonathan-teh, line 14: Is this the tag that will always be used in future?

.github/workflows/build.yaml
```diff
@@ -11,21 +11,22 @@ jobs:
...
-          toolchain: stable
+          toolchain: nightly
       - run: ln -s /host/bin/kmod /bin/modprobe
       - run: /bin/modprobe nbd
       - run: /bin/modprobe xfs
       - run: /bin/modprobe nvme_tcp
       - run: echo 8192 | tee /sys/kernel/mm/hugepages/hugepages-2048kB/nr_hugepages
       - run: rm mayastor/.cargo/config
       - run: rm nvmeadm/.cargo/config
+      - run: export RUST_BACKTRACE=1
```
tiagolobocastro, line 29: do we want to keep this one?

mayastor/src/core/reactor.rs
```diff
@@ -171,19 +216,36 @@ impl<'a> IntoIterator for &'a Reactors {
...
             flags: Cell::new(INIT),
             sx,
             rx,
-        }
+        };
+
+        // create a new main thread on which we run management tasks. As we are
+        // in the process of boot strapping the reactor is not active
+        // and thus we are not executing yet within a context of a valid
+        // thread. Therefor we dispatch a future which will get executed
```
paulyoong, line 235: Typo in "Therefore"

mayastor/src/core/thread.rs
```diff
@@ -76,11 +80,23 @@ impl Mthread {
...
-        unsafe { spdk_set_thread(self.0) };
-        unsafe { spdk_thread_exit(self.0) };
-        unsafe { spdk_thread_destroy(self.0) };
+        debug!("destroying thread...{:p}", self.0);
+        unsafe {
+            spdk_set_thread(self.0);
+            // set that we *want* to exit, but we have not exited yet
+            spdk_thread_exit(self.0);
+
+            // no wait until the thread is actually excited the internal
```
paulyoong, line 91: I think this meant to say "now wait..."

nix/pkgs/libspdk/default.nix
```diff
@@ -21,8 +22,8 @@ stdenv.mkDerivation rec {
   src = fetchFromGitHub {
     owner = "openebs";
     repo = "spdk";
-    rev = "bef8258c3116d8d427d46fab5985bb080ffb7666";
-    sha256 = "00xsfhlqiwnq3777xg8x6chcqr4nxhhqqmbsq1a2gng2iyhlpxzd";
+    rev = "f09e30d7789aed7f16317c865203ae5d1b9a5d5c";
```
paulyoong: There is a newer commit (as of today), should we point it at this version (hash - 0dd3398ace940bb291e18138a2535cf6b1e161b5)?

.github/workflows/build.yaml
gila: no, I will swap it out once the PR is approved

gila: Yes :+1:

.github/workflows/build.yaml
gila: I thnk so yes.

mayastor/src/core/thread.rs
gila: Yes thanks.

mayastor/src/core/reactor.rs
gila: :+1:

PR #268, opened 2020-06-08
Update terraform config

blaisedias:
Add more config settings to variables.tf
Setup hugepages in the node
Add support to load nvme modules ubuntu 20.04 only
Minor change in kube config handling, works with
 - ubuntu 20.04
 - ubuntu 18.04

### DIFF
--- a/terraform/mod/k8s/master.sh
+++ b/terraform/mod/k8s/master.sh
@@ -4,7 +4,7 @@ sudo kubeadm init --config /tmp/kubeadm_config.yaml \
   --ignore-preflight-errors=Swap,NumCPU,SystemVerification
 
 [ -d "$HOME"/.kube ] || mkdir -p "$HOME"/.kube
-sudo cp -i /etc/kubernetes/admin.conf $HOME/.kube/config
+sudo cp /etc/kubernetes/admin.conf $HOME/.kube/config
 sudo chown $(id -u):$(id -g) $HOME/.kube/config
 
 while ! nc -z localhost 6443; do
--- a/terraform/mod/k8s/master.sh
+++ b/terraform/mod/k8s/master.sh
@@ -1,14 +1,15 @@
 #!/bin/bash
-set -e
+set -ex
 sudo kubeadm init --config /tmp/kubeadm_config.yaml \
   --ignore-preflight-errors=Swap,NumCPU,SystemVerification
 
 [ -d "$HOME"/.kube ] || mkdir -p "$HOME"/.kube
-ln -s /etc/kubernetes/admin.conf "$HOME"/.kube/config
+sudo cp -i /etc/kubernetes/admin.conf $HOME/.kube/config
+sudo chown $(id -u):$(id -g) $HOME/.kube/config
 
 while ! nc -z localhost 6443; do
   echo "...Waiting on k8s API server to give a sign of life"
   sleep 5
 done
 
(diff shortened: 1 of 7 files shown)


terraform/mod/k8s/master.sh
```diff
@@ -1,14 +1,15 @@
 #!/bin/bash
-set -e
+set -ex
 sudo kubeadm init --config /tmp/kubeadm_config.yaml \
   --ignore-preflight-errors=Swap,NumCPU,SystemVerification
 
 [ -d "$HOME"/.kube ] || mkdir -p "$HOME"/.kube
-ln -s /etc/kubernetes/admin.conf "$HOME"/.kube/config
+sudo cp -i /etc/kubernetes/admin.conf $HOME/.kube/config
+sudo chown $(id -u):$(id -g) $HOME/.kube/config
```
jkryl, line 8: running cp in interactive mode in automation script sounds like a bad idea.

terraform/variables.tf
```diff
@@ -42,5 +42,25 @@ variable "num_nodes" {
...
+variable "overlay_cidr" {
+  type        = string
+  description = "CIDR, classless inter-domain routing"
+  #default     = "192.168.122.0/16"
+}
+
+variable "nr_hugepages" {
+  type        = string
+  description = "Number of Huge pages"
+  default     = "1024"
```
jkryl, line 57: I'm not sure if we should not use the minimum here that is 512. Given that these scripts are used to create test VMs and at least I create VMs with 4G of mem, taking half of it just for mayastor is maybe too much. Or is there any particular reason why we allocate 2x more than what we need?

terraform/variables.tf
```diff
@@ -42,5 +42,25 @@ variable "num_nodes" {
...
   type        = string
   description = "Ubuntu image for VMs - only needed for libvirt provider"
-  default     = "/bionic-server-cloudimg-amd64.img"
+  default     = "/ubuntu-18.04-server-cloudimg-amd64.img"
 }
+
+variable "overlay_cidr" {
+  type        = string
+  description = "CIDR, classless inter-domain routing"
+  #default     = "192.168.122.0/16"
```
jkryl, line 51: can we have the default here? The previous default worked fine for me. IMHO the fewer settings the user is required to tune the better.

terraform/mod/k8s/master.sh
blaisedias: Fixed https://github.com/openebs/Mayastor/pull/268/commits/5b38dc66d2408d7b84cd517462429a5cd21232c9
TBH I just copied what kubeadm spit out.

terraform/variables.tf
blaisedias: Fixed https://github.com/openebs/Mayastor/pull/268/commits/5b38dc66d2408d7b84cd517462429a5cd21232c9

terraform/variables.tf
blaisedias: Done https://github.com/openebs/Mayastor/pull/268/commits/5b38dc66d2408d7b84cd517462429a5cd21232c9

blaisedias:
In principle I agree with sentiment to support just latest ubuntu release.
That said, I erred on the side of caution, using ubuntu 20.04 has only had limited testing and I wanted to make it easy to fallback to the known working setup for those not working on nvme/nvmf

PR #271, opened 2020-06-09
Reduce the size of MOAC docker image (CAS-306)

jkryl:
Strip build-time dependencies of moac from docker image. It is a dirty
hack though I haven't found a better solution than patching generated
nix files. NPM and NIX aren't friends and this is what you get when
you invite both of them to your house :-) On a positive note we should
be able to remove the workaround when we won't have any modules with
bindings. That means when we get rid of grpc-uds that we plan to do
anyway soon.

The size of moac image dropped down from 981 MB to 271 MB.

Note 1: A lot of the changes are formatting changes as our nix style tool and
node2nix tool don't agree on what is the preferred style of nix files :-)

Note 2: we switch from production to development nix pkg generation, the only
difference is that packages from devDependencies are included as well (mocha,
...). It is unrelated change that is done to support typescript in future (that normally
is a devDependency). It increases the image size by ~10MB so not a big deal.

### DIFF
--- a/csi/moac/README.md
+++ b/csi/moac/README.md
@@ -134,10 +134,58 @@ updated too;
    nix-env -f '<nixpkgs>' -iA nodePackages.node2nix
    ```
 
-3. Generate nix package build files:
+3. Generate nix package build files. The flag `development` is needed for
+   typescript to get included in the package, because it is in `devDependencies`
+   (common case in JS world). The flag does not influence how things are built
+   (debug info, code optimisations, etc.).
    ```bash
    rm -rf node_modules
-   node2nix -l package-lock.json --nodejs-12 -c node-composition.nix
+   node2nix -l package-lock.json --development --nodejs-12 -c node-composition.nix
+   ```
+
+4. Patch generated nix files by following patch in order to reduce the size
+   of the package. This is a temporary workaround for
+   https://github.com/svanderburg/node2nix/issues/187.
+   ```diff
+   diff --git a/csi/moac/node-composition.nix b/csi/moac/node-composition.nix
+   index ac8de82..6441534 100644
+   --- a/csi/moac/node-composition.nix
+   +++ b/csi/moac/node-composition.nix
+   @@ -2,11 +2,12 @@
+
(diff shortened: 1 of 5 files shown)


tjoshum:
> switch from production to development nix pkg generation
Don't understand enough about nix pkg generation to have a detailed opinion, but is this our production container for moac? Feels wrong to be making that switch. How does the change support typescript?

jkryl:
@tjoshum: So the difference is not how modules are built or processed but just which of them are included and which aren't in the production package. With "production" type only packages from the dependencies section of package.json are included. With "development" type packages from dependencies and devDependencies section of package.json are included. So there are some extra packages used only for testing. That's the only difference. Now how does that relate to typescript? In one of my future PRs, I plan to convert a few JS files to TS, just to show how that can be done and as an example for the future work. Typescript compiles *.ts files to *.js files and it is common to be part of devDependencies because it is not needed at run time but only when the app gets built. However, due to the way how NIX builds the stuff (another example of NIX not playing well with JS style of doing things), the typescript compiler must be part of the final package used in production. NIX first makes a copy of the app with dependencies to nix store and only then it builds it. So there are two ways how to make it work: 1. put typescript module to devDependencies and set the build type to develop (that's what I do), 2. put typescript module to dependencies and leave the type of build set to production, however that violates common rules for typescript programs. To me, one is like the other, though I slightly more lean towards following the common way of doing things in JS/TS.

tjoshum:
@jkryl Thanks for the explanation. Sorry to be a pain but could you add a note to that effect (that we need --development for ts) to the readme?

jkryl:
@tjoshum : good point! I will add the comment and update the PR shortly.

@gila : So the dependencies of node_moac-0.1.0 derivation remain the same, because those are build dependencies used to build the output. However the runtime dependencies of the outcome (stored in out attr of node_moac-0.1.0 derivation) have decreased. Runtime dependencies are determined by constructing a NAR archive from out path and scanning the archive for hashes of buildInputs. You won't find any npm and gcc hashes with these changes there.

PR #273, opened 2020-06-10
Rework JS code checking and introduce support for TS

jkryl:
Semistandard should normally be part of nodejs project dependencies. The
reason why we relied on global installation was that it wasn't clear to
me how to call it without having to install all node module dependencies.
That is solved by using npx command now. npx uses either a local copy of
the program if it is in node_modules/ or downloads the tool to temporary
folder and runs it from there. As a result all complexity with creating
the semistandard package in nix files can go away.

As for the typescript, it is just showing the way forward. As time will
go we will convert more JS files to TS files as we will see a good fit.

Resolves: CAS-309


scripts/js-check.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+set -e
```
tjoshum, line 3: Since we're using pipes, we probably want `set -euo pipefail`

jkryl: thanks

csi/moac/pool.ts
```diff
@@ -86,7 +94,7 @@ class Pool {
   //
   // @param {object[]} replicas   New list of replicas properties for the pool.
   //
-  _mergeReplicas (replicas) {
+  _mergeReplicas(replicas: any[]) {
```
blaisedias, line 97: replica: Replica is used below so shouldn't this be replicas: Replica[]?

csi/moac/pool.ts
```diff
@@ -236,7 +244,7 @@ class Pool {
         `Failed to list new replica "${uuid}" on pool "${this}": ${err}`
       );
     }
-    var replicaInfo = resp.replicas.filter((r) => r.uuid === uuid)[0];
+    var replicaInfo = resp.replicas.filter((r: any) => r.uuid === uuid)[0];
```
blaisedias, line 247: r: any should be r: Replica?

jkryl: it's slightly confusing but replica here is object with replica properties returned by grpc method - not Replica object that represents replica internally in moac. So we must use any until we take a look at how to represent grpc replies by TS types.

csi/moac/pool.ts
jkryl: The same comment here. Those are not replica objects. replicasProperties might be a better name but too long :-)

PR #282, opened 2020-06-16
Merge release 0.2.0 staging changes back into master

GlennBullingham:
Return changes from the release 0.2.0 activities to the develop branch:

1) Fix-up for issues identified with rebuild; Add missing nvmf namespace identifier to the bdev name (cas-317) and additionally, fault a child if a rebuild action on it fails

2) nix: Changes to simplify making release builds:
-  point mayastor/default.nix to an actual commit we release
- bumped packages of nixpkgs itself and liburing
- bumped rust overlay such that are are on rust stable 1.44 (from 1.41)\
- bumped liburing 0.5 -> 0.6
- all images pkgs, by default now, are built in release mode
- fixed the libspdk/default.nix (not all variable are set by configure)

3) Update content of the deploy folder
- Remove tested on Azure AKS notice
- Update moac and mayator deployment yamls to target the v0.2.0 tagged images, and not :latest, which currently has no meaning for the develop branch (we don't push images on each build)

### DIFF
--- a/csi/src/node.rs
+++ b/csi/src/node.rs
@@ -11,7 +11,6 @@ use crate::{
     format::probed_format,
     mount::{match_mount, mount_fs, mount_opts_compare, unmount_fs},
 };
-use git_version::git_version;
 
 mod iscsiutil;
 use iscsiutil::{iscsi_attach_disk, iscsi_detach_disk, iscsi_find};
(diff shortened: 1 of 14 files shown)


GlennBullingham:
Closing - as agreed during stand-up, will push release to develop outside of PR context.

PR #285, opened 2020-06-17
CAS-321: add mayastorpool.yaml to deployment

tiagolobocastro:
When the timing is "just right" applying the pool yaml file fails
because it's up to moac to register the pool crd.

Add the pool crd to the deployment which might facilitate new installs

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -44,7 +44,7 @@ A worker node which will not host/contribute storage capacity to Mayastor does n
     HugePages_Surp:        0
 
     ```
-    
+
     If fewer than 512 pages are available, the page count should be configured as necessary, accounting for any other co-resident workloads which may require them. e.g.
 
     ```bash
@@ -57,7 +57,7 @@ A worker node which will not host/contribute storage capacity to Mayastor does n
     ```
 
     > If you modify the huge page configuration of a node, you *must* restart the kubelet or reboot the node
-    
+
 
 2.  Load the Network Block Device (NBD) kernel module.  This is necessary *only* if it is intended to use nbd transport for volume provisioning.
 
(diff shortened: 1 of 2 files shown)


deploy/README.md
```diff
@@ -85,10 +85,11 @@ The YAML files named below are to be found in the `deploy` folder of the Mayasto
 5.  Deploy the Mayastor and CSI components
     ```bash
     kubectl create -f nats-deployment.yaml
+    kubectl create -f mayastorpool.yaml
```
jkryl, line 88: not sure if it makes sense to add just one of them. There are two other yamls as well. Perhaps having just a note that CRD yamls are applied automatically by moac and a link to crds directory in moac would be sufficient. Whoever wants to apply them manually could do that then. But let's see what others think about it.

tiagolobocastro: My reasoning was that the pool instances are only thing which the user creates. The node and the volume instances are always created by moac.
Adding just the note also makes sense, whatever people think makes more sense / is clearer.

GlennBullingham: I think there is potential to confuse the user at to what is being performed in this step.  To me, taking mayastorpool.yaml at face value suggests that it's going to do just that - establish a working pool instance(s), which is not the case.

I'm inclined to agree with Jan, that if we are going to have the user create *some* necessary CRDs explicitly as part of the installation step, rather than leave this to moac (and/or some other kind of operator or supervisory process), then we should define them *all* there and have a single "CRDs" yaml file which creates all such entities required.

tiagolobocastro: it doesn't need to be called mayastorpool right, it could be mayastorpoolcrd, common.yaml, or crd.yaml.
The tricky thing with combining them all is that we'd have to release that as an artifact perhaps, since they're all separate yaml files in the tree.

GlennBullingham: Agreed.  Adding CRD to the filename would make sense.

PR #288, opened 2020-06-19
Update install-nix-action

tjoshum:
This seems to be the smallest change to get CI moving again.


tjoshum:
Maybe we should move certain actions to point to master. Where those actions are solely for interacting with external systems, the loss of stability from using master rather than a release might be offset by the gains from both the action and the external system moving in lock step.
I was hoping for a docker style 'latest' meaning the most recent release branch, but I don't think GitHub actions has such a thing. Anyway, the quickest way to unblock CI is just to bump the install-nix-action version, so that's what I've done.

PR #290, opened 2020-06-22
WIP build: default to a known good version of the rust toolchain

tjoshum:
This also contains changes to shell.ix required by the pre-commit hooks

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -10,25 +10,24 @@ jobs:
     name: Build and run cargo tests
     runs-on: self-hosted
     timeout-minutes: 20
-    container:
-      image: docker.io/mayadata/ms-buildenv:latest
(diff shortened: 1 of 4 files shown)


gila:
@tjoshum Thanks for having a stab at this, it turns out, however, that the problem is far more complicated to fix for various reasons. 

So, here the intent is to -- get shell.nix working again properly, right? I would suggest to simply remove the rust-channel entry from the build inputs such that you can use <insert distro> rustup to install stable, nightly and whatever.

jkryl:
I would suggest implementing what we all agreed upon as an intermediate solution, that is, simply using rust stable everywhere with exception of nix-shell where rust-stable would be the default, but in case of an explicit argument the user could avoid installing rust (in that case user is responsible for managing his own version of the rust with all negative consequences that it might have).

PR #291, opened 2020-06-24
Implement flush support for Nexus. CAS-329

blaisedias:
+ linter formatting changes to
mayastor/src/target/nvmf.rs

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_bdev.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev.rs
@@ -819,10 +819,10 @@ impl Nexus {
             .ch
             .iter()
             .map(|c| unsafe {
-                let (b, c) = c.io_tuple();
+                let (bdev, chan) = c.io_tuple();
                 spdk_bdev_writev_blocks(
-                    b,
-                    c,
+                    bdev,
+                    chan,
                     io.iovs(),
                     io.iov_count(),
                     io.offset() + io.nexus_as_ref().data_ent_offset,
@@ -854,10 +854,10 @@ impl Nexus {
             .ch
             .iter()
             .map(|c| unsafe {
-                let (b, c) = c.io_tuple();
+                let (bdev, chan) = c.io_tuple();
                 spdk_bdev_unmap_blocks(
-                    b,
-                    c,
+                    bdev,
+                    chan,
                     io.offset() + io.nexus_as_ref().data_ent_offset,
                     io.num_blocks(),
                     Some(Self::io_completion),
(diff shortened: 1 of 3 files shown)


mayastor/src/bdev/nexus/nexus_fn_table.rs
```diff
@@ -107,6 +107,10 @@ impl NexusFnTable {
                     trace!("{}: Dispatching RESET {:p}", nexus.bdev.name(), io);
                     nexus.reset(io, &ch)
                 }
+                io_type::FLUSH => {
+                    trace!("{}: Dispatching FLUSH {:p}", nexus.bdev.name(), io);
```
chriswldenyer, line 111: Maybe you want to comment-out the trace? I left the reset un-commented because resets were rare enough, but read and write don't generate trace output, is it needed for flush? 
(BTW the commented-out trace statements are wrong nexus.name -> nexus.bdev.name, could you fix?)

blaisedias: Done. https://github.com/openebs/Mayastor/pull/291/commits/2ffdf29d5786644710db887ecdac4c22e4d7b755

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -874,6 +875,38 @@ impl Nexus {
...
+        pio: *mut spdk_bdev_io,
+        channels: &NexusChannelInner,
+    ) {
+        let mut io = Bio(pio);
+        io.ctx_as_mut_ref().in_flight = channels.ch.len() as i8;
+        let results = channels
+            .ch
+            .iter()
+            .map(|c| unsafe {
+                let (b, c) = c.io_tuple();
```
jonathan-teh, line 889: Reusing 'c' is a tad confusing, (bdev, chan) as used in reset is more readable.

blaisedias: Fixed https://github.com/openebs/Mayastor/pull/291/commits/4b75f677cb11fb839585765031c99f207e4bd770

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -853,10 +854,42 @@ impl Nexus {
...
+                )
+            })
+            .collect::<Vec<_>>();
+
+        if results.iter().any(|r| *r != 0) {
+            error!(
+                "{}: Failed to submit dispatched IO {:?}",
+                io.nexus_as_ref().name,
+                pio
+            );
```
jkryl, line 874: Is it reasonable to log error messages in IO paths? I can image that if there is a bad device and app (fio) issues IOs at full speed then this will flood the log file. Perhaps the code that Chris is working on (error counters) is a better place for logging problems with IOs.

blaisedias: Yes that would be more appropriate.
But at the moment we are logging errors for write and read ops as well, they occur at a higher frequency than flush ops.
So I think IMHO changing the error handling to use counts instead of logging a message done in a separate change and for all such error logging.

PR #292, opened 2020-06-25
moac: msn is out of date (CAS-327)

tiagolobocastro:
Makes use of the existing work queue to serialise node updates

Sometimes the watcher is out of date with k8s. A temporary fix is to
always bypass the cache and get the information we need directly from
k8s and update. We also update the watcher cache with the new value.
(I've raised cas-328 to track this issue)

Also added the corresponding watcher tests

### DIFF
--- a/csi/moac/node_operator.js
+++ b/csi/moac/node_operator.js
@@ -14,6 +14,7 @@ const yaml = require('js-yaml');
 const EventStream = require('./event_stream');
 const log = require('./logger').Logger('node-operator');
 const Watcher = require('./watcher');
+const Workq = require('./workq');
 
 const crdNode = yaml.safeLoad(
   fs.readFileSync(path.join(__dirname, '/crds/mayastornode.yaml'), 'utf8')
@@ -31,6 +32,7 @@ class NodeOperator {
     this.watcher = null; // k8s resource watcher for CSI nodes resource
     this.registry = null;
     this.namespace = namespace;
+    this.workq = new Workq(); // for serializing node operations
   }
 
   // Create node CRD if it doesn't exist and augment client object so that CRD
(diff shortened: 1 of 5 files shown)


csi/moac/node_operator.js
```diff
@@ -119,36 +121,42 @@ class NodeOperator {
...
+            log.error(`Failed to update node resource "${name}": ${err}`);
             return;
           }
         }
-        await this._updateStatus(name, ev.object.isSynced() ? 'online' : 'offline');
-      } else if (ev.eventType === 'del') {
-        await self._deleteResource(ev.object.name);
-      } else {
-        assert.strictEqual(ev.eventType, 'sync');
+      } else if (ev.eventType === 'new' && !k8sNode) {
```
paulyoong, line 145: You don't need the check for '!k8sNode' here because this is already checked by the 'if' on line 135.

tiagolobocastro: Done

csi/moac/watcher.js
```diff
@@ -201,6 +201,51 @@ class Watcher extends EventEmitter {
     }
   }
 
+  // Fetches the latest object(s) from k8s and updates the cache; then it
+  // returns the k8s object(s) from the cache or null if it does not exist.
+  async getRawBypass (name) {
+    try {
+      const getObj = await this.getEp(name).get();
```
jkryl, line 208: would be good to narrow down the try-catch to just this line. Otherwise you catch all unexpected and programming errors not related to the GET request that normally should terminate the program.

csi/moac/watcher.js
```diff
@@ -201,6 +201,51 @@ class Watcher extends EventEmitter {
...
+    } catch (err) {
+      // an exception is thrown if the resource does not exist
+      if (err.code === 404) {
+        const cachedObj = this.objects[name];
+        if (cachedObj) {
+          // the object does not exist so delete it
+          this._processEvent({
+            type: 'DELETED',
+            object: cachedObj
+          });
```
jkryl, line 239: I would not rather handle 404 case and would return whatever we have in the cache instead. The reason is that if we delete the object from cache, the MODIFIED and DELETED events could still come in at some point. What happens when watcher sees MODIFIED and there is no object in the cache? It will create it and emit event that there is a new object (even though it does not exist at that point in time - confusing all listeners. More safe seems to be to return stale copy of the object from the cache and let the update operation done by the caller simply fail. It will also simplify the code in this function.

csi/moac/watcher.js
tiagolobocastro: Yep done

csi/moac/watcher.js
tiagolobocastro: I was trying to "short-circuit" the events, but I see your point, let's let the events settle on their own on this one.

PR #296, opened 2020-06-26
CAS-206 changes to support generic URIs for manipulating bdevs

cjones1024:
Created a trait called StorageBdev which the various device types must now implement, in order to create and delete bdevs from a URI. I have refactored the existing code so as to make each of the various implementations as "similar" to each other as possible.
There is still some work to do, particularly for iSCSI and NVMF, but as things currently stand the behaviour for these device types should be precisely what it was beforehand.


mayastor/src/nexus_uri.rs
```diff
@@ -89,104 +89,17 @@ impl RpcErrorCode for BdevCreateDestroy {
     }
 }
 
-/// enum type of URL to args we currently support
-#[derive(Debug)]
-pub enum BdevType {
-    /// you should not be using this other than for testing
```
jonathan-teh, line 95: Could these comments be moved to StorageUri::parse()? It'd be a shame to lose them.

mayastor/src/bdev/storage.rs
```diff
@@ -0,0 +1,69 @@
...
+mod aio;
+mod bdev;
+mod iscsi;
+mod nvmf;
+mod uring;
+
+#[async_trait]
+/// Definition of trait that any supported
+/// device type is required to implement.
+pub trait StorageBdev {
```
gila, line 34: To be honest, I don't really like the bdev StorageBdev, there are no hard requirements for naming them either but -- typically, there are some things considered to be best practice.

A copy-paste from a thread that gives very good examples:

```
Sending values to other threads:

    Send - can be sent
    Sender - can send something
    SendSafe - is safe to send (same as 1.)
 ```

Given that would the trait: bdev::Create not do just fine? that you have to implement fn destroy() is not readable from the trait but the compiler will let you know, doing CreateDestroy might also work.

File: mayastor/src/bdev/storage.rs [hunk 3]
```diff
@@ -0,0 +1,69 @@
...
+    type Error;
+    fn get_name(&self) -> String;
```
gila, line 36: This should not be part of the trait,  what would it? Who cares how you destroy the bdev right?

mayastor/src/bdev/storage/iscsi.rs
```diff
@@ -0,0 +1,167 @@
...
+    ffihelper::{cb_arg, done_errno_cb, errno_result_from_i32, ErrnoResult},
+    nexus_uri::{self, BdevCreateDestroy},
+};
+
+use super::StorageBdev;
+
+const ISCSI_IQN_PREFIX: &str = "iqn.1980-05.mayastor";
+
+#[derive(Debug)]
+pub struct Iscsi {
```
gila, line 28: Back to the naming thing, it was always iSCSI if I believe but the namespace was different, right?

it used to be Bdev::Iscsi or Bdev::Aio -- now its bdev::storage::Iscsi

mayastor/src/bdev/storage/iscsi.rs
```diff
@@ -0,0 +1,167 @@
...
+    bdev::util::uri,
+    core::Bdev,
+    ffihelper::{cb_arg, done_errno_cb, errno_result_from_i32, ErrnoResult},
+    nexus_uri::{self, BdevCreateDestroy},
+};
+
+use super::StorageBdev;
+
+const ISCSI_IQN_PREFIX: &str = "iqn.1980-05.mayastor";
+
```
gila, line 26: giving it this name makes it much easier to spot :+1:

mayastor/src/bdev/storage/bdev.rs
```diff
@@ -0,0 +1,56 @@
+use std::{collections::HashMap, convert::TryFrom};
```
gila, line 1: I'm not sure what the intent of this is i think its the bdev:/// thing right? There is an issue for that already I beleive, where the suggestion is to rename that to local:/// or loopback:///

mayastor/src/bdev/storage.rs
```diff
@@ -0,0 +1,69 @@
...
+    async fn create(&self) -> Result<String, Self::Error>;
+    async fn destroy(self: Box<Self>) -> Result<(), Self::Error>;
```
gila, line 38: I'm thinking about self: Box<Self> here.. If all want to destroy the bdev, do really need to go through the trouble of parsing etc? Could not simply destroy it? :thinking:

mayastor/src/nexus_uri.rs
cjones1024: Good Idea. Will do.

mayastor/src/bdev/storage.rs
cjones1024: We currently use different spdk function calls to delete the bdev based on its type - which we infer from the URI scheme.
So we need to do at least some parsing in order to determine the scheme.

mayastor/src/bdev/storage/bdev.rs
cjones1024: Yes this is exactly that. The name of the scheme can be easily changed from bdev to local if that is preferable.

cjones1024: This is in an actual fact CAS-1.  Will effect this change by replacing bdev:// with loopback://

mayastor/src/bdev/storage.rs
```diff
@@ -0,0 +1,65 @@
+//! Definition of Trait that defines how to create
+//! and destroy bdevs from an appropriate device URI.
+//! Note also that the create() and destroy() methods are async.
+//!
+//! Support for any new device type that we may wish to add effectively
+//! amounts to providing an appropriate implementation of this trait.
+//!
+//! Creating a bdev for any supported device type should now be a simple as
+//! ```
```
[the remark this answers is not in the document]
cjones1024: Done

cjones1024:
> 
> 
> Created a trait called StorageBdev which the various device types must now implement, in order to create and delete bdevs from a URI. I have refactored the existing code so as to make each of the various implementations as "similar" to each other as possible.
> 

Have renamed the trait to "BdevCreateDestroy", and the implementing modules are now rooted at dev rather than storage.
The "object" implementing the parse() function is now just called Uri rather than StorageUri.
So we have:
- crate:bdev::Uri
- crate::bdev::BdevCreateDestroy
- crate::bdev::dev::aio::Aio
- crate::bdev::dev::iscsi::Iscsi
  etc.
and you you create a bdev by calling bdev::Uri::parse("uri")?.create().await?;

PR #298, opened 2020-06-29
Nix develop

gila:
Yes, it worked! (the PR that is)

A lot of the nix file changes are mostly reordering of some bits, nothing too extreme really. Except for the fact that now we have a .json file that contains the URI's to our sources. 

Other than that:
 - ensure we compile to Nehalem all the way down
- disable/remove
- keep CI simple: everything runs in nix-shell, no caches no containers to push before making changes etc
- reworked the bindings a little bit which is needed for the upcoming upgrade

PS Clippy might fail on an issue already fixed by, that test case will be ignored (i.e push regardless)


.github/workflows/build.yaml
```diff
@@ -6,36 +6,23 @@ on:
...
-  Build_and_run_cargo:
+  nix-shell:
     name: Build and run cargo tests
     runs-on: self-hosted
-    timeout-minutes: 20
+    timeout-minutes: 30
     container:
-      image: docker.io/mayadata/ms-buildenv:latest
-      options: --privileged -v /dev:/dev -v /bin:/host/bin -v /lib/modules:/lib/modules
+      image: docker.io/mayadata/ms-buildenv:gila
```
tiagolobocastro, line 14: Is this the correct image? (thought we'd be using the base nix image now)

.github/workflows/pr.yaml
```diff
@@ -1,22 +1,17 @@
...
 on:
   pull_request:
+  push:
 jobs:
   Clippy:
     name: Clippy
     runs-on: ubuntu-latest
     container:
-      image: docker.io/mayadata/ms-buildenv:latest
+      image: docker.io/mayadata/ms-buildenv:gila
```
tiagolobocastro, line 10: same question as the previous and also on Js checks, should we update that image too?

.github/workflows/build.yaml
gila: I will change it at the end yes :)

nix/pkgs/mayastor/default.nix
```diff
@@ -43,91 +32,54 @@ rec {
             lib.hasPrefix (toString (src + "/${allowedPrefix}")) path)
           allowedPrefixes)
       src;
+in
+rustPlatform.buildRustPackage rec {
+  name = "mayastor";
+  cargoSha256 = "1g2yq30p5z15qr94bl22bhh4p6idm5jfmh6kh88w3ixlacjyk8pf";
+  #cargoSha256 = "0000000000000000000000000000000000000000000000000000";
```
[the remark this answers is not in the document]
gila: When you need to recalculate the sha256 you have to use all zero's. so you can swap them out easily

scripts/node-test.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env sh
+set -euxo pipefail
+export PATH=$PATH:${HOME}/.cargo/bin
+cargo build --all
+( cd mayastor-test && npm install && ./node_modules/mocha/bin/mocha test_cli.js )
```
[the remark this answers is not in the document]
gila: Shoot! you are right!

scripts/test.sh
```diff
@@ -1,12 +1,11 @@
-#!/usr/bin/env bash
```
jkryl, line 1: it looks like we can remove this file now as we have the test scripts under scripts/ doing the same ..

shell.nix
```diff
@@ -1,53 +1,58 @@
...
+  channel = import ./nix/lib/rust.nix { inherit sources; };
 in
 mkShell {
+
   # fortify does not work with -O0 which is used by spdk when --enable-debug
   hardeningDisable = [ "fortify" ];
-
   buildInputs = [
-    figlet
+    #    nats-server
```
jkryl, line 21: nats-server is needed by the tests. Was it removed by accident?

scripts/test.sh
gila: here all tests are run, and i like to run it before I do PRs etc. Unless you have a strong objection i'd like to keep it.

shell.nix
gila: Oh yes -- restored, tnx!

.github/workflows/pr.yaml
```diff
@@ -1,22 +1,17 @@
 name: "Lint and style Checks"
 on:
   pull_request:
+  push:
```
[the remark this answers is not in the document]
gila: Yes indeed, I did not want to open/close the PR -- removed now.

PR #299, opened 2020-06-29
Integrate rust client for NATS with mayastor/SPDK

jkryl:
New mayastor arg for NATS server endpoint,
registration & deregistration code. This has far implications for moac
and mayastor code as now the mayastor is independent on the CSI plugin
that was up until now still needed for mayastor registration. Further
code changes and removal of obsolete code in moac and mayastor will follow
once this PR is merged.

Resolves: CAS-288 and CAS-289


mayastor/src/nats.rs
```diff
@@ -0,0 +1,284 @@
...
+            Some(client) => client
+                .publish("register", serde_json::to_vec(&payload).unwrap())
+                .await
+                .map_err(|cause| Error::QueueRegister {
+                    cause,
+                })?,
+            None => return Err(Error::NotStarted {}),
+        }
+        // Note that the message was only queued and we don't know if it was
+        // really send to the NATS server (limitation of the nats lib)
```
tiagolobocastro, line 225: s/send/sent?

mayastor/src/nats.rs
```diff
@@ -0,0 +1,284 @@
...
+use tokio::{net::lookup_host, time::delay_for};
+use tokio_nats::{
+    connect,
+    Error as TokioNatsError,
+    NatsClient,
+    NatsConfigBuilder,
+};
+
+/// Mayastor sends registration messages in this interval (kind of heart-beat)
+const HB_INTERVAL: u64 = 10;
```
tiagolobocastro, line 28: what does this interval represent: s, m, h? Could we define it as a time::Duration?

mayastor/src/nats.rs
```diff
@@ -0,0 +1,284 @@
...
+                .publish("register", serde_json::to_vec(&payload).unwrap())
+                .await
+                .map_err(|cause| Error::QueueRegister {
+                    cause,
+                })?,
+            None => return Err(Error::NotStarted {}),
+        }
+        // Note that the message was only queued and we don't know if it was
+        // really send to the NATS server (limitation of the nats lib)
+        debug!(
```
tiagolobocastro, line 226: Is this printed every HB_INTERVAL? If so should it be a trace rather than a debug?

mayastor/src/nats.rs
```diff
@@ -0,0 +1,284 @@
...
+            if let Err(err) = self.register().await {
+                error!("Registration failed: {:?}", err);
+            };
+            let _res = select! {
+                () = delay_for(Duration::from_secs(self.hb_interval)).fuse() => (),
+                msg = receiver.next() => {
+                    match msg {
+                        Some(_) => warn!("Messages have not been implemented yet"),
+                        None => {
+                            debug!("Terminating the NATS client");
```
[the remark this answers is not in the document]
jkryl: done. thanks

mayastor/src/nats.rs
jkryl: fixed. thanks

mayastor/src/nats.rs
jkryl: Good point! I changed the code.

mayastor/src/nats.rs
jkryl: yes, it is printed every HB_INTERVAL that is in seconds and by default it is set to 10s. That's not very often compared to trace messages in IO path, but you are right that it is repeating information.

PR #301, opened 2020-07-01
moac: unpublish old volume (CAS-326)

tiagolobocastro:
After reinstall k8s keeps calling us to unpublish a volume which does
not exist anymore. To stop this forever loop return ok instead of
returning not found if the volume does not exist...


tiagolobocastro:
> I can see that it fixes leaking PVs on k8s.
> But still does not feel entirely correct.
> On balance this improves operation on k8s so I am approving.

It doesn't fix the leaks I don't think, it just stops k8s from calling us on a loop to unpublish a volume which does not exist.

PR #304, opened 2020-07-02
Fix CSI plugin crash. CAS-335

blaisedias:
Fail gracefully if an iscsadm executable
is not found.
Refactor iscsiutil and nvmfutil to use Error to
communicate failures instead of error strings.
This allows us to handle the lack of an iscsiadm
executable in a more rustacean manner.
Cleanup: remove nvmeadm binary.


csi/src/node/iscsiutil.rs
```diff
@@ -1,6 +1,9 @@
 use once_cell::sync::Lazy;
 use std::{env, path::Path, process::Command, thread, time};
 
+use crate::CSIError;
+use failure::Error;
+
 // The iscsiadm executable invoked is dependent on the environment.
 // For the container we set it using and environment variable,
```
paulyoong, line 8: I know this isn't part of your change but "and" should be "an".

csi/src/server.rs
```diff
@@ -45,6 +45,26 @@ mod identity;
 mod mount;
 mod node;
 
+#[macro_use]
+extern crate failure;
+#[derive(Debug, Fail)]
+pub enum CSIError {
+    #[fail(display = "IO error: {}", error)]
+    Iscsiadm { error: String },
```
paulyoong, line 53: If it's an Iscsiadm error, is it correct to be displaying "IO error"?

csi/src/node/iscsiutil.rs
blaisedias: fixed https://github.com/openebs/Mayastor/pull/304/commits/a8208afe6683479a8f40af63527b01a761ba19ef

csi/src/server.rs
blaisedias: fixed https://github.com/openebs/Mayastor/pull/304/commits/a8208afe6683479a8f40af63527b01a761ba19ef

PR #305, opened 2020-07-02
Tweak csi mocha test unknown device

blaisedias:
Replace hardcoded unknown uuid uri, with one derived
from an existing nexus.


gila:
@blaisedias you would have to resolve the conflicts before we can merge this.

PR #307, opened 2020-07-03
Correctly disconnect nvme devices CAS-337

blaisedias:
Use the correct nqn prefix
Handle the return value of nvmeadm::nvmf_discovery::disconnect

### DIFF
--- a/csi/src/node/nvmfutil.rs
+++ b/csi/src/node/nvmfutil.rs
@@ -102,7 +102,7 @@ fn nvmeadm_attach_disk(
             if let Some(errcode) = ioerr.raw_os_error() {
                 if errcode == 114 {
                     in_progress = true;
-                    debug!("nvmeadm operation already in progress for {}", uri);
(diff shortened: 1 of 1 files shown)


csi/src/node/nvmfutil.rs
```diff
@@ -127,10 +129,25 @@ fn nvmeadm_attach_disk(
...
+            }
+            1 => {
+                trace!("nvmf disconnected {}", nqn);
+                Ok(())
+            }
+            _ => {
+                debug!(
+                    "Warning: nvmf disconnect {} disconnected {} devices.",
+                    nqn, n
+                );
```
jkryl, line 147: it's a debug message beginning with "warning:". Perhaps warn! should be used instead (and `warning:` removed)?

blaisedias: https://github.com/openebs/Mayastor/pull/307/commits/0cc0a067896409e67197f6a1035499c0c68e4558

PR #313, opened 2020-07-07
bdev: add support for malloc_bdev

gila:
Note that memory is allocated from the hugepages and not the
from the heap


mayastor/src/bdev/dev/malloc.rs
```diff
@@ -0,0 +1,171 @@
...
+            };
+
+        if size != 0 && num_blocks != 0 {
+            return Err(NexusBdevError::UriInvalid {
+                uri: uri.to_string(),
+                message: "conflicting parameters num_blocks and size_mb are mutually exclusive"
+                    .to_string(),
+            });
+        }
+
```
cjones1024, line 95: Just a minor note.  the Implementations for the other types check and issue a warning if there are any unused parameters in the URI

PR #315, opened 2020-07-07
Auto assign reviewers

gila:


.github/auto_assign.yml
```diff
@@ -0,0 +1,24 @@
+# Set to true to add reviewers to pull requests
+addReviewers: true
+
+# Set to true to add assignees to pull requests
+addAssignees: true
```
tjoshum, line 5: What do we want assignee to mean in the context of PRs? I think we're covered with author and reviewers

gila: I'm not sure @tjoshum. I've never used it, it might be that it is gone before we even know it =)

PR #316, opened 2020-07-07
CAS-341: replica not replaced when node leaves

tiagolobocastro:
When a node leaves it should at least unbind its existing resources.
todo: when a node is removed gracefully should we also destroy the
pools?

Also we now replace children which have no underlying replica with new
ones during the volume fsa irregardless of the online count which seems
reasonable.

### DIFF
--- a/csi/moac/node.js
+++ b/csi/moac/node.js
@@ -88,6 +88,12 @@ class Node extends EventEmitter {
     this._offline();
   }
 
+  unbind () {
+    // todo: on user explicit removal should we destroy the pools as well?
+    this.pools.forEach((pool) => pool.unbind());
+    this.nexus.forEach((nexus) => nexus.unbind());
+  }
+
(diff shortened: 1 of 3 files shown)


csi/moac/node.js
```diff
@@ -88,6 +88,12 @@ class Node extends EventEmitter {
     this._offline();
   }
 
+  unbind () {
+    // todo: on graceful removal do we need to destroy the pools as well?
+    this.pools.forEach((pool) => pool.unbind());
```
gila, line 93: IMHO, no never. Deleting data is not something we should be doing. Who knows what (a) type of data it is and (b) what else the devices might be used for, right?

tiagolobocastro: Yeah we should only delete user data when we're really sure it's safe to do so. When a user removes a node explicitly (better wording than gracefully) is a potential candidate for it.

PR #320, opened 2020-07-09
CSI attach idempotency. CAS-340

blaisedias:
Make isci attach idempotent:
If a session for the target portal and iqn exists,
skip to just waiting for the device path to exist.

Refine the waiting for device path code,
1) specify granularity of sleeps,
2) reduce the timeout from 10 seconds to 1 second,
on the grounds that large timeouts can mask issues.

### DIFF
--- a/csi/src/node/iscsiutil.rs
+++ b/csi/src/node/iscsiutil.rs
@@ -41,16 +41,14 @@ fn get_iscsiadm() -> Result<&'static str, Error> {
 
 fn wait_for_path_to_exist(
     devpath: &str,
-    sleepmillis: u64,
-    millis: u64,
+    timeout: time::Duration,
+    max_retries: u32,
 ) -> bool {
-    let millisecond = time::Duration::from_millis(sleepmillis);
-    let max_retries = (millis + sleepmillis - 1) / sleepmillis;
     let device_path = Path::new(devpath);
-    let mut retries: u64 = 0;
+    let mut retries: u32 = 0;
     let now = time::Instant::now();
     while !device_path.exists() && retries < max_retries {
-        thread::sleep(millisecond);
+        thread::sleep(timeout);
         retries += 1;
     }
     trace!(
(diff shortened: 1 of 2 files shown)


csi/src/node/iscsiutil.rs
```diff
@@ -39,13 +39,18 @@ fn get_iscsiadm() -> Result<&'static str, Error> {
     }
 }
 
-fn wait_for_path_to_exist(devpath: &str, max_retries: i32) -> bool {
-    let second = time::Duration::from_millis(1000);
+fn wait_for_path_to_exist(
+    devpath: &str,
+    sleepmillis: u64,
+    millis: u64,
```
gila, line 45: Why not something like this:

```
fn wait_for_for_path(devpath: &str, timeout: Duration, retires: u32) {... } 
```

csi/src/node/iscsiutil.rs
```diff
@@ -79,67 +84,95 @@ fn attach_disk(
...
+        }
     }
 
-    if wait_for_path_to_exist(device_path.as_str(), 10) {
+    let millis: u64 = 1000;
+    if wait_for_path_to_exist(device_path.as_str(), 10, millis) {
         trace!("{} path exists!", device_path)
     } else {
-        trace!("{} path does not exist after 10s!", device_path);
+        trace!(
```
gila, line 169: So let us assume we timeout, do we clean up the session and the discovery entry?

blaisedias: No, that would be detrimental, I think. Because if the connection didn't succeed within an arbitrary time limit, it unlikely to succeed if everything is torn down and setup again with the same time limit.
The objective is to allow the connection process to proceed in parallel so that the next time attach is called it checks on the session that may have been initiated previously, in effect allowing the iscsi connection to proceed unimpeded across that time.

csi/src/node/iscsiutil.rs
blaisedias: Done. https://github.com/openebs/Mayastor/pull/320/commits/d9ab94c314e28a79663c8d9be0921f8e12834a77

csi/src/node/iscsiutil.rs
```diff
@@ -79,67 +82,97 @@ fn attach_disk(
...
-
-    // If the device path exists then a previous invocation of this
-    // method has succeeded.
-    if wait_for_path_to_exist(device_path.as_str(), 1) {
-        trace!("path already exists!");
-        return Ok(iscsi_realpath(device_path));
-    }
+    static RE_SESSION: Lazy<regex::Regex> = Lazy::new(|| {
+        regex::Regex::new(
+            r"\s*(?P<tp>\d+.\d+.\d+.\d+:\d+),0\s*(?P<iqn>iqn.2019-05.io.openebs:nexus-.*)\s"
```
chriswldenyer, line 87: Is there any way we can have this iqn defined in one place?

blaisedias: Thats the subject for another change, it's not just the iqn prefix but also the nqn prefix.

PR #322, opened 2020-07-13
Move rbac rules for moac to a separate yaml file

jkryl:
Depending on exact timing we could see that certain operations in moac
start to fail when doing:

  kubectl delete -f moac-deployment.yaml

That is because rbac rules for moac have already been destroyed but moac
still has work to do in the short time window before k8s destroys it too.
Now it is possible to enforce correct order when deleting moac by running:

  kubectl delete -f moac-deployment.yaml
  kubectl delete -f moac-rbac.yaml


jkryl:
you are right, it should be :-D I will update the message before the merge. thanks

jkryl:
I have updated the commit message

PR #323, opened 2020-07-13
moac: pool creation fails when storage node is imported (CAS-336)

jkryl:
The problem was that the node was marked as online only after calling
various merge object functions, so if those functions wanted to call
RPC method on the node, it failed.


csi/moac/test/node_test.js
```diff
@@ -12,7 +12,7 @@ const { MayastorServer } = require('./mayastor_mock');
 const enums = require('./grpc_enums');
 
 const UUID = 'ba5e39e9-0c0e-4973-8a3a-0dccada09cbb';
-const EGRESS_ENDPOINT = '127.0.0.1:12345';
```
gila, line 15: meh! thanks, for renaming this @jkryl of all bad names, this one is the worst :see_no_evil:

PR #324, opened 2020-07-13
CAS-171 Excessive errors take a child out of the IO path

chriswldenyer:
When an IO is added to error store, the error store is queried
to determine whether the number of errors meets the criterion that
should cause the child to be marked as faulted. If so, the child is
faulted.
The error store can also determine the number of repeated attempts
to send the same IO that have failed. This enables a per-IO error
count to be queried.
Configuration options exist to enable the error store and to enable
faulting the child due to number of retries or total number of IO
errors, and dependent on the age of the error.
Existing unit test to verify new behaviour and added unit tests for
child removal due to total error count and number of repeated errors.


mayastor/tests/error_fault_child_total.rs
```diff
@@ -0,0 +1,192 @@
...
+    common::delete_file(&[DISKNAME2.to_string()]);
+    common::delete_file(&[YAML_CONFIG_FILE.to_string()]);
+}
+
+fn check_nexus_state_is(expected_status: NexusStatus) {
+    let nexus = nexus_lookup(ERROR_COUNT_TEST_NEXUS).unwrap();
+    assert_eq!(nexus.status(), expected_status);
+}
+
+async fn create_error_bdev() {
```
jonathan-teh, line 103: There is a common version of this in common/error_bdev.rs.

mayastor/tests/error_fault_child_total.rs
```diff
@@ -0,0 +1,192 @@
...
+    let err_bdev_name_str =
+        CString::new(EE_ERROR_DEVICE).expect("Failed to create name string");
+    let raw = err_bdev_name_str.into_raw();
+
+    unsafe {
+        retval = spdk_vbdev_error_inject_error(raw, op, mode, count);
+    }
+    assert_eq!(retval, 0);
+}
+fn reactor_run_millis(milliseconds: u64) {
```
jonathan-teh, line 185: Nit: missing blank line.

mayastor/tests/error_fault_child_retries.rs
```diff
@@ -0,0 +1,193 @@
...
+    let err_bdev_name_str =
+        CString::new(EE_ERROR_DEVICE).expect("Failed to create name string");
+    let raw = err_bdev_name_str.into_raw();
+
+    unsafe {
+        retval = spdk_vbdev_error_inject_error(raw, op, mode, count);
+    }
+    assert_eq!(retval, 0);
+}
+fn reactor_run_millis(milliseconds: u64) {
```
jonathan-teh, line 186: Nit: blank line.

mayastor/src/bdev/nexus/nexus_child_error_store.rs
```diff
@@ -48,6 +51,11 @@ pub struct NexusErrStore {
     records: Vec<NexusChildErrorRecord>,
 }
 
+pub enum NexusErrStoreQuery {
```
gila, line 54: as all the code is already namespaced (i.e bdev::nexus::nexus_child_error_store) there is no need to make that:

bdev::nexus::nexus_child_error_store::NexusErrStoreQuery -- that is all very, very verbose, would you agree? Ideally it would not have been named nexus_child_error_store either but thats not part of this change so thats OK.

mayastor/src/bdev/nexus/nexus_child_error_store.rs
```diff
@@ -48,6 +51,11 @@ pub struct NexusErrStore {
...
+    Total,
```
gila, line 55: please document the members and the purpose

mayastor/src/bdev/nexus/nexus_child_error_store.rs
```diff
@@ -105,61 +113,138 @@ impl NexusErrStore {
         self.next_record_index =
              // consider using a mask if we restrict the size to a power-of-2
             (self.next_record_index + 1) % self.records.len();
-        // debug!("added record - buffer is {}", &self);
     }
 
-    pub fn query(
+    #[inline]
+    fn is_wanted_record(
```
gila, line 119: would "find" be better? or maybe, has_error(..) ?

mayastor/src/bdev/nexus/nexus_child_error_store.rs
```diff
@@ -105,61 +113,138 @@ impl NexusErrStore {
...
+        record: &NexusChildErrorRecord,
+        io_op_flags: u32,
+        io_error_flags: u32,
+        target_timestamp: Option<Instant>,
+    ) -> bool {
+        let found_op = match record.io_op {
+            io_type::READ => (io_op_flags & NexusErrStore::READ_FLAG) != 0,
```
gila, line 126: this can be refined using match guards: https://doc.rust-lang.org/reference/expressions/match-expr.html#match-guards

mayastor/src/bdev/nexus/nexus_child_error_store.rs
```diff
@@ -105,61 +113,138 @@ impl NexusErrStore {
...
+            io_type::WRITE => (io_op_flags & NexusErrStore::WRITE_FLAG) != 0,
+            io_type::UNMAP => (io_op_flags & NexusErrStore::UNMAP_FLAG) != 0,
+            io_type::FLUSH => (io_op_flags & NexusErrStore::FLUSH_FLAG) != 0,
+            io_type::RESET => (io_op_flags & NexusErrStore::RESET_FLAG) != 0,
+            _ => false,
```
gila, line 131: if you return here, the if condition on line 134 is not needed

mayastor/src/bdev/nexus/nexus_child_error_store.rs
```diff
@@ -105,61 +113,138 @@ impl NexusErrStore {
...
+        };
+
+        if found_op {
+            let found_err = match record.io_error {
+                io_status::FAILED => {
+                    (io_error_flags & NexusErrStore::IO_FAILED_FLAG) != 0
+                }
+                _ => false,
```
gila, line 139: the same here, an early return would make the code a bit cleaner?

mayastor/src/bdev/nexus/nexus_child_error_store.rs
```diff
@@ -105,61 +113,138 @@ impl NexusErrStore {
...
-                _ => false,
-            };
+        }
+        err_count
+    }
 
-            let found_err = match self.records[idx].io_error {
-                io_status::FAILED => {
-                    (io_error_flags & NexusErrStore::IO_FAILED_FLAG) != 0
+    #[inline]
```
gila, line 181: I'm not quite sure what we are trying to do here...  If we want to count the errors for a particular type, this seems rather expensive.

mayastor/src/subsys/opts.rs
```diff
@@ -482,15 +482,36 @@ impl GetOpts for IscsiTgtOpts {
 pub struct ErrStoreOpts {
     /// ring buffer size
     pub err_store_size: usize,
+
     /// NexusErrStore enabled
     pub enable_err_store: bool,
+
+    /// whether to fault the child due to failed re-tried IOs
+    pub fault_child_on_retry_count: bool,
```
gila, line 490: I think it would be better to name this a particular action we can do, i.e retries_exceed = "action" where action is ingore, reset, fault, or whatever.

mayastor/src/subsys/opts.rs
```diff
@@ -482,15 +482,36 @@ impl GetOpts for IscsiTgtOpts {
...
+
+    /// the maximum number of errors for the same retried IO
+    pub max_retry_errors: u32,
```
gila, line 493: max_retries?

mayastor/src/subsys/opts.rs
```diff
@@ -482,15 +482,36 @@ impl GetOpts for IscsiTgtOpts {
...
+
+    /// whether to fault the child due to the total number of failed IOs
+    pub fault_child_on_error_count: bool,
```
gila, line 496: I'm not sure we want to make our lives so hard here.. Whenever an IO failed WITH retires or not, it a failed IO that has been retried. 

Put differently, an IO to LBA0 that fails, retried, fails, retried OK, is still 2 IO errors. right? The Errorstore contains more detail as to what why how *exactly* but that's more for diagnosing the core problem?

To keep it simple, I would suggest to start out simple, getting the right behavior on failed IO's to a child, would be WIN-WIN, then we can start to think about special cases

mayastor/src/subsys/opts.rs
```diff
@@ -482,15 +482,36 @@ impl GetOpts for IscsiTgtOpts {
...
+
+    /// the maximum number of errors in total
+    pub max_total_errors: u32,
+
+    /// errors older than this are ignored
+    pub max_error_age_ns: u64,
```
gila, line 502: retention?

mayastor/src/subsys/opts.rs
```diff
@@ -482,15 +482,36 @@ impl GetOpts for IscsiTgtOpts {
...
 
 impl Default for ErrStoreOpts {
     fn default() -> Self {
         Self {
             err_store_size: 256,
-            enable_err_store: true,
+            enable_err_store: false,
+            fault_child_on_retry_count: false,
+            fault_child_on_error_count: false,
+            max_retry_errors: 0,
```
gila, line 512: is it a good idea to set these to 0? one retry and your out?

mayastor/tests/error_count.rs
```diff
@@ -35,12 +36,19 @@ static ERROR_DEVICE: &str = "error_device";
...
+
 #[test]
 fn nexus_error_count_test() {
     common::truncate_file(DISKNAME1, 64 * 1024);
     common::truncate_file(DISKNAME2, 64 * 1024);
 
-    test_init!();
+    let mut config = Config::default();
+    config.err_store_opts.enable_err_store = true;
+    config.err_store_opts.err_store_size = 256;
```
gila, line 48: if not part of this commit then please ignore, but "num_entries" would be a more descriptive name.

mayastor/tests/error_count.rs
```diff
@@ -35,12 +36,19 @@ static ERROR_DEVICE: &str = "error_device";
...
+    config.write(YAML_CONFIG_FILE).unwrap();
+
+    test_init!(Some(YAML_CONFIG_FILE.to_string()));
```
gila, line 51: Its not some to see Some(macro!()) perhaps, give the testing, you want to unwrap in the macro?

mayastor/src/subsys/opts.rs
chriswldenyer: My reasoning was that we would want to distinguish between, say, 10 different IOs which all fail on the first attempt then succeed, and one IO that fails 10 times in a row. It is easy to imagine wanting different thresholds for the two cases.
But I can take out the repeated IO error handling for now.

mayastor/src/bdev/nexus/nexus_child_error_store.rs
chriswldenyer: I tend to prefer verbose names, but if the house style is to keep them short, I'll do so.

mayastor/tests/error_fault_child_retries.rs
```diff
@@ -0,0 +1,193 @@
...
+    common::delete_file(&[DISKNAME2.to_string()]);
+    common::delete_file(&[YAML_CONFIG_FILE.to_string()]);
+}
+
+fn check_nexus_state_is(expected_status: NexusStatus) {
+    let nexus = nexus_lookup(ERROR_COUNT_TEST_NEXUS).unwrap();
+    assert_eq!(nexus.status(), expected_status);
+}
+
+async fn create_error_bdev() {
```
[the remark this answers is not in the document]
chriswldenyer: No, Jonathan pointed that out as well. I am changing it to use the common function.

mayastor/tests/error_fault_child_retries.rs
```diff
@@ -0,0 +1,193 @@
...
+        Ok(_) => {
+            assert_eq!(succeed, true);
+        }
+        Err(_) => {
+            assert_eq!(succeed, false);
+        }
+    };
+}
+
+async fn inject_error(op: u32, mode: u32, count: u32) {
```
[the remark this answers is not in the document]
chriswldenyer: Also being fixed as above

mayastor/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -259,6 +260,31 @@ impl Nexus {
...
+            return Err(Error::DestroyLastChild {
+                name: self.name.clone(),
+                child: name.to_owned(),
+            });
+        }
+        if let Some(child) = self.children.iter_mut().find(|c| c.name == name) {
+            if child.status() != ChildStatus::Faulted {
+                child.fault();
+                self.reconfigure(DREvent::ChildFault).await;
+                info!("Fault and reconfigure completed");
```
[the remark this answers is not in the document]
chriswldenyer: Ok, will remove this.

mayastor/src/bdev/nexus/nexus_child_error_store.rs
```diff
@@ -303,4 +424,22 @@ impl Nexus {
             Ok(None)
         }
     }
+
+    // returns false if the child is deemed faulted
```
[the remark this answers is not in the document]
chriswldenyer: Am expanding the comment.

mayastor/src/bdev/nexus/nexus_child_error_store.rs
```diff
@@ -241,23 +327,56 @@ impl Nexus {
...
+                            if nexus.fault_child(&child_name).await.is_err() {
+                                error!(
+                                    "Failed to fault the child {}",
+                                    child_name,
+                                );
+                            }
+                        }
+                    } else {
+                        error!(
+                            "Failed to record error - child has no error store"
```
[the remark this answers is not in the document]
chriswldenyer: Added child name to the message.

mayastor/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -259,6 +260,31 @@ impl Nexus {
         Ok(self.status())
     }
 
+    /// fault a child device and reconfigure the IO channels
+    pub async fn fault_child(&mut self, name: &str) -> Result<(), Error> {
+        trace!("{}: fault child request for {}", self.name, name);
+
+        if self.child_count < 2 {
+            return Err(Error::DestroyLastChild {
```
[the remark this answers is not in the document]
chriswldenyer: No, but following discussion, will rename error and change error message.

mayastor/tests/error_fault_child_total.rs
chriswldenyer: Now using the common version of this function (and inject_error()) in this and error_fault_child_retries.rs

mayastor/tests/error_fault_child_total.rs
chriswldenyer: fixed

mayastor/tests/error_fault_child_retries.rs
chriswldenyer: fixed

mayastor/src/subsys/opts.rs
chriswldenyer: changed to retention_ns

mayastor/src/bdev/nexus/nexus_child_error_store.rs
chriswldenyer: Enum purpose and values now described in comments.

mayastor/tests/error_fault_child_total.rs
```diff
@@ -0,0 +1,192 @@
+extern crate log;
```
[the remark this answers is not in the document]
chriswldenyer: Removing retry-count functionality means only one file is needed, so this is no longer an issue (for now).

mayastor/src/subsys/opts.rs
chriswldenyer: I'd expect this to be set to some non-zero value if the feature is enabled but I didn't want to assign an arbitrary number. 
On the other hand, a common retry value in SPDK is 5, which seems reasonable, so I'll set that as the default.

mayastor/src/bdev/nexus/nexus_child_error_store.rs
chriswldenyer: Refactored using match guards.

mayastor/src/bdev/nexus/nexus_child_error_store.rs
chriswldenyer: Added early return.

mayastor/src/bdev/nexus/nexus_child_error_store.rs
chriswldenyer: added early return

mayastor/src/bdev/nexus/nexus_child_error_store.rs
chriswldenyer: Ok, I'll go with has_error().

mayastor/src/subsys/opts.rs
chriswldenyer: This option is now not needed and removed.
But in the same vein renamed the other option max_total_errors to max_errors.

mayastor/tests/error_count.rs
chriswldenyer: It wasn't part of this commit. But I'll consider renaming it in a future iteration.

mayastor/src/subsys/opts.rs
chriswldenyer: Changed variable from a bool to an enum with currently 2 values, "fault" and "ignore"

mayastor/tests/error_count.rs
chriswldenyer: Moved the string handling into the macro.

mayastor/src/bdev/nexus/nexus_child_error_store.rs
chriswldenyer: count_repeats() now taken out of this PR as a result of taking out the repeated IO error handling, as suggested in this review.

PR #325, opened 2020-07-14
Nix test basic

gila:


nix/test/basic/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+TODO:
+[ ] We should write some high-level python code that we can reuse in the tests.
+[ ] We need to figure out how many tests we want to run within a single run
+[ ] We will need more gRPC methods to build more sophisticated tests
+[ ] A test where we set up and deploy, k8s first, and then install mayastor would be very nice but for sure should be done last.
+
+
+
+Note, this will only work after a "cargo build --all" as it the
```
blaisedias, line 23: missing word 'builds'

nix/test/basic/default.nix
```diff
@@ -0,0 +1,15 @@
+#
+# TODO: not sure if we need to import the sources
+#
+let
+  sources = import ./../../../nix/sources.nix;
```
blaisedias, line 5: why leading './..' instead of just '../'

nix/test/basic/fio_nvme_basic.nix
```diff
@@ -0,0 +1,109 @@
+{ pkgs, lib, ... }:
+let
+  targetIp = "192.168.0.1";
```
blaisedias, line 3: Doesn't a fixed ip address limit where this test can be deployed?

nix/test/basic/fio_nvme_basic.nix
```diff
@@ -0,0 +1,109 @@
...
+  initiatorIp = "192.168.0.2";
```
blaisedias, line 4: Doesn't a fixed ip address limit where this test can be deployed?

nix/test/basic/README.md
gila: fixed, thanks.

nix/test/basic/default.nix
gila: that is nix-ism. All imports that refer to a file always start with "./" regardless of the path itself.

nix/test/basic/fio_nvme_basic.nix
gila: No, the network is a VLAN -- and is isolated from the host. You can not reach this network with your own host/machine.

nix/test/basic/fio_nvme_basic.nix
gila: Ideally, we don't have fixed IPs however, mayastor needs to have an IP in order to run and we don't have code that can "auto-discover" the IPs (as they are passed in as ENV through k8s).

PR #326, opened 2020-07-14
ci: Add the nix tests to CI

tjoshum:


.github/workflows/nix-tests.yaml
```diff
@@ -0,0 +1,20 @@
...
+  main:
+    name: Run Nix Tests
+    runs-on: self-hosted
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          path: repo-under-test
+      - run: modprobe kvm_intel nested=1 # Could do this once persistently on the runner.
+      - run: echo "::set-env name=NIX_PATH::/home/gila/.nix-defexpr/channels"
```
gila, line 18: 🧐

tjoshum: That's just how the runners are setup

gila: Yeah, I figured as much -- thx for the confirmation!

PR #327, opened 2020-07-14
remove jsonrpc methods defined within mayastor (CAS-261)

jkryl:
```
    That's not the end of jsonrpc in mayastor. We remove just the methods
    that are already provided by gRPC API to avoid duplication. Private and
    experimental methods that should not be exposed through gRPC and that
    don't need to be exposed over the network, should still use jsonrpc.
    
    Related changes:
    * rename mctl to its original name "dumb"
    * move mctl/dumb into mayastor dir
```


mayastor/Cargo.toml
```diff
@@ -23,6 +24,10 @@ path = "src/bin/uring-support.rs"
 name = "mayastor-client"
 path = "src/bin/cli/cli.rs"
 
+[[bin]]
+name = "dumb"
```
gila, line 28: let us give it a better name @jkryl -- how about rpc? or jrpc (j for json?)

mayastor/src/replica.rs
```diff
@@ -1,30 +1,15 @@
...
 
 use std::ffi::{c_void, CStr, CString};
 
-use futures::{
-    channel::oneshot,
-    future::{self, FutureExt},
-};
+use futures::channel::oneshot;
 use nix::errno::Errno;
+use rpc::mayastor as grpc;
```
gila, line 10: is this a leftover from plowing through the errors, we just want to import all the structs we use here to avoid doing rpc::<whatever> in the code later, right?

mayastor/src/replica.rs
```diff
@@ -508,9 +438,9 @@ impl Iterator for ReplicaIter {
 }
 
 pub(crate) async fn create_replica(
-    args: CreateReplicaRequest,
-) -> Result<CreateReplicaReply, RpcError> {
-    let want_share = match ShareProtocolReplica::from_i32(args.share) {
+    args: grpc::CreateReplicaRequest,
```
gila, line 441: can you just add use grpc:: { .... } at the top?

doc/mcli.md
```diff
@@ -1,5 +1,9 @@
 # Overview
 
+!!! The doc is *stale* !!!
```
gila, line 3: mind removing it? nobody is going to update it and there nothing worse that old docs..

tiagolobocastro: I can update it while I still remember how to use mayastor-client :)

mayastor/Cargo.toml
jkryl: Tiago requested a better name too :-) So we decided to go forward with "jsonrpc" name.

mayastor/src/replica.rs
jkryl: I would have to explicitly import around 12 definitions that I find tedious and needlessly chatty. Another reason is that Replica definition from grpc conflicts with Replica object definition. That one I would have to rename to grpcReplica while importing it. All in all using grpc namespace looks like a better way in this case.

doc/mcli.md
jkryl: @tiagolobocastro : I would rather remove it too, but if you volunteer to go through the example commands, test them and update them to be accurate then I will leave it in 👍

mayastor/src/replica.rs
jkryl: AFAIK `module::*` imports are discouraged in rust and there is name conflict between these two modules as mentioned in my other post.

doc/mcli.md
tiagolobocastro: I've raised CAS-364 to do this

jkryl:
I updated the PR and renamed dumb to jsonrpc.

PR #331, opened 2020-07-20
Change CSI to use snafu

blaisedias:
Resolves cas-365 part of cas-338


csi/src/node/nvmfutil.rs
```diff
@@ -60,13 +59,14 @@ fn uuid_from_str(s: &str) -> String {
...
-    let second = time::Duration::from_secs(1);
-    let mut retries: i32 = 0;
+fn wait_for_path_to_exist(
+    uuid: String,
+    timeout: time::Duration,
+    max_retries: u32,
+) -> Option<String> {
+    let mut retries: u32 = 0;
     let now = time::Instant::now();
     while retries < max_retries {
```
tiagolobocastro, line 69: if max_**re**tries is 0 we don't even **try**?

blaisedias: fixed

PR #333, opened 2020-07-20
CAS-344: add UUID to URI format

tiagolobocastro:
Support for optional UUID query parameter for storage URIs, example:

aio:///disk1.img?blk_size=512&uuid=e40efd60-05a9-4f96-99af-1b6a25e109dd

If one is provided then we set the uuid of the bdev to that.

### DIFF
--- a/mayastor/src/bdev/dev.rs
+++ b/mayastor/src/bdev/dev.rs
@@ -11,8 +11,8 @@
 //!  - Arranging for the try_from() method to be called by the parse() function
 //!    below for a URI with the appropriate scheme.
 //!
-//! See mod.rs for the approriate trait definition(s), and refer
-//! to the files in the dev directoy for sample implementations.
+//! See mod.rs for the appropriate trait definition(s), and refer
+//! to the files in the dev directory for sample implementations.
 //!
 //! Creating a bdev for any supported device type is now as simple as:
 //! ```ignore
--- a/mayastor/src/bdev/dev/aio.rs
+++ b/mayastor/src/bdev/dev/aio.rs
@@ -19,6 +19,7 @@ pub(super) struct Aio {
     name: String,
     file: String,
     blk_size: u32,
+    uuid: Option<uuid::Uuid>,
 }
 
 /// Convert a URI to an Aio "object"
@@ -48,6 +49,12 @@ impl TryFrom<&Url> for Aio {
             None => 512,
         };
 
+        let uuid = uri::uuid(parameters.remove("uuid")).context(
+            nexus_uri::UuidParamParseError {
+                uri: url.to_string(),
+            },
+        )?;
+
         if let Some(keys) = uri::keys(parameters) {
             warn!("ignored parameters: {}", keys);
         }
@@ -56,6 +63,7 @@ impl TryFrom<&Url> for Aio {
             name: url.to_string(),
             file: format!("/{}", segments.join("/")),
             blk_size,
+            uuid,
         })
     }
 }
@@ -86,11 +94,17 @@ impl CreateDestroy for Aio {
         };
 
         async {
-            errno_result_from_i32(self.get_name(), errno).context(
-                nexus_uri::InvalidParams {
+            errno_result_from_i32(self.get_name(), errno)
+                .context(nexus_uri::InvalidParams {
                     name: self.get_name(),
-                },
-            )
+                })
+                .map(|name| {
+                    self.uuid.map(|u| {
+                        Bdev::lookup_by_name(&name)
+                            .map(|mut b| b.set_uuid(Some(u.to_string())))
+                    });
+                    name
+                })
         }
         .await
     }
--- a/mayastor/src/bdev/dev/iscsi.rs
+++ b/mayastor/src/bdev/dev/iscsi.rs
@@ -27,6 +27,7 @@ pub(super) struct Iscsi {
     name: String,
     iqn: String,
     url: String,
+    uuid: Option<uuid::Uuid>,
 }
 
 /// Convert a URI to an Iscsi "object"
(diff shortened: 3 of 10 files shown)


mayastor/src/bdev/dev/nvmf.rs
```diff
@@ -176,6 +181,8 @@ impl CreateDestroy for Nvmf {
                 name: self.name.clone(),
             })?;
 
+        uri::dev_set_uuid(&self.name, self.uuid);
```
gila, line 184: the bdev created, will already have a UUID defined by the target, so no point in setting it.

mayastor/src/bdev/dev/aio.rs
```diff
@@ -85,6 +89,10 @@ impl CreateDestroy for Aio {
             create_aio_bdev(cname.as_ptr(), filename.as_ptr(), self.blk_size)
         };
 
+        if errno == 0 {
+            uri::dev_set_uuid(&self.name, self.uuid);
```
gila, line 93: I guess its OK but, dev_set_uuid leaks the device abstraction into the URI abstraction which does not seem to be needed? We can just do something like:

```rust
bdev.lookup(&self.name).map(|b| b.set_uuid(self.uuid))
```
or something like that

mayastor/src/bdev/util/uri.rs
```diff
@@ -54,3 +60,32 @@ pub(crate) fn boolean(
     }
     value.parse::<bool>()
 }
+
+pub(crate) fn uuid(
```
gila, line 64: This does not seem to be needed ... we already have a Uri::parse() and from there we can just look up key/values by

mayastor/src/bdev/util/uri.rs
```diff
@@ -54,3 +60,32 @@ pub(crate) fn boolean(
...
+        None => None,
+    })
+}
+
+pub(crate) fn dev_set_uuid(name: &str, uuid: Option<uuid::Uuid>) {
+    match Bdev::lookup_by_name(&name) {
+        Some(mut bdev) => {
+            bdev.set_uuid(uuid.map(|u| u.to_string()));
+        }
+        // Should we fail to create a device if we can't set
```
gila, line 84: As mentioned I don't think this layer adds much value to working with bdevs however, if you add more "utils" then we can reassess when you're further down the road with it. However, we can not fail setting the UUID as do a simply memcopy

mayastor/src/bdev/util/uri.rs
```diff
@@ -54,3 +60,32 @@ pub(crate) fn boolean(
...
+                uri: url.to_string(),
+                parameter: "uuid".to_string(),
+            },
+        )?),
+        None => None,
+    })
+}
+
+pub(crate) fn dev_set_uuid(name: &str, uuid: Option<uuid::Uuid>) {
+    match Bdev::lookup_by_name(&name) {
```
gila, line 80: nit, you don't have to match here but simply do map directly.

mayastor/src/bdev/util/uri.rs
```diff
@@ -54,3 +60,32 @@ pub(crate) fn boolean(
...
+        Some(mut bdev) => {
+            bdev.set_uuid(uuid.map(|u| u.to_string()));
+        }
+        // Should we fail to create a device if we can't set
+        // the UUID? Then we need to return error here
+        None => error!(
```
gila, line 86: The error here is incorrect, this error would suggest we could not find the bdev.... not that we can not set the uri. The set_uuid() function returns void so it has to be the error from not being able to find it :)

mayastor/src/bdev/util/uri.rs
tiagolobocastro: From there where, the fn parse?
From what I can see every device "fetches" its own parameters from the url. So this was just added as a helper that also validates that the UUID is actually valid to avoid duplicating code on every device type.

mayastor/src/bdev/util/uri.rs
tiagolobocastro: yeah, it's more a failed to "find and set". But what happens if we fail to find? Or simply assume that since we've just created it it must have be found?

mayastor/src/bdev/util/uri.rs
gila: Yes but it does so only once.

Each bdev implements the TryFrom trait. This parses the URL, having functions that accept the URL, parse it, and return it seems, a bit inefficient in that respect, that's mostly the reason for the comment.

mayastor/src/bdev/util/uri.rs
gila: if we fail to find, then we have a bigggg problem ;)

mayastor/src/bdev/dev.rs
```diff
@@ -11,8 +11,8 @@
 //!  - Arranging for the try_from() method to be called by the parse() function
 //!    below for a URI with the appropriate scheme.
 //!
-//! See mod.rs for the approriate trait definition(s), and refer
-//! to the files in the dev directoy for sample implementations.
+//! See mod.rs for the appropriate trait definition(s), and refer
+//! to the files in the dev directly for sample implementations.
```
cjones1024, line 15: That was meant to have been "directory" rather "directly".

tiagolobocastro: that makes more sense :)

mayastor/src/bdev/dev/iscsi.rs
```diff
@@ -52,24 +53,31 @@ impl TryFrom<&Url> for Iscsi {
...
 
         Ok(Iscsi {
             name: url.to_string(),
             iqn: format!("{}:{}", ISCSI_IQN_PREFIX, Uuid::new_v4()),
-            url: format!("{}/0", url.to_string()),
+            url: if segments.len() == 2 {
+                url[.. url::Position::AfterPath].to_string()
+            } else {
+                format!("{}/0", &url[.. url::Position::AfterPath])
+            },
```
cjones1024, line 79: I would be tempted to also "parse" the  LUN segment to make sure it's actually an integer.

tiagolobocastro:
> Looks OK. Maybe you want to just mention the verbose error change in the description?

Thanks Chris, I've done it.

mayastor/src/bdev/dev/iscsi.rs
tiagolobocastro: I didn't do it because spdk already does this for us:
[2020-07-24T09:27:10.836709918Z ERROR bdev_iscsi.c:824] could not parse URL: Invalid URL iscsi://127.0.0.1:3262/iqn.2019-05.io.openebs:replica1/blahnotanumber
Could not parse <lun>
iSCSI URL must be of the form: "iscsi://[<username>[%<password>]@]<host>[:<port>]/<target-iqn>/<lun>"

PR #336, opened 2020-07-21
Split mayastor daemonset yaml into csi-agent and mayastor yamls (CAS-294)

jkryl:
csi-agent runs on all nodes so that iscsi and nvmf exported volumes can be mounted anywhere.

### DIFF
--- a/deploy/csi-daemonset.yaml
+++ b/deploy/csi-daemonset.yaml
@@ -1,0 +1,118 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  namespace: mayastor
+  name: mayastor-csi
+  labels:
+    openebs/engine: mayastor
+spec:
+  selector:
+    matchLabels:
+      app: mayastor-csi
+  updateStrategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 1
+  minReadySeconds: 10
+  template:
+    metadata:
+      labels:
+        app: mayastor-csi
+    spec:
+      hostNetwork: true
+      nodeSelector:
+        kubernetes.io/arch: amd64
+      # NOTE: Each container must have mem/cpu limits defined in order to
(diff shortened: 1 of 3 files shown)


deploy/csi-daemonset.yaml
```diff
@@ -0,0 +1,117 @@
...
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 1
+  minReadySeconds: 10
+  template:
+    metadata:
+      labels:
+        app: mayastor-csi
+    spec:
+      nodeSelector:
```
blaisedias, line 22: Don't we need hostnetwork: true here? as iscsiadm will run on the node which is running mayastor csi plugin/

deploy/csi-daemonset.yaml
```diff
@@ -0,0 +1,117 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  namespace: mayastor
+  name: mayastor-csi
+  labels:
+    openebs/engine: mayastor
```
GlennBullingham, line 7: (if) I read this correctly, deploying the csi plugin to a node will cause this label to be applied to that node?  If so, does this then not cause mayastor to be deployed there too (as this label is a selector clause for the mayastor daemonset)?

deploy/csi-daemonset.yaml
jkryl: thanks for reminding me! 😅  I had this change in a different workspace and accidentally did not include it in final PR ..

deploy/csi-daemonset.yaml
jkryl: it does not apply the label to the node but rather to the daemonset object. So we should be fine here 👍

GlennBullingham: Withdrawing my previous comment.  It is the Daemonset object definition which is being labeled, not the resultant pods, so there is no issue.

PR #337, opened 2020-07-23
CAS-372: Use read-only nexus descriptor in rebuild job

paulyoong:
The rebuild job requires the nexus descriptor to lock/unlock LBA ranges.
It is not necessary for the descriptor to be writable to use this
functionality. Furthermore, if the nexus is shared (i.e. over iscsi),
it is claimed, thereby prohibiting another writable descriptor being
opened against it.

### DIFF
--- a/mayastor/tests/add_child.rs
+++ b/mayastor/tests/add_child.rs
@@ -1,0 +1,129 @@
+use mayastor::{
+    bdev::{nexus_create, nexus_lookup, ChildStatus},
+    core::{mayastor_env_stop, MayastorCliArgs, MayastorEnvironment, Reactor},
+};
+use std::{ops::Deref, process::Command};
+
+static NEXUS_NAME: &str = "nexus";
+
+static DISKNAME1: &str = "/tmp/disk1.img";
+static BDEVNAME1: &str = "aio:///tmp/disk1.img?blk_size=512";
+
+static DISKNAME2: &str = "/tmp/disk2.img";
+static BDEVNAME2: &str = "aio:///tmp/disk2.img?blk_size=512";
+
+pub mod common;
+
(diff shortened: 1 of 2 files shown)


mayastor/tests/add_child.rs
```diff
@@ -0,0 +1,129 @@
...
+
+static DISKNAME1: &str = "/tmp/disk1.img";
+static BDEVNAME1: &str = "aio:///tmp/disk1.img?blk_size=512";
+
+static DISKNAME2: &str = "/tmp/disk2.img";
+static BDEVNAME2: &str = "aio:///tmp/disk2.img?blk_size=512";
+
+pub mod common;
+
+fn test_start() {
```
gila, line 17: We have common code for this in test/common.rs I think its best to use that unless there is a specific reason.

mayastor/tests/add_child.rs
```diff
@@ -0,0 +1,129 @@
...
+        .iter()
+        .map(|disk| {
+            let output = Command::new("truncate")
+                .args(&["-s", "64m", disk])
+                .output()
+                .expect("failed to create disk image");
+
+            assert_eq!(output.status.success(), true);
+        })
+        .for_each(drop);
```
gila, line 31: you can swap out the `.map()` for the `.for_each()` directly IOW:
```rust
disk.iter().for_each(|disk| { ... })
```

mayastor/tests/add_child.rs
paulyoong: Thanks - I've used the common code now.

mayastor/tests/add_child.rs
paulyoong: :+1:

PR #339, opened 2020-07-27
CAS-345: normalize the dev name

tiagolobocastro:
All "URI devices" are now created and set to their proper name and not
the full URI. The URI is still set as an alias which the rest of the
code currently relies on.

For example an aio device:
aio:///dev/ram0?blk_size=512 now has the name /dev/ram0

Add aliases to the output of mayastor-client bdev list subcommand.

Also, destroy the nexus children when failing to create a nexus with a
generic error. This fixes a JS test which was "passing" incorrectly.
Also temporarily "patch" that test which was not failing as expected.

### DIFF
--- a/mayastor/src/bdev/dev/aio.rs
+++ b/mayastor/src/bdev/dev/aio.rs
@@ -17,7 +17,7 @@ use crate::{
 #[derive(Debug)]
 pub(super) struct Aio {
     name: String,
-    aliases: Vec<String>,
+    alias: String,
     blk_size: u32,
     uuid: Option<uuid::Uuid>,
 }
@@ -61,7 +61,7 @@ impl TryFrom<&Url> for Aio {
 
         Ok(Aio {
             name: url.path().into(),
-            aliases: vec![url.to_string()],
+            alias: url.to_string(),
             blk_size,
             uuid,
         })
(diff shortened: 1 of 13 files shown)


mayastor/src/bdev/dev/aio.rs
```diff
@@ -17,7 +17,7 @@ use crate::{
 #[derive(Debug)]
 pub(super) struct Aio {
     name: String,
-    file: String,
+    aliases: Vec<String>,
```
gila, line 20: A nit but... We don't have to use a vec here because.... (explanation below)...

mayastor/src/bdev/dev/aio.rs
```diff
@@ -99,10 +98,17 @@ impl CreateDestroy for Aio {
...
                 .map(|name| {
-                    self.uuid.map(|u| {
-                        Bdev::lookup_by_name(&name)
-                            .map(|mut b| b.set_uuid(Some(u.to_string())))
-                    });
+                    if let Some(mut bdev) = Bdev::lookup_by_name(&self.name) {
+                        if let Some(uuid) = self.uuid {
+                            bdev.set_uuid(Some(uuid.to_string()));
+                        }
+                        if !bdev.add_aliases(&self.aliases) {
```
gila, line 105: we can simply call bdev.add_alias(&url.to_string()) and avoid the allocation of the Vec etc, right?

tiagolobocastro: yes that's right, the idea was to make it ready to support multiple aliases, should we want to but yes not really necessary for now...

PR #340, opened 2020-07-27
CAS-237 Refactor CSI code to use traits for attaching and detaching devices.

cjones1024:
Refactor csi code to use traits for attaching and detaching devices.
Use udev to determine location of attached devices.
Resolved issue where CSI plugin failed to detect if a volume had already been published.
Resolves CAS-237 and CAS-258


csi/src/dev.rs
```diff
@@ -0,0 +1,175 @@
...
+    ) -> Result<Option<Box<dyn Detach>>, DeviceError> {
+        let iscsi_key: String = format!("nexus-{}", uuid.to_string());
+        let nvmf_key: String = format!("uuid.{}", uuid.to_string());
+
+        let mut enumerator = Enumerator::new()?;
+
+        enumerator.match_subsystem("block")?;
+        enumerator.match_property("DEVTYPE", "disk")?;
+
+        for device in enumerator.scan_devices()? {
```
tiagolobocastro, line 91: I wonder if reversing the logic would simplify these long chains of "if's"

csi/src/dev.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    /// Wait for a device to show up in udev
+    /// once attach() has been called.
+    pub async fn wait_for_device(
+        device: Box<dyn Attach>,
+        retries: u32,
+    ) -> Result<DeviceName, DeviceError> {
+        const SECOND: Duration = Duration::from_secs(1);
+        for _ in 0 .. retries {
```
tiagolobocastro, line 167: should this be "..=" since it's a "retries" not "tries" ?

csi/src/dev/iscsi.rs
```diff
@@ -0,0 +1,249 @@
...
+        format!("ip-{}-iscsi-{}-lun-{}", self.portal, self.iqn, self.lun)
+    }
+
+    fn from_path(path: &str) -> Result<IscsiDevice, DeviceError> {
+        lazy_static! {
+            static ref PATTERN: Regex = Regex::new(r"^ip-(?P<host>[^:]+):(?P<port>[[:digit:]]+)-(?P<scheme>[[:alpha:]]+)-(?P<iqn>[[:alpha:]]+\.[[:digit:]]{4}-[[:digit:]]{2}\.[^:]+:(?P<suffix>[[:alnum:]]+(?:-[[:xdigit:]]+)+))-lun-(?P<lun>[[:digit:]]+)$").unwrap();
+        }
+
+        if let Some(captures) = PATTERN.captures(path) {
+            let host = captures.name("host").unwrap().as_str();
```
tiagolobocastro, line 42: I'm guessing these unwrap() won't fail because we've "captured" successfully?

csi/src/dev.rs
cjones1024: Maybe clearer just to rename it "tries" (or perhaps "attempts")

csi/src/dev.rs
cjones1024: I did give this some thought. Reversing the logic requires either more assignments or evaluating each expression twice.
However I think I can make this a bit more readable withought sacrificing efficency by creating a macro

csi/src/dev.rs
tiagolobocastro: yeah could be, although might be a bit to call a wait with 0 tries which won't do anything :)

csi/src/dev.rs
gila: Yeah reversing would be an option but.... there are some other things we can do here already:

(1) we can filter out (match_property) to only enumerate Nexus_CAS_Driver, this would would get rid of 3 levels already
(2) grab immediately what you are looking for (ID_ISCSI_SERIAL) its either Some or None, and if its some by virtue of the match_property of (1) we already know we are good, right?

If we do that, our statement would would reduce to:

```rust
match device_property_value("ID_SCSI_SERIAL") {
  Some(serial) => { /* rest of logic */ }
   _ => continue
  }
}

File: csi/src/dev.rs [hunk 4]
```diff
@@ -0,0 +1,175 @@
...
+                                    }
+                                }
+                            }
+                        }
+                    }
+                    continue;
+                }
+            }
+
+            if let Some(wwn) = device.property_value("ID_WWN") {
```
gila, line 131: the same match_property can be used here and then we can jump straight to the grabbing of the ID_WWN

csi/src/dev.rs
gila: We really should use Udev::Monitor for this (i.e get notified) rather then polling in a loop with a sleep. Can you create an issue for this @cjones1024 for us to work on that in the future? 

Also, what happens if the device attaches AFTER we timedout? Do we, when we timeout, call disconnect to ensure its not there when we are done waiting?

csi/src/dev/iscsi.rs
```diff
@@ -0,0 +1,249 @@
...
+
+    async fn find(&self) -> Result<Option<DeviceName>, DeviceError> {
+        let path = self.to_path();
+
+        let mut enumerator = Enumerator::new()?;
+
+        enumerator.match_subsystem("block")?;
+        enumerator.match_property("DEVTYPE", "disk")?;
+
+        for device in enumerator.scan_devices()? {
```
gila, line 161: Here the same thing applies, you practically know what you are looking for we can get rid of a lot of if's here.

csi/src/mount.rs
```diff
@@ -1,180 +1,263 @@
...
-                    },
-                    Some(false) => false,
-                    None => match dest_match {
-                        Some(true) => true,
-                        Some(false) => false,
-                        None => true,
-                    },
+    let mut found: Option<proc_mounts::MountInfo> = None;
+
+    for entry in MountIter::new().unwrap() {
```
gila, line 54: nit, but its nicer to use .find() on the iterator here I think.

csi/Cargo.toml
```diff
@@ -14,21 +14,25 @@ tonic-build = "0.1.0"
 prost-build = "0.6.0"
 
 [dependencies]
+async-trait = "0.1.36"
 bytes = "0.5"
 bytesize = "1.0.0"
 chrono = "0.4.9"
 clap = "2.33.0"
 env_logger = "0.7"
+failure = "0.1"
```
gila, line 23: @blaisedias added this one crate iirc but  after noticeing its deprecated changed to something else. So do we still need it?

csi/src/dev.rs
```diff
@@ -0,0 +1,175 @@
+//! Definition of traits required for attaching and detaching devices.
+//! Note that (unfortunately) the attach and detach operations are not
+//! quite symmetric. This is because it is not possible to discover all
+//! the information contained in a device URI from the entries in udev.
+//! Hence there are separate traits for attach and detach that each
+//! device type must implement.
```
gila, line 6: can you explain the problem? Its not a big issue ofcourse i just wonder why

csi/src/dev.rs
```diff
@@ -0,0 +1,175 @@
...
+mod util;
+
+const NVME_NQN_PREFIX: &str = "nqn.2019-05.io.openebs";
+
+pub use crate::error::DeviceError;
+pub type DeviceName = String;
+
+#[tonic::async_trait]
+pub trait Attach: Sync + Send {
+    async fn attach(&self) -> Result<(), DeviceError>;
```
gila, line 49: Why not -> Result<deviceName, DeviceError> ? The find (which must be implemented) is implicit for attach to succeed right? As we return from attach only when we found the device?

csi/src/dev.rs
```diff
@@ -0,0 +1,175 @@
...
+            scheme => Err(DeviceError::from(format!(
+                "unsupported device scheme: {}",
+                scheme
+            ))),
+        }
+    }
+
+    /// Lookup an existing device in udev matching the given UUID
+    /// to obtain a device implementing the Detach trait.
+    pub async fn lookup(
```
gila, line 80: I have difficultly understanding the difference between lookup and the find method defined with the trait, the both amount to the same thing. Why cant I attach and then do a lookup instead of doing an attach and then find?

csi/src/dev/iscsi.rs
gila: as mentioned i dont fully understand why we have .lookup() and .find() the both find devices... i get that this one is "iscsi specific" but the code is the same as in lookup. IOW it looks to me that .lookup can find your device if its a nvme OR iscsi device?

csi/src/dev/iscsi.rs
cjones1024: It is possible to create a regex with capture groups that are not used in a successful match (think disjunctions - ie. patterns of the form A|B). This can't happen in our pattern however, so yes it's safe to assume the capture groups all have Some() values. Even if the value of the group matches an empty string, this will still be a Some().

csi/src/dev.rs
cjones1024: You've convinced me will change to a ..=

csi/src/dev/iscsi.rs
cjones1024: When we call lookup() it's to find a device to detach. At that point we only have a UUID. We don't know if its iscsi/nvme or anything else and so we look for all possibilities in 1 pass. In the other case, when we use find(), we already have a device (derived from the URI) and we call find() after the attach() to find out what device gets assigned to it. Since we know what type of device it is already, we don't need to bother checking for the other possibilities which we know can't happen. Yes lookup() should work in that case, but at least half the comparisons are unnecessary.

csi/Cargo.toml
cjones1024: It's still used by our nvmeadm crate.

csi/src/dev.rs
cjones1024: I did try to narrow the search by adding multiple filters via match_property(), but found out (the hard way) that the filters do not narrow the search as I had assumed.  The iterator returns records that match *ANY* filter rather than *ALL* filters simultaneuosly.

csi/src/dev.rs
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        Ok(None)
+    }
+
+    /// Wait for a device to show up in udev
+    /// once attach() has been called.
+    pub async fn wait_for_device(
+        device: Box<dyn Attach>,
+        retries: u32,
```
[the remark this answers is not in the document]
cjones1024: Yes. Will do.

csi/src/dev.rs
cjones1024: Have created CAS-376 to use udev Monitors. Should also address appropriate failure response there as well.

csi/src/dev/iscsi.rs
cjones1024: Have refactored this avoid duplicating common code. As an added bonus, it is also hopefully now much clearer to see what we are matching against in udev.

csi/src/dev.rs
cjones1024: At present the attach() returns immediately without waiting to verify that the device has appeared in udev. The reason for this is that, according to Blaise, Kubernetes may abort (and then retry) the CSI call if they take to long. By being able to call find independently means that the CSI call has more chance of being successful on a retry. Again this should be revisited as part of CAS-376

csi/src/dev.rs
cjones1024: They do serve different purposes. find() is part of the trait while lookup() is not. lookup() is needed when you don't have an object or a URI with which to create one - you only have a UUID. This is inteded to be used when detaching.  find() is intended to be called once you have an object (obtained from URI) in order to attach. It's job is to tell you where to find the device referred to in the URI. There are also different implementations of find(), which need not use udev at all (such as  nbd - where find() doesn't involve a udev lookup  and does not implement detach at all).

PR #341, opened 2020-07-27
CAS-373: Add rebuild support to mayastor-client

paulyoong:
Added support for the following rebuild operations:
  - start
  - stop
  - pause
  - resume
  - state
  - progress

Added the optional "norebuild" argument to the "nexus add" command. This
determines whether or not a rebuild is started automatically for the
added child. If false, a rebuild is started. If true a rebuild is not
started. (Note: false is the default value)

### DIFF
--- a/mayastor/src/bin/cli/cli.rs
+++ b/mayastor/src/bin/cli/cli.rs
@@ -16,6 +16,7 @@ mod bdev_cli;
 mod context;
 mod nexus_cli;
 mod pool_cli;
+mod rebuild_cli;
 mod replica_cli;
 
 type MayaClient = MayastorClient<Channel>;
@@ -74,6 +75,7 @@ async fn main() -> Result<(), Status> {
         .subcommand(nexus_cli::subcommands())
         .subcommand(replica_cli::subcommands())
         .subcommand(bdev_cli::subcommands())
+        .subcommand(rebuild_cli::subcommands())
         .get_matches();
 
     let ctx = Context::new(&matches).await;
@@ -83,6 +85,7 @@ async fn main() -> Result<(), Status> {
         ("nexus", Some(args)) => nexus_cli::handler(ctx, args).await?,
         ("pool", Some(args)) => pool_cli::handler(ctx, args).await?,
         ("replica", Some(args)) => replica_cli::handler(ctx, args).await?,
+        ("rebuild", Some(args)) => rebuild_cli::handler(ctx, args).await?,
 
         _ => eprintln!("Internal Error: Not implemented"),
     };
--- a/mayastor/src/bin/cli/nexus_cli.rs
+++ b/mayastor/src/bin/cli/nexus_cli.rs
@@ -67,6 +67,12 @@ pub fn subcommands<'a, 'b>() -> App<'a, 'b> {
                 .required(true)
                 .index(2)
                 .help("uri of child to add"),
+        )
+        .arg(
+            Arg::with_name("norebuild")
+                .default_value("false")
+                .index(3)
+                .help("determines if a rebuild job runs automatically"),
         );
 
     let remove = SubCommand::with_name("remove")
--- a/mayastor/src/bin/cli/nexus_cli.rs
+++ b/mayastor/src/bin/cli/nexus_cli.rs
@@ -72,7 +72,7 @@ pub fn subcommands<'a, 'b>() -> App<'a, 'b> {
             Arg::with_name("norebuild")
                 .default_value("false")
                 .index(3)
-                .help("determines if a rebuild job runs automatically"),
+                .help("specify if a rebuild job runs automatically"),
         );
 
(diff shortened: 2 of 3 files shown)


mayastor/src/bin/cli/rebuild_cli.rs
```diff
@@ -0,0 +1,272 @@
...
+        "Stopping rebuild of child {} on nexus {}",
+        uri, uuid
+    ));
+    ctx.client
+        .stop_rebuild(rpc::StopRebuildRequest {
+            uuid: uuid.clone(),
+            uri: uri.clone(),
+        })
+        .await?;
+    ctx.v1(&format!(
```
tiagolobocastro, line 173: I wonder if these could be a bit misleading since it's not guaranteed that the action has been completed.

paulyoong: Good point - I've changed the messages slightly to address this.

mayastor/src/bin/cli/nexus_cli.rs
```diff
@@ -67,6 +67,12 @@ pub fn subcommands<'a, 'b>() -> App<'a, 'b> {
                 .required(true)
                 .index(2)
                 .help("uri of child to add"),
+        )
+        .arg(
+            Arg::with_name("norebuild")
```
gila, line 72: instead of calling this norebuild -- why not call it "rebuild"

mayastor/src/bin/cli/nexus_cli.rs
```diff
@@ -67,6 +67,12 @@ pub fn subcommands<'a, 'b>() -> App<'a, 'b> {
...
+                .default_value("false")
+                .index(3)
+                .help("determines if a rebuild job runs automatically"),
```
gila, line 75: determines -> specifiy?

mayastor/src/bin/cli/cli.rs
```diff
@@ -83,6 +85,7 @@ async fn main() -> Result<(), Status> {
         ("nexus", Some(args)) => nexus_cli::handler(ctx, args).await?,
         ("pool", Some(args)) => pool_cli::handler(ctx, args).await?,
         ("replica", Some(args)) => replica_cli::handler(ctx, args).await?,
+        ("rebuild", Some(args)) => rebuild_cli::handler(ctx, args).await?,
```
gila, line 88: I wonder as we can only rebuild a child that is part of a nexus (right?) if we should not put this as a subcommond *under*  the nexus? for example:

```bash
mayastor-client nexus $name rebuild pause
```
... or maybe not... we can always change our mind later (or add them to both)

mayastor/src/bin/cli/rebuild_cli.rs
```diff
@@ -0,0 +1,256 @@
...
+            Arg::with_name("uuid")
+                .required(true)
+                .index(1)
+                .help("uuid of the nexus"),
+        )
+        .arg(
+            Arg::with_name("uri")
+                .required(true)
+                .index(2)
+                .help("uri of child to start rebuilding"),
```
gila, line 39: @tiagolobocastro is working on changes of bdev names etc but IMHO, we should not specify the URI rather the name of the child? (we can leave it for now just a general comment for a todo -- it goes for all the subcommands)

tiagolobocastro: Yes but not quite yet because this goes through the Nexus stuff and not the bdevs directly. `NexusChild` names are still URIs.

mayastor/src/bin/cli/nexus_cli.rs
paulyoong: I chose this for consistency with the rpc call (where the naming also came up as a talking point). Below is a snippet from the proto file:

`message AddChildNexusRequest {
  string uuid = 1;    // uuid of the nexus
  string uri = 2;     // URI of the child device to be added
  bool norebuild = 3;   // auto start rebuilding
}`

IIRC 'norebuild' was chosen because the default rpc value for this is false and by default we wish to perform a rebuild.

mayastor/src/bin/cli/cli.rs
paulyoong: I did start with that but then thought it looked cleaner/more consistent for the command to look like:
`mayastor-client rebuild pause $uuid $uri`

I'd like to leave it like this for the time being, but as you say this can always be changed.

mayastor/src/bin/cli/rebuild_cli.rs
paulyoong: Agreed :+1: 

I've added a todo for this

mayastor/src/bin/cli/nexus_cli.rs
paulyoong: Done - thanks

PR #342, opened 2020-07-27
grpc create methods should return the created object (CAS-88)

jkryl:
This simplifies the code in the control plane and improves atomicity of the create methods (either everything succeeds or fails).

### DIFF
--- a/mayastor-test/test_nexus.js
+++ b/mayastor-test/test_nexus.js
@@ -112,7 +112,7 @@ var client;
 
 function controlPlaneTest (thisProtocol) {
   it('should publish the nexus', (done) => {
-    client.PublishNexus(
+    client.publishNexus(
       {
         uuid: UUID,
         share: thisProtocol
@@ -133,7 +133,7 @@ function controlPlaneTest (thisProtocol) {
   });
 
   it('should re-publish the nexus', (done) => {
-    client.PublishNexus(
+    client.publishNexus(
       {
         uuid: UUID,
         share: thisProtocol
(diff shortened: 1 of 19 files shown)


mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -278,16 +204,31 @@ impl mayastor_server::Mayastor for MayastorSvc {
...
+    ) -> GrpcResult<Child> {
         let args = request.into_inner();
         trace!("{:?}", args);
         let uuid = args.uuid.clone();
         debug!("Adding child {} to nexus {} ...", args.uri, uuid);
-        locally! { async move {
-            nexus_lookup(&args.uuid)?.add_child(&args.uri, args.norebuild).await.map(|_| ())
+        // XXX How to avoid following spaghetti code if we cannot
+        // use '?' inside 'locally' macro?
+        let child = locally! { async move {
```
tiagolobocastro, line 214: What if you put what's inside the locally into a function and then you just call it, would that work?

mayastor-test/test_nexus.js
```diff
@@ -385,11 +386,44 @@ describe('nexus', function () {
...
+          `iscsi://${externIp}:${iscsiReplicaPort}/iqn.2019-05.io.openebs:disk1`
+        );
+        assert.equal(nexus.children[2].state, 'CHILD_ONLINE');
+      }
+
+      if (doUring()) {
+        const uringIndex = 3 + doIscsiReplica;
+        assert.equal(
+          nexus.children[uringIndex].uri,
+          `uring:///${uringFile}?blk_size=4096`
```
tiagolobocastro, line 417: extra slash here

mayastor/src/grpc/mayastor_grpc.rs
jkryl: yes, I could do that. I can create a special function in nexus_grpc.rs and call it from here. Not the best solution but likely better than this code.

mayastor-test/test_nexus.js
jkryl: thanks. fixed

PR #348, opened 2020-07-29
Remove nbd storage class and tag moac image as develop in dev

blaisedias:
Remove nbd storage class from example storage class yaml
Make image tags in the develop branch consistent i.e. develop

### DIFF
--- a/csi/moac/default.nix
+++ b/csi/moac/default.nix
@@ -43,7 +43,7 @@ result // rec {
   #    of layering.
   buildImage = pkgs.dockerTools.buildLayeredImage {
     name = "mayadata/moac";
-    tag = "latest";
+    tag = "develop";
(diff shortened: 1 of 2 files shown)


csi/moac/default.nix
```diff
@@ -43,7 +43,7 @@ result // rec {
   #    of layering.
   buildImage = pkgs.dockerTools.buildLayeredImage {
     name = "mayadata/moac";
-    tag = "latest";
+    tag = "develop";
```
jkryl, line 46: I'm not sure it is necessary to make such change in the repo. You can set it to whatever you want when building your private image, but there is no reason to change it for everyone.

jkryl: I see what you mean now. Looks good 👍

PR #350, opened 2020-07-30
mayastor gRPC methods should be idempotent (CAS-232)

jkryl:


mayastor/src/grpc/nexus_grpc.rs
```diff
@@ -112,6 +112,17 @@ pub async fn nexus_add_child(
     args: rpc::AddChildNexusRequest,
 ) -> Result<rpc::Child, Error> {
     let n = nexus_lookup(&args.uuid)?;
+    // TODO: do not add child if it already exists (idempotency)
+    // For that we need api to check existence of child by name (not uri that
+    // contain parameters that may change).
```
tiagolobocastro, line 117: we wouldn't add the same child with different parameters though?

csi/moac/node.js
```diff
@@ -339,16 +339,8 @@ class Node extends EventEmitter {
   async createPool (name, disks) {
     log.debug(`Creating pool "${name}@${this.name}" ...`);
 
-    var poolInfo;
-    try {
-      poolInfo = await this.call('createPool', { name, disks });
```
tiagolobocastro, line 344: can we not hit other exceptions? Eg if the connection drops?

mayastor/src/grpc/nexus_grpc.rs
jkryl: It's difficult to say. In any case it would be consistent with what we do currently with pool, nexus and replica. In these cases, we don't check (and perhaps we should) parameters they are created with. Ideally, we should check all parameters for all objects and return EEXIST if the parameters are different. For example if I try to create a pool with disk /dev/sda and a pool with the same name but disk /dev/sdb, that's not the same thing, although we consider it to be the same currently.

csi/moac/node.js
jkryl: we can and those are caught at higher level in enclosing scope (in poolOperator)

mayastor/src/grpc/nexus_grpc.rs
tiagolobocastro: Yeah I guess we need to scrub for these things

csi/moac/node.js
tiagolobocastro: :+1:

PR #351, opened 2020-07-30
CAS-346: get uri from bdev

tiagolobocastro:
Get uri from a bdev by making use of the alias which is set during
device creation. The UUID is also added, if not present. This allows
the simplification of the config file.

Added missing get_share_uri for the nexus
Add a generic get_share_uri to the bdev by making use of the replica
code. todo: sharing code should be refactored allow us to add the uri
to the bdev create response
Add uri to the bdev list response


mayastor/src/core/bdev.rs
```diff
@@ -377,6 +377,13 @@ impl Bdev {
             Some(Bdev::from(bdev))
         }
     }
+    /// return share URI
+    pub fn get_share_uri(&self) -> Option<String> {
```
gila, line 381: How about extending the share trait? It already provides the ability to determine what it is shared with as well. This makes the code easier to reuse later on when we implement the trait for the nexus and replica's

mayastor/src/target/mod.rs
```diff
@@ -6,3 +8,29 @@ pub enum Side {
...
 }
+
+/// Types of remote access storage protocols and IDs for sharing replicas.
+pub enum ShareType {
+    Nvmf,
+    Iscsi,
+}
+
+/// Detect share protocol (if any)
+pub fn detect_share(
```
gila, line 19: If a bdev is shared or not is (as mentioned in the previous comment somewhere, sorry for repeating) is already possible with the Share trait you can see it implemented in bdev.rs

mayastor/src/nexus_uri.rs
```diff
@@ -76,3 +82,42 @@ pub async fn bdev_destroy(uri: &str) -> Result<(), NexusBdevError> {
 pub fn bdev_get_name(uri: &str) -> Result<String, NexusBdevError> {
     Ok(Uri::parse(uri)?.get_name())
 }
+
+impl std::cmp::PartialEq<url::Url> for &Bdev {
+    fn eq(&self, uri: &url::Url) -> bool {
+        match Uri::parse(&uri.to_string()) {
+            Ok(device) => {
+                let uri_driver = if uri.scheme() == "nvmf" {
```
gila, line 90: it looks like where we converting  a nvmf scheme into nvme? why is that needed?

mayastor/src/nexus_uri.rs
```diff
@@ -76,3 +82,42 @@ pub async fn bdev_destroy(uri: &str) -> Result<(), NexusBdevError> {
...
+                    uri.scheme()
+                };
+                uri_driver == self.driver() && device.get_name() == self.name()
+            }
+            _ => false,
+        }
+    }
+}
+
+impl TryFrom<&Bdev> for url::Url {
```
gila, line 102: This is nice .. however..... the From<T> and TryFrom<T> typically take a T not &T. This alsi imlies that calling TryFrom<T> consumes T. This is not a hard rule but according to the rust docs, what is supposed to happen. so when calling this, what about, if we we do try_from(bdev.clone())?

mayastor/src/core/bdev.rs
tiagolobocastro: yeah that's a good idea :+1:

mayastor/src/nexus_uri.rs
tiagolobocastro: because the driver name reported as nvme and not nvmf. I guess because it's indeed nvme but with a "fabrics" transport.

mayastor/src/nexus_uri.rs
tiagolobocastro: Yeah that would not be ideal indeed, I'll have a look!

tiagolobocastro: hmm [here](https://doc.rust-lang.org/std/ops/index.html) seems to say that we should add &T|T as required?
Regardless, T with clone does work as well so we can go around consuming T with that.

PR #353, opened 2020-07-31
CAS-377: Increase the rebuild data size

paulyoong:
The amount of data that can be rebuilt at any given time has been
increased from 40KiB to 1MiB.

The number of concurrent copy tasks is increased to 16, with each task
copying 64KiB of data.

### DIFF
--- a/mayastor/src/rebuild/rebuild_impl.rs
+++ b/mayastor/src/rebuild/rebuild_impl.rs
@@ -10,7 +10,7 @@ use futures::{
 use once_cell::sync::OnceCell;
 use snafu::ResultExt;
 
-use spdk_sys::spdk_get_thread;
+use spdk_sys::{spdk_get_thread, SPDK_BDEV_LARGE_BUF_MAX_SIZE};
 
 use crate::{
     bdev::VerboseError,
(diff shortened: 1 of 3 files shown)


mayastor/src/rebuild/rebuild_impl.rs
```diff
@@ -42,9 +42,10 @@ struct TaskResult {
 }
 
 /// Number of concurrent copy tasks per rebuild job
-const SEGMENT_TASKS: usize = 4;
+const SEGMENT_TASKS: usize = 16;
 /// Size of each segment used by the copy task
-pub const SEGMENT_SIZE: u64 = 10 * 1024; // 10KiB
+/// Set to 64KiB to match SPDK's max buffer size (SPDK_BDEV_LARGE_BUF_MAX_SIZE)
+pub const SEGMENT_SIZE: u64 = 64 * 1024; // 64KiB
```
tiagolobocastro, line 48: Could we just set it to SPDK_BDEV_LARGE_BUF_MAX_SIZE then?

paulyoong: Good point - done

PR #354, opened 2020-08-03
V20.07.x

gila:


mayastor/src/target/iscsi.rs
```diff
@@ -165,51 +165,45 @@ fn destroy_iscsi_groups() {
...
+            &mut bdev.name().into_cstring().as_ptr(),
+            lun_ids,
+            1,
+            128,
+            true,
+            false,
+            false,
+            0,
+            false,
+            false,
```
jonathan-teh, line 202: Could the comments be preserved? It's less readable now.

nix/pkgs/libspdk/default.nix
```diff
@@ -55,8 +64,8 @@ stdenv.mkDerivation rec {
     "${enableFeature enableDebug "unit-tests"}"
     "--target-arch=nehalem"
     "--without-isal"
+    "--without-vhost"
```
jonathan-teh, line 67: Presumably we don't use/need vhost at the moment, is there another reason this is being disabled?

mayastor/tests/common/error_bdev.rs
```diff
@@ -22,7 +18,9 @@ pub fn create_error_bdev(error_device: &str, backing_device: &str) {
     let err_bdev_name_str = std::ffi::CString::new(error_device.to_string())
         .expect("Failed to create name string");
     unsafe {
-        retval = spdk_vbdev_error_create(err_bdev_name_str.as_ptr()); // create the error bdev around it
+        retval = vbdev_error_create(err_bdev_name_str.as_ptr()); // create the
```
tiagolobocastro, line 21: nit: I'd suggest adding the comment as one line above

mayastor/src/target/iscsi.rs
```diff
@@ -165,51 +165,45 @@ fn destroy_iscsi_groups() {
...
-            &mut lun_id as *mut _, // lun id
-            1,                     // length of lun id list
-            128,                   // max queue depth
-            false,                 // disable chap
-            false,                 // require chap
-            false,                 // mutual chap
-            0,                     // chap group
-            false,                 // header digest
-            false,                 // data digest
+            -1,
```
tiagolobocastro, line 187: it'd be nice to keep/update the comments

mayastor/tests/common/error_bdev.rs
gila: I had that but rust-fmt disagrees.

mayastor/src/target/iscsi.rs
gila: (1) the comments are not checked and or updated (biggest issue)
(2) the function arguments are pretty self-explanatory 
(3) we don't do that anywhere else

mayastor/src/target/iscsi.rs
gila: We don't comment function arguments anywhere else and they have to be updated/maintained. IMHO it is better this way as it forces you to check/validate the prototype of the function and not just assume the comments are (still) correct.

PR #355, opened 2020-08-03
CAS-382 CSI plugin should not assume "ID_SCSI_SERIAL" attribute is present in udev records for iSCSI devices

cjones1024:
Modified match_iscsi_device() so that it no longer refers to the "ID_SCSI_SERIAL" attribute when searching for iSCSI devices, as there is no guarantee that it will be present in the udev device record.

Fixed bug in extract_uuid().


gila:
This change itself looks good but I cant see how/why the serial property is missing/absent?

PR #358, opened 2020-08-04
Env fix

gila:


mayastor/src/bdev/dev/nvme.rs
```diff
@@ -0,0 +1,180 @@
+use std::{
```
tiagolobocastro, line 1: should this be called pcie.rs?

mayastor/src/bdev/dev/nvme.rs
```diff
@@ -0,0 +1,180 @@
...
+
+use crate::{
+    bdev::{CreateDestroy, GetName},
+    core::Bdev,
+    ffihelper::{cb_arg, errno_result_from_i32, ErrnoResult, IntoCString},
+    nexus_uri::{self, NexusBdevError},
+};
+
+#[derive(Debug)]
+pub(super) struct NVMe {
```
tiagolobocastro, line 29: same here, should this be PCIe?

mayastor/src/bdev/dev/nvme.rs
gila: Yeah hmm well yeah dunno =) 

I've been in limbo with this somewhat. The scheme refers to transport only so perhaps instead of having a PCIe scheme, it should be nvme scheme, but that would clash with nvmf.. As the device is a nvme device nvme makes more sense to me then pcie as that refers to the transport. 

I think, in the long run we might want to refactor this and have a single nvme.rs file that parses nvmf-tcp, nvmf-rdma, nvmf-fc,and pcie?

PR #359, opened 2020-08-04
Add support for block volumes. CAS-14

blaisedias:
To support block volumes.
  at node stage volume
    - attach to the device
  at node publish volume
    - create a block special file
    - bind mount the device path to the block special file
  at node unpublish volume
    - unmount the block special file
    - delete the block special file
  at node unstage volume
    - detach from the device

To facilitate this
At staging ensure the sequence of operations is
 1 Attach the device.
 2 For fs volumes only, mount the device on the staging path.

At unstaging ensure the sequence of operations is
 1 For fs volumes only, unmount staging path from the device.
 2 Detach the device.

Ensure that each of the above operations are idempotent.

For clarity move
 - filesystem volume functions (un/stage, un/publish),
 - block volume (un/publish)
to separarate files.

### DIFF
--- a/csi/src/mount.rs
+++ b/csi/src/mount.rs
@@ -267,3 +267,47 @@ pub fn bind_unmount(target: &str) -> Result<(), Error> {
 
     Ok(())
 }
+
+/// Mount a block device
+pub fn blockdevice_mount(
+    source: &str,
+    target: &str,
+    readonly: bool,
+) -> Result<Mount, Error> {
+    debug!("Mounting {} ...", source);
+
+    let mut flags = MountFlags::empty();
+    flags.insert(MountFlags::BIND);
+
+    let mount = Mount::new(
+        source,
+        target,
+        FilesystemType::Manual("none"),
+        flags,
+        None,
+    )?;
+    info!("Block device {} mounted to {}", source, target,);
+
+    if readonly {
+        flags.insert(MountFlags::REMOUNT);
+        flags.insert(MountFlags::RDONLY);
+
+        let mount =
+            Mount::new("", target, FilesystemType::Manual(""), flags, None)?;
+        info!("Remounted block device {} (readonly) to {}", source, target,);
+        return Ok(mount);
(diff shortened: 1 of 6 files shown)


csi/src/node.rs
```diff
@@ -33,6 +31,11 @@ pub struct Node {
     pub filesystems: Vec<String>,
 }
 
+// Timeout settings for device attach
+// 10 retries at 100ms intervals
```
tiagolobocastro, line 35: comment seems out of date. I'd say just remove it since the variable names are good.

csi/src/node.rs
```diff
@@ -526,194 +361,92 @@ impl node_server::Node for Node {
...
-                return Err(failure!(
-                    Code::AlreadyExists,
-                    "Failed to stage volume {}: device {} is already mounted elsewhere",
-                    volume_id,
-                    device_path
-                ));
-            }
+            Some(devpath) => devpath,
+            None => {
+                debug!("Attaching volume");
```
tiagolobocastro, line 404: maybe add volume ID?

csi/src/mount.rs
```diff
@@ -267,3 +267,47 @@ pub fn bind_unmount(target: &str) -> Result<(), Error> {
...
+    let mount = Mount::new(
+        source,
+        target,
+        FilesystemType::Manual("none"),
+        flags,
+        None,
+    )?;
+    info!("Block device {} mounted to {}", source, target,);
+
+    if readonly {
```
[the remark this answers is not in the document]
blaisedias: Yes the remount is necessary for block devices. The RDONLY flag is co-opted for other reasons at bind mount, not good but it is what it is.

csi/src/node.rs
blaisedias: Done.

csi/src/node.rs
blaisedias: Done.

PR #362, opened 2020-08-05
CAS-348: export config on grpc / CAS-349: persist the config in k8s

tiagolobocastro:
CAS-348: Export the config file on "state-changing" grpc calls
If we fail to export (unexpected) then we fail the grpc and rely on the
control plane to retry - methods should be idempotent

Change the config device creation order:
1. pools
2. share pool replicas (if shared)
3. nexus (may be using the pool)
only if IMPORT_NEXUSES is true as when a node comes back up it may have
outdated view of the volumes and it would be dangerous to blindly
recreate the nexuses
4. base bdevs (last, otherwise the nexus fails to create)

CAS-349: persist the config ...
for k8s in the following host path: /var/local/mayastor/config.yaml
This is achieve using the host path volume:
directoryOrCreate type mount to create the path
fileOrCreate type mount to create the file

set IMPORT_NEXUSES env variable to false as we don't want to
automatically import/recreate nexuses after a crash as this information
might be outdated

### DIFF
--- a/mayastor/src/grpc/mod.rs
+++ b/mayastor/src/grpc/mod.rs
@@ -33,6 +33,30 @@ mod mayastor_grpc;
 mod nexus_grpc;
 mod server;
 
+use crate::subsys::Config;
+use futures::Future;
 pub use server::MayastorGrpcServer;
 use tonic::{Response, Status};
+
 pub type GrpcResult<T> = std::result::Result<Response<T>, Status>;
+
+/// Used by the gRPC method implementations to sync the current configuration by
+/// exporting it to a config file
+/// If `sync_config` fails then the method should return a failure
+/// requiring the gRPC caller to retry the method, which should be idempotent
+pub async fn sync_config<F, T>(future: F) -> GrpcResult<T>
+where
+    F: Future<Output = GrpcResult<T>>,
+{
+    let result = future.await;
+    if result.is_ok() {
+        match Config::export_config() {
+            Ok(_) => {}
+            Err(e) => {
+                error!("Failed to export config file: {}", e);
+                return Err(Status::data_loss("Failed to export config"));
+            }
+        }
+    }
+    result
+}
--- a/deploy/mayastor-daemonset.yaml
+++ b/deploy/mayastor-daemonset.yaml
@@ -42,10 +42,13 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: status.podIP
+        - name: IMPORT_NEXUSES
+          value: "false"
         args:
         - "-N$(MY_NODE_NAME)"
         - "-g$(MY_POD_IP)"
         - "-nnats"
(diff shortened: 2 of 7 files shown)


mayastor/src/grpc/mod.rs
```diff
@@ -33,6 +33,24 @@ mod mayastor_grpc;
...
 mod server;
 
+use crate::{bdev::VerboseError, subsys};
 pub use server::MayastorGrpcServer;
 use tonic::{Response, Status};
+
 pub type GrpcResult<T> = std::result::Result<Response<T>, Status>;
+
+/// exports the current configuration to the optional config file
+pub fn export_config() -> std::result::Result<(), Status> {
```
gila, line 43: What about, if we were to put this function within the config subsystem itself and create a function that accepts a closure that writes on success

```rust
async fn sync<F, T, E>( future: F) -> Result<T, E>
where
    F: Future<Output = Result<T, E>>,
{
    let cfg = ....;
    let result = future.await;
    if result.is_ok() {
        cfg.write();
    }
    result
}
```

Then we can use it like:
```rust
async fn grpc_thing(&self, request: Request<Null>) -> GrpcResult<Null> {
  Config::sync(  async { /* body that was there already */  } ).await
}
```

This way we can easily see which functions require/want to save?

tiagolobocastro: yeah that's much better :+1:

mayastor/src/core/env.rs
```diff
@@ -597,6 +597,13 @@ impl MayastorEnvironment {
         }
     }
 
+    pub(crate) fn import_nexuses() -> bool {
```
gila, line 600: I could not spot where this is used but -- Could we move this to src/subsys/config ?

mayastor/src/grpc/mod.rs
```diff
@@ -33,6 +33,40 @@ mod mayastor_grpc;
...
 
+use crate::{bdev::VerboseError, subsys};
+use futures::Future;
 pub use server::MayastorGrpcServer;
 use tonic::{Response, Status};
+
 pub type GrpcResult<T> = std::result::Result<Response<T>, Status>;
+
+/// exports the current configuration to the mayastor config file
+fn export_config() -> std::result::Result<(), Status> {
```
gila, line 44: this function could also be moved to subsys/config -- just to keep it all organized and clean. Look at how to do that in  mod file, look for  pub use XXX -- thats where the visiblity is controled mostly.

mayastor/src/core/env.rs
tiagolobocastro: Moved there, where it's used

mayastor/src/grpc/mod.rs
tiagolobocastro: Moved

PR #363, opened 2020-08-05
pool: support creation by means of URIs

gila:
This change will create base bdevs using URIs if we are passed
a valid scheme. If we use the legacy model where we specify
a device path like /dev/sdx -- we will use that

### DIFF
--- a/mayastor/src/core/bdev.rs
+++ b/mayastor/src/core/bdev.rs
@@ -115,12 +115,27 @@ impl Share for Bdev {
     }
 
     /// return share URI for nvmf and iscsi
-    fn get_share_uri(&self) -> Option<String> {
+    fn share_uri(&self) -> Option<String> {
         match self.shared() {
             Some(Protocol::Nvmf) => nvmf::get_uri(&self.name()),
             Some(Protocol::Iscsi) => iscsi::get_uri(Side::Nexus, &self.name()),
             None => None,
         }
+    }
+
+    fn bdev_uri(&self) -> Option<String> {
+        for alias in self.aliases().iter() {
+            if let Ok(mut uri) = url::Url::parse(alias) {
+                if self == uri {
+                    if uri.query_pairs().find(|e| e.0 == "uuid").is_none() {
+                        uri.query_pairs_mut()
+                            .append_pair("uuid", &self.uuid_as_string());
+                    }
+                    return Some(uri.to_string());
+                }
+            }
+        }
+        None
     }
 }
 
--- a/mayastor/src/nexus_uri.rs
+++ b/mayastor/src/nexus_uri.rs
@@ -83,6 +83,21 @@ pub fn bdev_get_name(uri: &str) -> Result<String, NexusBdevError> {
     Ok(Uri::parse(uri)?.get_name())
 }
 
+impl std::cmp::PartialEq<url::Url> for &Bdev {
+    fn eq(&self, uri: &url::Url) -> bool {
+        match Uri::parse(&uri.to_string()) {
+            Ok(device) if device.get_name() == self.name() => {
+                self.driver()
+                    == match uri.scheme() {
+                        "nvmf" | "pcie" => "nvme",
+                        scheme => scheme,
+                    }
+            }
+            _ => false,
+        }
+    }
+}
+
 impl std::cmp::PartialEq<url::Url> for Bdev {
     fn eq(&self, uri: &url::Url) -> bool {
         match Uri::parse(&uri.to_string()) {
--- a/mayastor/src/pool.rs
+++ b/mayastor/src/pool.rs
@@ -10,6 +10,7 @@ use std::{
 
 use futures::channel::oneshot;
 use snafu::Snafu;
+use url::Url;
 
 use rpc::mayastor as rpc;
 use spdk_sys::{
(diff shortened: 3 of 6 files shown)


mayastor/src/pool.rs
```diff
@@ -419,42 +420,60 @@ impl Pool {
...
 
-        let (sender, receiver) = oneshot::channel::<i32>();
-        if base_bdev_type == "aio" {
-            unsafe {
-                bdev_aio_delete(
-                    base_bdev.as_ptr(),
-                    Some(done_cb),
-                    cb_arg(sender),
-                );
+            let (sender, receiver) = oneshot::channel::<i32>();
```
tiagolobocastro, line 442: Do we need this "legacy" stuff or is this just a stop gap until moac creates pools with URIs?

mayastor/src/core/bdev.rs
```diff
@@ -115,13 +115,28 @@ impl Share for Bdev {
...
-    fn get_share_uri(&self) -> Option<String> {
+    fn share_uri(&self) -> Option<String> {
         match self.shared() {
             Some(Protocol::Nvmf) => nvmf::get_uri(&self.name()),
             Some(Protocol::Iscsi) => iscsi::get_uri(Side::Nexus, &self.name()),
             None => None,
         }
     }
+
+    fn bdev_uri(&self) -> Option<String> {
```
tiagolobocastro, line 126: hmm I was indeed torn between adding it here or as a trait when I added the Bdev->Url converter.

mayastor/src/nexus_uri.rs
```diff
@@ -83,6 +83,21 @@ pub fn bdev_get_name(uri: &str) -> Result<String, NexusBdevError> {
...
+                    }
+            }
+            _ => false,
+        }
+    }
+}
+
 impl std::cmp::PartialEq<url::Url> for Bdev {
     fn eq(&self, uri: &url::Url) -> bool {
         match Uri::parse(&uri.to_string()) {
```
tiagolobocastro, line 103: since you've added the trait above, this can just become self == uri ?

mayastor/src/pool.rs
```diff
@@ -560,6 +583,63 @@ pub(crate) async fn create_pool(
...
+        bdev_if: e.to_string(),
+        name: args.disks[0].clone(),
+    })?;
+
+    if let Some(pool) = Pool::lookup(&args.name) {
+        return if pool.get_base_bdev().name() == parsed.get_name() {
+            Ok(pool.into())
+        } else {
+            Err(Error::AlreadyExists {
+                name: args.name,
```
tiagolobocastro, line 607: should this error include the existing disk as well?

mayastor/src/pool.rs
gila: Right. We will remove the "legacy" stuff altogether but later one, but for that the whole pool creation would need to change somewhat, beginning with examining the bdevs. I want to postpone that until later.

mayastor/src/nexus_uri.rs
gila: I'm not sure what you are suggesting..

mayastor/src/pool.rs
gila: Likely yes, but all these error variants, all get smooshed down into the same error anyway (internal or not found) so I'd like to address that in the follow-up releases were we reduce the error variants.

mayastor/src/core/bdev.rs
gila: Yeah arguably, the share trait is not the right place for it and it should be just a method in the Impl block but im not sure how that will playout with replica's later

mayastor/src/nexus_uri.rs
tiagolobocastro: You've added the trait PartialEq for &Bdev. So now the trait PartialEq for Bdev could use make use of the &Bdev version becoming a simple bdev == uri on line 103?

gila: as discussed we cant :+1:

gila:
I've tested the following with k8s:

```yaml
 $ cat pool.yaml
apiVersion: "openebs.io/v1alpha1"
kind: MayastorPool
metadata:
  name: pool-2
  namespace: mayastor
spec:
  node: ksnode-2
  disks: ["/dev/sdb"]
---
apiVersion: "openebs.io/v1alpha1"
kind: MayastorPool
metadata:
  name: pool-3
  namespace: mayastor
spec:
  node: ksnode-3
  disks: ["aio:///dev/sdb"]
```

```bash
./bin/mayastor-client -a ksnode-2 bdev list
{
  "bdevs": [
    {
      "aliases": "",
      "blk_size": 512,
      "claimed": true,
      "claimed_by": "lvol",
      "name": "/dev/sdb",
      "num_blocks": 2097152,
      "uri": "",
      "uuid": "565e822a-6993-47f0-bb1b-c9668fff90b0"
    }
  ]
}
/bin/mayastor-client -a ksnode-3 bdev list
{
  "bdevs": [
    {
      "aliases": "aio:///dev/sdb",
      "blk_size": 512,
      "claimed": true,
      "claimed_by": "lvol",
      "name": "/dev/sdb",
      "num_blocks": 2097152,
      "uri": "aio:///dev/sdb?uuid=08ed0dc0-7028-4186-8e7a-5eb4e10b05a5",
      "uuid": "08ed0dc0-7028-4186-8e7a-5eb4e10b05a5"
    }
  ]
}
 ./bin/mayastor-client -a ksnode-3 pool list
NAME   STATE    CAPACITY USED DISKS
pool-3 online 1069547520    0 aio:///dev/sdb
```

PR #372, opened 2020-08-14
ci: add tests and start to move out common code

tjoshum:
There's still plenty to do, but I think there's value in merging this as it is and it'll already make developing further tests easier.

In particular, we need to move the import of the mayastor package to the same basis as the other 'system' nix ones, and find a better way of setting default values for the nodes.


nix/test/mayastorLib/MayastorUtils.py
```diff
@@ -0,0 +1,11 @@
+DEFAULT_NEXUS_PORT = "8420"
+DEFAULT_REPLICA_PORT = "8430"
+
+def subsystem_is_discoverable(host, ip, port, subsys):
```
paulyoong, line 4: Maybe prefix function name with "nvme" as this is nvme specific. We could possibly want to discover other subsystems in the future.

gila:
Nice @tjoshum it is starting to look like something now that can simply be used rather than struggled with. To make it even more agnostic, could you look into the actual test code being outside of the nix file as well? testScript='' <script>'' becomes a testScript = ./actualtest.py; 

this would make reviews in particlar easier in the future.

paulyoong (approved):
lgtm

tjoshum:
> Nice @tjoshum it is starting to look like something now that can simply use rather than struggled with. To make it even more agnostic, could you look into the actual test code being outside of the nix file as well? testScript='' <script>'' becomes a testScript = ./actualtest.py;
> 
> this would make reviews in particular easier in the future.

Tried with `builtins.readFile ./actualytest.py` and it doesn't allow us to substitute in things like targetIp. Will have to see if there's a way of getting nix to evaluate the string to get the substitutions to work properly.

nix/test/mayastorLib/MayastorUtils.py
tjoshum: Even better, I moved it out into an nvmfUtils file

tjoshum:
@gila I'm going to give up on getting the tests in .py files under this PR, in order to get these changes in.

tjoshum:
That said, you probably will be able to define a lot of stuff in the .py files under mayastorUtils.

tiagolobocastro (approved):

PR #374, opened 2020-08-17
CAS-332: rework the pool layer

gila:
In principle nothing changes, however the creation
of base_bdevs is implicitly done through bdev URI's
and lvols now Impl Share.

One new feature that has been added is that we can
can now set properties on lvols. This means that when
we create a lvol, and then share it, the fact that it
is shared is persisted on disk. On import the lvols are
automatically shared.

### DIFF
--- a/mayastor/src/lvs/pool.rs
+++ b/mayastor/src/lvs/pool.rs
@@ -1,0 +1,511 @@
+use std::{convert::TryFrom, fmt::Debug, os::raw::c_void, ptr::NonNull};
+
+use futures::channel::oneshot;
+use nix::errno::Errno;
+use pin_utils::core_reexport::fmt::Formatter;
+use tracing::instrument;
+
+use rpc::mayastor::CreatePoolRequest;
+use spdk_sys::{
+    lvol_store_bdev,
+    spdk_bs_free_cluster_count,
+    spdk_bs_get_cluster_size,
+    spdk_bs_total_data_cluster_count,
+    spdk_lvol,
+    spdk_lvol_store,
+    vbdev_get_lvol_store_by_name,
+    vbdev_get_lvs_bdev_by_lvs,
+    vbdev_lvol_create,
+    vbdev_lvol_store_first,
+    vbdev_lvol_store_next,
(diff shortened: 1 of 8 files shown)


gila:
BTW this does not change the existing grpc pool interaction yet, thats can only be done after this PR.

mayastor/src/lvs/pool.rs
```diff
@@ -0,0 +1,511 @@
...
+        match Self::import(&args.name, &bdev).await {
+            Ok(pool) => Ok(pool),
+            Err(_) => Self::create(&args.name, &bdev).await,
+        }
+    }
+
+    /// export the given lvl
+    #[allow(clippy::unit_arg)] // here to silence the () argument
+    #[instrument(level = "debug", err)]
+    pub async fn export(self) -> Result<(), Error> {
```
tiagolobocastro, line 327: I'm confused about this export... seems more like we're shutting down the pool?

mayastor/src/lvs/pool.rs
```diff
@@ -0,0 +1,511 @@
...
+                Some(Self::lvs_op_cb),
+                cb_arg(s),
+            )
+        };
+
+        r.await
+            .expect("callback gone while destroying lvs")
+            .to_result(|e| Error::Export {
+                err: Errno::from_i32(e),
+                msg: format!("failed to export pool {}", pool),
```
tiagolobocastro, line 421: should this be "failed to destroy the pool {}"?

mayastor/src/lvs/pool.rs
gila: Hmm, sorry @tiagolobocastro I just saw this comment while I was closing the pool request... Sorry about that.

export means, unload the pool, destroy the base_bdev -- but leave the data on disk intact such that it can be re-imported.

destroy means the same as above but you cant import it anymore

mayastor/src/lvs/pool.rs
gila: Yes, I will fix that later, thanks for spotting.

PR #375, opened 2020-08-17
CAS-378: Rebuild nix tests

paulyoong:
The test cases cover scenarios such as:
- network failure
- node crash

Once @tjoshum has created the structure for the nix tests (i.e. a common Python library) the rebuild tests can be restructured accordingly.

### DIFF
--- a/nix/test/rebuild/rebuild.nix
+++ b/nix/test/rebuild/rebuild.nix
@@ -1,0 +1,331 @@
+{ pkgs, lib, ... }:
+let
+  node1ip = "192.168.0.1";
+  node2ip = "192.168.0.2";
+  nexus_uuid = "19b98ac8-c1ea-11ea-8e3b-d74f5d324a22";
+  replica1_uuid = "9a9843db-f715-4f52-8aa4-119f5df3d05d";
+  replica2_uuid = "9a9843db-f715-4f52-8aa4-119f5df3d06e";
+in
+{
+  name = "rebuild";
+  meta = with pkgs.stdenv.lib.maintainers; {
+    maintainers = [ paulyoong ];
+  };
+
+  nodes = {
+    node1 = { config, lib, ... }: {
+
+      virtualisation = {
+        memorySize = 2048;
+        emptyDiskImages = [ 512 ];
+        vlans = [ 1 ];
+      };
+
+      boot = {
+        kernel.sysctl = {
+          "vm.nr_hugepages" = 512;
+        };
+      };
+
+      networking.firewall.enable = false;
+      networking.interfaces.eth1.ipv4.addresses = pkgs.lib.mkOverride 0 [
+        { address = node1ip; prefixLength = 24; }
+      ];
+
+      environment = {
+        systemPackages = with pkgs; [
+          images.mayastor-develop
+        ];
+
+        etc."mayastor-config.yaml" = {
+          mode = "0664";
+          source = ./node1-mayastor-config.yaml;
+        };
+      };
+
+      systemd.services.mayastor = {
+        enable = true;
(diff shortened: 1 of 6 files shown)


nix/test/rebuild/rebuild.nix
```diff
@@ -0,0 +1,331 @@
...
+  replica2_uuid = "9a9843db-f715-4f52-8aa4-119f5df3d06e";
+in
+{
+  name = "rebuild";
+  meta = with pkgs.stdenv.lib.maintainers; {
+    maintainers = [ paulyoong ];
+  };
+
+  nodes = {
+    node1 = { config, lib, ... }: {
```
tiagolobocastro, line 16: could you have a single node expression and pass the yaml as an input?

nix/test/rebuild/rebuild.nix
```diff
@@ -0,0 +1,331 @@
...
+
+    with subtest("resume a rebuild"):
+        exec_rebuild_operation("resume", replica2_uri)
+        rebuild_state = get_rebuild_state(replica2_uri)
+        expected_state = "running"
+        assert rebuild_state == expected_state, "Rebuild state {}, expected {}".format(
+            rebuild_state, expected_state
+        )
+
+    with subtest("get number rebuilds"):
```
tiagolobocastro, line 241: get number **of** rebuilds?

nix/test/rebuild/rebuild.nix
```diff
@@ -0,0 +1,331 @@
...
+        num_rebuilds = get_num_rebuilds()
+        expected_num_rebuilds = 1
+        assert (
+            num_rebuilds == expected_num_rebuilds
+        ), "Number of rebuilds {}, expected {}".format(num_rebuilds, expected_num_rebuilds)
+
+    with subtest("network fault"):
+        # Wait for some rebuild progress
+        while get_rebuild_progress_percentage() < 5:
+            sleep(1)
```
tiagolobocastro, line 251: Should this sleep be smaller, eg is it possible that the rebuild completes within this 1 second?

nix/test/rebuild/rebuild.nix
paulyoong: Maybe, but there are a number of node specific settings there also, so I'm not sure that would be appropriate. 

Having looked at a number of tests in the NixOS repo (e.g. https://github.com/NixOS/nixpkgs/blob/master/nixos/tests/3proxy.nix) the convention seems to be to specify each node configuration. Also the NixOS example seemed to suggest specifying individual machine configurations (https://nixos.org/nixos/manual/index.html#sec-writing-nixos-tests)

nix/test/rebuild/rebuild.nix
paulyoong: :+1: done

nix/test/rebuild/rebuild.nix
paulyoong: The volume has been chosen to be 100MiB so that the rebuild take sufficiently long for the tests. The quickest I have seen a volume this size being rebuilt is approximately 30 seconds.

paulyoong:
> I noticed you haven't added the test to the gh actions.
> It might be interesting to add this a scheduled workflow on the develop branch: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions ?

Good point - I've added it alongside the fio_nvme_basic test in nix-tests.yaml

nix/test/rebuild/rebuild.nix
tjoshum: The common/default stuff is part of what I'm looking at, hopefully I'll be able to get it generic enough to be useful here too, whilst still actually being a useful way of avoiding duplicate code.

PR #376, opened 2020-08-17
Proper mayastor debug package and code cleanup.

jkryl:
There are two flavours of mayastor nix package: mayastor and
mayastor-dev. The same for csi plugin package mayastor-csi and
mayastor-csi-dev. They don't get accidentaly confused by the build
system as was happening before with "release" parameter that made
creating the release packages very error-prone.

Overhaul of nix code for creating docker images. No code duplication
and intuitive target names. All image targets end with "-image" suffix.
The flavour of the package that the image is based on (if any) preceeds
the suffix (i.e. dev or adhoc). Nix code for moac image is part of
images now and follows the same rules.

We are able to derive the accurate version number of mayastor package
in fully automated way. It uses a git command to lookup a tag of the
top-most commit and if the tag does not exist, it uses short commit hash.
We don't need to update the files by hand to change the tag anymore.

Example of how to create release images for v0.3.0:
```
git checkout v0.3.0
nix-build --no-out-link -A images.mayastor-image
nix-build --no-out-link -A images.mayastor-csi-image
nix-build --no-out-link -A images.moac-image
```

and debug images for the same release:
```
git checkout v0.3.0
nix-build --no-out-link -A images.mayastor-dev-image
nix-build --no-out-link -A images.mayastor-csi-dev-image
nix-build --no-out-link -A images.moac-image
```

### DIFF
--- a/doc/build.md
+++ b/doc/build.md
@@ -209,6 +209,32 @@ Or, you can copy over the .so to `/usr/local/lib` or something similar.
 
 One this is done, you should be able to run `cargo build --all`
 
+## Building docker images
+
+Use NIX to build the images. Note that the images are based on NIX packages
+that are built as part of building the image. The tag of the image will be
+short commit hash of the top-most commit or a tag name of the commit if
+present. Example of building a moac package:
+
+```bash
+nix-build -A images.moac-image
+```
+
+At the end of the build is printed path to docker image tar archive. Load the
+image into Docker (don't use _import_ command) and run bash to poke around:
+
+```bash
+docker load -i /nix/store/hash-docker-image-moac.tar.gz
+docker run --rm -it image-hash /bin/bash
+```
+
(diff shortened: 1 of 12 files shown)


doc/build.md
```diff
@@ -209,6 +209,32 @@ Or, you can copy over the .so to `/usr/local/lib` or something similar.
...
+that are built as part of building the image. The tag of the image will be
+short commit hash of the top-most commit or a tag name of the commit if
+present. Example of building a moac package:
+
+```bash
+nix-build -A images.moac-image
+```
+
+At the end of the build is printed path to docker image tar archive. Import
+it to a docker (don't use _import_ command) and run bash to poke around:
```
jonathan-teh, line 224: Perhaps rephrase as "Load the image into Docker".

jkryl: done. thanks! 👍

PR #378, opened 2020-08-19
CAS-399 Ensure status field in NioCtx is always defined.

chriswldenyer:
Ensure the status field in the NioCtx gets defined when the Bio is created. 
This change also reduces some code duplication by adding a constructor for the Bio.

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_io.rs
+++ b/mayastor/src/bdev/nexus/nexus_io.rs
@@ -87,6 +87,14 @@ impl Bio {
         unsafe { Bdev::from((*self.0).bdev) }
     }
 
+    /// create and return a new, initialized Bio object
+    pub fn new(pio: *mut spdk_bdev_io, in_flight: i8) -> Bio {
+        let mut io = Bio(pio);
+        io.ctx_as_mut_ref().in_flight = in_flight;
+        io.ctx_as_mut_ref().status = io_status::SUCCESS;
+        io
+    }
+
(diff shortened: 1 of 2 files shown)


paulyoong (approved):

mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -87,6 +87,13 @@ impl Bio {
         unsafe { Bdev::from((*self.0).bdev) }
     }
 
+    pub fn new(pio: *mut spdk_bdev_io, in_flight: i8) -> Bio {
```
jonathan-teh, line 90: Should have a comment for each function.

chriswldenyer: Ok, I'll fix that.

jonathan-teh (approved):
Looks good although I wonder if it should be fixed in the caller, io_submit, instead.

chriswldenyer:
> Looks good although I wonder if it should be fixed in the caller, io_submit, instead.

Maybe, but the problem with that is that the in_flight value varies per io_type, so the caller would have to take the io_type into account. This way the individual io functions look after that.

jonathan-teh:
> Maybe, but the problem with that is that the in_flight value varies per io_type, so the caller would have to take the io_type into account. This way the individual io functions look after that.

True, but I should have been more precise, I meant the initialisation of the status field could be fixed in the caller.

PR #370, opened 2020-08-13
CAS-330: Expose lvol snapshot functionality

jonathan-teh:
This implements basic functionality to allow a Nexus to create
a snapshot on a replica using an NVMe Admin command.

Add NVME_ADMIN as a permitted IO type and implement NVMe Admin
Passthru starting from BdevHandle as a standalone function until it
reaches nexus_bdev. Creating a snapshot sends an admin command with
opcode c0h and the current Unix time in cdw10/11.

Register a custom command handler for NVMe Admin opcode c0h with SPDK
to handle snapshot creation when initialising the NVMf subsystem.
The handler creates a snapshot against the replica (lvol) by calling
vbdev_lvol_create_snapshot with a snapshot name derived from the
timestamp. Complete the NVMf request in the snapshot done callback.
The snapshot is then available in the list of replicas.

Add a test of NVMe Admin commands first with an unsupported opcode,
followed by creating a snapshot, given a BdevHandle, on a Nexus.

PR #379, opened 2020-08-20
snapshots: Add regression test via initiator

jonathan-teh:
Add a create-snapshot command for the standalone initiator binary
and have test-replica use that to create a snapshot and verify that it
appears in the list of replicas.
Extends #370.

### DIFF
--- a/mayastor-test/test_replica.js
+++ b/mayastor-test/test_replica.js
@@ -691,6 +691,31 @@ describe('replica', function () {
       );
     });
 
+    it('should take snapshot on nvmf replica', (done) => {
+      common.execAsRoot(
+        common.getCmdPath('initiator'),
+        [uri, 'create-snapshot'],
+        done
+      );
+    });
+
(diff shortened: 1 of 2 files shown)


mayastor-test/test_replica.js
```diff
@@ -691,6 +691,31 @@ describe('replica', function () {
...
+      client.listReplicas({}, (err, res) => {
+        if (err) return done(err);
+        res = res.replicas.filter((ent) => {
+          return ent.uuid.startsWith(UUID + '-snap-');
+        });
+        assert.lengthOf(res, 1);
+        res = res[0];
+        assert.equal(res.pool, POOL);
+        assert.equal(res.thin, false);
+        assert.equal(res.size, 96 * 1024 * 1024);
```
chriswldenyer, line 712: Someone should define this as a constant somewhere, but maybe not now.

chriswldenyer (approved):
LGTM

tiagolobocastro (approved):

PR #381, opened 2020-08-20
grpc: make use of the new lvs and lvol implementation

gila:
This reworks the mayastor svc layer to make use of the new lvs and lvol layer.

In short it makes the code far more specific and aligned to what we do from an
upper layer. This makes things less flexible, but is a good thing in terms of
what we do with it.

For example, lvols are always shared over nvmf, so creating an replica will do
this now implicitly and the ability to create a replica, that is not shared has
been removed. If we create the lvol, but fail to share, we delete the lvol,
before we return the error.

The grpc methods themselves are implemented in pool_grpc this it to keep the
registration of the events to the grpc server itself clean, but also, it allows
us to test the grpc methods without the need of a grpc client as can bee seen
in lvs_pool_rpc

lvols that are shared (should) now be re-shared automatically when
we load/import the lvs again.


mayastor-test/test_csi.js
```diff
@@ -29,10 +29,13 @@ var csiSock = common.CSI_ENDPOINT;
...
-NumberOfLuns 1
-LunSizeInMB  150
-BlockSize    4096
+sync_disable: true
+base_bdevs:
+  - uri: "malloc:///malloc0?size_mb=64&11111111-0000-0000-0000-00000000000"
+  - uri: "malloc:///malloc1?size_mb=64&11111111-0000-0000-0000-00000000001"
+  - uri: "malloc:///malloc2?size_mb=64&11111111-0000-0000-0000-00000000002"
+  - uri: "malloc:///malloc3?size_mb=64&11111111-0000-0000-0000-00000000003"
+  - uri: "malloc:///malloc4?size_mb=64&11111111-0000-0000-0000-00000000004"
```
cjones1024, line 38: Shouldn't these include a `uuid=` like so:  `uri: "malloc:///malloc0?size_mb=64&uuid=11111111-0000-0000-0000-00000000000"`

PR #386, opened 2020-08-23
gRPC lvs layer

gila:
3 commits, please see the respective commit messages for details. But this is revised PR based on some backchannel, chats with @jkryl and @tiagolobocastro 

Running this on k8s works as expected.

### DIFF
--- a/mayastor-test/test_common.js
+++ b/mayastor-test/test_common.js
@@ -193,7 +193,7 @@ function startMayastor (config, args, env, yaml) {
   env = env || {};
 
   if (yaml) {
-    fs.writeFileSync(MS_CONFIG_PATH, yaml,  { mode : 0o777});
+    fs.writeFileSync(MS_CONFIG_PATH, yaml);
     args = args.concat(['-y', MS_CONFIG_PATH]);
   }
 
--- a/mayastor-test/test_common.js
+++ b/mayastor-test/test_common.js
@@ -188,9 +188,14 @@ function startSpdk (config, args, env) {
 }
 
 // Start mayastor process and return immediately.
-function startMayastor (config, args, env) {
+function startMayastor (config, args, env, yaml) {
   args = args || ['-r', SOCK, '-g', grpcEndpoint];
   env = env || {};
+
+  if (yaml) {
+    fs.writeFileSync(MS_CONFIG_PATH, yaml,  { mode : 0o777});
+    args = args.concat(['-y', MS_CONFIG_PATH]);
+  }
 
   if (config) {
     fs.writeFileSync(MS_CONFIG_PATH, config);
--- a/mayastor-test/test_csi.js
+++ b/mayastor-test/test_csi.js
@@ -29,10 +29,13 @@ var csiSock = common.CSI_ENDPOINT;
 
 // One big malloc bdev which we put lvol store on.
 const CONFIG = `
-[Malloc]
-NumberOfLuns 1
-LunSizeInMB  150
-BlockSize    4096
+sync_disable: true
+base_bdevs:
+  - uri: "malloc:///malloc0?size_mb=64&uuid=11111111-0000-0000-0000-000000000000&blk_size=4096"
+  - uri: "malloc:///malloc1?size_mb=64&uuid=11111111-0000-0000-0000-000000000001&blk_size=4096"
+  - uri: "malloc:///malloc2?size_mb=64&uuid=11111111-0000-0000-0000-000000000002&blk_size=4096"
+  - uri: "malloc:///malloc3?size_mb=64&uuid=11111111-0000-0000-0000-000000000003&blk_size=4096"
+  - uri: "malloc:///malloc4?size_mb=64&uuid=11111111-0000-0000-0000-000000000004&blk_size=4096"
 `;
 // uuid without the last digit
 const BASE_UUID = '11111111-0000-0000-0000-00000000000';
@@ -110,7 +113,7 @@ describe('csi', function () {
   // NOTE: Don't use mayastor in setup - we test CSI interface and we don't want
   // to depend on correct function of mayastor iface in order to test CSI.
   before((done) => {
-    common.startMayastor(CONFIG);
+    common.startMayastor(null, null, null,  CONFIG);
     common.startMayastorCsi();
 
     var client = common.createGrpcClient();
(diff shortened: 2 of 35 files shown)


mayastor/src/lvs/lvs_pool.rs
```diff
@@ -309,15 +349,35 @@ impl Lvs {
...
+                Err(Error::Import {
+                    source,
+                    name,
+                })
+            }
+            // else some other error, try to create the the pool
+            Err(Error::Import {
+                source, ..
+            }) if source == Errno::EILSEQ => {
+                Self::create(&args.name, &bdev).await
```
tiagolobocastro, line 375: if the import or the create fails should we destroy the base bdev (though only for the case where we've just created it) ?

tiagolobocastro (approved):

mayastor/src/core/thread.rs
```diff
@@ -148,7 +147,35 @@ impl Mthread {
...
+                .into_iter()
+                .for_each(|i| libc::CPU_CLR(i as usize, &mut set));
+
+            libc::sched_setaffinity(
+                0,
+                std::mem::size_of::<libc::cpu_set_t>(),
+                &set,
+            );
+
+            warn!("thread pinned to core {}", libc::sched_getcpu());
```
tiagolobocastro, line 178: should this be <warn ?

mayastor/src/lvs/lvs_pool.rs
gila: Yes good idea, I've added the deletion it and error!() in the logs. If we fail to delete, not sure what else we can do...

mayastor/src/core/thread.rs
gila: change the level and message a bit :+1:

mayastor/src/lvs/lvs_pool.rs
```diff
@@ -45,6 +45,7 @@ impl From<*mut spdk_lvol_store> for Lvs {
 /// iterator over all lvol stores
 pub struct LvsIterator(*mut lvol_store_bdev);
 
+/// returns a new
```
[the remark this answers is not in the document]
gila: :see_no_evil:

mayastor-test/test_common.js
```diff
@@ -188,10 +188,15 @@ function startSpdk (config, args, env) {
 }
 
 // Start mayastor process and return immediately.
-function startMayastor (config, args, env) {
+function startMayastor (config, args, env, yaml) {
   args = args || ['-r', SOCK, '-g', grpcEndpoint];
   env = env || {};
 
+  if (yaml) {
+    fs.writeFileSync(MS_CONFIG_PATH, yaml,  { mode : 0o777});
```
jonathan-teh, line 196: Config files don't need to be executable.

jonathan-teh (approved):
Looks fine, just a minor nit.

mayastor/src/grpc/pool_grpc.rs
```diff
@@ -0,0 +1,238 @@
...
+            }
+        } else {
+            Err(LvsError::InvalidBdev {
+                source: NexusBdevError::BdevNotFound {
+                    name: args.uuid.clone(),
+                },
+                name: args.uuid,
+            })
+        }
+    })
```
jkryl, line 237: looks good except minor issues with logging. We used to log a debug message at the start of each grpc method and info message for all methods that change the state of mayastor when completed. That is handy for seeing operations that either hang or take too long easily in the log file. I suppose that the debug message at the beginning is replaced by instrument macro now so that is probably ok. But I'm missing the info message. Or are we sure that messages get logged for all replica and pool methods somewhere deeper in the modules that get called? Also the previous log messages were nice to users and speak their language  (about pools, replicas) as opposed using terms lvols and lvs that people not familiar with SPDK won't have idea what they mean.

mayastor/src/replica.rs
```diff
@@ -2,7 +2,7 @@
 //!
 //! Replica is a logical data volume exported over nvmf (in SPDK terminology
 //! an lvol). Here we define methods for easy management of replicas.
-
+#![allow(dead_code)]
```
jkryl, line 5: is it that the whole file is not used or just part of it?

mayastor-test/test_csi.js
```diff
@@ -310,7 +287,6 @@ describe('csi', function () {
   });
 
   csiProtocolTest('NBD', enums.NEXUS_NBD, 10000);
-  // TODO: find out why it takes 2 minutes to execute the tests
```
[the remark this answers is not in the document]
gila: It should not Jan...

mayastor/src/grpc/pool_grpc.rs
gila: That's still there when the debug level is set, let me know what you are missing.

mayastor/src/replica.rs
gila: Once this PR lands, it should be removed.

mayastor/src/grpc/pool_grpc.rs
gila: I misread it the OK messages as what you are referring to, I'll have a good look :+1: 

As for the terms, the messages still speak of pools and replica's although I would argue those concepts are somehow more familiar to them unless they are familiar with zfs :)

mayastor/src/core/bdev.rs
```diff
@@ -127,7 +130,7 @@ impl Share for Bdev {
         match self.shared() {
             Some(Protocol::Nvmf) => nvmf::get_uri(&self.name()),
             Some(Protocol::Iscsi) => iscsi::get_uri(Side::Nexus, &self.name()),
-            None => None,
+            _ => Some(format!("bdev:///{}", self.name())),
```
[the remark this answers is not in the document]
gila: According to the RFC https://www.ietf.org/rfc/rfc1738.txt it should be there. The reasoning is if its "localhost" the "hostname" can be omitted. IOW; bdev://localhost/whatever => bdev:///whatever.

mayastor-test/test_replica.js
```diff
@@ -346,7 +305,6 @@ describe('replica', function () {
           assert.lengthOf(res, 1);
           res = res[0];
           assert.equal(res.share, 'REPLICA_NVMF');
-          assert.match(res.uri, NVMF_URI);
```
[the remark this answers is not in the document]
gila: will fix :+1:

cjones1024 (approved):

PR #387, opened 2020-08-24
CAS-402: Fix cancel_child_rebuild_jobs

paulyoong:
The cancel_child_rebuild_jobs function was, among other things,
cancelling all rebuild jobs with the child as a source, and then
restarting them with the intention that they would use a different
source child. In reality, the same source child could be chosen as the
state of the child was typically updated after the call to this function.

This fix splits the function into two separate cancel and start
functions. The caller is responsible for first cancelling the rebuilds,
then updating the child state appropriately before finally restarting
any previously cancelled rebuild jobs.

Additional fixes:
1. Fix the rebuild_lookup test to iterate over all "non-added" children.
2. Fix the uri in the destination handle error message.

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_bdev_children.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev_children.rs
@@ -217,7 +217,8 @@ impl Nexus {
             });
         }
 
-        self.cancel_child_rebuild_jobs(uri).await;
+        let cancelled_rebuilding_children =
+            self.cancel_child_rebuild_jobs(uri).await;
 
         let idx = match self.children.iter().position(|c| c.name == uri) {
             None => return Ok(()),
@@ -234,7 +235,10 @@ impl Nexus {
         child.destroy().await.context(DestroyChild {
             name: self.name.clone(),
             child: uri,
-        })
+        })?;
+
+        self.start_rebuild_jobs(cancelled_rebuilding_children).await;
+        Ok(())
     }
 
     /// offline a child device and reconfigure the IO channels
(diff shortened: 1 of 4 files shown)


blaisedias (approved):
Looks okay to me.

mayastor/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -217,7 +217,8 @@ impl Nexus {
             });
         }
 
-        self.cancel_child_rebuild_jobs(uri).await;
+        let cancelled_rebuilding_children =
+            self.cancel_child_rebuild_jobs(uri).await;
 
         let idx = match self.children.iter().position(|c| c.name == uri) {
             None => return Ok(()),
```
tiagolobocastro, line 224: if we return here then we don't restart the rebuilds? though I guess that's ok as we wouldn't have cancelled anything if the child is not present

mayastor/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -234,7 +235,10 @@ impl Nexus {
         child.destroy().await.context(DestroyChild {
             name: self.name.clone(),
             child: uri,
-        })
+        })?;
```
tiagolobocastro, line 238: if we fail to destroy here we don't restart the rebuilds?

jonathan-teh (approved):

mayastor/src/bdev/nexus/nexus_bdev_children.rs
paulyoong: Good spot - fixed

mayastor/src/bdev/nexus/nexus_bdev_children.rs
paulyoong: Agreed - if the child isn't present we shouldn't be using it in any rebuilds.

tiagolobocastro (approved):

PR #388, opened 2020-08-27
CAS-424: import volumes on startup

tiagolobocastro:
Moac should not attempt to recreate new volumes based on the existing
CRDs on startup as this could lead to having a completely new data.
Rather, it now simply imports them and attaches existing replicas.

If a replica is currently not online then at the moment it will be
removed. Unfortunately there is an issue when trying to add a replica
from the same node again, see cas-426

### DIFF
--- a/csi/moac/volume.js
+++ b/csi/moac/volume.js
@@ -23,8 +23,10 @@ class Volume {
   // @params {string[]} spec.requiredNodes   Replicas must be on these nodes.
   // @params {number}   spec.requiredBytes   The volume must have at least this size.
   // @params {number}   spec.limitBytes      The volume should not be bigger than this.
+  // @params {string}   spec.protocol        The share protocol for the nexus.
+  // @params {object}   [size=0]             Current properties of the volume.
   //
-  constructor (uuid, registry, spec) {
+  constructor (uuid, registry, spec, size = 0) {
     assert(spec);
     // specification of the volume
     this.uuid = uuid;
@@ -34,7 +36,8 @@ class Volume {
     this.requiredNodes = _.clone(spec.requiredNodes || []).sort();
     this.requiredBytes = spec.requiredBytes;
     this.limitBytes = spec.limitBytes;
-    this.size = 0;
+    this.protocol = spec.protocol;
+    this.size = size;
     // state variables of the volume
     this.nexus = null;
     this.replicas = {}; // replicas indexed by node name
@@ -545,6 +548,7 @@ class Volume {
   // @params {string[]} spec.requiredNodes   Replicas must be on these nodes.
   // @params {number}   spec.requiredBytes   The volume must have at least this size.
   // @params {number}   spec.limitBytes      The volume should not be bigger than this.
+  // @params {string}   spec.protocol        The share protocol for the nexus.
   // @returns {boolean} True if the volume spec has changed, false otherwise.
   //
   update (spec) {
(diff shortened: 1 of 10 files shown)


csi/moac/volume.js
```diff
@@ -23,8 +23,10 @@ class Volume {
   // @params {string[]} spec.requiredNodes   Replicas must be on these nodes.
   // @params {number}   spec.requiredBytes   The volume must have at least this size.
   // @params {number}   spec.limitBytes      The volume should not be bigger than this.
+  // @params {string}   spec.protocol        The share protocol for the nexus.
+  // @params  {object}  status               Current properties of the volume
   //
-  constructor (uuid, registry, spec) {
+  constructor (uuid, registry, spec, status) {
```
jkryl, line 29: just for your consideration: if we pass an optional argument size (instead of the whole status), then apis become more clear and we don't leak internal structures from k8s to the rest of the code. If this function is changed then Import volume function can use the size instead of status too.

csi/moac/volume.js
```diff
@@ -23,8 +23,10 @@ class Volume {
   // @params {string[]} spec.requiredNodes   Replicas must be on these nodes.
   // @params {number}   spec.requiredBytes   The volume must have at least this size.
   // @params {number}   spec.limitBytes      The volume should not be bigger than this.
+  // @params {string}   spec.protocol        The share protocol for the nexus.
+  // @params  {object}  status               Current properties of the volume
```
jkryl, line 27: Optional arg should have `[...]` around it.

csi/moac/pool.ts
```diff
@@ -129,7 +129,9 @@ export class Pool {
   // @param {object} replica      New replica object.
   //
   registerReplica(replica: Replica) {
-    assert(!this.replicas.find((r) => r.uuid === replica.uuid));
+    if (this.replicas.find((r) => r.uuid === replica.uuid)) {
+      this.unregisterReplica(replica)
+    }
```
jkryl, line 134: Is there a particular reason for handling this gracefully? Adding a replica that already exists to a pool could be a programming mistake and in that case, assert would be a better option for dealing with that.

csi/moac/volume.js
```diff
@@ -34,12 +36,17 @@ class Volume {
...
+    this.protocol = spec.protocol;
     this.size = 0;
     // state variables of the volume
     this.nexus = null;
     this.replicas = {}; // replicas indexed by node name
     this.state = 'pending';
     this.runFsa = 0; // number of requests to run FSA
+
+    if (status) {
+      this.size = status.size;
```
jkryl, line 48: it seems we don't have a test case where we would actually pass the size of the volume and test it is what it is expected to be.

csi/moac/volume.js
```diff
@@ -562,6 +570,12 @@ class Volume {
         `Shrinking the volume "${this}" is not supported`
       );
     }
+    if (this.protocol !== spec.protocol) {
+      throw new GrpcError(
+        GrpcCode.INVALID_ARGUMENT,
+        `Changing the protocol for volume "${this}" is not supported`
+      );
+    }
```
jkryl, line 578: Can you add a test case for this case that ends up with an exception?

jkryl (approved):
The changes look good to me. Just a couple of nits ..

csi/moac/volume.js
tiagolobocastro: yep that makes sense, will do.

csi/moac/volume.js
tiagolobocastro: thanks will do

csi/moac/pool.ts
tiagolobocastro: I noticed that this can happen when a node/mayastor is brought up and moac tries to recreate the replica. Seems like we add the replica through the response of create replica but we also pick it up again from the events?

csi/moac/volume.js
tiagolobocastro: good catch I'll add it

csi/moac/volume.js
tiagolobocastro: :+1:

gila (approved):

PR #389, opened 2020-08-31
Want simple performance tool

gila:
This tool allows us to pump IO full speed to supported storage URI's.
The intent of this is to be able to test performance without the need
for setting up any targets and, subsequently, benchmark tools.

Its nowhere near as sophisticated as fio, but that's also not the
intent. The intent is to get good numbers but also, to assist in
testing/developing the handling of IO errors and rebuild.

The PR is a bit noisy as I removed some of the (needless) cast operations. This was an artifact still of the many upgraded we have done in the past. 

Right now, only READs are implemented (it is safer to develop)

### DIFF
--- a/mayastor/src/bdev/dev.rs
+++ b/mayastor/src/bdev/dev.rs
@@ -56,12 +56,10 @@ impl Uri {
 
             // retain this for the time being for backwards compatibility
             "bdev" => Ok(Box::new(loopback::Loopback::try_from(&url)?)),
-
-            // backend iSCSI target - most stable
-            "iscsi" => Ok(Box::new(iscsi::Iscsi::try_from(&url)?)),
-
             // arbitrary bdev found in spdk (used for local replicas)
             "loopback" => Ok(Box::new(loopback::Loopback::try_from(&url)?)),
+            // backend iSCSI target - most stable
+            "iscsi" => Ok(Box::new(iscsi::Iscsi::try_from(&url)?)),
 
             // backend NVMF target - fairly unstable (as of Linux 5.2)
             "nvmf" => Ok(Box::new(nvmf::Nvmf::try_from(&url)?)),
--- a/mayastor/src/bin/casperf.rs
+++ b/mayastor/src/bin/casperf.rs
@@ -1,0 +1,403 @@
+use std::{cell::RefCell, os::raw::c_void, ptr::NonNull};
+
+use clap::{value_t, App, AppSettings, Arg};
+use rand::Rng;
+
+use mayastor::{
+    core::{
+        mayastor_env_stop,
+        Bdev,
+        Cores,
+        Descriptor,
+        DmaBuf,
+        IoChannel,
+        MayastorCliArgs,
+        MayastorEnvironment,
+        Mthread,
+        Reactors,
+    },
+    logger,
+    nexus_uri::bdev_create,
+    subsys::Config,
+};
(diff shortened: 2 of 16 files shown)


jkryl (approved):
neat! 🎉

mayastor/src/bin/casperf.rs
```diff
@@ -0,0 +1,405 @@
...
+                        NonNull::new(spdk_sys::spdk_poller_register(
+                            Some(perf_tick),
+                            std::ptr::null_mut(),
+                            1_000_000,
+                        ))
+                });
+            }
+        });
+
+        // kick of a task for each core if wanted here
```
paulyoong, line 402: typo - "kick **off**"

paulyoong (approved):
Thanks for this tool :+1:

mayastor/src/bdev/dev.rs
```diff
@@ -56,12 +56,10 @@ impl Uri {
 
             // retain this for the time being for backwards compatibility
             "bdev" => Ok(Box::new(loopback::Loopback::try_from(&url)?)),
-
-            // backend iSCSI target - most stable
-            "iscsi" => Ok(Box::new(iscsi::Iscsi::try_from(&url)?)),
-
             // arbitrary bdev found in spdk (used for local replicas)
             "loopback" => Ok(Box::new(loopback::Loopback::try_from(&url)?)),
+            // backend iSCSI target - most stable
```
tiagolobocastro, line 61: tiny nit - newline between loopback and iscsi?

mayastor/src/bin/casperf.rs
```diff
@@ -0,0 +1,405 @@
...
+    spdk_poller,
+    spdk_poller_unregister,
+};
+
+#[derive(Debug)]
+enum IoType {
+    /// perform random read operations
+    READ,
+    /// perform random write operations
+    #[allow(dead_code)]
```
tiagolobocastro, line 36: `Write` seems to be used, is this dead_code allow needed?

tiagolobocastro (approved):

mayastor/src/bin/casperf.rs
gila: Ha! -- you cant fool the compiler =) "used" means actually used not just typed in the code somewhere, so no, it will complain :)

mayastor/src/core/thread.rs
```diff
@@ -147,6 +146,43 @@ impl Mthread {
...
+
+        // helper routine to unpack the closure and its arguments
+        extern "C" fn trampoline<F, T>(arg: *mut c_void)
+        where
+            F: FnMut(T),
+            T: 'static + std::fmt::Debug,
+        {
+            let mut ctx = unsafe { Box::from_raw(arg as *mut Ctx<F, T>) };
+            (ctx.closure)(ctx.args);
+        }
```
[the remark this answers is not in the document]
gila: Maybe -- but I don't think so, or at least, not sure how that would look. The function is scoped within the msg() method itself, and its generics and the trait bounds are for now alteast, specific to the send_message() FFI function. If we see more if this then perhaps we can move it to FFI helpers.

mayastor/src/bin/casperf.rs
gila: i removed the whole comment as it was a leftover thanks :+1:

PR #390, opened 2020-09-01
CAS-407: moac pool resource ignored

tiagolobocastro:
Sync Nodes with the pools on the 'mod' event as well as the 'sync'.

'sync' covers the case where the node has already been added to the
registry and synced which. In this case the event stream will replay
this and so we need to pre-populated the resource list.

'mod' is the other case where the pool custom resource gets processed
before the node is synced or when it is down.

Added a log message when resources are pre-populated which as otherwise
we might not see any indication of pools being processed.


jkryl (approved):
thanks!

cjones1024 (approved):

PR #391, opened 2020-09-01
Add mocha tests for CSI block volumes. CAS-379

blaisedias:
Also add an idempotency fix to block volume publishing.
We now use the findmnt utility to check if the device is already
mounted.
A side effect of that change is that we return an error in the
unlikely case that another device is mounted at the same location.

### DIFF
--- a/csi/src/block_vol.rs
+++ b/csi/src/block_vol.rs
@@ -84,22 +84,7 @@ pub async fn publish_block_volume(
             }
         }
 
-        let devt = unsafe { libc::makedev(259, 254) };
-
-        let cstr_dst = std::ffi::CString::new(target_path.as_str()).unwrap();
-        let res =
-            unsafe { libc::mknod(cstr_dst.as_ptr(), libc::S_IFBLK, devt) };
-
-        if res != 0 {
-            let e = nix::errno::errno();
-            return Err(failure!(
-                Code::Internal,
-                "Failed to publish volume {}: mknod at {} failed: {}",
-                volume_id,
-                target_path,
-                e
-            ));
-        }
+        std::fs::File::create(&target_path)?;
 
         if let Err(error) = mount::blockdevice_mount(
             &device_path,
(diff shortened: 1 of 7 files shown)


mayastor-test/test_csi.js
```diff
@@ -80,6 +80,16 @@ function createPublishDir (mountTarget) {
...
 }
 
+function cleanBlockMount (blockfile, done) {
+  const proc = common.runAsRoot('umount', ['-f', blockfile]);
+  proc.once('close', (code, signal) => {
+    const proc = common.runAsRoot('rm', ['-f', blockfile]);
+    proc.once('close', (code, signal) => {
+      done();
+    });
+  });
```
jkryl, line 90: consider using execAsRoot for these that are simpler to use for one-time commands. runAsRoot is meant for long running processes that run in parallel to the test code (i.e. running mayastor).

jkryl (approved):
looks good

csi/src/block_vol.rs
```diff
@@ -49,21 +53,41 @@ pub async fn publish_block_volume(
...
-
-        if res != 0 {
-            return Err(failure!(
-                Code::Internal,
-                "Failed to publish volume {}: mknod at {} failed",
-                volume_id,
-                target_path
-            ));
+        // Idempotency, if we have done this already just return success.
+        match findmnt_device(target_path) {
```
cjones1024, line 57: I think that we should only perform this check if target_path actually exists - and then only if it is a device special and not a regular file. Also, we should create the file only if it did not exist previously - rather than implicitly truncating an existing file. If the file exists already and has size 0 we can just bind mount it. If it exists and has non-zero size we should probably abort as truncating such a file means potential data loss.

mayastor-test/test_csi.js
blaisedias: Done. https://github.com/openebs/Mayastor/pull/391/commits/72f4419b538bb6f8d967e948843a065087212eba

cjones1024 (approved):

PR #392, opened 2020-09-02
CAS-434: fix pool status information

tiagolobocastro:
the pools were being exported in "legacy" mode and therefore when
imported some parameters were missing (uuid, alias)..

### DIFF
--- a/mayastor/src/subsys/config/mod.rs
+++ b/mayastor/src/subsys/config/mod.rs
@@ -30,7 +30,7 @@ use spdk_sys::{
 
 use crate::{
     bdev::{nexus::instances, nexus_create, VerboseError},
-    core::{Bdev, Cores, Reactor},
+    core::{Bdev, Cores, Reactor, Share},
     jsonrpc::{jsonrpc_register, Code, RpcErrorCode},
     nexus_uri::bdev_create,
     pool::{create_pool, PoolsIter},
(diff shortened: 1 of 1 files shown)


gila (approved):

mayastor/src/subsys/config/mod.rs
```diff
@@ -286,7 +286,10 @@ impl Config {
         let pools = PoolsIter::new()
             .map(|p| Pool {
                 name: p.get_name().into(),
-                disks: vec![p.get_base_bdev().name()],
+                disks: vec![p
+                    .get_base_bdev()
+                    .bdev_uri()
+                    .unwrap_or_else(|| p.get_base_bdev().name())],
                 blk_size: p.get_base_bdev().block_len(),
```
cjones1024, line 293: is it worth abstracting out the repeated calls to get_base_bdev()
ie.
`.map(|p| { let base = p.get_base_bdev() ; Pool { ... } }`

tiagolobocastro: I avoided it because it makes the change a bit more "verbose" but maybe it's better. I've pushed it.

mayastor/src/subsys/config/mod.rs
```diff
@@ -284,17 +284,20 @@ impl Config {
...
-                        name: p.get_uuid().to_string(),
-                        share: p.get_share_type(),
-                    })
-                    .collect::<Vec<_>>(),
+            .map(|p| {
+                let base = p.get_base_bdev();
+                Pool {
+                    name: p.get_name().into(),
+                    disks: vec![base.bdev_uri().unwrap_or_else(|| base.name())],
+                    blk_size: p.get_base_bdev().block_len(),
```
cjones1024, line 292: use base.block_len() as well?

cjones1024 (approved):

PR #393, opened 2020-09-02
New debug package libspdk-dev.

jkryl:
We used "debugEnable" parameter passed to libspdk's default.nix, but
that did not work quite well. Sometimes it could have happened that
both debug and production derivations had the same output hash which
lead to unpredictable build results (don't ask me how). By having two
packages (one for production and another for debugging) it became
clearer which one is used where.

Notable change is that nix-shell starts with libspdk-dev which is what
most people expect for development environment.


tiagolobocastro (approved):

paulyoong (approved):
lgtm

PR #394, opened 2020-09-03
Add nix network disconnection test.

chriswldenyer:
Test verifies that IOs to the nexus (from fio) hang for at least
60s when the network connection between the nexus and a replica
remains disconnected.
This test is there for development purposes and is not included in the CI tests.


nix/test/disconnect/disconnect.nix
```diff
@@ -0,0 +1,60 @@
+{ pkgs, lib, ... }:
+let
+  targetIp = "192.168.0.1";
+  initiatorIp = "192.168.0.2";
+  common = import ../common.nix { inherit pkgs; };
+in
+{
+  name = "fio_against_nvmf_target";
+  meta = with pkgs.stdenv.lib.maintainers; {
+    maintainers = [ gila ];
```
paulyoong, line 10: should the maintainer name be you?

jkryl (approved):

paulyoong (approved):

nix/test/disconnect/disconnect.nix
tiagolobocastro: can we remove those entirely? we should all be the maintainers

chriswldenyer: Ok, I'll remove that.

paulyoong (approved):

PR #396, opened 2020-09-04
Lazy instance has previously been poisoned in rust tests (CAS-432)

jkryl:
Reactor test required the test system to have more than 4 cpus
otherwise test threads panic'd but not the main thread. Most of
the time the test passed while producing strange panic stacks in
the log, but sometimes the test failed with cryptic error message
in the description.

The requirement has changed to having at least 2 vcpus that most
likely all test machines will have. The problem with an error that
is not propagated to the main thread will be solved on behalf of
a different ticket.

### DIFF
--- a/mayastor/tests/reactor.rs
+++ b/mayastor/tests/reactor.rs
@@ -14,11 +14,12 @@ use pin_utils::core_reexport::time::Duration;
 
 pub mod common;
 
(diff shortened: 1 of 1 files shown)


blaisedias (approved):
LGTM

gila (approved):

PR #398, opened 2020-09-04
CAS-17: auto pool names

tiagolobocastro:
Use `generateName` in the msp example in the deployment directory.
With this users won't have to give their pools unique names

### DIFF
--- a/deploy/pool.yaml
+++ b/deploy/pool.yaml
@@ -1,7 +1,8 @@
 apiVersion: "openebs.io/v1alpha1"
 kind: MayastorPool
 metadata:
-  name: pool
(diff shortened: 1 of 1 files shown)


paulyoong (approved):

cjones1024 (approved):
Do we need to document anywhere that this also means that "kubectl apply" cannot now be used to create the pool and "kubectl create" must be used instead.

tiagolobocastro:
> Do we need to document anywhere that this also means that "kubectl apply" cannot now be used to create the pool and "kubectl create" must be used instead.

The docs in deploy already use create so I guess we're ok

PR #401, opened 2020-09-04
Detach on failed fs mount. CAS-429

blaisedias:
To stage a file system volume we first attach the nexus,
then do a file system mount.
Now if the file system mount fails, detach the nexus.


csi/src/node.rs
```diff
@@ -96,6 +96,32 @@ fn get_access_type(
...
+/// Detach the nexus device from the system, either at volume unstage,
+/// or after failed filesystem mount at volume stage.
+async fn detach(uuid: &Uuid, errheader: String) -> Result<(), Status> {
+    if let Some(device) = Device::lookup(uuid).await.map_err(|error| {
+        failure!(
+            Code::Internal,
+            "{} error locating device: {}",
+            &errheader,
+            error
+        )
```
paulyoong, line 108: Should we include the uuid of the device in this error?

cjones1024 (approved):

paulyoong (approved):

csi/src/node.rs
blaisedias: That information is included in the `errheader` string, i.e. `msg.volume_id`.  Invocations of the `detach` function on lines 485 and 546.

tiagolobocastro (approved):

PR #404, opened 2020-09-07
Added retry handling for the nexus IO path.

chriswldenyer:
This is set by the configuration option max_io_attempts.
If greater than 1 errors returned to io_completion()
will cause the same IO to be retried until the total
number of attempts would exceed max_io_attempts.
The default value is 1 which does not change the existing behaviour.
A cargo-based test is added to verify this function.

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_channel.rs
+++ b/mayastor/src/bdev/nexus/nexus_channel.rs
@@ -130,6 +130,8 @@ impl NexusChannel {
         let nexus = unsafe { Nexus::from_raw(device) };
         debug!("{}: Creating IO channels at {:p}", nexus.bdev.name(), ctx);
 
+        let max_io_attempts = nexus.max_io_attempts;
+
         let ch = NexusChannel::from_raw(ctx);
         let mut channels = Box::new(NexusChannelInner {
             ch: Vec::new(),
@@ -144,7 +146,9 @@ impl NexusChannel {
             .filter(|c| c.status() == ChildStatus::Online)
             .map(|c| {
                 channels.ch.push(
-                    BdevHandle::try_from(c.get_descriptor().unwrap()).unwrap(),
+                    BdevHandle::try_from(c.get_descriptor().unwrap())
+                        .unwrap()
+                        .set_retries(max_io_attempts),
                 )
             })
             .for_each(drop);
--- a/mayastor/src/bdev/nexus/nexus_channel.rs
+++ b/mayastor/src/bdev/nexus/nexus_channel.rs
@@ -130,8 +130,6 @@ impl NexusChannel {
         let nexus = unsafe { Nexus::from_raw(device) };
         debug!("{}: Creating IO channels at {:p}", nexus.bdev.name(), ctx);
 
-        let max_io_attempts = nexus.max_io_attempts;
-
         let ch = NexusChannel::from_raw(ctx);
         let mut channels = Box::new(NexusChannelInner {
             ch: Vec::new(),
@@ -146,9 +144,7 @@ impl NexusChannel {
             .filter(|c| c.status() == ChildStatus::Online)
             .map(|c| {
                 channels.ch.push(
-                    BdevHandle::try_from(c.get_descriptor().unwrap())
-                        .unwrap()
-                        .set_retries(max_io_attempts),
+                    BdevHandle::try_from(c.get_descriptor().unwrap()).unwrap(),
                 )
             })
             .for_each(drop);
--- a/mayastor/src/bdev/nexus/nexus_io.rs
+++ b/mayastor/src/bdev/nexus/nexus_io.rs
@@ -154,9 +154,7 @@ impl Bio {
     ) {
         self.ctx_as_mut_ref().in_flight -= 1;
 
-        if cfg!(debug_assertions) {
-            assert!(self.ctx_as_mut_ref().in_flight >= 0);
-        }
+        debug_assert!(self.ctx_as_mut_ref().in_flight >= 0);
 
         if !success && !child_io.is_null() {
             let io_type = Bio::io_type(self.0).unwrap();
--- a/mayastor/src/bdev/nexus/nexus_io.rs
+++ b/mayastor/src/bdev/nexus/nexus_io.rs
@@ -13,15 +13,10 @@ use spdk_sys::{
 use crate::{
     bdev::nexus::{
         nexus_bdev::{Nexus, NEXUS_PRODUCT_ID},
-        nexus_channel::NexusChannel,
+        nexus_fn_table::NexusFnTable,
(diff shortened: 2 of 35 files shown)


gila (commented):
I wonder if we can not simply the bookkeeping, of IO's that require a retry. I see that you extended the zero-sized array of the IO field, but I don't understand why we want to put a pointer there that then holds (u32) pointers. I'd rather stick something static in there because, among others, I wonder about cache flushes when we need to get this from the heap. (for which I have zero benchmarks but its just a gut feeling I have).

What if we make/keep things simple first: if *one* IO fails we consider them all failed, and we retry the IO to each, even if one of them already succeeded (we can optimize that later). 


So as far as I can tell, having as simple retry counter as part of the NioCtx would be sufficient? For reads we already RR, so the "handing there" would be somewhat different?

mayastor/src/bdev/nexus/nexus_channel.rs
```diff
@@ -144,7 +146,9 @@ impl NexusChannel {
             .filter(|c| c.status() == ChildStatus::Online)
             .map(|c| {
                 channels.ch.push(
-                    BdevHandle::try_from(c.get_descriptor().unwrap()).unwrap(),
+                    BdevHandle::try_from(c.get_descriptor().unwrap())
+                        .unwrap()
+                        .set_retries(max_io_attempts),
```
gila, line 151: Im not quite sure why this is added here....

mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -88,10 +105,16 @@ impl Bio {
     }
 
     /// create and return a new, initialized Bio object
-    pub fn new(pio: *mut spdk_bdev_io, in_flight: i8) -> Bio {
+    pub fn new(
```
gila, line 108: I think a name like `bio_prep()` would sounce better, would you agree? new to me sounds a lot like an actual new IO, its not really...

mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -111,13 +134,58 @@ impl Bio {
...
-        unsafe { spdk_bdev_io_complete(self.0, io_status::FAILED) };
+        self.drop_retries();
+        unsafe {
+            spdk_bdev_io_complete(self.0, io_status::FAILED);
+        }
+    }
+
+    /// ensure the retry_counts box is freed
+    #[inline]
+    pub(crate) fn drop_retries(&mut self) {
```
gila, line 153: IO's are always put back on the queue they were borrowed from, and they are fixed in size, and don't require any (de)allocation in the "hot path". I think we should try *hard* to keep it that way. Moreover, this can lead to leaks and we are throwing away, the benefits of rust in this case.

chriswldenyer: If it's OK to retry all child IOs when one fails that would simplify things a lot and the allocation could be removed. I was trying to limit it to just the failed child. I'll make the change.

mayastor/src/bdev/nexus/nexus_channel.rs
chriswldenyer: Jonathan's change was to have the max number of retries stored in the BdevHandle. I wanted to be consistent with that, rather than having one place for nexus IO and another for internal IO.

mayastor/src/bdev/nexus/nexus_io.rs
chriswldenyer: Ok, I can change that.

chriswldenyer:
> I wonder if we can not simply the bookkeeping, of IO's that require a retry. I see that you extended the zero-sized array of the IO field, but I don't understand why we want to put a pointer there that then holds (u32) pointers. I'd rather stick something static in there because, among others, I wonder about cache flushes when we need to get this from the heap. (for which I have zero benchmarks but its just a gut feeling I have).
> 
> What if we make/keep things simple first: if _one_ IO fails we consider them all failed, and we retry the IO to each, even if one of them already succeeded (we can optimize that later).
> 
> So as far as I can tell, having as simple retry counter as part of the NioCtx would be sufficient? For reads we already RR, so the "handing there" would be somewhat different?

I felt it was necessary to retry just the failed IO, which means keeping track of attempts per-child, which means some kind of vector to keep the number of attempts in. If we retry all IOs for each failure, then that simplifies things a lot. If you are happy with that approach then I'll change it.

mayastor/src/bdev/nexus/nexus_channel.rs
chriswldenyer: Changed to use just the value stored in the nexus.

paulyoong (approved):

mayastor/tests/error_count_retry.rs
```diff
@@ -0,0 +1,187 @@
...
+        Ok(_) => {
+            assert_eq!(succeed, true);
+        }
+        Err(_) => {
+            assert_eq!(succeed, false);
+        }
+    };
+}
+
+fn reactor_run_millis(milliseconds: u64) {
```
tiagolobocastro, line 180: IIRC you've used this kind of function before, maybe worth moving it into common?

tiagolobocastro (approved):

mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -130,7 +155,7 @@ impl Bio {
         self.ctx_as_mut_ref().in_flight -= 1;
 
         if cfg!(debug_assertions) {
-            assert_ne!(self.ctx_as_mut_ref().in_flight, -1);
+            assert!(self.ctx_as_mut_ref().in_flight >= 0);
```
jonathan-teh, line 158: This block can just be a debug_assert!().

jonathan-teh (approved):

mayastor/src/subsys/config/opts.rs
```diff
@@ -497,6 +497,9 @@ pub struct ErrStoreOpts {
 
     /// errors older than this are ignored
     pub retention_ns: u64,
+
+    /// the maximum number IO attempts per IO
```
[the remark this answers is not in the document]
chriswldenyer: Yes, thanks.

mayastor/tests/error_count_retry.rs
chriswldenyer: True, will fix.

mayastor/src/bdev/nexus/nexus_io.rs
chriswldenyer: Ok, didn't know you could do that.

jonathan-teh (approved):

tiagolobocastro (approved):

PR #405, opened 2020-09-07
CAS-392: Add create snapshot support to RPC and CLI

jonathan-teh:
Implement create_snapshot against a nexus over gRPC and the
mayastor-client CLI. This basically calls BdevHandle::create_snapshot
which sends our vendor-specific NVMe Admin command to all children.

Returns name of the snapshot over gRPC, derived from the nexus name
but is not used anywhere else at the moment.

Add a mocha test for create_snapshot against a nexus with 1 child
replica over gRPC. This requires changes to functions in test_common
to allow more than 1 instance of Mayastor to be started and to ensure
that those multiple instances are stopped.

### DIFF
--- a/mayastor-test/test_common.js
+++ b/mayastor-test/test_common.js
@@ -137,8 +137,12 @@ function getMyIp () {
 }
 
 // Common code for starting mayastor, mayastor-csi and spdk processes.
-function startProcess (command, args, env, closeCb, psName) {
-  assert(!procs[command]);
+function startProcess (command, args, env, closeCb, psName, suffix) {
+  let procsIndex = command;
+  if (suffix) {
+    procsIndex += suffix;
+  }
+  assert(!procs[procsIndex]);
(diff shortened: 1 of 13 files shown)


jkryl (approved):

jkryl:
There is a test_cli.js that tests CLI commands. You may want to contribute a new test for create snapshot command there. But it can be done later.

tiagolobocastro (approved):

jonathan-teh:
I'm going to hold off merging because the first run hung and I think there's a bug in killSudoedProcess if the first process whose name matches doesn't have a matching PID it will simply hang the test.

jonathan-teh:
Turns out it's nothing to do with killSudoedProcess, the Mayastor process hosting the nexus was ignoring the SIGTERM (probably as it's somehow trying to maintain the connection to its children over nvmf even though the other Mayastor instance has terminated) so it now destroys the nexus as the last step of the test.

PR #406, opened 2020-09-08
Detach on failed fs mount. CAS-429

blaisedias:
Make staging volumes more idempotent.
Staging  a file system volume consists 3 phases
1) validation of arguments (return on fail)
2) attaching to the nexus,
3) file system mount.

This change
 i) moves the accesstype check to phase 1.
ii) detaches the nexus if the file system mount fails.

### DIFF
--- a/csi/src/node.rs
+++ b/csi/src/node.rs
@@ -96,6 +96,32 @@ fn get_access_type(
     }
 }
 
+/// Detach the nexus device from the system, either at volume unstage,
+/// or after failed filesystem mount at volume stage.
+async fn detach(uuid: &Uuid, errheader: String) -> Result<(), Status> {
+    if let Some(device) = Device::lookup(uuid).await.map_err(|error| {
+        failure!(
+            Code::Internal,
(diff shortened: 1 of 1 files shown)


csi/src/node.rs
```diff
@@ -96,6 +96,32 @@ fn get_access_type(
     }
 }
 
+/// Detach the nexus device from the system, either at volume unstage,
+/// or after failed filesystem mount at volume stage.
+async fn detach(uuid: &Uuid, errheader: String) -> Result<(), Status> {
+    if let Some(device) = Device::lookup(uuid).await.map_err(|error| {
+        failure!(
```
tiagolobocastro, line 103: if we're idempotent shouldn't we return OK here?

tiagolobocastro: The code is a bit confusing but seems like in this case it would be an actual error and thus why we have Result<Option,> I guess...

tiagolobocastro (approved):

cjones1024 (approved):

PR #409, opened 2020-09-09
CAS-125: Restart rebuilds when Mayastor restarts

paulyoong:
Child status information is now persisted to a configuration file (a
yaml file currently stored in the /tmp directory). Whenever the child
status changes, the configuration file is updated to reflect this.

When Mayastor starts, it reads in the configuration file and sets the
child status accordingly. Any children in the "degraded" state will be
automatically rebuilt (regardless of the state of the rebuild when
Mayastor was stopped).

Support has been added to the mayastor-client to fault a child. The
command looks like:

	mayastor-client nexus child fault <nexus_uuid> <child_uri>

**Note: For now the configuration file is saved in the /tmp directory. Where this should actually be saved will be subsequently addressed.**

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_bdev_children.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev_children.rs
@@ -234,10 +234,10 @@ impl Nexus {
 
         let mut child = self.children.remove(idx);
         self.child_count -= 1;
-        self.reconfigure(DREvent::ChildRemove).await;
 
         // Update child status to remove this child
         NexusChild::save_state_change();
+        self.reconfigure(DREvent::ChildRemove).await;
 
         let result = child.destroy().await.context(DestroyChild {
             name: self.name.clone(),
@@ -321,7 +321,6 @@ impl Nexus {
                 child: name.to_owned(),
                 name: self.name.clone(),
             })?;
-            child.out_of_sync(true);
             self.start_rebuild(name).await.map(|_| {})?;
             Ok(self.status())
         } else {
--- a/mayastor/src/bdev/nexus/nexus_bdev_children.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev_children.rs
@@ -40,6 +40,7 @@ use crate::{
             },
             nexus_channel::DREvent,
             nexus_child::{ChildState, ChildStatus, NexusChild},
+            nexus_child_status_config::ChildStatusConfig,
             nexus_label::{
                 LabelError,
                 NexusChildLabel,
@@ -181,6 +182,9 @@ impl Nexus {
                 // it can never take part in the IO path
                 // of the nexus until it's rebuilt from a healthy child.
                 child.out_of_sync(true);
+                if ChildStatusConfig::add(&child).is_err() {
+                    error!("Failed to add child status information");
+                }
 
                 self.children.push(child);
                 self.child_count += 1;
@@ -231,6 +235,9 @@ impl Nexus {
         let mut child = self.children.remove(idx);
         self.child_count -= 1;
         self.reconfigure(DREvent::ChildRemove).await;
+
+        // Update child status to remove this child
+        NexusChild::save_state_change();
 
         let result = child.destroy().await.context(DestroyChild {
             name: self.name.clone(),
--- a/mayastor/src/bdev/nexus/nexus_channel.rs
+++ b/mayastor/src/bdev/nexus/nexus_channel.rs
@@ -45,6 +45,8 @@ pub enum DREvent {
     ChildRemove,
     /// Child rebuild event
     ChildRebuild,
+    /// Child status information is being applied
+    ChildApplyStatus,
 }
 
 impl NexusChannelInner {
@@ -167,7 +169,8 @@ impl NexusChannel {
             | DREvent::ChildOnline
             | DREvent::ChildRemove
             | DREvent::ChildFault
-            | DREvent::ChildRebuild => unsafe {
+            | DREvent::ChildRebuild
+            | DREvent::ChildApplyStatus => unsafe {
                 spdk_for_each_channel(
                     device,
                     Some(NexusChannel::refresh_io_channels),
--- a/mayastor/src/bdev/nexus/nexus_channel.rs
+++ b/mayastor/src/bdev/nexus/nexus_channel.rs
@@ -46,7 +46,7 @@ pub enum DREvent {
     /// Child rebuild event
     ChildRebuild,
     /// Child status information is being applied
-    ChildApplyStatus,
+    ChildStatusSync,
 }
 
 impl NexusChannelInner {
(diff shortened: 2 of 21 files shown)


mayastor/src/bdev/nexus/nexus_channel.rs
```diff
@@ -45,6 +45,8 @@ pub enum DREvent {
     ChildRemove,
     /// Child rebuild event
     ChildRebuild,
+    /// Child status information is being applied
+    ChildApplyStatus,
```
tiagolobocastro, line 49: Not sure about the name but also don't have good suggestions... maybe something like NexusStartup?

mayastor/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -232,6 +236,9 @@ impl Nexus {
         self.child_count -= 1;
         self.reconfigure(DREvent::ChildRemove).await;
 
+        // Update child status to remove this child
+        NexusChild::save_state_change();
```
tiagolobocastro, line 240: Should this be done before the child is removed or at least before we reconfigure the IO path?

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -266,10 +272,18 @@ impl NexusChild {
         }
         self.open(parent_size).map(|s| {
             self.status_reasons.offline(false);
+            NexusChild::save_state_change();
```
tiagolobocastro, line 275: I wonder if this is "too early", because right after we online it we need to mark it as degraded so it can be rebuilt? Eg, if we crashed in between we'd think that this child was healthy?

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -266,10 +272,18 @@ impl NexusChild {
...
             s
         })
     }
 
+    /// Save the state of the children to the config file
+    pub(crate) fn save_state_change() {
+        if ChildStatusConfig::save().is_err() {
+            error!("Failed to save child status information");
```
tiagolobocastro, line 283: should this be a fatal error?

mayastor/src/bdev/nexus/nexus_child_status_config.rs
```diff
@@ -0,0 +1,139 @@
...
+            Err(_) => Err(std::io::Error::new(
+                std::io::ErrorKind::Other,
+                "failed to serialize status config",
+            )),
+        }
+    }
+
+    /// Add the child to the configuration and then save it.
+    /// The configuration is updated on a status change and expects the child to
+    /// already be listed as a nexus child. However, When a child is added,
```
tiagolobocastro, line 128: s/**W**hen/**w**hen ?

mayastor/src/bin/cli/nexus_child_cli.rs
```diff
@@ -0,0 +1,64 @@
...
+) -> Result<(), Status> {
+    match matches.subcommand() {
+        ("fault", Some(args)) => fault(ctx, &args).await,
+        (cmd, _) => {
+            Err(Status::not_found(format!("command {} does not exist", cmd)))
+        }
+    }
+}
+
+// TODO: Make subcommands use the name of the child rather than the uri
```
tiagolobocastro, line 21: I don't think we can do this because all replicas in k8s have the same name?

mayastor/src/core/env.rs
```diff
@@ -630,6 +631,18 @@ impl MayastorEnvironment {
...
     }
 
+    // load the child status file
+    fn load_child_status(&self) {
+        ChildStatusConfig::get_or_init(|| {
+            if let Ok(cfg) = ChildStatusConfig::load() {
+                cfg
+            } else {
+                // if the configuration is invalid exit early
+                std::process::exit(-1);
```
tiagolobocastro, line 641: maybe replace by a panic with an error message?

mayastor/src/bdev/nexus/nexus_child_status_config.rs
```diff
@@ -0,0 +1,139 @@
...
+    instances,
+    nexus_channel::DREvent,
+    nexus_child::{NexusChild, StatusReasons},
+};
+use once_cell::sync::OnceCell;
+use serde::{Deserialize, Serialize};
+use std::{collections::HashMap, fs, fs::File, io::Write};
+
+type ChildName = String;
+pub const CONFIG_FILE: &str = "/tmp/child_status.yaml";
```
tiagolobocastro, line 25: might be worth considering getting this a CLI parameter? We do the same with the other config file.

mayastor/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -181,6 +182,9 @@ impl Nexus {
                 // it can never take part in the IO path
                 // of the nexus until it's rebuilt from a healthy child.
                 child.out_of_sync(true);
+                if ChildStatusConfig::add(&child).is_err() {
+                    error!("Failed to add child status information");
```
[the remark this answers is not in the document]
paulyoong: I was favouring keeping the system running given that a subsequent status update would update everything again anyway. Some thought needs to go into how we validate operations and unwind on errors anyway, so I'll leave things as they are for now and raise a work item to address this type of thing.

mayastor/src/bdev/nexus/nexus_child.rs
paulyoong: See response above about handling error conditions. (Defer the work to a future work item)

mayastor/src/bdev/nexus/nexus_channel.rs
paulyoong: I've left it as it is for now. I couldn't think of a better name.

mayastor/src/bdev/nexus/nexus_bdev_children.rs
paulyoong: Done.

mayastor/src/bdev/nexus/nexus_child.rs
paulyoong: Done. I've moved the setting of the out-of-sync flag to where we change the offline flag to false, so it is more of an atomic operation and prevents things like this.

mayastor/src/bdev/nexus/nexus_child_status_config.rs
paulyoong: :+1: Done

mayastor/src/bin/cli/nexus_child_cli.rs
paulyoong: This was a copy-paste error. I've removed this comment together with where I copied it from given that we can't do this.

mayastor/src/core/env.rs
paulyoong: Done

mayastor/src/bdev/nexus/nexus_child_status_config.rs
paulyoong: Done.

mayastor/src/bdev/nexus/nexus_channel.rs
cjones1024: ChildStatusSync maybe?

cjones1024 (approved):

mayastor/src/bdev/nexus/nexus_channel.rs
paulyoong: Good suggestion, done.

tiagolobocastro (approved):

PR #410, opened 2020-09-09
CAS-444: replace tokio-nats with nats.rs

tiagolobocastro:
Replace the tokio-nats library with nats.io/nats.rs official library.
They share similar API so there's not many code changes other than
not needing to lookup the name of the nats server as the library
handles that for us.


paulyoong (approved):

jkryl (approved):
Looks nice 👍

PR #411, opened 2020-09-09
mount rust unit tests hang if assertion is triggered (CAS-451)

jkryl:
The scope of the fix is limited to rust mount tests.


mayastor/tests/mount_fs.rs
```diff
@@ -169,9 +183,14 @@ fn mount_fs_2() {
...
 
         Mthread::spawn_unaffinitized(move || {
-            s.send(common::fio_run_verify(&device))
+            s.send(match common::fio_run_verify(&device) {
+                Ok(_) => "".to_owned(),
+                Err(err) => err,
+            })
         });
-        reactor_poll!(r);
+        let result;
```
tiagolobocastro, line 191: I see we use now this pattern a lot, maybe we can add it as a function in the test library?

tiagolobocastro (approved):

chriswldenyer (approved):
LGTM

blaisedias (approved):

mayastor/tests/mount_fs.rs
gila: Yeah, but this whole reactor_poll!() stuff isn't really needed anymore, to begin with. I think @jkryl is focussing on getting the tests to pass reliably on the new environment.

jkryl: I have contributed a new macro which is a copy of reactor_poll!() and does also the assert - so all combined together. So that will be a small convenience to eliminate repeating code. I think that more fundamental changes can be done later.

tiagolobocastro: :+1: Let's get the tests running reliably first then we'll be able to refactor with more confidence :)

PR #412, opened 2020-09-10
nexus publish hangs if kernel inits nbd dev for too long (CAS-453)

jkryl:
If the initialization takes time or never finishes (i.e. a bug in code
or misconfiguration) the atomic variable was never updated and the
receiver in the main thread waited forever. Operation was hung - unable
to complete.

Besides the fix of the hang, the fix improves the situation in a
couple of ways:
* the timeout has been increased from 100ms to ~1s
* exponential backoff prevents crazy many loops for slow systems
* log messages give insight without flooding the screen


jonathan-teh (approved):
Looks fine. The fix for the CLI help text should really be a separate commit but that's probably more hassle that it's worth.

blaisedias (approved):

PR #416, opened 2020-09-15
produce nice test reports from mocha in jenkins (CAS-448)

jkryl:
It makes use of Junit (standard xml test output).


jkryl:
This is how it looks like in Jenkins then:
image

blaisedias (approved):
LGTM

PR #420, opened 2020-09-16
nexus: flush should always succeed

gila:
During testing and performance benchmarking, it showed that when we
create an lvol, it does not support flush. The bdev then is exported as
a target -- over nvmf and consumed as replica.

By virtue of it being a nvme bdev, it implicitly supports flush, which
our lvol does not handle.

We must support flush and thus, simply complete the IO.

While here, i've removed some of the raw pointer deferences.

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_fn_table.rs
+++ b/mayastor/src/bdev/nexus/nexus_fn_table.rs
@@ -96,71 +96,67 @@ impl NexusFnTable {
         io: *mut spdk_bdev_io,
     ) {
         // only set the number of IO attempts before the first attempt
-        Bio::init(io, channel, Bio(io).nexus_as_ref().max_io_attempts);
-        Self::io_submit_or_resubmit(channel, io);
+        let mut bio = Bio::from(io);
+        bio.init();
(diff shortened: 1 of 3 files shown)


mayastor/src/bdev/nexus/nexus_fn_table.rs
```diff
@@ -96,71 +96,64 @@ impl NexusFnTable {
...
                 }
-                io_type::WRITE_ZEROES => {
-                    if nexus.io_is_supported(io_type) {
-                        nexus.write_zeroes(&nio, &ch)
-                    } else {
-                        nio.fail()
-                    }
+            }
+            io_type::FLUSH => {
+                nio.reset(0);
```
jonathan-teh, line 135: Might be worth a code comment as to why it appears that the flush is essentially not doing anything.

jonathan-teh (approved):

mayastor/src/bdev/nexus/nexus_fn_table.rs
gila: Good point

chriswldenyer (approved):
Looks good, and a lot neater.

PR #422, opened 2020-09-17
CAS-410: Allow snapshots to be shared

jonathan-teh:
The pool layer rework in e6f7517 introduced lvs and is the preferred
home for new functionality on pools/replicas so move snapshot creation
there.

Sharing a snapshot, which is essentially a replica or lvol, causes
the shared property to be written to the blob's extended attributes,
which fails because a snapshot is read-only. Simply return early and
log a warning instead of an error and failing the RPC call.

Update the Rust integration test to share the snapshot, create a nexus
with the snapshot as its child, and extend the IO tests to verify that
the snapshot behaves as one.

### DIFF
--- a/mayastor/src/lvs/lvol.rs
+++ b/mayastor/src/lvs/lvol.rs
@@ -19,9 +19,6 @@ use spdk_sys::{
     spdk_blob_set_xattr,
     spdk_blob_sync_md,
     spdk_lvol,
-    spdk_nvme_cpl,
-    spdk_nvme_status,
-    spdk_nvmf_request,
     vbdev_lvol_create_snapshot,
     vbdev_lvol_destroy,
     vbdev_lvol_get_from_bdev,
@@ -38,6 +35,7 @@ use crate::{
         IntoCString,
     },
     lvs::{error::Error, lvs_pool::Lvs},
+    subsys::NvmfReq,
 };
 
 /// properties we allow for being set on the lvol, this information is stored on
@@ -362,8 +360,8 @@ impl Lvol {
 
     /// Create a snapshot
     pub async fn create_snapshot(
-        self,
-        nvmf_req: *mut spdk_nvmf_request,
+        &self,
+        nvmf_req: &NvmfReq,
         snapshot_name: &str,
     ) {
         extern "C" fn snapshot_done_cb(
@@ -371,13 +369,9 @@ impl Lvol {
             _lvol_ptr: *mut spdk_lvol,
             errno: i32,
         ) {
-            let rsp: &mut spdk_nvme_cpl = unsafe {
-                &mut *spdk_sys::spdk_nvmf_request_get_response(
-                    nvmf_req_ptr as *mut spdk_nvmf_request,
-                )
-            };
-            let nvme_status: &mut spdk_nvme_status =
-                unsafe { &mut rsp.__bindgen_anon_1.status };
+            let nvmf_req = NvmfReq::from(nvmf_req_ptr);
+            let mut rsp = nvmf_req.response();
+            let nvme_status = rsp.status();
 
             nvme_status.set_sct(0); // SPDK_NVME_SCT_GENERIC
             nvme_status.set_sc(match errno {
@@ -390,9 +384,7 @@ impl Lvol {
 
             // From nvmf_bdev_ctrlr_complete_cmd
             unsafe {
-                spdk_sys::spdk_nvmf_request_complete(
-                    nvmf_req_ptr as *mut spdk_nvmf_request,
-                );
+                spdk_sys::spdk_nvmf_request_complete(nvmf_req.0.as_ptr());
             }
         }
 
@@ -402,7 +394,7 @@ impl Lvol {
                 self.0.as_ptr(),
                 c_snapshot_name.as_ptr(),
                 Some(snapshot_done_cb),
-                nvmf_req as *mut c_void,
+                nvmf_req.0.as_ptr().cast(),
             )
         };
 
--- a/mayastor/src/lvs/lvol.rs
+++ b/mayastor/src/lvs/lvol.rs
@@ -17,6 +17,10 @@ use spdk_sys::{
     spdk_blob_set_xattr,
     spdk_blob_sync_md,
     spdk_lvol,
+    spdk_nvme_cpl,
+    spdk_nvme_status,
+    spdk_nvmf_request,
+    vbdev_lvol_create_snapshot,
     vbdev_lvol_destroy,
     vbdev_lvol_get_from_bdev,
 };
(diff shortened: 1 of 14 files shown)


mayastor/src/lvs/lvol.rs
```diff
@@ -330,4 +353,59 @@ impl Lvol {
...
+    /// Format snapshot name
+    /// base_name is the nexus or replica UUID
+    pub fn format_snapshot_name(base_name: &str, snapshot_time: u64) -> String {
+        format!("{}-snap-{}", base_name, snapshot_time)
+    }
+
+    /// Create a snapshot
+    pub async fn create_snapshot(
+        self,
+        nvmf_req: *mut spdk_nvmf_request,
```
gila, line 366: I'd like to avoid the usage of raw pointers as much as possible. Perhaps we can do something like 

```rust
#[derive(Clone)]
struct NvmfReq(NonNull<*mut spdk_nvmf_request>)
```

Then we can also do nicer things in the future like Impl Debug?

mayastor/src/lvs/lvol.rs
```diff
@@ -330,4 +353,59 @@ impl Lvol {
...
+
+    /// Format snapshot name
+    /// base_name is the nexus or replica UUID
+    pub fn format_snapshot_name(base_name: &str, snapshot_time: u64) -> String {
+        format!("{}-snap-{}", base_name, snapshot_time)
+    }
+
+    /// Create a snapshot
+    pub async fn create_snapshot(
+        self,
```
gila, line 365: create snapshot consumes self, is that intentional?

mayastor/src/lvs/lvol.rs
```diff
@@ -330,4 +353,59 @@ impl Lvol {
...
+        nvmf_req: *mut spdk_nvmf_request,
+        snapshot_name: &str,
+    ) {
+        extern "C" fn snapshot_done_cb(
+            nvmf_req_ptr: *mut c_void,
+            _lvol_ptr: *mut spdk_lvol,
+            errno: i32,
+        ) {
+            let rsp: &mut spdk_nvme_cpl = unsafe {
+                &mut *spdk_sys::spdk_nvmf_request_get_response(
```
gila, line 375: If we have a NvmfReq here, we can (as an example) replace this rather hard to read stuff into something like

```rust
let &mut respose = NvmfReq::from(nvmf_req_ptr).get_response()
```

mayastor/src/lvs/lvol.rs
```diff
@@ -330,4 +353,59 @@ impl Lvol {
...
+                    nvmf_req_ptr as *mut spdk_nvmf_request,
+                )
+            };
+            let nvme_status: &mut spdk_nvme_status =
+                unsafe { &mut rsp.__bindgen_anon_1.status };
```
gila, line 380: I really *really* do not like this ` __bindgen_anon_1` these are subject to change if we bump things in Cargo.toml. Have you looked into the actual structs (i have not) and see if they are either (a) not whitelisted properly or (b) if they if we can create them "by hand"? 

If we can create them by hand if rather have some small C wrappers (like in logwrapper.c for example) that would return the values we need? 

What do you think? We don't have to do this all at once of course.

mayastor/src/lvs/lvol.rs
```diff
@@ -330,4 +353,59 @@ impl Lvol {
...
+                _ => {
+                    debug!("vbdev_lvol_create_snapshot errno {}", errno);
+                    0x06 // SPDK_NVME_SC_INTERNAL_DEVICE_ERROR
+                }
+            });
+
+            // From nvmf_bdev_ctrlr_complete_cmd
+            unsafe {
+                spdk_sys::spdk_nvmf_request_complete(
+                    nvmf_req_ptr as *mut spdk_nvmf_request,
```
gila, line 394: :thinking: we complete the request here... but still create the snapshot?

mayastor/src/lvs/lvol.rs
```diff
@@ -259,6 +275,13 @@ impl Lvol {
         let blob = unsafe { self.0.as_ref().blob };
         assert_ne!(blob.is_null(), true);
 
+        if self.is_snapshot() {
```
gila, line 278: It's good we check errors here but perhaps we should be more strict and assert? in lvs_pool.rs we have code something like:

```rust
pub fn lvols(&self) -> Option<impl Iterator<Item = Lvol>> 
``` 

there we filter for out the lvols we belonging to a pool, what if we add a new add a filter there? i.e where we check if its a snapshot or not? then we can assert here and leave out the warning as it seems to me that never want to call `set()`  on a snapshot?

mayastor/src/lvs/lvol.rs
```diff
@@ -259,6 +275,13 @@ impl Lvol {
...
+            warn!("ignoring set property on snapshot {}", self.name());
+            return Ok(());
+        }
+        if self.is_read_only() {
+            warn!("{} is read-only", self.name());
```
gila, line 283: we fall through here, is that intentional?

mayastor/src/lvs/lvol.rs
```diff
@@ -330,4 +353,59 @@ impl Lvol {
...
+            }
+        }
+
+        let c_snapshot_name = snapshot_name.into_cstring();
+        unsafe {
+            vbdev_lvol_create_snapshot(
+                self.0.as_ptr(),
+                c_snapshot_name.as_ptr(),
+                Some(snapshot_done_cb),
+                nvmf_req as *mut c_void,
```
gila, line 405: I've learned recently that we rust now also supports `.cast()` makes it look nicer IMHO :)

mayastor/tests/replica_snapshot.rs
```diff
@@ -99,41 +121,59 @@ fn replica_snapshot() {
     test_init!();
 
     Reactor::block_on(async {
-        create_nexus().await;
-        bdev_io::write_some(NXNAME).await.unwrap();
+        create_nexus(0).await;
+        bdev_io::write_some(NXNAME, 0, 0xff).await.unwrap();
         custom_nvme_admin(0xc1)
```
chriswldenyer, line 126: Not part of this change, but it would be nice to replace 0xc1 with a symbol

mayastor/src/lvs/lvol.rs
jonathan-teh: Sort of. The blob can become read-only for various reasons, such as if it becomes corrupt, somehow.

mayastor/src/lvs/lvol.rs
jonathan-teh: The problem is the status field is part of an anonymous union with the spdk_nvme_cpl struct so there doesn't seem to be another way to reach it. I figured the struct was stable enough as it's part of the NVMe spec.

mayastor/src/lvs/lvol.rs
jonathan-teh: No, now fixed.

mayastor/src/lvs/lvol.rs
jonathan-teh: This is part of the callback so the request to create the snapshot will have already been sent.

mayastor/src/lvs/lvol.rs
jonathan-teh: Fixed, we agreed it should be `NonNull<spdk_nvmf_request>`.

mayastor/src/lvs/lvol.rs
jonathan-teh: Fixed.

mayastor/src/lvs/lvol.rs
jonathan-teh: Fixed

mayastor/tests/replica_snapshot.rs
jonathan-teh: I've opted for a code comment.

mayastor/src/lvs/lvol.rs
jonathan-teh: This filter is used in [un]share_all so it won't help when only a single replica is being shared via `pool_grpc::share_replica`, unless I'm missing something.

mayastor/src/lvs/lvol.rs
```diff
@@ -330,4 +352,55 @@ impl Lvol {
...
+        &self,
+        nvmf_req: &NvmfReq,
+        snapshot_name: &str,
+    ) {
+        extern "C" fn snapshot_done_cb(
+            nvmf_req_ptr: *mut c_void,
+            _lvol_ptr: *mut spdk_lvol,
+            errno: i32,
+        ) {
+            let nvmf_req = NvmfReq(
```
gila, line 373: If you do a `Impl From<T> for NvmfRequest {}` the NonNull::new().unwrap() part can be handled there.

mayastor/src/lvs/lvol.rs
```diff
@@ -330,4 +352,55 @@ impl Lvol {
...
+            }
+        }
+
+        let c_snapshot_name = snapshot_name.into_cstring();
+        unsafe {
+            vbdev_lvol_create_snapshot(
+                self.0.as_ptr(),
+                c_snapshot_name.as_ptr(),
+                Some(snapshot_done_cb),
+                nvmf_req.0.as_ptr().cast::<c_void>(),
```
gila, line 400: Is the turbofish needed? would .cast() not infer void from the function itself?

gila (commented):
Just a few more nitpickings...

mayastor/src/lvs/lvol.rs
jonathan-teh: Thanks, fixed.

mayastor/src/lvs/lvol.rs
jonathan-teh: Indeed it isn't, spot the C++ coder.

mayastor/src/lvs/lvol.rs
jonathan-teh: I've introduced a C helper function to address this.

gila (approved):
Thanks @jonathan-teh and also for putting up with me wrt the __anon stuff.

chriswldenyer (approved):

PR #423, opened 2020-09-17
Update deployment YAML files to use 'latest' tag

GlennBullingham:
From v0.4.0, deployment files should use the 'latest' tag on Mayastor container image specifications, to avoid the need to update these with each release.  A note to users to that effect, and the need to change the tag as maybe required (e.g. if seeking to run an earlier version) has been added to the README.md file.

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -1,7 +1,13 @@
-# Quickstart
+# Quickstart Guide
 
-### Relocated
+### Has been relocated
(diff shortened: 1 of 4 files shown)


jkryl (approved):
👍

GlennBullingham:
Merging on one approval, since this action was already discussed with Jeffry.

PR #434, opened 2020-09-23
Jenkins should not update status of commit if all stages were skipped (CAS-459)

jkryl:
Bonus feature are slack messages sent to mayastor-backend channel
when nightly tests break/recover. Otherwise it would be very easy
to go unnoticed.


paulyoong (approved):

GlennBullingham (approved):
:beers: indeed!

jkryl:
bors merge

PR #435, opened 2020-09-24
CAS-430: Use mayastor instead of spdk binary in mocha tests

jonathan-teh:
Change test_nexus to start a 2nd instance of Mayastor to act as an
NVMf target instead of starting the SPDK binary. As this was the only
user, remove common.startSpdk.

This allows the default nvmf target in SPDK to be disabled.


jkryl (approved):

gila (approved):
Thanks @jonathan-teh , 

I cant see from the PR what instances are running what core, as this in debug mode it should not be a problem but its something to consider; if we assume at least two cores, it is probably better to run each instance on a different core.

PR #441, opened 2020-09-28
tests: Add a ginkgo for mayastor install

tjoshum:
Given a running 3 node cluster, install mayastor and verify that it
comes up.

Very basic first steps into testing against a kubernetes cluster.


gila:
@tjoshum can we move all this under `mayastor-test/e2e`?

paulyoong (approved):

e2e/install/install_test.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// Install mayastor on the cluster under test.
+// We deliberately call out to kubectl, rather than constructing the client-go
+// objects, so that we can verfiy the local deploy yamls are correct.
+func installMayastor() {
+	deployDir := getDeployYamlDir()
+	applyYaml(deployDir, "namespace.yaml")
+	applyYaml(deployDir, "moac-rbac.yaml")
+	applyYaml(getCsiYamlDir(), "mayastorpool.yaml")
```
blaisedias:
```
$ ls -l deploy/mayastorpoolcrd.yaml 
lrwxrwxrwx 1 xxxx xxxx 34 Sep  8 13:51 deploy/mayastorpoolcrd.yaml -> ../csi/moac/crds/mayastorpool.yaml
```
May be simpler to just use files only from the deploy directory.

tjoshum:
> @tjoshum can we move all this under `mayastor-test/e2e`?

Done

blaisedias (changes requested):

blaisedias:
Perhaps also install `deploy/storage-class.yaml` as prt of the test?

mayastor-test/e2e/install/install_test.go
```diff
@@ -0,0 +1,138 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+var cfg *rest.Config
+var k8sClient client.Client
+var k8sManager ctrl.Manager
+var testEnv *envtest.Environment
+
+// Helper for passing yaml to kubectl
+func applyYaml(rootDir string, filename string) {
```
gila, line 32: One of the reasons I suggested to make use of the k8s ecosystem is "k8s.io/kubernetes/test/e2e/framework" which instead of doing our own thing, allows replacing this with:

```go
framework.RunKubectl("-n", nameSpace, "delete", "-f", configMapPath)
```

https://pkg.go.dev/k8s.io/kubernetes/test/e2e/framework provides a better overview.

question: Is there a reason we are *not* using that right now?

tjoshum: e2e/framework is part of k8s.io/kubernetes and isn't intended to be used as a standalone library. Seemingly by design, it therefore can't be imported as a go module https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-505725449 (referenced closer to home https://github.com/openebs/zfs-localpv/pull/131).
I can hack it in, but I think that RunKubectl is the only function in that library that we'd really want at this point, and it's ~20 lines of mess in the mod files to save 1 line in the code.

tjoshum:
> Perhaps also install `deploy/storage-class.yaml` as part of the test?

I can bring it in easily enough if you'd like? Though it's not referenced in the quick start (which talks about using custom classes: https://mayastor.gitbook.io/introduction/quickstart/configure-mayastor#create-mayastor-storageclass-s)

e2e/install/install_test.go
tjoshum: Oh, didn't see that. Done.

blaisedias:
Depends on the objective I guess. My thinking was  mayastor is unusable without storage classes. Since we do provide 2 examples, it would be good to check that they work as expected. Though this test doesn't do that. So I'm fine if they are not loaded as part of the this change.

blaisedias (approved):

jkryl (commented):
nice! Just a nit but I think we should move the e2e directory to the top level not to interfere with npm package containing mocha tests that are testing mayastor grpc API. So that the distinction is clear.

tiagolobocastro:
> nice! Just a nit but I think we should move the e2e directory to the top level not to interfere with npm package containing mocha tests that are testing mayastor grpc API. So that the distinction is clear.

what about moving the mocha tests to mayastor-test/mocha or something like that and keep all the tests inside mayastor-test? And could we move the nix tests there as well?

tjoshum:
I originally had the tests in the top level directory, for the reason @jkryl says. I think organising mocha, ginkgo and nix tests under mayastor-test is a good idea, but might do that in a separate PR to keep it contained (especially since bors now handles the pain of merging for us).

tjoshum:
bors retry

tjoshum:
bors merge

PR #442, opened 2020-09-28
terraform: Require the libvirt provider

tjoshum:
We need this field so that we work with terraform v0.13

### DIFF
--- a/terraform/mod/libvirt/main.tf
+++ b/terraform/mod/libvirt/main.tf
@@ -147,5 +147,12 @@ output "node_list" {
 }
 
 terraform {
-  required_version = ">= 0.12"
+  required_version = ">= 0.13"
+
+  required_providers {
+    libvirt = {
+      source  = "dmacvicar/libvirt"
+      version = "0.6.2"
+    }
+  }
 }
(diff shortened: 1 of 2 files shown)


paulyoong (approved):

blaisedias:
Does this work with terraform versions < 0.13?

tjoshum:
I think you have v0.12? Could you run terraform validate for me, rather than me trying to downgrade?

blaisedias:
> I think you have v0.12? Could you run terraform validate for me, rather than me trying to downgrade?
Sorry, I moved to 0.13 ages ago, and have this change locally, but didn't check it in cause I was concerned that it would break earlier versions.

gila:
I would argue that most people can figure out how to install terraform properly if they consider to use it. The alternative is, that we provide a version in our shell.nix which I think would be overkill. IOW, im fine with this as it is.

blaisedias:
@tjoshum please can you update `terraform/README.adoc` to say that the files work with 0.13 or words to to that effect.

blaisedias (approved):

terraform/mod/libvirt/main.tf
```diff
@@ -148,4 +148,11 @@ output "node_list" {
 
 terraform {
   required_version = ">= 0.12"
```
jonathan-teh, line 150: Suggest bumping this up to 0.13.

blaisedias: Done.

blaisedias:
bors try

jonathan-teh (approved):

blaisedias:
bors merge

PR #444, opened 2020-09-28
CAS-466: Simplify the nexus child state management

paulyoong:
The nexus child state information is now managed simply by the
ChildState enum. The faulted state makes use of enum variants to
distinguish between the different types of fault conditions.

The use of ChildStatus has been completely removed. This was meant to
map internal Mayastor child state information to states required by the
control plane. This conversion is now performed when the child object is
mapped to its grpc representation.

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_bdev_rebuild.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev_rebuild.rs
@@ -16,7 +16,7 @@ use crate::{
                 RemoveRebuildJob,
             },
             nexus_channel::DREvent,
-            nexus_child::{ChildState, ChildStatus},
+            nexus_child::{ChildState, NexusChild, Reason},
         },
         VerboseError,
     },
@@ -36,7 +36,7 @@ impl Nexus {
         let src_child_name = match self
             .children
             .iter()
-            .find(|c| c.state == ChildState::Open && c.name != name)
+            .find(|c| c.state() == ChildState::Open && c.name != name)
         {
             Some(child) => Ok(child.name.clone()),
             None => Err(Error::NoRebuildSource {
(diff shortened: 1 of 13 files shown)


mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -168,7 +154,7 @@ impl Display for NexusChild {
                 "{}: {:?}/{:?}, blk_cnt: {}, blk_size: {}",
                 self.name,
                 self.state,
-                self.status(),
+                self.state(),
```
chriswldenyer, line 157: Looks like you can remove one of these, also on line 167 below.

mayastor/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -236,13 +238,14 @@ impl Nexus {
...
-                recovering_child.out_of_sync(false);
+                recovering_child.set_state(ChildState::Open);
+                NexusChild::save_state_change();
                 info!(
                     "Child {} has been rebuilt successfully",
                     recovering_child.name
                 );
 
-                assert_eq!(recovering_child.status(), ChildStatus::Online);
+                assert_eq!(recovering_child.state(), ChildState::Open);
```
chriswldenyer, line 248: Looks like this assert isn't needed any more as the state has just been set on line 241.

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -149,8 +135,8 @@ pub struct NexusChild {
     #[serde(skip_serializing)]
     pub(crate) desc: Option<Arc<Descriptor>>,
     /// current state of the child
-    pub(crate) state: ChildState,
-    pub(crate) status_reasons: StatusReasons,
+    #[serde(skip_serializing)]
+    state: RefCell<ChildState>,
```
tiagolobocastro, line 139: Why do we need RefCell here?

mayastor/src/bdev/nexus/nexus_child.rs
paulyoong: Thanks, made the changes.

mayastor/src/bdev/nexus/nexus_bdev_rebuild.rs
paulyoong: I've removed it

mayastor/src/bdev/nexus/nexus_child.rs
paulyoong: I've removed it - it was for a previous version of the changes but I don't think it is necessary now.

gila (approved):

mayastor/src/bdev/nexus/nexus_child.rs
tiagolobocastro: :+1:

tiagolobocastro (approved):

paulyoong (commented):
bors merge

paulyoong (commented):
bors retry

paulyoong (commented):
bors retry

PR #445, opened 2020-09-28
CAS-445: Move the mbus into a subsys

tiagolobocastro:
Move the message bus registration into a subsystem.

A connection to the server is ensured during the environment startup.
Further reconnects are then handled by the message bus.

Decouple the registration process from the message bus making the
registration process just another consumer of the message bus.

The message bus is currently just global static OnceCell which can be
used immutably to send/receive async messages.


mayastor/src/subsys/mbus/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+};
+use structopt::StructOpt;
+
+use crate::{
+    core::{MayastorCliArgs, MayastorEnvironment},
+    subsys::mbus::{mbus_nats::NATS_MSG_BUS, registration::Registration},
+};
+
+// wrapper around our MBUS subsystem used for registration
+pub struct MessageBusSubsystem(pub(crate) *mut spdk_subsystem);
```
gila, line 26: is the suffix "subsystem" needed? IOW MessageBus() would be sufficient?

mayastor/src/subsys/mbus/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+        Self::new()
+    }
+}
+
+impl MessageBusSubsystem {
+    /// initialize a new subsystem that handles the control plane
+    /// message bus registration process
+    extern "C" fn init() {
+        debug!("mayastor mbus subsystem init");
+        let args = MayastorEnvironment::new(MayastorCliArgs::from_args());
```
gila, line 39: Wow, I did not know this was even possible..

mayastor/src/subsys/mbus/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+        if let (Some(_), Some(grpc)) = (args.mbus_endpoint, args.grpc_endpoint)
+        {
```
gila, line 41: :thinking:

mayastor/src/subsys/mbus/registration.rs
```diff
@@ -0,0 +1,201 @@
...
+}
+
+static MESSAGE_BUS_REG: OnceCell<Registration> = OnceCell::new();
+impl Registration {
+    pub(super) fn init(node: &str, grpc_endpoint: &str) {
+        MESSAGE_BUS_REG.get_or_init(|| Registration::new(node, grpc_endpoint));
+
+        // spawn a runner thread responsible for registering and
+        // deregistering the mayastor instance on shutdown
+        std::thread::spawn(|| {
```
gila, line 89: hmpf. Why do we need a thread for this? and if we do, we should set its mask such that it does not run on the coremask.

mayastor/src/subsys/mbus/registration.rs
```diff
@@ -0,0 +1,201 @@
...
+        });
+    }
+
+    fn new(node: &str, grpc_endpoint: &str) -> Registration {
+        let (msg_sender, msg_receiver) = smol::channel::unbounded::<()>();
+        let config = Configuration {
+            node: node.to_owned(),
+            grpc_endpoint: grpc_endpoint.to_owned(),
+            hb_interval: match env::var("MAYASTOR_HB_INTERVAL")
+                .map(|v| v.parse::<u64>())
```
gila, line 102: map_or_else() or unwrap_or_default might be useful here?

mayastor/src/subsys/mbus/registration.rs
```diff
@@ -0,0 +1,201 @@
...
+        };
+        Self {
+            config,
+            rcv_chan: msg_receiver,
+            fini_chan: msg_sender,
+        }
+    }
+
+    pub(super) fn get() -> &'static Registration {
+        MESSAGE_BUS_REG.get().unwrap()
```
gila, line 116: lets replace this with .expect("msg") where msg explains why we crash here ( get called for init?)

mayastor/src/subsys/mbus/mod.rs
tiagolobocastro: then it conflicts with the actual MessageBus type?

mayastor/src/subsys/mbus/mod.rs
tiagolobocastro: The registration requires the grpc endpoint so we can give it to moac, so it can only run if both nats and grpc are enabled.

mayastor/src/subsys/mbus/registration.rs
tiagolobocastro: The registration process keeps running waking up on an interval to do work so we can't run on the "subsystem init context". 
Good catch, guess we could use `spawn_unaffinitized`.

mayastor/src/subsys/mbus/registration.rs
tiagolobocastro: Had trouble making it "neat" that way because env::var returns a result and parse also returns a result so ended up having to do two unwraps, eg:

mayastor/src/subsys/mbus/registration.rs
tiagolobocastro: :+1:

mayastor/src/subsys/mbus/mod.rs
gila: Okay, if that is taken, and renaming it does not work then Ok. BTW, the subsystem pointer should ideally not be pub though.

mayastor/src/subsys/mbus/registration.rs
gila: Yeah but as we use futures, can't we use a timer to ping pong here for us?

mayastor/src/subsys/mbus/mod.rs
tiagolobocastro: hmm yeah that would make sense, maybe the `MessageBusSubsystem` can expose a pub static register_subsystem or something like that.

mayastor/src/subsys/mbus/registration.rs
tiagolobocastro: What do you mean by ping pong?

mayastor/src/subsys/mbus/mod.rs
tiagolobocastro: I've exposed it via add which is kept under a generic `add_subsystem` where we could add more subsystems that start after main. Not the best naming I know...

mayastor/src/subsys/mbus/registration.rs
```diff
@@ -0,0 +1,203 @@
...
+            grpc_endpoint: self.config.grpc_endpoint.clone(),
+        };
+        super::message_bus()
+            .publish("register", serde_json::to_vec(&payload).unwrap())
+            .await
+            .map_err(|cause| Error::QueueRegister {
+                cause,
+            })?;
+
+        // Note that the message was only queued and we don't know if it was
```
gila, line 171: having a way to figure out that we are registered, seems to make a whole lot of sense to me. In particular, during production issues where there seems to be a disconnect somehow somewhere. 

If we set some variable somehow and make that readable/gettable trough RPC we can then ensure: "yes -- this instance is regisitered" -- having this in the logs then would also help.

mayastor/src/subsys/mbus/registration.rs
```diff
@@ -0,0 +1,203 @@
...
+        Ok(())
+    }
+
+    /// Send a deregister message to the MessageBus.
+    async fn deregister(&self) -> Result<(), Error> {
+        let payload = DeregisterArgs {
+            id: self.config.node.clone(),
+        };
+        super::message_bus()
+            .publish("deregister", serde_json::to_vec(&payload).unwrap())
```
gila, line 188: Same as above, we can make the argument have a trait guard so we can do the json calls inside the publish function

mayastor/src/subsys/mod.rs
```diff
@@ -38,3 +42,8 @@ pub(crate) fn register_subsystem() {
         spdk_add_subsystem_depend(Box::into_raw(depend));
     }
 }
+
+/// Add subsystems after the main entrypoint
+pub(crate) fn add_subsystems() {
+    MessageBusSubsystem::add();
```
gila, line 48: The whole idea of this is that the subsystems run *before* main right? This way, it does not do that.

mayastor/src/subsys/mbus/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+        }
+        unsafe { spdk_subsystem_fini_next() }
+    }
+
+    fn new() -> Self {
+        info!("creating Mayastor mbus subsystem...");
+        let args = MayastorEnvironment::new(MayastorCliArgs::from_args());
+        if let Some(nats) = args.mbus_endpoint {
+            mbus_nats::message_bus_init(nats);
+        }
```
gila, line 60: I don't think we have to call this here.  The init/fini is to ensure you get called before we start up fully, but it is not needed to start the bus here. The reason I say this that I'm a bit concerned about what creating the runtime in the init call does.

Having that said, this runtime mess is not something of our doing, but.. I do have an idea to make it less horrible but... i might get to that later...

mayastor/src/subsys/mbus/registration.rs
```diff
@@ -0,0 +1,203 @@
...
+    rcv_chan: smol::channel::Receiver<()>,
+    /// Termination channel
+    fini_chan: smol::channel::Sender<()>,
+}
+
+static MESSAGE_BUS_REG: OnceCell<Registration> = OnceCell::new();
+impl Registration {
+    /// initialise the global registration instance
+    pub(super) fn init(node: &str, grpc_endpoint: &str) {
+        MESSAGE_BUS_REG.get_or_init(|| Registration::new(node, grpc_endpoint));
```
gila, line 85: This is why we don't need to init it inside the subsystem init, new() should call message_bus() which calls the OnceCell closure and starts the bus. 

Also are you sure about the pub here? I dont think you want anyone else to call this?

mayastor/src/subsys/mbus/registration.rs
```diff
@@ -0,0 +1,203 @@
...
+        };
+        Self {
+            config,
+            rcv_chan: msg_receiver,
+            fini_chan: msg_sender,
+        }
+    }
+}
+
+impl Registration {
```
gila, line 126: why 2 impl blocks?

mayastor/src/subsys/mbus/mbus_nats.rs
```diff
@@ -0,0 +1,100 @@
...
+                        server
+                    );
+                    return connection;
+                }
+                Err(error) => {
+                    if log_error {
+                        warn!(
+                            "Error connection: {}. Quietly retrying...",
+                            error
+                        );
```
jkryl, line 45: If I understand correctly we will have two of these warning messages per second printed to log file if nats is not available. Perhaps it would have been better to print it only once (the first time).

mayastor/src/subsys/mbus/mbus_nats.rs
```diff
@@ -0,0 +1,100 @@
...
+use once_cell::sync::OnceCell;
+use smol::io;
+
+pub(super) static NATS_MSG_BUS: OnceCell<NatsMessageBus> = OnceCell::new();
+pub(super) fn message_bus_init(server: String) {
+    NATS_MSG_BUS.get_or_init(|| {
+        // Waits for the message bus to become ready
+        tokio::runtime::Runtime::new()
+            .unwrap()
+            .block_on(async { NatsMessageBus::new(&server).await })
```
jkryl, line 16: So here we block the reactor until mayastor connects to the message bus, right? That's probably ok, just asking.

mayastor/src/core/env.rs
```diff
@@ -110,8 +109,10 @@ pub struct MayastorCliArgs {
     /// Name of the node where mayastor is running (ID used by control plane)
     pub node_name: Option<String>,
     #[structopt(short = "n")]
-    /// IP address and port of the NATS server
-    pub nats_endpoint: Option<String>,
+    /// URL of the message bus, example formats:
+    /// nats://${name_or_ip}:{optional_port}
+    /// ${name_or_ip}:{optional_port}
+    pub mbus_endpoint: Option<String>,
```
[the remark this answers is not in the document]
tiagolobocastro: hmm, at the moment the nats client library handles the resolution for us. So do you mean pre-validating it before we "ask" nats?

mayastor/src/core/env.rs
```diff
@@ -209,8 +210,8 @@ type Result<T, E = EnvError> = std::result::Result<T, E>;
 pub struct MayastorEnvironment {
     pub config: Option<String>,
     node_name: String,
-    nats_endpoint: Option<String>,
-    grpc_endpoint: Option<String>,
+    pub mbus_endpoint: Option<String>,
+    pub grpc_endpoint: Option<String>,
```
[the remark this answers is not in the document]
tiagolobocastro: that may potentially be a little bit tricky if the nats server is specified as a name and not IP. For example, if mayastor manages to start before the nats server service is configured we may not be able to perform this conversion and would have to add another retry loop?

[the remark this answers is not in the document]
tiagolobocastro: :+1: I much prefer that tbh and will also simply the nats checking code which atm also checks if grpc is there...

mayastor/src/subsys/mbus/mbus_nats.rs
tiagolobocastro: Yes that's right, this was to avoid having the message bus deal with it not being connected the first time. After the first connect then the library handles that for us.

mayastor/src/subsys/mbus/mbus_nats.rs
tiagolobocastro: hmm, actually it only logs it once, see the log_error bool?

mayastor/src/subsys/mbus/registration.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    /// Send a register message to the MessageBus.
+    async fn register(&self) -> Result<(), Error> {
+        let payload = RegisterArgs {
+            id: self.config.node.clone(),
+            grpc_endpoint: self.config.grpc_endpoint.clone(),
+        };
+        super::message_bus()
+            .publish("register", serde_json::to_vec(&payload).unwrap())
```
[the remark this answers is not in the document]
tiagolobocastro: Yes I hit this yesterday writing code to get the config from a service. We can also have a trait to have that done automatically I guess, though I don't know if we're really going to keep this json stuff or not.

mayastor/src/subsys/mbus/mbus_nats.rs
jkryl: ah yes. thanks 👍

jkryl (approved):

mayastor/src/subsys/mbus/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+    }
+}
+
+#[async_trait]
+pub trait MessageBus {
+    /// publish a message - not guaranteed to be sent or received (fire and
+    /// forget)
+    async fn publish(
+        &self,
+        channel: &str,
```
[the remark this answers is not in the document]
tiagolobocastro: :+1:

mayastor/src/subsys/mbus/registration.rs
tiagolobocastro: We can do that via the request I think. We'll have to figure out how to implement that, probably using request/reply with a timeout retry loop.
The way it works is you send to channel X, message Y and wait on channel Z for the reply. The other side then receives Y on channel X and then replies to channel Z. 
We have to retry in case the other side was not connected when the message was sent of if it crashed after receiving it and before replying.

mayastor/src/subsys/mod.rs
tiagolobocastro: I tried to start it after main but before we start initializing things because there is some kind of issue when connecting the nats client before main:
thread '<unnamed>' panicked at 'assertion failed: c.borrow().is_none()', src/libstd/sys_common/thread_info.rs:40:26

mayastor/src/subsys/mbus/mod.rs
tiagolobocastro: It's was here for convenience only. If you remember the registration "run" was being spawned in the init call so I tried to make sure that nats was already ready then by having it connect in the "new" as it is now. But we now run the registration in the "env" tokio loop so we can indeed move the message bus init elsewhere...

mayastor/src/subsys/mbus/registration.rs
tiagolobocastro: It's a pub(super) so it's accessible from mbus.rs and that's the only place where it's called. The idea was that if we had other things other than registration that we need to setup, we can do them from the mbus and have the "things" in different files.

mayastor/src/subsys/mbus/registration.rs
tiagolobocastro: no particular reason really, I'll fix that.

mayastor/src/subsys/mbus/registration.rs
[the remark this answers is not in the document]
tiagolobocastro: I've added the Serialize trait to the message argument

mayastor/src/core/env.rs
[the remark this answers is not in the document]
tiagolobocastro: As discussed, the nats endpoint will remain a name for now (though I now validate the hostname on startup). We can later decide if we want to convert it to an actual IP address.

mayastor/src/subsys/mod.rs
tiagolobocastro: Moved the actual bus init into later so now the subsystem starts before main.

tiagolobocastro:
bors try

tiagolobocastro:
bors ping

tiagolobocastro:
bors try

gila (approved):

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

PR #446, opened 2020-09-29
CAS-474: Use child status config in Mayastor deployment

paulyoong:
Add support for using the child status config file when deploying Mayastor
on K8s. This allows child states to be correctly set when a Mayastor pod
is restarted on the same node.

Bug fix:
    - Update child states when a nexus is destroyed to ensure that their
      entries in the child status config file are removed.

### DIFF
--- a/deploy/mayastor-daemonset.yaml
+++ b/deploy/mayastor-daemonset.yaml
@@ -48,7 +48,6 @@ spec:
         - "-N$(MY_NODE_NAME)"
         - "-g$(MY_POD_IP)"
         - "-nnats"
-        - "-y/var/local/mayastor/config.yaml"
         securityContext:
           privileged: true
         volumeMounts:
@@ -56,10 +55,6 @@ spec:
           mountPath: /dev
         - name: dshm
           mountPath: /dev/shm
-        - name: configlocation
-          mountPath: /var/local/mayastor/
-        - name: config
-          mountPath: /var/local/mayastor/config.yaml
         resources:
           limits:
             cpu: "1"
@@ -85,11 +80,3 @@ spec:
       - name: hugepage
         emptyDir:
           medium: HugePages
-      - name: configlocation
-        hostPath:
-          path: /var/local/mayastor/
-          type: DirectoryOrCreate
-      - name: config
-        hostPath:
-          path: /var/local/mayastor/config.yaml
-          type: FileOrCreate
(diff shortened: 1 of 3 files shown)


tiagolobocastro (approved):

blaisedias (approved):

paulyoong (commented):
bors merge

paulyoong (commented):
bors retry

paulyoong (commented):
bors retry

paulyoong (commented):
bors retry

paulyoong (commented):
bors cancel

paulyoong (commented):
bors r-

paulyoong (commented):
bors try

deploy/mayastor-daemonset.yaml
```diff
@@ -60,6 +61,8 @@ spec:
           mountPath: /var/local/mayastor/
         - name: config
           mountPath: /var/local/mayastor/config.yaml
+        - name: child-status-config
```
gila, line 64: let us not put this stuff in the deployment files right now as it will end up in people configs when doing a release. It creates issues when troubleshooting among others. In fact, could you remove (for now) both config files? I'm fine if we create a separate file here, that does include them.

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -567,6 +567,8 @@ impl Nexus {
         }
 
         if r.await.unwrap() {
+            // Update the child states to remove them from the config file.
+            NexusChild::save_state_change();
```
gila, line 571: Forgot to ask last time around but why does this not take &self? i.e nexus.config_sync() would be more appropriate?

paulyoong: Each call to save_state_change gets the state of every child for every nexus in the running system and updates the config file accordingly. IOW it overwrites the config file in its entirety rather than updating it for a specific nexus. That was why I didn't use _nexus.config_sync()_ as it implies we are only updating the config for that nexus.

The reason I went for this approach was that it mirrors what we do for the mayastor config.

deploy/mayastor-daemonset.yaml
paulyoong: Done - I've removed the config from here and created a separate mayastor-daemonset-config.yaml file which does use them

paulyoong (commented):
bors try

paulyoong (commented):
bors r+

paulyoong (commented):
bors r-

paulyoong (commented):
bors try

paulyoong (commented):
bors merge

PR #448, opened 2020-09-30
447 Initialise tracing so that default subscriber filtering may be configured via environment variable.

cjones1024:
Create an EnvFilter for the default subscriber that allows the filter to be configured via the RUST_LOG environment variable.
This reproduces as closely as possible the old behaviour using the logging crate.
The default logging level is honoured *only* if RUST_LOG is not set (to a valid value).

### DIFF
--- a/mayastor/src/logger.rs
+++ b/mayastor/src/logger.rs
@@ -1,7 +1,10 @@
 use std::{ffi::CStr, os::raw::c_char, str::FromStr};
 
 use tracing_log::format_trace;
-use tracing_subscriber::fmt::{format::FmtSpan, time::FormatTime, Subscriber};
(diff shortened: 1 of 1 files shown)


jonathan-teh (approved):
Looks fine but you should remove the "CAS-" prefix from the commit message as that's a different numbering system.

tiagolobocastro (approved):

cjones1024:
> 
> 
> Looks fine but you should remove the "CAS-" prefix from the commit message as that's a different numbering system.

Yes, have realized that. will amend accordingly when merging into develop

jonathan-teh:
Commit messages need to be changed and pushed first though as bors does a merge commit.

PR #453, opened 2020-10-01
Prevent faulting the last healthy child

paulyoong:
If a nexus has a single healthy child, prevent that child from being
faulted (otherwise it would make the nexus unusable).


tiagolobocastro (approved):

paulyoong:
> Is there a regression test for this?

Good point - I have added a test

jonathan-teh (approved):
Looks good, thanks for adding the test.

paulyoong (commented):
bors merge

PR #455, opened 2020-10-05
Enhance check before writing labels to children

paulyoong:
Before writing a label to a child we must first ensure that it is
read-writable. The previous check did not take into account that a child
can service I/O whilst in the out-of-sync state. As a result, when
adding a child to a nexus a "Failed to sync labels" error message was
erroneously reported (even though the label was updated).


jonathan-teh (approved):

gila (approved):

gila:
borse r+

PR #457, opened 2020-10-07
Add common message bus code and kiiss example...

tiagolobocastro:
...service

Common message bus api to be shared between services and mayastor
Common service api for the services

Kiiss service as an example to show case the service api.

I've split this in 3 commits, which should make reviewing easier.
I also added a few cargo examples which might be useful.

### DIFF
--- a/mayastor/src/core/env.rs
+++ b/mayastor/src/core/env.rs
@@ -14,6 +14,7 @@ use std::{
 
 use byte_unit::{Byte, ByteUnit};
 use futures::{channel::oneshot, future};
+use mbus_api::{ConfigGetCurrent, Message, ReplyConfig};
 use once_cell::sync::{Lazy, OnceCell};
 use snafu::Snafu;
 use structopt::StructOpt;
@@ -601,7 +602,7 @@ impl MayastorEnvironment {
         }
     }
 
-    /// start the  JSON rpc server which listens only to a local path
+    /// start the JSON rpc server which listens only to a local path
     extern "C" fn start_rpc(rc: i32, arg: *mut c_void) {
         let ctx = unsafe { Box::from_raw(arg as *mut SubsystemCtx) };
 
@@ -637,6 +638,21 @@ impl MayastorEnvironment {
             Config::get_or_init(Config::default)
         };
         cfg.apply();
+    }
+
+    #[allow(dead_code)]
+    async fn get_service_config(&self) -> ReplyConfig {
+        if self.mbus_endpoint.is_some() {
+            ConfigGetCurrent {
+                kind: mbus_api::Config::MayastorConfig,
+            }
+            .request()
+            .await
+            // we need the library to be able to retry
+            .unwrap()
+        } else {
+            Default::default()
+        }
     }
 
     // load the child status file
--- a/mayastor/src/subsys/mbus/registration.rs
+++ b/mayastor/src/subsys/mbus/registration.rs
@@ -7,11 +7,9 @@
 //! but can be overridden by the `MAYASTOR_HB_INTERVAL` environment variable.
 //! containing the node name and the grpc endpoint.
 
-use super::MessageBus;
-use crate::subsys::mbus::Channel;
 use futures::{select, FutureExt, StreamExt};
+use mbus_api::*;
 use once_cell::sync::OnceCell;
-use serde::{Deserialize, Serialize};
 use snafu::Snafu;
 use std::{env, time::Duration};
 
@@ -43,20 +41,6 @@ pub enum Error {
     QueueRegister { cause: std::io::Error },
     #[snafu(display("Failed to queue deregister request: {:?}", cause))]
     QueueDeregister { cause: std::io::Error },
-}
-
-/// Register message payload
-#[derive(Serialize, Deserialize, Debug)]
-struct RegisterArgs {
-    id: String,
-    #[serde(rename = "grpcEndpoint")]
-    grpc_endpoint: String,
-}
-
-/// Deregister message payload
-#[derive(Serialize, Deserialize, Debug)]
-struct DeregisterArgs {
-    id: String,
 }
 
 #[derive(Clone)]
@@ -156,12 +140,13 @@ impl Registration {
 
     /// Send a register message to the MessageBus.
     async fn register(&self) -> Result<(), Error> {
-        let payload = RegisterArgs {
+        let payload = Register {
             id: self.config.node.clone(),
             grpc_endpoint: self.config.grpc_endpoint.clone(),
         };
-        super::message_bus()
-            .publish(Channel::Register, &payload)
+
+        payload
+            .publish()
             .await
             .map_err(|cause| Error::QueueRegister {
                 cause,
@@ -180,16 +165,18 @@ impl Registration {
 
     /// Send a deregister message to the MessageBus.
     async fn deregister(&self) -> Result<(), Error> {
-        let payload = DeregisterArgs {
+        let payload = Deregister {
             id: self.config.node.clone(),
         };
-        super::message_bus()
-            .publish(Channel::DeRegister, &payload)
+
+        payload
+            .publish()
             .await
-            .map_err(|cause| Error::QueueDeregister {
+            .map_err(|cause| Error::QueueRegister {
                 cause,
             })?;
-        if let Err(e) = super::message_bus().flush().await {
+
+        if let Err(e) = bus().flush().await {
             error!("Failed to explicitly flush: {}", e);
         }
 
--- a/mbus-api/src/lib.rs
+++ b/mbus-api/src/lib.rs
@@ -1,0 +1,242 @@
+#![warn(missing_docs)]
+//! All the different messages which can be sent/received to/from the control
+//! plane services and mayastor
+//! We could split these out further into categories when they start to grow
+
+mod mbus_nats;
+/// received message traits
+pub mod receive;
+/// send messages traits
+pub mod send;
+
+use async_trait::async_trait;
+use dyn_clonable::clonable;
+pub use mbus_nats::{bus, message_bus_init, message_bus_init_tokio};
+pub use receive::*;
+pub use send::*;
+use serde::{Deserialize, Serialize};
+use smol::io;
+use snafu::Snafu;
+use std::{fmt::Debug, marker::PhantomData, str::FromStr};
+
+/// Available Message Bus channels
+#[derive(Clone, Debug)]
+pub enum Channel {
+    /// Default
+    Default,
+    /// Registration of mayastor instances with the control plane
+    Registry,
+    /// Keep it In Sync Service
+    Kiiss,
+    /// Reply to requested Channel
+    Reply(String),
+}
+
+impl FromStr for Channel {
+    type Err = String;
+
+    fn from_str(source: &str) -> Result<Self, Self::Err> {
+        match source {
+            "default" => Ok(Self::Default),
+            "registry" => Ok(Self::Registry),
+            "kiiss" => Ok(Self::Kiiss),
+            _ => Err(format!("Could not parse the channel: {}", source)),
+        }
+    }
+}
+
+impl Default for Channel {
+    fn default() -> Self {
+        Channel::Default
+    }
+}
+
+impl std::fmt::Display for Channel {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match self {
+            Channel::Default => write!(f, "default"),
+            Channel::Registry => write!(f, "registry"),
+            Channel::Kiiss => write!(f, "kiiss"),
+            Channel::Reply(ch) => write!(f, "{}", ch),
+        }
+    }
+}
+
+/// Message id which uniquely identifies every type of unsolicited message
+/// The solicited (replies) message do not currently carry an id as they
+/// are sent to a specific requested channel
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+#[serde(rename_all = "camelCase")]
+pub enum MessageId {
+    /// Default
+    Default,
+    /// Update Config
+    ConfigUpdate,
+    /// Request current Config
+    ConfigGetCurrent,
+    /// Register mayastor
+    Register,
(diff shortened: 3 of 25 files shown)


mayastor/src/core/env.rs
```diff
@@ -639,6 +640,22 @@ impl MayastorEnvironment {
...
+            tokio::runtime::Runtime::new().unwrap().block_on(async {
+                let _reply = GetConfig::request(
+                    RequestConfig {
+                        kind: ConfigKind::MayastorConfig,
+                    },
+                    Channel::Kiiss,
+                    &subsys::message_bus(),
+                );
+                //.await;
+                //let _config = _reply.unwrap();
```
cjones1024, line 654: comments should be removed?

mbus-api/src/receive.rs
```diff
@@ -0,0 +1,166 @@
...
+    for<'de> S: Deserialize<'de> + 'a + Debug + Clone + MessageIdTrait,
+    R: Serialize,
+{
+    /// Get a copy of the actual payload data which was received.
+    pub fn payload(&self) -> S {
+        self.request.data.clone()
+    }
+    /// Get a reference of the actual payload data which was received.
+    pub fn payload_as_ref(&self) -> &S {
+        &self.request.data
```
cjones1024, line 33: Would it to better to use the AsRef trait here?

mbus-api/src/receive.rs
```diff
@@ -0,0 +1,166 @@
...
+    fn try_from(value: ReceivedRawMessage<'a>) -> Result<Self, Self::Error> {
+        ReceivedMessage::<S, R>::new(value.bus_msg)
+    }
+}
+
+impl<T> From<T> for ReplyPayload<T> {
+    fn from(val: T) -> Self {
+        ReplyPayload {
+            0: Ok(val),
+        }
```
cjones1024, line 156: I wasn't even aware this was legal - but isn't it better to just use ReplyPayload(Ok(val))

mbus-api/src/receive.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+}
+
+impl<T> From<Result<T, Error>> for ReplyPayload<T> {
+    fn from(val: Result<T, Error>) -> Self {
+        ReplyPayload {
+            0: val,
+        }
+    }
```
cjones1024, line 165: see above comment

mayastor/src/core/env.rs
tiagolobocastro: Yeah though I'm tempted to remove it altogether, that was just to leave as an example of how to call into the message bus

mbus-api/src/receive.rs
tiagolobocastro: Yeah, it was originally a "normal" struct, so when I changed the type, the job got half done only :)

mbus-api/src/receive.rs
tiagolobocastro: I guess S needs to implement the trait. I've removed it for now. If we use proc-macros we might be able to automatically add that trait and any others to the message bus types.

mayastor/src/core/env.rs
tiagolobocastro: I opted to leave the code in, but just not call the function so this way we don't need to comment out things that would fail without the service being deployed.

mbus-api/src/receive.rs
cjones1024: Yeah. this is still harder than it should be to get right.

cjones1024 (approved):

mbus-api/src/receive.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// a Result-like type.
+    /// May fail if serialization of the reply fails or if the
+    /// message bus fails to respond.
+    /// Can receive either `R`, `Err()` or `ReplyPayload`.
+    pub async fn respond<T: Into<ReplyPayload<R>>>(
+        &self,
+        reply: T,
+    ) -> io::Result<()> {
+        let reply: ReplyPayload<R> = reply.into();
+        let payload = serde_json::to_vec(&reply)?;
```
gila, line 46: If the ReplyPayload would have the guard Serialize -- the conversion can be done in one blow?

mbus-api/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+/// be uniquely identifiable
+pub trait MessageIdTrait {
+    /// identification if this object according to the `MessageId`
+    fn id(&self) -> MessageId;
+}
+
+/// The preamble is used to peek into messages so allowing for them to be routed
+/// by their identifier
+#[derive(Serialize, Deserialize, Debug)]
+struct MessagePreamble {
```
gila, line 171: MetaData or Header?

mbus-api/src/receive.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// May fail if the raw data cannot be deserialized into `S`
+    pub fn payload<S: Deserialize<'a>>(&self) -> io::Result<S> {
+        let request: SendPayload<S> =
+            serde_json::from_slice(&self.bus_msg.data)?;
+        Ok(request.data)
+    }
+
+    /// Get the identifier of this message.
+    /// May fail if the raw data cannot be deserialized into the preamble.
+    pub fn id(&self) -> io::Result<MessageId> {
```
gila, line 91: why are we not using the messageID trait here?

mbus-api/src/send.rs
```diff
@@ -0,0 +1,230 @@
...
+    /// todo: add timeout with retry logic?
+    pub(crate) async fn request(&self) -> io::Result<R> {
+        let payload = serde_json::to_vec(&self.payload)?;
+        let reply = self
+            .bus
+            .request(&self.channel.to_string(), &payload)
+            .await?
+            .data;
+        let reply: ReplyPayload<R> = serde_json::from_slice(&reply)?;
+        match reply.0 {
```
gila, line 223: you could use map_err() here and get rid of the match.

services/common/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+            let subscriptions = subscriptions.1.clone();
+
+            let handle = tokio::spawn(async move {
+                let mut handle = bus
+                    .subscribe(&channel.to_string())
+                    .await
+                    .expect("Subscription failed");
+
+                loop {
+                    // todo: tidy this up...
```
gila, line 153: ok let me know when thats done 👍

gila (commented):
@tiagolobocastro i've added several comments where some might not make any sense given the fact that there is a certain way of doing things with the messages themselves. 

Overall, this looks very good!

mayastor/src/core/env.rs
```diff
@@ -639,6 +640,26 @@ impl MayastorEnvironment {
         cfg.apply();
     }
 
+    #[allow(dead_code)]
+    fn get_service_config(&self) -> ReplyConfig {
+        if self.mbus_endpoint.is_some() {
+            tokio::runtime::Runtime::new().unwrap().block_on(async {
```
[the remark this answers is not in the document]
tiagolobocastro: Yeah... maybe we can follow similar kind of strategy to the reactors and send a future to be run by the tokio runtime we create?

mayastor/src/subsys/mbus/registration.rs
```diff
@@ -156,12 +140,12 @@ impl Registration {
...
     async fn register(&self) -> Result<(), Error> {
-        let payload = RegisterArgs {
+        let payload = Register {
             id: self.config.node.clone(),
             grpc_endpoint: self.config.grpc_endpoint.clone(),
         };
-        super::message_bus()
-            .publish(Channel::Register, &payload)
+
+        Register::publish(payload, Channel::Registry, super::message_bus())
```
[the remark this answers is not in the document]
tiagolobocastro: I was split between the two and ended up following this approach in an attempt to constraint the types, specifically for the request which expects a specific return.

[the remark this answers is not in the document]
tiagolobocastro: I was split between the two and ended up following this approach in an attempt to constraint the types, specifically for the request which expects a specific return.
I think we can get similar syntax by adding the bus() to the trait, so we'd have:
`Register::publish(payload,Channel)` with the bus hidden away.

mayastor/src/subsys/mbus/mbus_nats.rs
```diff
@@ -18,9 +14,16 @@ pub(super) fn message_bus_init(server: String) {
     });
 }
 
+pub fn message_bus<'a>() -> &'a nats::asynk::Connection {
```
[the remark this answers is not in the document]
tiagolobocastro: This was originally to due with the bus trait which I've omitted for the moment to make things simpler and so the bus was to remind me that I needed to return a bus here :) 
Once we have the bus, if we have it as a trait then we can't do mbus::get() because async traits don't play well with static methods :(

mbus-api/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+
+/// Message id which uniquely identifies every type of unsolicited message
+/// The solicited (replies) message do not currently carry an id as they
+/// are sent to a specific requested channel
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+pub enum MessageId {
+    /// Default
+    Default,
+    /// Update Config
+    UpdateConfig,
```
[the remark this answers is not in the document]
tiagolobocastro: yeah perhaps, these are mainly just dummies for me to try out the api.
Another thing I was split between... I think it's doable... by having it in the actual types so you'd get the Id almost for free as well :) 
Though I didn't like how my naming was turning out to be, eg Change(ChangeStruct)?
The other disadvantage is that I wouldn't be able to create examples with dummy types, but that's not a big deal.

mbus-api/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+
+/// Sender identification (eg which mayastor instance sent the message)
+pub type SenderId = String;
+
+/// Mayastor configurations
+/// Currently, we have the global mayastor config and the child states config
+#[derive(
+    Serialize, Deserialize, Debug, Clone, Eq, PartialEq, std::hash::Hash,
+)]
+pub enum ConfigKind {
```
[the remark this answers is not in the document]
tiagolobocastro: yeah, all these namings need scrubing tbh :(

mbus-api/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+bus_impl_all!(GetConfig, RequestConfig, ReplyConfig);
+
+/// Registration
+
+/// Register message payload
+#[derive(Serialize, Deserialize, Default, Debug, Clone)]
+pub struct Register {
+    /// id of the mayastor instance
+    pub id: String,
```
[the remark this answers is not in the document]
tiagolobocastro: These were just moved out from the mayastor into the lib.
I think the id is a string because that's how mayastor names the mayastor instances, though I guess we could type alias it.

mbus-api/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+#[derive(Serialize, Deserialize, Debug, Default, Clone)]
+pub struct ReplySvcName {
+    /// name of the service
+    pub name: String,
+}
+bus_impl_all!(GetSvcName, ReplySvcName);
+
+/// This trait must be implemented by every message type so it can
+/// be uniquely identifiable
+pub trait MessageIdTrait {
```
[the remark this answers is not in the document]
tiagolobocastro: The only reason to add the trait was because otherwise it'd collide with the actual MessageId type. Message seems appropriate.

mbus-api/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+#[derive(Serialize, Deserialize)]
+struct SendPayload<T> {
+    pub(crate) id: MessageId,
+    pub(crate) sender: SenderId,
+    pub(crate) data: T,
+}
+
+/// Simple Error type placeholder
+#[derive(Serialize, Deserialize, Debug)]
+pub enum Error {
```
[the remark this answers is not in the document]
tiagolobocastro: yes :)

mbus-api/src/receive.rs
```diff
@@ -0,0 +1,158 @@
...
+    pub fn sender(&self) -> SenderId {
+        self.request.sender.clone()
+    }
+
+    /// Respond back to the sender with the `reply` payload wrapped by
+    /// a Result-like type.
+    /// May fail if serialization of the reply fails or if the
+    /// message bus fails to respond.
+    /// Can receive either `R`, `Err()` or `ReplyPayload`.
+    pub async fn respond<T: Into<ReplyPayload<R>>>(
```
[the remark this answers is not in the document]
tiagolobocastro: that's a good question :)
Because nats respondes with a respond, so I renamed reply() to respond() but left the reply type as reply. I guess I can rename that to response. What sounds better to you, respond or reply?

mbus-api/src/receive.rs
tiagolobocastro: Because the MessageID trait is not fallible... though I guess we can make it fallible

mbus-api/src/receive.rs
tiagolobocastro: hmm but then it depends on what you pass on to the fn. Eg, If you call the fn with R rather than Ok(R) then we'd serialize R because we didn't do the .into to the ReplyPayload<R>?

mbus-api/src/receive.rs
```diff
@@ -0,0 +1,158 @@
...
+        &self,
+        reply: T,
+    ) -> io::Result<()> {
+        let reply: ReplyPayload<R> = reply.into();
```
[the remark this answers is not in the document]
tiagolobocastro: hmm why must it be Send+Sync?

services/common/src/lib.rs
tiagolobocastro: It's done :)

tiagolobocastro:
bors try

gila (approved):
:+1:

tiagolobocastro:
bors try

PR #458, opened 2020-10-07
CAS-461 support enumerating host block devices

cjones1024:
Initial support for enumerating block devices on a host.


mayastor/src/host/device.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    enumerator.match_subsystem("block")?;
+    enumerator.match_property("DEVTYPE", "disk")?;
+
+    for entry in enumerator.scan_devices()? {
+        let partitions = get_partitions(true, &entry, mounts)?;
+        let include = partitions.is_empty();
+
+        if let Some(device) = new_device(include, &entry, &mounts) {
+            if all || device.available {
```
tiagolobocastro, line 85: could we just push devices to the list and just "filter" them on the return?

also rather than having the "manual check" I wonder if we can have a list of filters which is constructed after parsing the cli arguments.

tiagolobocastro (commented):
I've ran it locally and seems to be working ok.
It is however identifying a reserved partition with an unknown file system as available.
It's also picking up ramdisks, which we might want to filter out?
I seem to have on my system some device mapper disks which do indeed have some size and are used for the LXD pool. We can actually find out if they're open, so maybe we should exclude open devices?

jonathan-teh (commented):
On my system it's marking `/dev/sdb1` as available but it's actually an encrypted swap partition, which is available as `/dev/dm-0` and is correctly marked as swap and unavailable with `-a`. The other partitions (2 ext4 and a vfat EFI system partition) and block devices (unused loop, nbd and card reader devices) have been correctly marked as unavailable. `dmsetup -o blkdevname deps` shows the underlying block device(s) each DM device depends on.

paulyoong (commented):
On my system it was showing the cdrom drive (/dev/sr0) as available

mayastor/src/host/device.rs
```diff
@@ -0,0 +1,124 @@
...
+        });
+    }
+    None
+}
+
+fn get_mounts() -> Result<HashMap<OsString, MountInfo>, Error> {
+    let mut table: HashMap<OsString, MountInfo> = HashMap::new();
+
+    for entry in MountIter::new()? {
+        if let Ok(mount) = entry {
```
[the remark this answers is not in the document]
cjones1024: Each iteration involves effectively parsing a mtab line. It is conceivable that the line fails to be parsed correctly, although I suspect if that happens you've got bigger problems.

mayastor/src/host/device.rs
cjones1024: That is a nice idea, although given the current time constraints is probably something to look at further down the line.

mayastor/Cargo.toml
```diff
@@ -72,6 +73,7 @@ tracing = "0.1"
 tracing-futures = "0.2.4"
 tracing-log = "0.1.1"
 tracing-subscriber = "0.2.0"
+udev = "0.4"
```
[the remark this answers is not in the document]
cjones1024: I just used the same version as we're using the csi plugin.
I don't see any reason for not upgrading but should be done for csi as well.

mayastor/src/host/device.rs
tiagolobocastro: :+1:

gila:
> On my system it's marking `/dev/sdb1` as available but it's actually an encrypted swap partition, which is available as `/dev/dm-0` and is correctly marked as swap and unavailable with `-a`. The other partitions (2 ext4 and a vfat EFI system partition) and block devices (unused loop, nbd and card reader devices) have been correctly marked as unavailable. `dmsetup -o blkdevname deps` shows the underlying block device(s) each DM device depends on.

If a device is in use by device mapper can easily be obtained by looking at the sysfs holder attribute. @cjones1024 the example code I shared does that as well iirc.

mayastor/src/bin/cli/device_cli.rs
```diff
@@ -0,0 +1,118 @@
+use super::context::Context;
+use ::rpc::mayastor as rpc;
+use clap::{App, AppSettings, Arg, ArgMatches, SubCommand};
+use tonic::Status;
+
+pub fn subcommands<'a, 'b>() -> App<'a, 'b> {
+    let list = SubCommand::with_name("list")
+        .about("List host block devices")
```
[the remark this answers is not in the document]
jkryl: right. Perhaps adding something like "suitable for holding storage pool". Or we could make the help message for "all" option more descriptive to make the difference clear.

rpc/proto/mayastor.proto
```diff
@@ -64,6 +64,9 @@ service Mayastor {
 
   // Snapshot operations
   rpc CreateSnapshot (CreateSnapshotRequest) returns (CreateSnapshotReply) {}
+
+  // Enumerate block devices on current host
+  rpc GetHostBlockDevices (GetHostBlockDevicesRequest) returns (GetHostBlockDevicesReply) {}
```
jkryl, line 69: Minor issue related to consistent names: isn't the "Host" word in method and struct names a bit useless? All methods that are defined in this file deal with objects that are on the Host so that is implicitly assumed. Just like we don't have ListHostReplicas. Shorter method and struct names -> shorter code :-)

On the same note according to rules used in this file so far: "get" means give me one object. "list" means give me an array of objects. So more appropriate name for this method would be "ListBlockDevices" (and similar name change for parameters).

rpc/proto/mayastor.proto
```diff
@@ -368,3 +371,38 @@ message BdevUri {
...
+  message Filesystem {
+    string fstype = 1;          // filesystem type: ext3, ntfs, ...
+    string label = 2;           // volume label
+    string uuid = 3;            // UUID identifying the volume (filesystem)
+    string mountpoint = 4;      // path where filesystem is currently mounted
+  }
+  string devname = 1;           // entry in /dev associated with device
+  string devtype = 2;           // currently "disk" or "partition"
+  uint64 devmajor = 3;          // major device number
+  uint64 devminor = 4;          // minor device number
```
jkryl, line 393: devmajor and devminor definitely don't need to be 64 bit ints. IIRC these are normally represented by a single byte in C. Would be good to keep the same size as the OS.

rpc/proto/mayastor.proto
```diff
@@ -368,3 +371,38 @@ message BdevUri {
...
+    string scheme = 4;          // partition scheme: gpt, dos, ...
+    string typeid = 5;          // partition type identifier
+    string uuid = 6;            // UUID identifying partition
+  }
+  message Filesystem {
+    string fstype = 1;          // filesystem type: ext3, ntfs, ...
+    string label = 2;           // volume label
+    string uuid = 3;            // UUID identifying the volume (filesystem)
+    string mountpoint = 4;      // path where filesystem is currently mounted
+  }
```
jkryl, line 389: is there a reason why these structs are defined nested? It seems that if one needs to refer to the struct name, he/she has to do BlockDevice.Filesystem instead of just Filesystem. That's probably ok. Though we don't do that anywhere else so just asking what is the intent here.

cjones1024: The Partition struct contains information pertaining *only* to partitions and likewise Filesystem contains *only* information related to the file system. In many cases one or both of these will contain no data - and indeed rust compiles these to Option<Filesystem>, etc . In our use case in particular, we are mainly interested in the case when this object has a value "None". In fact this is precisely one of the criteria for deciding whether or not the device is available for use.

rpc/proto/mayastor.proto
cjones1024: Not quite a single byte.
Linux currently uses a 32-bit dev_t with 12 bits for major and 20 bits for minor.
Apparently glibc uses a 64-bit dev_t with 32 bits each for major and minor.
So probably best to change to u32 here.

rpc/proto/mayastor.proto
cjones1024: Will change these

rpc/proto/mayastor.proto
jkryl: agree and thanks for looking up the right size 👍

jkryl (approved):

mayastor/src/host/device.rs
```diff
@@ -0,0 +1,305 @@
...
+        }
+    }
+
+    Ok(list)
+}
+
+/// Return a list of block devices on the current host.
+/// The <all> parameter controls whether to return list containing
+/// all matching devices, or just those deemed to be available.
+pub async fn get_block_devices(all: bool) -> Result<Vec<BlockDevice>, Error> {
```
tiagolobocastro, line 302: this doesn't need to be async?

tiagolobocastro (approved):

mayastor/src/host/device.rs
```diff
@@ -0,0 +1,305 @@
...
+impl From<Property<'_>> for u64 {
+    fn from(property: Property) -> Self {
+        Option::<Self>::from(property).unwrap_or(0)
+    }
+}
+
+// Determine the type of devices which may be potentially presented
+// as "available" for use
+fn usable_device(devmajor: &u64) -> bool {
+    const DEVICE_TYPES: [u64; 4] = [7, 8, 43, 259];
```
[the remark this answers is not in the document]
cjones1024: Possibly, although have added comments to identify the meaning of each number.

mayastor/src/bin/cli/device_cli.rs
```diff
@@ -0,0 +1,182 @@
...
+        .arg(
+            Arg::with_name("all")
+                .short("a")
+                .long("all")
+                .takes_value(false)
+                .help("list all devices"),
+        )
+        .arg(
```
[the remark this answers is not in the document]
cjones1024: I may raise a Jira to address this later, as if we are to do it we probably should do it consistently accross all commands.

mayastor/src/host/device.rs
```diff
@@ -0,0 +1,327 @@
+//!
+//! This module implements the list_block_devices() gRPC method
```
gila, line 2: Can we please rename this to blk_device, more devices will come in the future.

mayastor/src/host/device.rs
```diff
@@ -0,0 +1,327 @@
...
+
+impl From<Property<'_>> for u64 {
+    fn from(property: Property) -> Self {
+        Option::<Self>::from(property).unwrap_or(0)
+    }
+}
+
+// Determine the type of devices which may be potentially presented
+// as "available" for use.
+fn usable_device(devmajor: &u32) -> bool {
```
gila, line 84: what is the major for nvme? Also, have you considered leveraging https://github.com/stratis-storage/libblkid-rs it is we already have liblkid iirc. Not sure if it brings anything for our use case right now, perhaps something for the future?

mayastor/src/host/device.rs
```diff
@@ -0,0 +1,327 @@
...
+fn usable_partition(partition: &Option<Partition>) -> bool {
+    const GPT_PARTITION_TYPES: [&str; 1] = [
+        "0fc63daf-8483-4772-8e79-3d69d8477de4", // Linux
+    ];
+
+    const MBR_PARTITION_TYPES: [&str; 1] = [
+        "0x83", // Linux
+    ];
+
+    if let Some(part) = partition {
```
gila, line 113: you can write this as const ["UUID1", "UUID2"].iter().any(...) this avoids the if statements

cjones1024:
bors merge

mayastor/src/host/device.rs
cjones1024: nvme ends up with major 259.
Will investigate libblkid-rs as possible future enhancement, although the goal here was to keep it as simpe as possible at least in the first instance.

cjones1024:
bors merge

cjones1024:
bors retry

PR #461, opened 2020-10-13
Remove unnecessary force flag which...

tiagolobocastro:
... applies only to NFS during volume unpublish calls
fixes an issue where deleting a pod using our pvc was stuck
because the umount force call was failing.

### DIFF
--- a/csi/src/mount.rs
+++ b/csi/src/mount.rs
@@ -257,9 +257,7 @@ pub fn bind_remount(target: &str, options: &[String]) -> Result<Mount, Error> {
 /// Unmounts a path that has previously been bind mounted.
 /// Should not be used for unmounting devices.
 pub fn bind_unmount(target: &str) -> Result<(), Error> {
-    let mut flags = UnmountFlags::empty();
-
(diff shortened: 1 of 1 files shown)


blaisedias (approved):
Maybe also remove force from   blockdevice_unmount()?

csi/src/mount.rs
```diff
@@ -303,7 +301,7 @@ pub fn blockdevice_mount(
 
 /// Unmount a block device.
 pub fn blockdevice_unmount(target: &str) -> Result<(), Error> {
-    let flags = UnmountFlags::FORCE;
+    let flags = UnmountFlags::empty();
 
     debug!("Unmounting block device {} (flags={:?}) ...", target, flags);
```
cjones1024, line 306: There's no real point in having the flags value in the debug message now either

cjones1024 (approved):

csi/src/mount.rs
tiagolobocastro: Yeah, I kept it just in case it changes in the future but can remove

tiagolobocastro:
bors merge

PR #462, opened 2020-10-14
Create service associated with CSI plugin. CAS-480

blaisedias:
Create service associated with CSI plugin to service
filesystem freeze unfreeze requests.

To properly snapshot an filesystem volume,
the FS must be frozen prior to creating the
lvol snapshot and then be unfrozen.
To do this use the linux utililty fsfreeze which
must be run on the node on which the volume has
been mounted.
MOAC will make remote calls to the service running
on the node to do this.

### DIFF
--- a/csi/build.rs
+++ b/csi/build.rs
@@ -4,5 +4,9 @@ fn main() {
     tonic_build::configure()
         .build_server(true)
         .compile(&["proto/csi.proto"], &["proto"])
-        .unwrap_or_else(|e| panic!("csi protobuf compilation failed: {}", e));
+        .expect("csi protobuf compilation failed");
+    tonic_build::configure()
+        .build_server(true)
+        .compile(&["proto/mayastornodeplugin.proto"], &["proto"])
+        .expect("mayastor node grpc service protobuf compilation failed");
 }
--- a/csi/proto/mayastornodeplugin.proto
+++ b/csi/proto/mayastornodeplugin.proto
@@ -1,0 +1,44 @@
+// The definition of mayastor node plugin gRPC interface.
+// The node plugin service runs on all nodes running
+// the Mayastor CSI node plugin, and is complementary
+// to the CSI node plugin service.
+
+// This interface is supposed to be independent on particular computing
+// environment (i.e. kubernetes).
+
+syntax = "proto3";
+
+option java_multiple_files = true;
+option java_package = "io.grpc.examples.mayastornodeplugin";
+option java_outer_classname = "MayastorNodePluginProto";
+
+package mayastornodeplugin;
+
+// Service for freezing and unfreezing file systems.
+service MayastorNodePlugin {
+  // Freeze the file system identified by the volume ID
+  // no check is made if the file system had been previously frozen.
+  rpc FreezeFS (FreezeFSRequest) returns (FreezeFSReply) {}
+  // Unfreeze the file system identified by the volume ID,
+  // no check is made if the file system had been previously frozen.
+  rpc UnfreezeFS (UnfreezeFSRequest) returns (UnfreezeFSReply) {}
+}
+
+// The request message containing ID of the volume to be frozen
+message FreezeFSRequest {
+  string volume_id = 1;
+}
+
+// The response message for the freeze request.
+message FreezeFSReply {
+}
+
+// The request message containing ID of the volume to be unfrozen
+message UnfreezeFSRequest {
+
+  string volume_id = 1;
+}
+
+// The response message for the unfreeze request.
+message UnfreezeFSReply {
+}
--- a/csi/src/freezefs.rs
+++ b/csi/src/freezefs.rs
@@ -1,0 +1,75 @@
+//! The files system freeze support using linux utility fsfreeze
+use crate::{
+    dev::{Device, DeviceError},
+    mount,
+};
+use snafu::{ResultExt, Snafu};
+use tokio::process::Command;
+use uuid::Uuid;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility = "pub(crate)")]
+pub enum FreezeFsError {
+    #[snafu(display("Cannot find volume: volume ID: {}", volid))]
+    VolumeNotFound { volid: String },
+    #[snafu(display("Invalid volume ID: {}, {}", volid, source))]
+    InvalidVolumeId {
+        source: uuid::parser::ParseError,
+        volid: String,
+    },
+    #[snafu(display("fsfreeze failed: volume ID: {}, {}", volid, error))]
+    FsfreezeFailed { volid: String, error: String },
+    #[snafu(display("Internal failure: volume ID:{}, {}", volid, source))]
+    InternalFailure { source: DeviceError, volid: String },
+    #[snafu(display("IO error: volume ID: {}, {}", volid, source))]
+    IOError {
+        source: std::io::Error,
+        volid: String,
+    },
+}
+
+const FSFREEZE: &str = "fsfreeze";
+
+async fn fsfreeze(
+    volume_id: &str,
+    freeze_op: &str,
+) -> Result<(), FreezeFsError> {
+    let uuid = Uuid::parse_str(volume_id).context(InvalidVolumeId {
+        volid: volume_id.to_string(),
+    })?;
+
+    if let Some(device) =
+        Device::lookup(&uuid).await.context(InternalFailure {
+            volid: volume_id.to_string(),
+        })?
+    {
+        let device_path = device.devname();
+        if let Some(mnt) = mount::find_mount(Some(&device_path), None) {
+            let args = [freeze_op, &mnt.dest];
+            let output =
+                Command::new(FSFREEZE).args(&args).output().await.context(
+                    IOError {
+                        volid: volume_id.to_string(),
+                    },
+                )?;
+            if output.status.success() {
+                return Ok(());
+            } else {
+                return Err(FreezeFsError::FsfreezeFailed {
+                    volid: volume_id.to_string(),
+                    error: String::from_utf8(output.stderr).unwrap(),
+                });
+            }
+        }
+    }
+    Err(FreezeFsError::VolumeNotFound {
+        volid: volume_id.to_string(),
+    })
+}
+pub async fn freeze_volume(volume_id: &str) -> Result<(), FreezeFsError> {
+    fsfreeze(volume_id, "--freeze").await
+}
+
+pub async fn unfreeze_volume(volume_id: &str) -> Result<(), FreezeFsError> {
+    fsfreeze(volume_id, "--unfreeze").await
+}
--- a/csi/src/nodeplugin_grpc.rs
+++ b/csi/src/nodeplugin_grpc.rs
@@ -1,0 +1,85 @@
+//! The mayastor node plugin gRPC service
+//! This provides access to functionality that needs to be executed on the same
+//! node as a Mayastor CSI node plugin, but it is not possible to do so within
+//! the CSI framework. This service must be deployed on all nodes the
+//! Mayastor CSI node plugin is deployed.
+use crate::freezefs;
+use freezefs::{freeze_volume, unfreeze_volume, FreezeFsError};
+use mayastor_node_plugin::*;
+use tonic::{transport::Server, Code, Request, Response, Status};
+
+pub mod mayastor_node_plugin {
+    tonic::include_proto!("mayastornodeplugin");
+}
+
+#[derive(Debug, Default)]
+pub struct MayastorNodePluginSvc {}
+
+impl From<FreezeFsError> for Status {
+    fn from(err: FreezeFsError) -> Self {
+        match err {
+            FreezeFsError::VolumeNotFound {
+                ..
+            } => Status::new(Code::NotFound, err.to_string()),
+            FreezeFsError::FsfreezeFailed {
+                ..
+            } => Status::new(Code::Internal, err.to_string()),
+            FreezeFsError::InvalidVolumeId {
+                ..
+            } => Status::new(Code::InvalidArgument, err.to_string()),
+            FreezeFsError::InternalFailure {
+                ..
+            } => Status::new(Code::Internal, err.to_string()),
+            FreezeFsError::IOError {
+                ..
+            } => Status::new(Code::Unknown, err.to_string()),
+        }
+    }
+}
+
+#[tonic::async_trait]
+impl mayastor_node_plugin_server::MayastorNodePlugin for MayastorNodePluginSvc {
+    async fn freeze_fs(
+        &self,
+        request: Request<FreezeFsRequest>,
+    ) -> Result<Response<FreezeFsReply>, Status> {
+        let volume_id = request.into_inner().volume_id;
+        debug!("freeze_fs({})", volume_id);
+        freeze_volume(&volume_id).await?;
+        Ok(Response::new(mayastor_node_plugin::FreezeFsReply {}))
+    }
+
+    async fn unfreeze_fs(
+        &self,
+        request: Request<UnfreezeFsRequest>,
+    ) -> Result<Response<UnfreezeFsReply>, Status> {
+        let volume_id = request.into_inner().volume_id;
+        debug!("unfreeze_fs({})", volume_id);
+        unfreeze_volume(&volume_id).await?;
+        Ok(Response::new(mayastor_node_plugin::UnfreezeFsReply {}))
+    }
+}
+
+pub struct MayastorNodePluginGrpcServer {}
+
+impl MayastorNodePluginGrpcServer {
+    pub async fn run(endpoint: std::net::SocketAddr) -> Result<(), ()> {
+        info!(
+            "Mayastor node plugin gRPC server configured at address {:?}",
+            endpoint
+        );
+        if let Err(e) = Server::builder()
+            .add_service(
+                mayastor_node_plugin_server::MayastorNodePluginServer::new(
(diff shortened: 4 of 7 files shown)


blaisedias:
bors try

blaisedias:
bors try

csi/build.rs
```diff
@@ -5,4 +5,13 @@ fn main() {
...
         .compile(&["proto/csi.proto"], &["proto"])
         .unwrap_or_else(|e| panic!("csi protobuf compilation failed: {}", e));
+    tonic_build::configure()
+        .build_server(true)
+        .compile(&["proto/mayastornodeplugin.proto"], &["proto"])
+        .unwrap_or_else(|e| {
+            panic!(
+                "mayastor node grpc service protobuf compilation failed: {}",
+                e
+            )
```
cjones1024, line 15: should use expect() here

csi/src/server.rs
```diff
@@ -109,8 +111,18 @@ impl AsyncWrite for UnixStream {
...
 }
 
-#[tokio::main]
-async fn main() -> Result<(), String> {
+macro_rules! default_port {
+    () => {
+        10199
+    };
+}
+
```
cjones1024, line 119: Not sure why we need a macro here. Why not just define a constant?

csi/src/nodeplugin_grpc.rs
```diff
@@ -0,0 +1,97 @@
...
+
+pub mod mayastor_node_plugin {
+    tonic::include_proto!("mayastornodeplugin");
+}
+
+#[derive(Debug, Default)]
+pub struct MayastorNodePluginSvc {}
+
+fn freeze_fs_err_to_status(err: FreezeFsError) -> Status {
+    let msg = format!("{:?}", err);
```
cjones1024, line 19: It would be better to implement From trait here.

csi/src/server.rs
```diff
@@ -191,16 +213,37 @@ async fn main() -> Result<(), String> {
...
+                node_name: node_name.into(),
+                filesystems: probe_filesystems(),
+            }))
+            .add_service(IdentityServer::new(Identity {}))
+            .serve_with_incoming(uds_sock.incoming().map_ok(UnixStream));
+
+        match uds.await {
+            Ok(_) => Ok(()),
+            Err(e) => {
+                println!("CSI node server failed with error: {}", e);
```
cjones1024, line 244: probably should be using error! here

csi/src/freezefs.rs
```diff
@@ -0,0 +1,73 @@
...
+    IOError { error: String },
+}
+
+impl From<std::io::Error> for FreezeFsError {
+    fn from(error: std::io::Error) -> FreezeFsError {
+        FreezeFsError::IOError {
+            error: format!("{}", error),
+        }
+    }
+}
```
cjones1024, line 28: If you create an appropriate FreezeFsError variant for std::Io:Error, there is no real need for this.

blaisedias: Done. https://github.com/openebs/Mayastor/pull/462/commits/7c714899d911a2f659d5a7861968ba08b1aa354b

csi/src/server.rs
blaisedias: Done. https://github.com/openebs/Mayastor/pull/462/commits/7c714899d911a2f659d5a7861968ba08b1aa354b

csi/src/nodeplugin_grpc.rs
blaisedias: Done. https://github.com/openebs/Mayastor/pull/462/commits/7c714899d911a2f659d5a7861968ba08b1aa354b

csi/src/server.rs
blaisedias: Done. https://github.com/openebs/Mayastor/pull/462/commits/7c714899d911a2f659d5a7861968ba08b1aa354b

csi/build.rs
blaisedias: I'll leave it as it is, as we use panic! elsewhere.

blaisedias:
bors try

csi/src/freezefs.rs
```diff
@@ -0,0 +1,75 @@
...
+async fn fsfreeze(
+    volume_id: &str,
+    freeze_op: &str,
+) -> Result<(), FreezeFsError> {
+    let uuid = Uuid::parse_str(volume_id).context(InvalidVolumeId {
+        volid: volume_id.to_string(),
+    })?;
+
+    if let Some(device) =
+        Device::lookup(&uuid).await.context(InternalFailure {
```
gila, line 42: is device::lookup returning an Option<Result<T,E>>? looks to me we want to fix that (create an issue for it?)

csi/src/nodeplugin_grpc.rs
```diff
@@ -0,0 +1,100 @@
...
+
+#[derive(Debug, Default)]
+pub struct MayastorNodePluginSvc {}
+
+impl From<FreezeFsError> for Status {
+    fn from(err: FreezeFsError) -> Self {
+        let msg = format!("{:?}", err);
+        match err {
+            FreezeFsError::VolumeNotFound {
+                volid: _,
```
gila, line 23: what is this _, about? does it mean: I don't really care what the value is?  If so it is likely nice to write something like:

FreezeFsError::VolumeNotfound { .. } => {  }, so use dots..

csi/src/nodeplugin_grpc.rs
```diff
@@ -0,0 +1,100 @@
...
+pub mod mayastor_node_plugin {
+    tonic::include_proto!("mayastornodeplugin");
+}
+
+#[derive(Debug, Default)]
+pub struct MayastorNodePluginSvc {}
+
+impl From<FreezeFsError> for Status {
+    fn from(err: FreezeFsError) -> Self {
+        let msg = format!("{:?}", err);
```
gila, line 20: This can also be done as Status::new(xxx, err.into())

csi/src/nodeplugin_grpc.rs
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+
+    async fn unfreeze_fs(
+        &self,
+        request: Request<UnfreezeFsRequest>,
+    ) -> Result<Response<UnfreezeFsReply>, Status> {
+        let volume_id = request.into_inner().volume_id;
+        debug!("unfreeze_fs({})", volume_id);
+        match unfreeze_volume(&volume_id).await {
```
gila, line 65: what about

```rust
unfreeze_volume(&volume_id).await?;
Ok(Repsponse::new( ... ))
```

csi/src/server.rs
```diff
@@ -109,8 +111,19 @@ impl AsyncWrite for UnixStream {
     }
 }
 
-#[tokio::main]
-async fn main() -> Result<(), String> {
+const GRPC_PORT: u16 = 10199;
+/// If endpoint is missing a port number then add the default one.
+pub fn grpc_endpoint(endpoint: String) -> std::net::SocketAddr {
+    (if endpoint.contains(':') {
```
gila, line 117: it is not common to use anonymous functions this way (i did not even know it was possible to be honest) what about something like this:
```
let s_addr = if endpoint.containers(':') { endpoint } else { format!( ... ) };
s_addr.parse().expect("...")
```

csi/src/nodeplugin_grpc.rs
```diff
@@ -0,0 +1,100 @@
...
+            FreezeFsError::IOError {
+                volid: _,
+                source: _,
+            } => Status::new(Code::Unknown, msg),
+        }
+    }
+}
+
+#[tonic::async_trait]
+impl mayastor_node_plugin_server::MayastorNodePlugin for MayastorNodePluginSvc {
```
gila, line 46: theseAreRatherLongNamesIsThatNeeded?

csi/src/nodeplugin_grpc.rs
```diff
@@ -0,0 +1,100 @@
...
+            endpoint
+        );
+        let svc = Server::builder()
+            .add_service(
+                mayastor_node_plugin_server::MayastorNodePluginServer::new(
+                    MayastorNodePluginSvc {},
+                ),
+            )
+            .serve(endpoint);
+        match svc.await {
```
gila, line 89: This seems like a copy-paste from the mayastor service as the match svc.await is unless. We can simply do svc.await; or even remove the svc. This gets rid of a useless match here.

csi/src/server.rs
```diff
@@ -191,16 +214,36 @@ async fn main() -> Result<(), String> {
...
-            filesystems: probe_filesystems(),
-        }))
-        .add_service(IdentityServer::new(Identity {}))
-        .serve_with_incoming(uds_sock.incoming().map_ok(UnixStream));
-    let _ = uds.await;
+    let fut_csi = CSIServer::run(csi_socket, node_name);
+    let fut_grpc =
+        MayastorNodePluginGrpcServer::run(grpc_endpoint(endpoint.to_string()));
+    let futs = join(fut_csi, fut_grpc);
+    let mut rt = Runtime::new().expect("failed to obtain a new Runtime object");
```
gila, line 221: So its not clear what type of runtime we are building here. Please check the docs, because let us say I have a 96 core server, we might now have 96 threads ... a little much too much, don't you agree? So you can specify what you want exactly i.e:

let mut rt = Runtime::Builder::new().basic_scheduler().enable_io().whatever().build().unwrap();
rt_block_on( ....) 

its not clear though -- as to why you removed tokio::main?

csi/src/server.rs
```diff
@@ -191,16 +214,36 @@ async fn main() -> Result<(), String> {
...
+
+        let uds = Server::builder()
+            .add_service(NodeServer::new(Node {
+                node_name: node_name.into(),
+                filesystems: probe_filesystems(),
+            }))
+            .add_service(IdentityServer::new(Identity {}))
+            .serve_with_incoming(uds_sock.incoming().map_ok(UnixStream));
+
+        match uds.await {
```
gila, line 241: same as the previous comment

csi/src/freezefs.rs
blaisedias: Yes, the function is defined as `pub async fn lookup( uuid: &Uuid,) -> Result<Option<Box<dyn Detach>>, DeviceError>`

csi/src/server.rs
blaisedias: Done.

csi/src/nodeplugin_grpc.rs
blaisedias: Done.

csi/src/server.rs
blaisedias: Done

csi/src/nodeplugin_grpc.rs
blaisedias: Done.

csi/src/nodeplugin_grpc.rs
blaisedias: Done

csi/src/nodeplugin_grpc.rs
blaisedias: The long name is a consequence of this line in the .proto file 
```
service MayastorNodePlugin { 
```
I elected for a descriptive name for the service as opposed to a shorter more cryptic one.

csi/build.rs
tiagolobocastro: I suspect because it's printing e rather than just an expect message

csi/src/freezefs.rs
```diff
@@ -0,0 +1,75 @@
...
+        source: std::io::Error,
+        volid: String,
+    },
+}
+
+const FSFREEZE: &str = "fsfreeze";
+
+async fn fsfreeze(
+    volume_id: &str,
+    freeze_op: &str,
```
tiagolobocastro, line 35: should this be an enum?

csi/src/nodeplugin_grpc.rs
```diff
@@ -0,0 +1,82 @@
...
+
+pub struct MayastorNodePluginGrpcServer {}
+
+impl MayastorNodePluginGrpcServer {
+    pub async fn run(endpoint: std::net::SocketAddr) -> Result<(), ()> {
+        info!(
+            "Mayastor node plugin gRPC server configured at address {:?}",
+            endpoint
+        );
+        let _ = Server::builder()
```
tiagolobocastro, line 72: don't we need to check `_` for errors?

csi/src/server.rs
```diff
@@ -191,16 +204,37 @@ async fn main() -> Result<(), String> {
...
 }
+
+struct CSIServer {}
+
+impl CSIServer {
+    pub async fn run(csi_socket: &str, node_name: &str) -> Result<(), String> {
+        let mut uds_sock = UnixListener::bind(csi_socket).unwrap();
+        info!("CSI plugin bound to {}", csi_socket);
+
+        let _ = Server::builder()
```
tiagolobocastro, line 229: same comment as before for `_`.

csi/build.rs
tiagolobocastro: I suspect because it's printing e rather than just an expect message.

csi/src/nodeplugin_grpc.rs
blaisedias: Done.

csi/src/server.rs
blaisedias: Done

csi/src/server.rs
blaisedias: Fixed, used tokio::main and tokio::join

csi/src/freezefs.rs
blaisedias: Depends, I didn't see any point in doing that. The function is  internal to the freezefs, and I'd have to convert the enum to a string for the command line args, so elected to just pass the string itself.

csi/src/nodeplugin_grpc.rs
blaisedias: Done.

csi/src/server.rs
blaisedias: Done.

csi/src/nodeplugin_grpc.rs
```diff
@@ -0,0 +1,85 @@
+//! The mayastor node plugin gRPC service
+//! This provides access to functionality that needs to be executed on the same
+//! node as a Mayastor CSI node plugin, but it is not possible to do so within
+//! the CSI framework. This service must be deployed on all nodes the the
```
paulyoong, line 4: The additional "the" should be removed from the end of the line.

blaisedias: Fixed.

csi/proto/mayastornodeplugin.proto
```diff
@@ -0,0 +1,44 @@
...
+message FreezeFSReply {
+}
+
+// The request message containing ID of the volume to be unfrozen
+message UnfreezeFSRequest {
+
+  string volume_id = 1;
+}
+
+// The response message for the freeze request.
```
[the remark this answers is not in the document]
blaisedias: Fixed

csi/build.rs
cjones1024: This is from the definition of expect
`pub fn expect(self, msg: &str) -> T
Returns the contained Ok value, consuming the self value.
Panics if the value is an Err, with a panic message including the passed message, and the content of the Err.`

So expect() *will* generate a panic with exactly what we want: a message that includes the value passed to expect() followed by the error message. In fact the above is exactly equivalent to
`.expect("mayastor node grpc service protobuf compilation failed")`

csi/src/freezefs.rs
cjones1024: This is correct. It is possible that the lookup will fail because of an error. Or it may just fail to find what we're looking for because it is not there - which is not an error.

csi/src/nodeplugin_grpc.rs
blaisedias: Fixed.

csi/build.rs
blaisedias: Done

gila (approved):

csi/build.rs
```diff
@@ -4,5 +4,9 @@ fn main() {
     tonic_build::configure()
         .build_server(true)
         .compile(&["proto/csi.proto"], &["proto"])
-        .unwrap_or_else(|e| panic!("csi protobuf compilation failed: {}", e));
+        .expect("csi protobuf compilation failed.");
```
cjones1024, line 7: minor nit: should leave out the full stop at the end, as the message in the panic will follow this with a ':'

csi/build.rs
```diff
@@ -4,5 +4,9 @@ fn main() {
...
+    tonic_build::configure()
+        .build_server(true)
+        .compile(&["proto/mayastornodeplugin.proto"], &["proto"])
+        .expect("mayastor node grpc service protobuf compilation failed.");
```
cjones1024, line 11: and here

csi/proto/mayastornodeplugin.proto
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// The request message containing ID of the volume to be frozen
+message FreezeFSRequest {
+  string volume_id = 1;
+}
+
+// The response message for the freeze request.
+message FreezeFSReply {
+}
```
cjones1024, line 34: Might be better to create an explicit Null reply message as we do for the mayastor API (see Mayastor/rpc/proto/mayastor.proto)

blaisedias:
bors try

cjones1024 (approved):

blaisedias:
bors merge

PR #463, opened 2020-10-16
core: rework main.rs

gila:
A lot of the "main" code is done with in the start closure of
mayastor. This PR moves this into main.rs where it should be. The
older .start() method is still present as they are used by tests but
over time we can phase it out.

The need for this - is mostly to avoid spawning runtimes during init

### DIFF
--- a/mayastor/src/bin/main.rs
+++ b/mayastor/src/bin/main.rs
@@ -1,20 +1,25 @@
 #[macro_use]
 extern crate tracing;
 
-use std::path::Path;
-
-use structopt::StructOpt;
-
+use futures::FutureExt;
 use mayastor::{
     bdev::util::uring,
-    core::{MayastorCliArgs, MayastorEnvironment},
+    core::{MayastorCliArgs, MayastorEnvironment, Reactors},
+    grpc,
     logger,
+    subsys,
 };
-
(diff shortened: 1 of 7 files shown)


gila:
bors try

mayastor/src/bin/main.rs
```diff
@@ -44,10 +49,21 @@ fn main() -> Result<(), std::io::Error> {
         if uring_supported { "yes" } else { "no" }
     );
     info!("free_pages: {} nr_pages: {}", free_pages, nr_pages);
-    let env = MayastorEnvironment::new(args);
-    env.start(|| {
-        info!("Mayastor started {} ...", '\u{1F680}');
```
jonathan-teh, line 49: Is there a replacement for this message?

gila: Ah good point

jonathan-teh (approved):

tiagolobocastro (approved):
:+1: 
Should the fn start() comment be updated to reflect that it's now used for test only?

gila:
> +1
> Should the fn start() comment be updated to reflect that it's now used for test only?

Im redoing the tests so it should be gone within the next couple of PRs :+1:

PR #464, opened 2020-10-19
CAS-477: snapshot creation fails on replicas which are not shared

jonathan-teh:
Implement create snapshots for local replicas by having a separate codepath for lvol bdevs that creates the snapshot.

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_fn_table.rs
+++ b/mayastor/src/bdev/nexus/nexus_fn_table.rs
@@ -61,11 +61,11 @@ impl NexusFnTable {
         let nexus = unsafe { Nexus::from_raw(ctx) };
         match io_type {
             // we always assume the device supports read/write commands
-            io_type::READ | io_type::WRITE => true,
+            // allow NVMe Admin as it is needed for local replicas
+            io_type::READ | io_type::WRITE | io_type::NVME_ADMIN => true,
             io_type::FLUSH
             | io_type::RESET
             | io_type::UNMAP
-            | io_type::NVME_ADMIN
             | io_type::WRITE_ZEROES => {
                 let supported = nexus.io_is_supported(io_type);
                 if !supported {
@@ -145,13 +145,7 @@ impl NexusFnTable {
                     nio.fail()
                 }
             }
-            io_type::NVME_ADMIN => {
-                if nexus.io_is_supported(io_type) {
-                    nexus.nvme_admin(&nio, &ch)
-                } else {
-                    nio.fail()
-                }
-            }
+            io_type::NVME_ADMIN => nexus.nvme_admin(&nio, &ch),
             _ => panic!(
                 "{} Received unsupported IO! type {}",
                 nexus.name, io_type
--- a/mayastor/src/lvs/lvol.rs
+++ b/mayastor/src/lvs/lvol.rs
@@ -25,7 +25,8 @@ use spdk_sys::{
 };
 
 use crate::{
-    core::{Bdev, CoreError, Protocol, Share},
+    bdev::nexus::nexus_bdev::Nexus,
+    core::{Bdev, CoreError, Mthread, Protocol, Share},
     ffihelper::{
         cb_arg,
         errno_result_from_i32,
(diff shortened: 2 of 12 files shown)


jonathan-teh:
bors try

mayastor/src/bdev/nexus/nexus_fn_table.rs
```diff
@@ -61,11 +61,11 @@ impl NexusFnTable {
         let nexus = unsafe { Nexus::from_raw(ctx) };
         match io_type {
             // we always assume the device supports read/write commands
-            io_type::READ | io_type::WRITE => true,
+            // allow NVMe Admin as it is needed for local replicas
+            io_type::READ | io_type::WRITE | io_type::NVME_ADMIN => true,
```
tiagolobocastro, line 65: do we need a local replica test below? This way it's opening it up for anything, eg what if a child does not support it (eg iscsi) ?

mayastor/src/lvs/lvol.rs
```diff
@@ -398,6 +399,42 @@ impl Lvol {
...
+            errno: i32,
+        ) {
+            if errno != 0 {
+                error!("vbdev_lvol_create_snapshot errno {}", errno);
+            }
+            // Must complete IO on thread IO was submitted from
+            let thread = Mthread::from_null_checked(unsafe {
+                spdk_sys::spdk_bdev_io_get_thread(bio_ptr.cast())
+            })
+            .unwrap();
```
tiagolobocastro, line 423: should this be an expect?

mayastor/src/lvs/lvol.rs
```diff
@@ -398,6 +399,42 @@ impl Lvol {
...
+        io: *mut spdk_sys::spdk_bdev_io,
+        snapshot_name: &str,
+    ) {
+        extern "C" fn snapshot_done_cb(
+            bio_ptr: *mut c_void,
+            _lvol_ptr: *mut spdk_lvol,
+            errno: i32,
+        ) {
+            if errno != 0 {
+                error!("vbdev_lvol_create_snapshot errno {}", errno);
```
tiagolobocastro, line 417: if there can be an error I guess we should fail the create_snapshot?

mayastor/src/subsys/nvmf/admin_cmd.rs
```diff
@@ -134,6 +133,22 @@ extern "C" fn nvmf_create_snapshot_hdlr(req: *mut spdk_nvmf_request) -> i32 {
...
+) {
+    let snapshot_time = unsafe {
+        spdk_sys::nvme_cmd_cdw10_get_val(&*cmd) as u64
+            | (spdk_sys::nvme_cmd_cdw11_get_val(&*cmd) as u64) << 32
+    };
+    let snapshot_name = Lvol::format_snapshot_name(&lvol.name(), snapshot_time);
+    // Blobfs operations must be on md_thread
+    Reactors::master().send_future(async move {
+        lvol.create_snapshot_local(io, &snapshot_name).await;
+    });
```
tiagolobocastro, line 149: do we need to wait for the inner call to complete? Otherwise when we leave `create_snapshot` the snapshot might not be created yet and we don't know if it failed?

mayastor/src/bdev/nexus/nexus_fn_table.rs
jonathan-teh: When it reaches nexus_bdev::Nexus::nvme_admin, the call to spdk_bdev_nvme_admin_passthru should fail and hence fail the snapshot operation. It's not ideal but we don't support, or use, iSCSI children in k8s anyway.

mayastor/src/lvs/lvol.rs
jonathan-teh: I guess it should, I'll fix it.

mayastor/src/lvs/lvol.rs
jonathan-teh: That's handled later when io_completion_local is called with success = false.

mayastor/src/subsys/nvmf/admin_cmd.rs
jonathan-teh: Also handled with io_completion_local. It's basically simulating the effect of a child IO without actually issuing a child IO (in the usual sense) as it's issuing a vbdev_lvol operation instead.

gila (approved):

mayastor/src/lvs/lvol.rs
jonathan-teh: Fixed.

tiagolobocastro (approved):

jonathan-teh:
bors merge

PR #469, opened 2020-10-20
CAS-488: message bus timeout and retry

tiagolobocastro:
Message bus requests can go unanswered. In fact, if no receiver was
listening in when the request was sent, it will not get it altogether.

This a default timeout of 5s with a default max number of 5 retries.
To avoid complicating things for the moment, each retry it spaced out by
 1s * nr of retries and capped at 10s - further down the line we might
want to customize this a bit.

By using request_ext one can specify timeout and max_retries explicitly.

### DIFF
--- a/mbus-api/src/lib.rs
+++ b/mbus-api/src/lib.rs
@@ -17,7 +17,7 @@ pub use send::*;
 use serde::{Deserialize, Serialize};
 use smol::io;
 use snafu::Snafu;
-use std::{fmt::Debug, marker::PhantomData, str::FromStr};
+use std::{fmt::Debug, marker::PhantomData, str::FromStr, time::Duration};
 
 /// Available Message Bus channels
 #[derive(Clone, Debug)]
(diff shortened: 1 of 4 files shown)


mbus-api/src/lib.rs
```diff
@@ -214,6 +220,55 @@ pub type BusOptions = nats::Options;
...
+pub struct TimeoutOptions {
+    /// initial request message timeout
+    pub(crate) timeout: std::time::Duration,
+    /// max number of retries following the initial attempt's timeout
+    pub(crate) max_retries: Option<u32>,
+}
+
+impl TimeoutOptions {
+    pub(crate) fn default_timeout() -> Duration {
+        Duration::from_secs(6)
```
gila, line 235: 😋 I think indeed that after careful experimentation has demonstrated it's best to use 6.

mbus-api/src/mbus_nats.rs
```diff
@@ -99,8 +112,37 @@ impl Bus for NatsMessageBus {
...
+                return Err(io::ErrorKind::TimedOut.into());
+            }
+            retries += 1;
+            timeout = std::cmp::min(
+                Duration::from_secs(1) * retries,
+                Duration::from_secs(10),
+            );
+
+            log::trace!(
+                "Retrying message bus message on {} with timeout of {:?}",
```
gila, line 141: Nitpicking but how about ”timeout on {}....”, channel and add the num of retries left? 

Whatever the case bus message on {channel} sounds odd?

gila (approved):

paulyoong (approved):
lgtm :+1:

mbus-api/src/mbus_nats.rs
tiagolobocastro: yeah, sounds a bit like inception :)

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #471, opened 2020-10-22
Add a function to check if a nexus child is local

paulyoong:
A nexus child is identified as being local if its bdev driver is neither
nvme nor iscsi. Therefore, nexus children with bdev drivers of type
"lvol", "aio" or "malloc" will be identified as local.

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_child.rs
+++ b/mayastor/src/bdev/nexus/nexus_child.rs
@@ -439,4 +439,16 @@ impl NexusChild {
             .map(|j| j.stats().progress as i32)
             .unwrap_or_else(|| -1)
     }
+
+    /// Determines if a child is local to the nexus (i.e. on the same node)
+    pub fn is_local(&self) -> Option<bool> {
+        match &self.bdev {
+            Some(bdev) => {
+                // A local child is not exported over nvme or iscsi
+                let local = bdev.driver() != "nvme" && bdev.driver() != "iscsi";
+                Some(local)
+            }
+            None => None,
+        }
+    }
 }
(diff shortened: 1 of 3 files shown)


mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -439,4 +439,16 @@ impl NexusChild {
             .map(|j| j.stats().progress as i32)
             .unwrap_or_else(|| -1)
     }
+
+    /// Returns true if the child is on the same node as the nexus.
```
tiagolobocastro, line 443: Returns Some(true) or None if the bdev is not open?

mayastor/tests/nexus_child_location.rs
```diff
@@ -0,0 +1,94 @@
...
+            )
+            .await
+            .unwrap();
+
+            let nexus = nexus_lookup(NEXUS_NAME).expect("Failed to find nexus");
+            let children = &nexus.children;
+            assert_eq!(children.len(), 3);
+            assert_eq!(children[0].is_local().unwrap(), true);
+            assert_eq!(children[1].is_local().unwrap(), false);
+            assert_eq!(children[2].is_local().unwrap(), false);
```
tiagolobocastro, line 91: missing the test for when the bdev is closed?

tiagolobocastro (approved):

mayastor/src/bdev/nexus/nexus_child.rs
paulyoong: I have updated the comments - thanks

mayastor/tests/nexus_child_location.rs
paulyoong: The bdev in NexusChild never actually gets set to None, so I can't test this condition currently.

chriswldenyer (approved):
LGTM

paulyoong (commented):
bors merge

PR #473, opened 2020-10-22
mayastor: Add flag for polling Reactor incoming thread queue

### LINKED ISSUE #438: Incoming queue polled to often
While running performance benchmarks, it observed that for each loop, we poll the reactor(s) for incoming threads. The queue on which we insert new threads is atomic which is rather expensive and could be polled less frequently. 

Instead of doing `while let Ok(i) = self.incomming.pop()` every iteration it is better to set a flag "incoming" to true when we schedule a new thread. When this flag is set, we should while loop over the queue.

daniel-shuy:
Closes #438 

Add `has_incoming` flag to `Reactor` to indicate that there are incoming threads to poll. This avoids polling the incoming thread queue too frequently.

### DIFF
--- a/mayastor/src/core/reactor.rs
+++ b/mayastor/src/core/reactor.rs
@@ -103,6 +103,9 @@ pub struct Reactor {
     /// incoming threads that have been scheduled to this core but are not
     /// polled yet
     incoming: crossbeam::queue::SegQueue<Mthread>,
+    /// flag to indicate that there are incoming threads to be polled.
+    /// This is to avoid polling the incoming thread queue too frequently.
+    has_incoming: Cell<bool>,
     /// the logical core this reactor is created on
     lcore: u32,
     /// represents the state of the reactor
@@ -179,6 +182,7 @@ impl Reactors {
                     r.lcore
                 );
                 r.incoming.push(Mthread(thread));
+                r.has_incoming.set(true);
                 return true;
             }
             false
@@ -267,6 +271,7 @@ impl Reactor {
         Self {
             threads: RefCell::new(VecDeque::new()),
             incoming: crossbeam::queue::SegQueue::new(),
+            has_incoming: Cell::new(false),
             lcore: core,
             flags: Cell::new(ReactorState::Init),
             sx,
@@ -455,16 +460,19 @@ impl Reactor {
             t.poll();
         });
 
-        while let Ok(i) = self.incoming.pop() {
-            self.threads.borrow_mut().push_back(i);
+        if self.has_incoming.get() {
+            self.has_incoming.set(false);
+
+            while let Ok(i) = self.incoming.pop() {
+                self.threads.borrow_mut().push_back(i);
+            }
         }
     }
 
     /// poll the threads n times but only poll the futures queue once and look
     /// for incoming only once.
     ///
-    /// We might want to set a flag that we need to run futures and or incoming
-    /// queues
+    /// We might want to set a flag that we need to run futures
     pub fn poll_times(&self, times: u32) {
         let threads = self.threads.borrow();
         for _ in 0 .. times {
@@ -477,8 +485,12 @@ impl Reactor {
         self.run_futures();
         drop(threads);
 
-        while let Ok(i) = self.incoming.pop() {
-            self.threads.borrow_mut().push_back(i);
+        if self.has_incoming.get() {
+            self.has_incoming.set(false);
+
+            while let Ok(i) = self.incoming.pop() {
+                self.threads.borrow_mut().push_back(i);
+            }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


mayastor/src/core/reactor.rs
```diff
@@ -455,16 +460,19 @@ impl Reactor {
             t.poll();
         });
 
-        while let Ok(i) = self.incoming.pop() {
-            self.threads.borrow_mut().push_back(i);
+        if self.has_incoming.get() {
```
tiagolobocastro, line 463: I think this could be simplified into something like if self.has_incoming.replace(false) == true { ... }

tiagolobocastro (commented):
Thanks for working on this @daniel-shuy !
This is where my rust knowledge is still not up to scratch yet but I wonder if has_incoming needs to be "volatile" here (or whatever the rust construct is), otherwise it could be optimized into a register since the compiler does not "see it change" in the polling loop?

mayastor/src/core/reactor.rs
daniel-shuy: Done, thanks!

daniel-shuy:
> This is where my rust knowledge is still not up to scratch yet but I wonder if has_incoming needs to be "volatile" here (or whatever the rust construct is), otherwise it could be optimized into a register since the compiler does not "see it change" in the polling loop?

@tiagolobocastro You're right, `Cell` is not thread-safe, I switched to using `AtomicBool` instead

tiagolobocastro:
> > This is where my rust knowledge is still not up to scratch yet but I wonder if has_incoming needs to be "volatile" here (or whatever the rust construct is), otherwise it could be optimized into a register since the compiler does not "see it change" in the polling loop?
> 
> @tiagolobocastro You're right, `Cell` is not thread-safe, I switched to using `AtomicBool` instead.
> 
> `volatile` is not sufficient as it only prevents compiler optimization, but doesn't provide shared memory between threads, which means that `poll_times()` could read `has_incoming` as `false` even though it has been set to `true` in `schedule()`. `volatile` in Rust is only useful for accessing memory that can be modified outside the application (eg. memory-mapped I/O registers), is it useless for multithreading.

In this case the code is actually run by the same thread... but it's called through the `spdk_thread_create` "in C" and so the rust compiler might not have "visibility" over this which is why I suggested the volatile.

I did not raise this issue but as I understand it, what we're trying to avoid here is the cost of polling the atomic incoming queue and thus why we're trying to avoid atomics here so we just need the compiler to check the actual memory everytime (changed by the same thread) and not "cache" the flag in a register.

gila:
The original patch -- i.e. without the atomic bool was indeed correct. I would be interested in the numbers pre/post-patch. 

We can obtain this with ”casperf -b 512 -q 64 ’malloc:///disk0?size_mb=100’” I think that even with debug build it would already show an improvement

daniel-shuy:
@tiagolobocastro @gila thanks, I now understand that `poll_times()` and `schedule()` are called from the same thread. I've dropped the `AtomicBool` change.

```
In other words, your original code was what we want I think, just need to check that the code generated is correct 👍
```
Sorry I'm kinda new to Rust, how can I see the generated code?

```
The original patch -- i.e. without the atomic bool was indeed correct. I would be interested in the numbers pre/post-patch.

We can obtain this with ”casperf -b 512 -q 64 ’malloc:///disk0?size_mb=100’” I think that even with debug build it would already show an improvement
```
Unfortunately I am unable to fully compile and run the project, it fails with "version `GLIBC_2.32' not found" (with `nix-shell`).

tiagolobocastro:
IIRC @blaisedias hit similar glibc issue recently?

gila:
We recently bumped nix packages so perhaps there is some overlap. Could you exit (all of them!) nix-shell and run the garbage collector and re-enter the shell?

daniel-shuy:
> We recently bumped nix packages so perhaps there is some overlap. Could you exit (all of them!) nix-shell and run the garbage collector and re-enter the shell?

Thanks, that worked.

However, when I run `./casperf -b 512 -q 64 'malloc:///disk0?size_mb=100'`, it fails with:
`EAL: No available hugepages reported in hugepages-2048kB`

After running `echo 512 | sudo tee  /sys/kernel/mm/hugepages/hugepages-2048kB/nr_hugepages`, `casperf` fails with:
`EAL: 512 hugepages of size 2097152 reserved, but no mounted hugetlbfs found for that size`

I then tried running:
```
mkdir -p /mnt/huge
sudo mount -t hugetlbfs nodev /mnt/huge
```
and it now fails with:
`EAL: Couldn't get fd on hugepage file`

I'm not sure if this has to do with me running on WSL.

daniel-shuy:
I tried the Docker approach as well, but running `cargo`/`rustc` throws `SegmentationFault`

jkryl:
I'm closing this PR as it hasn't been active for quite some time already. If the author wishes to continue in this effort, feel free to re-open. thanks

PR #475, opened 2020-10-24
doc: update faulty links in build document

de-sh:
A few links in `doc/build.md` corrected on reading the docs to get a better understanding of the codebase

### DIFF
--- a/doc/build.md
+++ b/doc/build.md
@@ -45,7 +45,7 @@ Installation of a [nix package manager](https://nixos.org/nix/download.html) on
 other distros:
 
 ```bash
-curl https://nixos.org/nix/install | sh
+curl -L https://nixos.org/nix/install | sh
 ```
 
 We have provided a `shell.nix` file that can be used to build and compile
--- a/doc/build.md
+++ b/doc/build.md
@@ -5,9 +5,9 @@ use the nightly compiler. Nightly is required in all of the provided build possi
 
 Build options
 ==================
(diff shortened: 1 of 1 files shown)


de-sh:
I would also like to note that I am unable to build on my pc with these features turned on, either with `nix-shell` or `docker`.

PC specs: Ubuntu 20.04, 8GB i5 7th Gen

tiagolobocastro (approved):
Thanks @de-sh!
That first error related to udev, were you running within nix-shell? I'm asking because it's a build input to the nix-shell so wouldn't expect to see that error.
Not sure why the docker build fails so badly either as we used to run our build in a docker container within github actions...

de-sh:
I am unsure why, but nix shell doesn't seem to work. I have another update, let me push that as well.

jkryl (approved):

jkryl:
We have updated all nix packages on behalf of commit dfccad1179ef747aa261c3c64b9a440b1d21bbba, though that does not explain your failure. But it might not be a bad idea to exit all nix-shell's, run `nix-collect-garbage -d` and try to enter nix-shell again. That will refetch/rebuild all dependencies of mayastor again. Let us know if that helped.  Would be good to discuss the issue on slack (`kubernetes.slack.com #openebs`) if you can join.

jkryl:
bors merge

gila:
bors retry

PR #479, opened 2020-10-27
CAS-493 Separate Nexus read and write channels.

chriswldenyer:
To make the logic for controlling write-only child bdevs more straightforward.
The write-only count in the NexusChannel is removed and now we use a two separate vectors of BdevHandles,
one for read operations and one for write (or other modifying) operations.
Included new check for attempted reads when no readable channels are present.

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_bdev.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev.rs
@@ -708,7 +708,7 @@ impl Nexus {
         }
 
         let ch = NexusChannel::inner_from_channel(ch);
-        let (desc, ch) = ch.ch[ch.previous].io_tuple();
+        let (desc, ch) = ch.read_ch[ch.previous].io_tuple();
         let ret = Self::readv_impl(io, desc, ch);
         if ret != 0 {
             let bio = Bio::from(io);
@@ -719,8 +719,7 @@ impl Nexus {
 
     /// read vectored io from the underlying children.
     pub(crate) fn readv(&self, io: &Bio, channels: &mut NexusChannelInner) {
-        // we use RR to read from the children also, set that we only need
-        // to read from one child before we complete the IO to the callee.
+        // we use RR to read from the children.
         let child = channels.child_select();
 
         // if there is no buffer space for us allocated within the request
@@ -736,7 +735,7 @@ impl Nexus {
             return;
         }
 
-        let (desc, ch) = channels.ch[child].io_tuple();
+        let (desc, ch) = channels.read_ch[child].io_tuple();
 
         let ret = Self::readv_impl(io.as_ptr(), desc, ch);
 
@@ -778,7 +777,7 @@ impl Nexus {
     pub(crate) fn reset(&self, io: &Bio, channels: &NexusChannelInner) {
         // in case of resets, we want to reset all underlying children
         let results = channels
-            .ch
+            .write_ch
             .iter()
             .map(|c| unsafe {
                 let (bdev, chan) = c.io_tuple();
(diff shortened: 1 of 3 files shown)


mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -719,8 +719,12 @@ impl Nexus {
 
     /// read vectored io from the underlying children.
     pub(crate) fn readv(&self, io: &Bio, channels: &mut NexusChannelInner) {
-        // we use RR to read from the children also, set that we only need
-        // to read from one child before we complete the IO to the callee.
+        // we use RR to read from the children.
+        if channels.read_ch.len() == 0 {
```
tiagolobocastro, line 723: don't think this can ever happen, should it be an assert?

mayastor/src/bdev/nexus/nexus_channel.rs
```diff
@@ -26,8 +26,8 @@ pub(crate) struct NexusChannel {
 #[repr(C)]
 #[derive(Debug)]
 pub(crate) struct NexusChannelInner {
-    pub(crate) ch: Vec<BdevHandle>,
-    pub(crate) write_only: usize,
+    pub(crate) write_ch: Vec<BdevHandle>,
```
tiagolobocastro, line 29: write_only was a poor name (my bad) so I wonder if write_ch should just be name something else, since it's actually used for different operations and not just write. Maybe even just ch?

tiagolobocastro (approved):
:+1:

mayastor/src/bdev/nexus/nexus_channel.rs
jonathan-teh: I think `write_ch` is fine, it's sometimes used in certain contexts to mean any modifying operation.

jonathan-teh (approved):

mayastor/src/bdev/nexus/nexus_bdev.rs
chriswldenyer: Ok maybe an assert will do. But we do clear the vectors when we reconfigure and I wouldn't bet there couldn't be a window when an IO gets processed at that moment.

tiagolobocastro: that's a good point in favour of keeping this one in then though I guess we shouldn't be processing IO's while the channels are being reconfigured otherwise we'd open the doors to data corruption?

chriswldenyer: Actually I think you are right. refresh_io_channels() happens in the same thread that the IO belongs to so refresh() is effectively atomic as far as IO is concerned. I'll make it an assert.

gila (approved):
One (non PR blocking) suggestion, instead of write_ch and read_ch, what about writers and readers ?

mayastor/src/bdev/nexus/nexus_channel.rs
tiagolobocastro: that's true, although in this case it's not that we have a read and a write handle which are opened differently. The read/write here is simply to indicate that a child is not up to date so cannot be read from as it might return invalid user data.
I wonder actually, when we take snapshots, maybe that should be done only on the read_ch?

jonathan-teh: I suppose creating a snapshot should be done on `read_ch` though `nexus_bdev::Nexus::nvme_admin`, as written, applies to any NVMe Admin command. You could select the channel based on the opcode, or just leave a fixme comment behind until we get round to addressing snapshots.

chriswldenyer:
> One (non PR blocking) suggestion, instead of write_ch and read_ch, what about writers and readers ?

Ok, I have to resubmit anyway because of the clippy problem so I'll do that as well.

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

PR #482, opened 2020-10-27
Add FindVolume to node plugin service

blaisedias:
The FindVolume call can be use to
1. Identify which mayastor CSI node plugin
   has mounted a volume
2. The type of mount: filesystem or raw block

This call in now in addition to FreezeFS and UnfreezeFS,
so rename csi/src/freezefs.rs -> csi/src/nodeplugin_svc.rs
refactor some symbols to more generic names.

The mount crate does not show mounts for block devices,
use findmnt for that purpose.
Split out the utility code using findmnt to a separate mod,
and refactor it to support searching for raw block device mounts.

FreezeFS and UnfreezeFS calls now fail with FailedPrecondition
for raw block volumes instead of NotFound


csi/proto/mayastornodeplugin.proto
```diff
@@ -22,8 +22,15 @@ service MayastorNodePlugin {
   // Unfreeze the file system identified by the volume ID,
   // no check is made if the file system had been previously frozen.
   rpc UnfreezeFS (UnfreezeFSRequest) returns (UnfreezeFSReply) {}
+  // Find the volume identified by the volume ID, and return the mount type:
+  // raw block or filesystem
+  rpc FindVolume (FindVolumeRequest) returns (FindVolumeReply) {}
 }
 
+enum MountType {
+  MOUNT_TYPE_FILESYSTEM = 0; // File system mount
```
gila, line 31: would VolumeType not be more appropriate? IIRC that's what CSI calls it? In any case, we don't need this long name, MountType::FileSystem and MountType::Block would be sufficient, right?

csi/src/findmnt.rs
```diff
@@ -0,0 +1,223 @@
...
+const FSTYPE_KEY: &str = "fstype";
+
+#[derive(Debug)]
+pub struct DeviceMount {
+    pub mount_path: String,
+    pub fstype: String,
+}
+
+#[derive(Debug)]
+struct FindmntFilter<'a> {
```
gila, line 18: nit; FindMntFilter but as it is confined to this file (which is good thing) Filter would work as well +)

csi/src/findmnt.rs
```diff
@@ -0,0 +1,223 @@
...
+    value: &'a str,
+}
+
+impl PartialEq<Value> for FindmntFilter<'_> {
+    /// Special case the comparison for the source field returned
+    /// by findmnt.
+    fn eq(&self, value: &Value) -> bool {
+        if self.key == SOURCE_KEY {
+            if let Some(strvalue) = value.as_str() {
+                let devpath = convert_findmnt_devicepath(strvalue);
```
gila, line 29: We have TryFrom() for this.

csi/src/findmnt.rs
```diff
@@ -0,0 +1,223 @@
...
+///  2. An object may contain nested arrays of objects.
+///
+/// The search is deliberately generic (and hence slower) in an attempt to
+/// be more robust to future changes in findmnt.
+fn filter_findmnt(
+    json_val: &serde_json::value::Value,
+    filter: &FindmntFilter,
+    results: &mut Vec<HashMap<String, String>>,
+) {
+    if let Some(json_array) = json_val.as_array() {
```
gila, line 101: I think we should use a match here.

csi/src/findmnt.rs
```diff
@@ -0,0 +1,223 @@
...
+    match RE_UDEVPATH.captures(devpath) {
+        Some(caps) => format!("/dev{}", &caps["device"]),
+        _ => devpath.to_string(),
+    }
+}
+
+/// Convert the json map entry to a hashmap of strings
+/// The source field returned from findmnt is converted
+/// to the /dev/xxx form if required.
+fn jsonmap_to_hashmap(
```
gila, line 63: Again TryFrom() is more appropriate here. Also, I'm not sure why we need to convert here? It already is a Map<String, Value>

csi/src/findmnt.rs
```diff
@@ -0,0 +1,223 @@
...
+        })
+    }
+}
+
+/// Use the Linux utility findmnt to find the name of the device mounted at a
+/// directory or block special file, if any.
+/// mount_path is the path a device is mounted on.
+pub(crate) fn findmnt_get_devicepath(
+    mount_path: &str,
+) -> Result<Option<String>, DeviceError> {
```
gila, line 151: I think an Option here would be sufficient as apposed to Result<Option>

csi/src/findmnt.rs
```diff
@@ -0,0 +1,223 @@
...
+    let tgt_filter = FindmntFilter {
+        key: TARGET_KEY,
+        value: mount_path,
+    };
+    match findmnt(tgt_filter) {
```
gila, line 156: if you use ? here you can dispense of the Err(e) => Err(e) thing

csi/src/findmnt.rs
```diff
@@ -0,0 +1,223 @@
...
+            }
+        }
+        Err(e) => Err(e),
+    }
+}
+
+/// Use the Linux utility findmnt to find the mount paths for a block device,
+/// if any.
+/// device_path is the path to the device for example "/dev/sda1"
+pub(crate) fn findmnt_get_mountpaths(
```
gila, line 191: Since we have namespaces, why do we prefix everything with findmnt? why not get_mounts() as it would then be

findmnt::get_mounts() rather then findmnt::findmnt_get_mount_paths() ?

csi/src/findmnt.rs
blaisedias: Done.

csi/src/findmnt.rs
blaisedias: Done.

csi/proto/mayastornodeplugin.proto
blaisedias: Will change it to VolumeType
If you use MOUNT_TYPE_BLAH, then in the rust code it becomes MountType::blah, see csi/src/nodeplugin_grpc.rs find_volume.

csi/src/findmnt.rs
blaisedias: I replaced the use of json map clone because  the callers always want strings, and not having to deal with Value makes those bits of code simpler to read.

csi/proto/mayastornodeplugin.proto
blaisedias: Done.

csi/src/findmnt.rs
blaisedias: Done.

csi/src/findmnt.rs
blaisedias: Done.

csi/src/findmnt.rs
blaisedias: Refactored to not require this.

csi/src/findmnt.rs
blaisedias: Cannot do TryFrom since neither struct is defined in findmnt. I have simplified the function.

csi/src/findmnt.rs
blaisedias: If I change to Option, then we would have to either 1). translate none to an error or 2) ignore the error
So we would lose precision in error reporting.

csi/src/findmnt.rs
```diff
@@ -0,0 +1,215 @@
...
+            // If the object has arrays, then the assumption is that they are
+            // arrays of objects.
+            for (_, jsonvalue) in json_map {
+                if jsonvalue.is_array() {
+                    filter_findmnt(jsonvalue, filter, results);
+                }
+            }
+        }
+        _ => {
+            warn!("Unexpected json type");
```
tiagolobocastro, line 111: add the actual json value to the log?

csi/src/nodeplugin_svc.rs
```diff
@@ -0,0 +1,141 @@
...
+                    error: String::from_utf8(output.stderr).unwrap(),
+                })
+            };
+        } else {
+            let mountpaths = findmnt::get_mountpaths(&device_path).context(
+                InternalFailure {
+                    volid: volume_id.to_string(),
+                },
+            )?;
+            // if mount::find_mount didn't return any matches,
```
tiagolobocastro, line 81: might be worth saying what the difference between these find*mnt things is and why we need both.

csi/src/nodeplugin_svc.rs
```diff
@@ -0,0 +1,141 @@
...
+        Device::lookup(&uuid).await.context(InternalFailure {
+            volid: volume_id.to_string(),
+        })?
+    {
+        let device_path = device.devname();
+        let mountpaths =
+            findmnt::get_mountpaths(&device_path).context(InternalFailure {
+                volid: volume_id.to_string(),
+            })?;
+        debug!("mountpaths : {:?}", mountpaths);
```
tiagolobocastro, line 119: would it be useful to add the volume_id to the log?

csi/src/findmnt.rs
blaisedias: Done.

csi/src/nodeplugin_svc.rs
blaisedias: Done

csi/src/nodeplugin_svc.rs
```diff
@@ -0,0 +1,164 @@
...
+                    "{} for volume_id :{} : failed, {}",
+                    freeze_op, volume_id, errmsg
+                );
+                Err(ServiceError::FsfreezeFailed {
+                    volid: volume_id.to_string(),
+                    error: errmsg,
+                })
+            };
+        } else {
+            // mount::find_mount dees not return any matches,
```
tiagolobocastro, line 81: dees/does

tiagolobocastro (approved):
Please squash the "review fix" commits otherwise bors will merge with them.

csi/src/nodeplugin_svc.rs
blaisedias: Done.

csi/src/nodeplugin_svc.rs
blaisedias: Done.

blaisedias:
bors try

blaisedias:
bors try

csi/src/nodeplugin_svc.rs
```diff
@@ -0,0 +1,164 @@
...
+    #[snafu(display("Cannot find volume: volume ID: {}", volid))]
+    VolumeNotFound { volid: String },
+    #[snafu(display("Invalid volume ID: {}, {}", volid, source))]
+    InvalidVolumeId {
+        source: uuid::parser::ParseError,
+        volid: String,
+    },
+    #[snafu(display("fsfreeze failed: volume ID: {}, {}", volid, error))]
+    FsfreezeFailed { volid: String, error: String },
+    #[snafu(display("Internal failure: volume ID:{}, {}", volid, source))]
```
jonathan-teh, line 25: Space after the second colon.

jonathan-teh (approved):

blaisedias:
bors merge

PR #488, opened 2020-10-30
CAS-492 Serialise reconfigure

chriswldenyer:
Ensure only one reconfigure task is active per-nexus at a time.
This is achieved by adding a oneshot channel for each extra task
and getting it to wait for a completion message from the earlier task.

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_bdev.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev.rs
@@ -329,6 +329,8 @@ pub struct Nexus {
     pub nexus_target: Option<NexusTarget>,
     /// the maximum number of times to attempt to send an IO
     pub(crate) max_io_attempts: i32,
+    /// queue of channels for futures waiting to call reconfigure
+    reconfigure_queue: Vec<oneshot::Sender<()>>,
 }
 
 unsafe impl core::marker::Sync for Nexus {}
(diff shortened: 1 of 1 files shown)


mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -467,11 +491,11 @@ impl Nexus {
         NexusChannel::reconfigure(self.as_ptr(), &event);
 
         let result = r.await;
-
         info!(
             "{}: Dynamic reconfiguration event: {:?} completed {:?}",
             self.name, event, result
         );
+        self.release_reconfigure();
```
tiagolobocastro, line 498: Is there a race condition here as we clear the dr_complete_notify before releasing the reconfigure. Can another reconfigure get in between? Then it'd get stuck?

chriswldenyer: No, because if dr_complete_notify isn't set, then it won't wait on the channel either.
Also, if there is no await in between those points, I don't think another reconfigure can happen.

tiagolobocastro: Yeah true but what if more than 2 extra reconfigure come along? the first goes in waits on r.await (493); then the callback "takes" the dr_complete_notify and does s.send, but before the first r.await another 2 reconfigure come along, the first one goes through and the second awaits on the oneshot (465)... meanwhile the initial reconfigure gets to run again and releases the oneshot (465) channel and now we have 2 reconfigure running?

tiagolobocastro: I may be misunderstanding this but then again concurrency is hard :) I had a play with the async mutex and it "seems" to work, so maybe worth giving investing in that one a bit more?

Issue #468, opened 2020-10-20
Failed to stage volume: attach failed: Invalid argument (os error 22)


stoneshi-yunify:
**Describe the bug**
Was trying the mayastor and run [dbench](https://github.com/leeliu/dbench) on a mayastor nvme volume. The dbench pod can not start due to pvc mount error. The k8s event is `MountVolume.MountDevice failed for volume "pvc-ac2b83cd-3489-43fe-80c0-697694a7a72d" : rpc error: code = Internal desc = Failed to stage volume ac2b83cd-3489-43fe-80c0-697694a7a72d: attach failed: Invalid argument (os error 22)`.

```
root@testr01n01:~/stone# kubectl get pod
NAME                         READY   STATUS              RESTARTS   AGE
dbench-fbrnv                 0/1     Completed           0          15h
dbench-mayastor-nvmf-xdjx5   0/1     ContainerCreating   0          23s
root@testr01n01:~/stone# kubectl describe pod dbench-mayastor-nvmf-xdjx5
Name:           dbench-mayastor-nvmf-xdjx5
Namespace:      default
Priority:       0
Node:           testr01n01/192.168.101.174
Start Time:     Tue, 20 Oct 2020 09:47:28 +0800
Labels:         controller-uid=6d363037-5b85-4bef-b046-0d1457708d88
                job-name=dbench-mayastor-nvmf
Annotations:    <none>
Status:         Pending
IP:
IPs:            <none>
Controlled By:  Job/dbench-mayastor-nvmf
Containers:
  dbench:
    Container ID:
    Image:          fbuchmeier/dbench:latest
    Image ID:
    Port:           <none>
    Host Port:      <none>
    State:          Waiting
      Reason:       ContainerCreating
    Ready:          False
    Restart Count:  0
    Environment:
      DBENCH_MOUNTPOINT:  /data
    Mounts:
      /data from dbench-pv (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-gjxb5 (ro)
Conditions:
  Type              Status
  Initialized       True
  Ready             False
  ContainersReady   False
  PodScheduled      True
Volumes:
  dbench-pv:
    Type:       PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:  dbench-pv-claim-mayastor-nvmf
    ReadOnly:   false
  default-token-gjxb5:
    Type:        Secret (a volume populated by a Secret)
[...]

PR #489, opened 2020-10-30
nvmeadm: use the snafu crate to generate improved errors messages CAS…

blaisedias:
…-338

Replace use of the failure crate with snafu.
Fixes remaining part of https://github.com/openebs/Mayastor/issues/468

Add more context to errors generate, for example on IO errors,
add filename being accessed.

Change how connection in progress is handled by returning a specific error
so that the CSI driver can check explicitly, and use error kind for detection
instead of the magic number 114.


gila (changes requested):
I seem to me this commit contains things which are not intended for merging?

PR #496, opened 2020-11-05
tests: Add scripts to bringup a test environment

tjoshum:
I think this is now at a good point for other people to take a look. example-parallel.sh and example-simple.sh in mayastor-test/e2e show how to run the three stages (bring up the cluster, build/publish the images, run the tests).
I still want to move out the mayastor install to a library (rather than its current place of just messing up the yaml install test), and I want to get everything to run in our nix environment.
We should also think about how we want to manage our yaml files, I've added templates for deployments using our images (mayastor, csi and moac) which differ only from the existing yamls in deploy/ in that they don't hard code the docker image version as "latest".


blaisedias (approved):
LGTM

mayastor-test/e2e/example-simple.sh
```diff
@@ -0,0 +1,15 @@
+#! /bin/bash
+
+set -euxo pipefail
+cd "$(dirname ${BASH_SOURCE[0]})"
+
+pushd mayastor-test/e2e/setup
```
chriswldenyer, line 6: I think the directory-handling is wrong here. Because of the cd on line 4

scripts/release.sh
```diff
@@ -18,6 +19,22 @@ get_tag() {
...
+  -d, --dry-run                 Output actions that would be taken, but don't run them.
+  -h, --help                    Display this text.
+  --private-registry <address>  Push the built images to the provided registry.
+  --skip-build                  Don't perform nix-build.
+  --skip-publish-to-dockerhub   Don't publish to Dockerhub.
+
+Examples:
+  $(basename $0) --private-registry 127.0.0.1:5000
+EOF
+}
```
jkryl, line 36: thanks for contributing usage message and arg parser - nice improvement! I would change the parameters to make them even more useful in general and to avoid code duplication.

`--private-registry` can become just `--registry` (if none is specified then dockerhub registry is assumed). Then you can leave the code as it is and just add registry prefix to built images when pushing to a particular registry. In docker speech, the target registry is nothing more than a prefix in the image name so we can keep that idea here.

`--skip-publish-to-dockerhub` can then be removed entirely or you could rename it to `--skip-publish` to skip the entire step of publishing images. But it is not needed now so perhaps better to remove it.

We can introduce a new `--tag` option to explicitly set a tag of built images and override the default logic when the tag is set based on the source branch. For ci that can be set to `ci`. For other use cases to something else.

mayastor-test/e2e/example-simple.sh
```diff
@@ -0,0 +1,15 @@
+#! /bin/bash
```
jkryl, line 1: just a nit. using env with bash would be more portable (applies to all bash scripts in this PR).

mayastor-test/e2e/setup/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Pre-requisites
+
+1. Install libvirt plugin
+```
+# From https://github.com/vagrant-libvirt/vagrant-libvirt#installation
+#apt-get build-dep vagrant ruby-libvirt
+sudo apt-get install -y qemu libvirt-daemon-system libvirt-clients ebtables dnsmasq-base ruby-libvirt
+sudo apt-get install -y libxslt-dev libxml2-dev libvirt-dev zlib1g-dev ruby-dev
+vagrant plugin install vagrant-libvirt
```
jkryl, line 14: would be cool to have a nix-shell file in e2e directory and everyone who would run it would get all needed dependencies for running the e2e tests.

jkryl (commented):
Thanks. Looks great! I have left a couple of comments and I have one more general comment about yaml template files. Ideally we would like to use yamls in deploy directory to deploy mayastor for two reasons:

1. That way we can verify that they are correct on behalf of our ci/cd pipeline.
2. We don't have to duplicate them here which is error-prone when making the changes to yaml files in the future.

That said I'm not sure if it is possible. If it's only about changing the image names, then I suppose we could use sed to do that until we have something better in place (i.e. kustomize). What do you think?

mayastor-test/e2e/example-simple.sh
```diff
@@ -0,0 +1,15 @@
...
+
+set -euxo pipefail
+cd "$(dirname ${BASH_SOURCE[0]})"
+
+pushd setup
+  ./bringup-cluster.sh
+popd
+
+./scripts/release.sh --private-registry "172.18.8.101:30291" --skip-publish-to-dockerhub
```
blaisedias, line 10: Should be `../../scripts` not `./scripts`

scripts/release.sh
tjoshum: I tried altering the layout of the arguments as you specified, but I think it makes the code (and behaviour) more confusing than just having arguments to skip individual stages. Eg It's clear if you're just running locally you don't want to go near dockerhub so just run with --skip-push-to-dockerhub.

mayastor-test/e2e/setup/README.md
tjoshum: I'll look at it outside of this PR

scripts/release.sh
jkryl: Correct me if I'm wrong but for the purpose of producing CI images the way how it is used would change from:
```
release.sh --private-registry=x.y.z.t --skip-publish-to-dockerhub
```
to
```
release.sh --registry=x.y.z.t --tag=ci
```
I don't think that the later is more confusing in any way, though opinions on that might differ. Also with the original args there are combinations that don't make sense and there are combinations that would be valid use cases but are not possible. Examples of each follow.

This is certainly not what you want to do if you  publish images to two registries (with the new arguments that is impossible to do):
```
release.sh --private-registry=x.y.z.t
```

I want to build the images with the tag "watcher" and publish them to dockerhub registry so that someone else can test them too:
```
release.sh --tag=watcher
```

So it is basically turning the tool from CI specific tool to a swiss-knife. Though if you don't agree, we can leave that for later. I don't want to delay the PR if this is the last thing holding it.

jkryl (approved):
I'm leaving my remaining comments to your consideration. I don't want to block this PR as other people are waiting for it. Please, feel free to merge when you feel that changes are ready for it.

tjoshum:
bors merge

PR #498, opened 2020-11-06
test: need ability to (easily) drive IO to a nexus

gila:
This change adds the ability to create workloads within a given mayastor
instance. This is for testing -- not just during mayastor development but
also for e2e tests where we can invoke job creations form within k8s in
the near future.

An example is given `io_job.rs` where we create 2 nvmf targets through
the compose framework, and then use `struct MayastorTest` to do
some IO to it.

### DIFF
--- a/mayastor/src/bdev/dev/nvmf.rs
+++ b/mayastor/src/bdev/dev/nvmf.rs
@@ -9,6 +9,7 @@ use std::{
 use async_trait::async_trait;
 use futures::channel::oneshot;
 use snafu::ResultExt;
+use tracing::instrument;
 use url::Url;
 
 use spdk_sys::{
@@ -137,6 +138,7 @@ impl CreateDestroy for Nvmf {
     type Error = NexusBdevError;
 
     /// Create an NVMF bdev
+    #[instrument(err)]
     async fn create(&self) -> Result<String, Self::Error> {
         if Bdev::lookup_by_name(&self.get_name()).is_some() {
             return Err(NexusBdevError::BdevExists {
--- a/mayastor/src/core/io_driver.rs
+++ b/mayastor/src/core/io_driver.rs
@@ -1,0 +1,403 @@
+//! helper routines to drive IO to the nexus for testing purposes
+use futures::channel::oneshot;
+use rand::Rng;
+use std::{ptr::NonNull, sync::Mutex};
+
+use spdk_sys::{
+    spdk_bdev_free_io,
+    spdk_bdev_read,
+    spdk_bdev_reset,
+    spdk_bdev_write,
+};
+
+use crate::{
+    core::{Bdev, Cores, Descriptor, DmaBuf, IoChannel, Mthread},
+    ffihelper::pair,
+    nexus_uri::bdev_create,
+};
+
+#[derive(Debug)]
+pub enum IoType {
+    /// perform random read operations
+    READ,
+    /// perform random write operations
+    WRITE,
+}
+
+impl Default for IoType {
+    fn default() -> Self {
+        Self::READ
+    }
+}
+
+#[derive(Debug)]
+struct Io {
+    /// buffer we read/write from/to
+    buf: DmaBuf,
+    /// type of IO we are supposed to issue
+    iot: IoType,
+    /// current offset where we are reading or writing
+    offset: u64,
+    /// pointer to our the job we belong too
+    job: NonNull<Job>,
+}
+
+impl Io {
+    /// start submitting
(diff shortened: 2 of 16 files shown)


mayastor/src/core/io_driver.rs
```diff
@@ -0,0 +1,403 @@
...
+    }
+}
+
+#[derive(Debug)]
+struct Io {
+    /// buffer we read/write from/to
+    buf: DmaBuf,
+    /// type of IO we are supposed to issue
+    iot: IoType,
+    /// current offset where we are reading from
```
paulyoong, line 39: The comment says "reading" specifically, it could be writing also.

mayastor/tests/common/compose.rs
```diff
@@ -566,10 +566,31 @@ impl ComposeTest {
...
+                    .output()
+                    .unwrap();
+                std::process::Command::new("docker")
+                    .args(&["rm", c])
+                    .output()
+                    .unwrap();
+            });
+
+            std::process::Command::new("docker")
+                .args(&["network", "rm", &self.name])
```
paulyoong, line 584: :+1: thanks, this should solve the issue I was seeing

mayastor/tests/io_job.rs
```diff
@@ -0,0 +1,130 @@
...
+        ..Default::default()
+    });
+
+    DOCKER_COMPOSE.set(compose).unwrap();
+    let ms = MAYASTOR.get_or_init(|| mayastor_test);
+
+    create_work(queue.clone()).await;
+    tokio::time::delay_for(Duration::from_secs(1)).await;
+    tokio::time::delay_for(Duration::from_secs(2)).await;
+    // create and share a bdev on each container
```
paulyoong, line 121: Is this comment in the wrong place?

paulyoong (approved):

mayastor/tests/io_job.rs
chriswldenyer: The two calls to delay_for() is a typo I suppose?

mayastor/tests/io_job.rs
```diff
@@ -0,0 +1,130 @@
+use std::sync::Arc;
+
+use once_cell::sync::OnceCell;
+use tokio::time::Duration;
+
+use common::{Builder as ComposeBuilder, ComposeTest, MayastorTest};
```
chriswldenyer, line 6: Maybe the Builder in compose should be called ComposeBuilder ?

mayastor/tests/common/mod.rs
```diff
@@ -133,7 +134,7 @@ pub fn mayastor_test_init() {
                 panic!("binary: {} not present in path", binary);
             }
         });
-    logger::init("DEBUG");
+    logger::init("INFO");
```
tiagolobocastro, line 137: was this intended?

mayastor/tests/common/compose.rs
gila: It should already during Drop constructor .. but sometimes we need to control the order so I added a "manual" drop there as well... called down() +)

mayastor/src/bdev/dev/nvmf.rs
```diff
@@ -9,6 +9,7 @@ use std::{
 use async_trait::async_trait;
 use futures::channel::oneshot;
 use snafu::ResultExt;
+use tracing::instrument;
```
chriswldenyer, line 12: Maybe mention the tracing change in the pull request comment.

mayastor/tests/io_job.rs
gila: fixed

mayastor/tests/io_job.rs
gila: Right yeah, the name clash let me fix that. :+1:

mayastor/src/core/io_driver.rs
```diff
@@ -0,0 +1,403 @@
...
+        let ioq: &mut Io = unsafe { &mut *arg.cast() };
+        let job = unsafe { ioq.job.as_mut() };
+
+        if !success {
+            error!("{}: {:#?}", job.thread.as_ref().unwrap().name(), bdev_io);
+        }
+
+        assert_eq!(Cores::current(), job.core);
+        job.n_io += 1;
+        job.n_inflight -= 1;
```
chriswldenyer, line 196: Should probably assert that `job.n_inflight` is not zero at this point

chriswldenyer (approved):

cjones1024 (approved):

PR #505, opened 2020-11-11
CAS-512 Refactor mount_fs testcases to use NVMe protocol instead of NBD

mtzaurus:
Existing mount tests are refactored to not use Mthreads and to use NVMF protocol for testing nexus connectivity instead of NBD.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1742,6 +1742,7 @@ dependencies = [
  "mbus_api",
  "nats",
  "nix 0.16.1",
+ "nvmeadm",
  "once_cell",
  "pin-utils",
  "proc-mounts",
--- a/mayastor/Cargo.toml
+++ b/mayastor/Cargo.toml
@@ -81,6 +81,7 @@ dns-lookup = "1.0.4"
 ipnetwork = "0.17.0"
 bollard = "0.8.0"
 mbus_api = { path = "../mbus-api" }
+nvmeadm = {path = "../nvmeadm", version = "0.1.0"}
 
 [dependencies.rpc]
 path = "../rpc"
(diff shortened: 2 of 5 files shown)


gila (approved):
Much much cleaner @mtzaurus :shipit:

gila:
I suspect one CI failure will happen which has something to do with the cargosha256... lets see what bors will do here.

gila:
bors try

tiagolobocastro (approved):
Do we need a "catch and unwind" on each test or a test fn which runs last to try and disconnect any leftover connected nvmf devices (eg if a test fn crashes and doesn't get to disconnect the nvmf device)?

gila:
> Do we need a "catch and unwind" on each test or a test fn which runs last to try and disconnect any leftover connected nvmf devices (eg if a test fn crashes and doesn't get to disconnect the nvmf device)?

We panic if one of the tests fails, and we can not unwind any more. Perhaps an idea is to, in the future, add a disconnect --all in the jenkins file?

mtzaurus:
> > Do we need a "catch and unwind" on each test or a test fn which runs last to try and disconnect any leftover connected nvmf devices (eg if a test fn crashes and doesn't get to disconnect the nvmf device)?
> 
> We panic if one of the tests fails, and we can not unwind any more. Perhaps an idea is to, in the future, add a disconnect --all in the jenkins file?

@gila Good point. Do

tiagolobocastro:
bors cancel

tiagolobocastro:
@mtzaurus there's a bug which is failing CI, should be fixed soon with 503.

jkryl (approved):

PR #520, opened 2020-11-18
env: add -l option to specify cores

gila:
The -l option serves the same purpose as the -m option however it accepts
a list of cores with ranges i.e 1,2,10-15.

Also two default values have been updated while here as the number of
shared buffers was to low to test this change and some notes to the YAML
file has been added.

### DIFF
--- a/deploy/mayastor-daemonset.yaml
+++ b/deploy/mayastor-daemonset.yaml
@@ -25,10 +25,6 @@ spec:
       nodeSelector:
         openebs.io/engine: mayastor
         kubernetes.io/arch: amd64
-      # NOTE: Each container must have mem/cpu limits defined in order to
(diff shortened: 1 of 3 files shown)


jkryl (approved):
@GlennBullingham I think we will need to update our docs. At least to state that mayastor will take 2 cores from the system by default now and potentially instructions on how to set it back to just one core for people who are just trying out the mayastor (on a poor VM).

gila:
> @GlennBullingham I think we will need to update our docs. At least to state that mayastor will take 2 cores from the system by default now and potentially instructions on how to set it back to just one core for people who are just trying out the mayastor (on a poor VM).

We already mention we the two CPU requirement IIRC. I hope, that with CPU: "2" it will simply not schedule there either. :crossed_fingers:

mtzaurus (approved):

PR #524, opened 2020-11-20
test: add IO failure test to validate admin queue timeouts

gila:


gila:
bors try

mayastor/src/core/descriptor.rs
```diff
@@ -192,12 +192,16 @@ impl Drop for Descriptor {
...
-        )
+        if self.0 != std::ptr::null_mut() {
+            write!(
+                f,
+                "Descriptor {:p} for bdev: {}",
+                self.as_ptr(),
+                self.get_bdev().name()
+            )
+        } else {
+            write!(f, "not alloacted")
```
paulyoong, line 203: Typo "allocated". Also should we say "Descriptor not allocated"?

mayastor/tests/bdev_test.rs
```diff
@@ -0,0 +1,337 @@
...
+//! possible.
+use std::{
+    sync::{atomic::Ordering, Arc},
+    time::Duration,
+};
+
+use once_cell::sync::OnceCell;
+
+const NEXUS_UUID: &str = "00000000-0000-0000-0000-000000000001";
+//const NEXUS_NAME: &str = "nexus-00000000-0000-0000-0000-000000000001";
```
paulyoong, line 45: Can we remove this line?

paulyoong (approved):

gila:
bors retry

mayastor/src/core/descriptor.rs
gila: hmm we should not print at all actually. if self.0 != null_mut() { ... } else { Ok()) } would be sufficient IMHO.

mayastor/tests/bdev_test.rs
gila: Sure.

chriswldenyer (approved):
LGTM
We should remove the error-store unit tests soon as they are not likely to be needed (in that form) again.

mayastor/tests/bdev_test.rs
```diff
@@ -0,0 +1,337 @@
...
+        .unwrap();
+    dbg!(h.bdev.list(Null {}).await.unwrap());
+}
+
+#[tokio::test]
+async fn nvmf_bdev_test() {
+    let queue = Arc::new(JobQueue::new());
+
+    Config::get_or_init(|| Config {
+        nvme_bdev_opts: NvmeBdevOpts {
```
tiagolobocastro, line 251: why are we using a non default config? so we won't spend as long in the test waiting for the timeout?

tiagolobocastro (approved):

gila:
> LGTM
> We should remove the error-store unit tests soon as they are not likely to be needed (in that form) again.

done

mayastor/src/bdev/nexus/nexus_child_error_store.rs
```diff
@@ -279,55 +279,57 @@ impl Nexus {
...
-                                error!(
-                                    "Failed to fault the child {}",
-                                    child_name,
-                                );
+                                let child_name = child.name.clone();
+                                info!("Faulting child {}", child_name);
+                                if nexus
+                                    .fault_child(&child_name, Reason::IoError)
+                                    .await
+                                    .is_err()
```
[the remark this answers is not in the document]
gila: That code is not used right now, I only added an if condition, we need to rework these parts anyway.

mayastor/tests/bdev_test.rs
gila: Yes, as im working (and others) on these test it's easy to change things here rather than in opts.rs as it requires a recompile of mayastor. this requires just a recompile of the test itself.

gila:
bors retry

tiagolobocastro:
bors try

PR #525, opened 2020-11-20
Pin io_driver targets to cores and docker cleanup wrapper for cargo-test

bobek:
I've packed two changes into one PR as they are quite small a loosely related: 

1. pins `io_driver` test target containers to improve likelihood of succeeding as suggested by @gila 
2. traps signals in the wrapper `cargo-test.sh` to allow for clean up. This is related to CAS-524, if we want to make it nice and driven by tests themselves.

~Cleanup is currently a bit brutal as it goes after all containers.~ I've done that as the other alternative would be listing names which is not very future-proof. Also our CI/CD runs on dedicated machines. Possible changes:
- explicitly list names used throughout the tests
- prefix all names in tests with a selector (e.g. `mayadata-`) and filter for it in the loop at handler
- prefix all names in tests with a selector passed as an environment variable to allow non conflicting names for parallel test runs + being able to select out containers related to this run in the handler.
- based on @tiagolobocastro 's suggestion, to use labels and mark our resources. Name of the label is `io.mayastor.test.name` and it holds value of the test which created the resource. This still means, that we can run only one mayastor on the machine, but will not kill other containers / networks.

### DIFF
--- a/mayastor/tests/io_job.rs
+++ b/mayastor/tests/io_job.rs
@@ -10,7 +10,7 @@ use mayastor::{
 use rpc::mayastor::{BdevShareRequest, BdevUri};
 
 pub mod common;
-use common::compose::{self, ComposeTest, MayastorTest};
+use common::compose::{self, Binary, ComposeTest, MayastorTest};
 use mayastor::core::io_driver::JobQueue;
 
 static DOCKER_COMPOSE: OnceCell<ComposeTest> = OnceCell::new();
@@ -137,11 +137,19 @@ async fn io_driver() {
     let queue = Arc::new(JobQueue::new());
 
     // create the docker containers
+    // we are pinning them to 3rd and 4th core spectively to improve stability
+    // of the test. Be aware that default docker container cpuset is 0-3!
     let compose = compose::Builder::new()
         .name("cargo-test")
         .network("10.1.0.0/16")
-        .add_container("nvmf-target1")
-        .add_container("nvmf-target2")
+        .add_container_bin(
+            "nvmf-target1",
+            Binary::from_dbg("mayastor").with_args(vec!["-l", "2"]),
+        )
+        .add_container_bin(
+            "nvmf-target2",
+            Binary::from_dbg("mayastor").with_args(vec!["-l", "3"]),
+        )
         .with_clean(true)
         .build()
         .await
--- a/scripts/cargo-test.sh
+++ b/scripts/cargo-test.sh
@@ -1,9 +1,13 @@
 #!/usr/bin/env bash
 
 cleanup_handler() {
-  for c in $(docker ps -a --format '{{.ID}}') ; do
+  for c in $(docker ps -a --filter "label=io.mayastor.test.name" --format '{{.ID}}') ; do
     docker kill "$c" || true
     docker rm "$c" || true
+  done
+
+  for n in $(docker network ls --filter "label=io.mayastor.test.name" --format '{{.ID}}') ; do
+    docker network rm "$n" || true
   done
 }
 
--- a/scripts/cargo-test.sh
+++ b/scripts/cargo-test.sh
@@ -1,4 +1,14 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
+
+cleanup_handler() {
+  for c in $(docker ps -a --format '{{.ID}}') ; do
+    docker kill "$c" || true
+    docker rm "$c" || true
+  done
+}
+
+trap cleanup_handler ERR INT QUIT TERM HUP
+
 set -euxo pipefail
 export PATH=$PATH:${HOME}/.cargo/bin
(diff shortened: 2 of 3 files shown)


arne-rusek:
code LGTM, not 100% sure about the blast radius of docker kill/rm on our jenkins. Please allow for another reviewer to give it a tick.

arne-rusek (approved):

mayastor/tests/io_job.rs
```diff
@@ -137,11 +137,19 @@ async fn io_driver() {
...
 
     // create the docker containers
+    // we are pinning them to 3rd and 4th core spectively to improve stability
+    // of the test. Be aware that default docker container cpuset is 0-3!
     let compose = compose::Builder::new()
         .name("cargo-test")
         .network("10.1.0.0/16")
-        .add_container("nvmf-target1")
-        .add_container("nvmf-target2")
+        .add_container_bin(
```
gila, line 145: sweet! that looks great @bobek thank you!

scripts/cargo-test.sh
```diff
@@ -1,4 +1,14 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
+
+cleanup_handler() {
+  for c in $(docker ps -a --format '{{.ID}}') ; do
+    docker kill "$c" || true
```
tiagolobocastro, line 5: you might be able to use the label: io.mayastor.test.$name to kill only our containers. also do we need to remove the network too?

scripts/cargo-test.sh
```diff
@@ -1,4 +1,14 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
+
+cleanup_handler() {
```
gila, line 3: The composer now also has an option .with_prune() that when enabled, will clean up any existing containers before starting a new test. The upside of that is that if there is an issue, we can docker logs it. With this change, we kill the container (and thus lose the logs). Would that help your situation?

bobek: I was thinking about the logs, but as they are captured by the test, I hoped that we can forget about them.

> The composer now also has an option .with_prune() that when enabled, will clean up any existing containers before starting a new test.

I dunno, it seems that it is only a marginal improvement as a consecutive run will prune containers anyway (and we will lose logs). I would say that it will be easier from the developers' perspective as she can just re-run in Jenkins. I think that really robust one will be the 3rd option from the description:

> prefix all names in tests with a selector passed as an environment variable to allow non conflicting names for parallel test runs + being able to select out containers related to this run in the handler.

Which I can actually try to implement, if you see any value in it.

scripts/cargo-test.sh
bobek: Nice, I've overlooked that one (network is also labeled):

```json
            "Labels": {
                "io.mayastor.test.cargo-test": "true"
            }
```

Network is not mission critical as it will not break the test. But yes, it would be nicer to clean it up as well.

bobek: I have so far discovered `cargo-test rest child_location_test lvs-pool-grpc mayastor_master composer node rest_backend` as potential values for the `$name`. Unfortunately docker doesn't have a partial/regex match support on labels. So either `jq` or list or changing the semantics:

```
            "Labels": {
                "io.mayastor.test": "cargo-test"
            }
```

Thus `docker ps -a --filter "label=io.mayastor.test"` would return all containers with label set to any value.

scripts/cargo-test.sh
tiagolobocastro: hmm might not necessarily loose the logs in either case, we can access them via journalctl -u docker?

scripts/cargo-test.sh
bobek: @tiagolobocastro can you please take a look at b96eb81? I've replaced `label` with `label_prefix` and introduced a new "static" label `$label_prefix.name` with value of `name`. This will look like 

```
        "Labels": {
            "io.mayastor.test.name": "cargo-test"
        }
```

scripts/cargo-test.sh
bobek: > hmm might not necessarily loose the logs in either case, we can access them via journalctl -u docker?

Yes, with little bit of configuration, thanks. It gives you only `dockerd` logs by default. But one can configure [journald logging driver](https://docs.docker.com/config/containers/logging/journald/) quite easily if we feel the need for it.

gila (approved):

tiagolobocastro (approved):

tiagolobocastro:
bors merge

bobek:
borse merge

bobek:
bors merge

jkryl (approved):

bobek:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #526, opened 2020-11-20
Fix rebuild bug and add new subcommand

tiagolobocastro:
Check if there are any active tasks when completing the rebuild
Add new stats subcommand to retrieve the rebuild stats

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_bdev_rebuild.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev_rebuild.rs
@@ -1,7 +1,11 @@
 use futures::channel::oneshot::Receiver;
 use snafu::ResultExt;
 
-use rpc::mayastor::{RebuildProgressReply, RebuildStateReply};
+use rpc::mayastor::{
+    RebuildProgressReply,
+    RebuildStateReply,
+    RebuildStatsReply,
+};
 
 use crate::{
     bdev::{
@@ -21,7 +25,13 @@ use crate::{
         VerboseError,
     },
     core::Reactors,
-    rebuild::{ClientOperations, RebuildError, RebuildJob, RebuildState},
+    rebuild::{
+        ClientOperations,
+        RebuildError,
+        RebuildJob,
+        RebuildState,
+        RebuildStats,
+    },
 };
 
 impl Nexus {
@@ -155,6 +165,15 @@ impl Nexus {
         Ok(RebuildStateReply {
             state: rj.state().to_string(),
         })
+    }
+
+    /// Return the stats of a rebuild job
+    pub async fn get_rebuild_stats(
+        &mut self,
+        name: &str,
+    ) -> Result<RebuildStatsReply, Error> {
+        let rj = self.get_rebuild_job(name)?;
+        Ok(rj.stats().into())
     }
 
     /// Returns the rebuild progress of child target `name`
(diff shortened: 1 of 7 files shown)


tiagolobocastro:
bors try

mayastor/src/rebuild/rebuild_impl.rs
```diff
@@ -597,11 +599,21 @@ impl RebuildJob {
...
-        while self.await_one_task().await.is_some() {
-            if self.task_pool.active == 0 {
-                break;
+        debug!(
+            "Awaiting all active tasks({}) for rebuild {}",
+            self.task_pool.active, self.destination
+        );
+        if self.task_pool.active > 0 {
+            while self.await_one_task().await.is_some() {
+                if self.task_pool.active == 0 {
```
jkryl, line 608: just a nit. The loop could be more compact:
```
while self.task_pool.active > 0 {
   self.await_one_task().await...
}
```

jkryl (approved):
looks good! We used to have a test for each CLI command. Those commands are using fake gRPC server so they don't test the real thing, but on the other hand it might be better to have at least some test compared to having none.

mayastor/src/rebuild/rebuild_impl.rs
tiagolobocastro: Thanks done, also added an error condition which although unlikely as we never close the channels.

tiagolobocastro:
> looks good! We used to have a test for each CLI command. Those commands are using fake gRPC server so they don't test the real thing, but on the other hand it might be better to have at least some test compared to having none.

Good point, I'll try to add a test, do you know where the existing ones are?

mayastor/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -324,3 +343,17 @@ impl Nexus {
         }
     }
 }
+
+impl From<RebuildStats> for RebuildStatsReply {
+    fn from(stats: RebuildStats) -> Self {
+        RebuildStatsReply {
+            blocks_total: stats.block_size,
```
paulyoong, line 350: I don't think you meant stats.block_size for the total here.

mayastor/src/bin/cli/rebuild_cli.rs
```diff
@@ -99,6 +100,21 @@ pub fn subcommands<'a, 'b>() -> App<'a, 'b> {
...
+            Arg::with_name("uuid")
+                .required(true)
+                .index(1)
+                .help("uuid of the nexus"),
+        )
+        .arg(
+            Arg::with_name("uri")
+                .required(true)
+                .index(2)
+                .help("uri of child to get the rebuild state from"),
```
paulyoong, line 115: "state" should be "stats"

rpc/proto/mayastor.proto
```diff
@@ -292,6 +293,21 @@ message RebuildStateReply {
...
 
+message RebuildStatsRequest {
+  string uuid = 1;  // uuid of the nexus
+  string uri = 2;   // uri of the destination child
+}
+
+message RebuildStatsReply {
+  uint64 blocks_total = 1; // total number of blocks to recover
+  uint64 blocks_recovered = 2; // number of blocks recovered
+  uint64 progress = 3; // rebuild progress in % (0-100)
```
paulyoong, line 304: nit - do you need "(0-100)" on the end? It's a % so it's implied

tiagolobocastro: Done, also fixed the same comment from rebuild_api

paulyoong (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #529, opened 2020-11-23
CAS-511: Test nexuses with common child

jonathan-teh:
New test with 2 identical Nexuses on 2 nodes with the same replica on
one node as their child. This requires creating the nexus with a local
replica on one node first, then sharing the replica over nvmf before
creating the nexus on the other node with the remote replica as its
child.

Use the kernel initiator to connect to both nexuses and check that
both are disconnected when disconnecting by NQN.

On Jenkins, clean up all nvmf connections after cargo tests in case
this test fails to clean up. While I'm here, remove the temporary
workaround to clean up state spdk_iscsi_conns shm files as SPDK 20.10
no longer creates such files.

### DIFF
--- a/mayastor/tests/nexus_multipath.rs
+++ b/mayastor/tests/nexus_multipath.rs
@@ -1,0 +1,173 @@
+//! Multipath NVMf tests
+//! Create the same nexus on both nodes with a replica on 1 node their child.
+use mayastor::{
+    bdev::{nexus_create, nexus_lookup},
+    core::MayastorCliArgs,
+};
+use rpc::mayastor::{
+    CreateNexusRequest,
+    CreatePoolRequest,
+    CreateReplicaRequest,
+    PublishNexusRequest,
+    ShareProtocolNexus,
+    ShareReplicaRequest,
+};
+use std::process::Command;
+
+pub mod common;
+use common::{compose::Builder, MayastorTest};
+
+static POOL_NAME: &str = "tpool";
+static UUID: &str = "cdc2a7db-3ac3-403a-af80-7fadc1581c47";
+static HOSTNQN: &str = "nqn.2019-05.io.openebs";
+
+#[tokio::test]
+async fn nexus_multipath() {
+    // create a new composeTest
+    let test = Builder::new()
+        .name("nexus_shared_replica_test")
+        .network("10.1.0.0/16")
+        .add_container("ms1")
+        .with_clean(true)
+        .build()
+        .await
+        .unwrap();
+
+    let mut hdls = test.grpc_handles().await.unwrap();
+
+    // create a pool on remote node
+    hdls[0]
+        .mayastor
+        .create_pool(CreatePoolRequest {
+            name: POOL_NAME.to_string(),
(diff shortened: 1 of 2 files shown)


jonathan-teh:
bors try

mayastor/tests/nexus_multipath.rs
```diff
@@ -0,0 +1,173 @@
...
+        .args(&["disconnect"])
+        .args(&["-n", &nqn])
+        .output()
+        .unwrap();
+    assert!(
+        output_dis.status.success(),
+        "failed to disconnect from nexuses, {}",
+        output_dis.status
+    );
+    let s = String::from_utf8(output_dis.stdout).unwrap();
```
gila, line 163: what not use JSON output?

jonathan-teh: There's no such option.

gila: bummer :(

mayastor/tests/nexus_multipath.rs
```diff
@@ -0,0 +1,173 @@
...
+    let v: Vec<&str> = s.split(' ').collect();
+    tracing::info!("nvme disonnected: {:?}", v);
```
tiagolobocastro, line 165: disconnected?

jonathan-teh: Doh, fixed.

mayastor/tests/nexus_multipath.rs
```diff
@@ -0,0 +1,173 @@
...
+        .args(&["-n", &nqn])
+        .status()
+        .unwrap();
+    assert!(
+        status.success(),
+        "failed to connect to local nexus, {}",
+        status
+    );
+
+    // The first attempt often fails with "Duplicate cntlid x with y" error from
```
tiagolobocastro, line 131: is this a kernel bug then?

jonathan-teh: It's unclear. There are a few mentions of this error message in the wild but they involve users trying to use 2 NVMe SSDs of the same model and a firmware bug means they both present the same subnqn. In this case, having identical subnqns is necessary for ANA to work. It doesn't always happen, I just tried it manually on the CLI and the second connect worked just fine. I did try adding a delay in the cargo test and steadily increased it to 5s but that didn't seem to fix it.

tiagolobocastro (approved):

mayastor/tests/nexus_multipath.rs
tiagolobocastro: yeah, seems like it's confused somehow, guess this means our CSI plugin will have to do some kind of retry when connecting as well.

gila (approved):
We don't seem to actually test the IO path here, is that a follow-up PR?

jonathan-teh:
> We don't seem to actually test the IO path here, is that a follow-up PR?

The plan was to leave IO to another longer test, though I could try a small amount of IO here (have to find the block device it connected to first).

jonathan-teh:
bors merge

PR #532, opened 2020-11-23
CAS-514: Make nexus children wait for bdev removal on close

paulyoong:
A channel is used to allow a nexus child to wait on the removal of the
underlying bdev before completing the close call.

The previous assumption had been that a call to destroy() would only
return when the bdev had actually been removed. However, in some cases,
such as for NVMf bdevs, this was not the case.

When closing a local (loopback) child, the underlying bdev must not be
destroyed. Doing so would result in the lvol and data being deleted.
Instead, the child is only removed, but the underlying bdev remains
untouched.

### DIFF
--- a/mayastor/src/bdev/dev/loopback.rs
+++ b/mayastor/src/bdev/dev/loopback.rs
@@ -5,7 +5,7 @@ use snafu::ResultExt;
 use url::Url;
 
 use crate::{
-    bdev::{util::uri, CreateDestroy, GetName},
+    bdev::{lookup_child_from_bdev, util::uri, CreateDestroy, GetName},
     core::Bdev,
     nexus_uri::{self, NexusBdevError},
 };
@@ -78,6 +78,9 @@ impl CreateDestroy for Loopback {
     }
 
     async fn destroy(self: Box<Self>) -> Result<(), Self::Error> {
+        if let Some(child) = lookup_child_from_bdev(&self.name) {
+            child.remove();
+        }
         Ok(())
     }
 }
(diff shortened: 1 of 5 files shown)


mayastor/src/bdev/dev/loopback.rs
```diff
@@ -78,6 +78,18 @@ impl CreateDestroy for Loopback {
     }
 
     async fn destroy(self: Box<Self>) -> Result<(), Self::Error> {
+        instances().iter_mut().for_each(|n| {
```
gila, line 81: I see code like this at various places, it might be worth it to extract this into a method?

gila (commented):
This seems to replicate the nexus.reconfigure() approach (albeit without FFI) why can't we use the reconfigure approach here? note that we can per DRevent, can change the behaviour as needed

tiagolobocastro (approved):

mayastor/src/bdev/dev/loopback.rs
paulyoong: I have extracted this into the _lookup_child_from_bdev()_ function

paulyoong:
> This seems to replicate the nexus.reconfigure() approach (albeit without FFI) why can't we use the reconfigure approach here? note that we can per DRevent, can change the behaviour as needed

@gila based on our earlier discussion we know that we can probably do better with the way we handle closing down and reconfiguring channels. However, this PR does fix some issues seen during recent testing, so shall we commit this now and I will raise some additional work items to revisit this later?

gila (approved):

paulyoong:
bors merge

PR #537, opened 2020-11-25
Only set a child to closed on removal if open

paulyoong:
When removing a child only set the state to closed if it is open. In all
other cases, do not modify the state.

This change should fix those issues where the child changes to the 
closed state on removal, when it had previously had its state set appropriately. 
This results in the control plane being unable to make the correct decisions.

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_child.rs
+++ b/mayastor/src/bdev/nexus/nexus_child.rs
@@ -350,8 +350,19 @@ impl NexusChild {
         // The bdev is being removed, so ensure we don't use it again.
         self.bdev = None;
 
+        match self.state.load() {
+            ChildState::Open | Faulted(Reason::OutOfSync) => {
+                // Change the state of the child to ensure it is taken out of
+                // the I/O path when the nexus is reconfigured.
+                self.set_state(ChildState::Closed)
+            }
+            ChildState::Init
+            | ChildState::ConfigInvalid
+            | ChildState::Closed
+            | Faulted(_) => {}
+        }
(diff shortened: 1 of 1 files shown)


chriswldenyer (approved):
LGTM

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -350,8 +350,14 @@ impl NexusChild {
         // The bdev is being removed, so ensure we don't use it again.
         self.bdev = None;
 
+        // Ensure the child is not in the Open state. This guarantees that the
+        // child will be removed from the I/O path when the nexus is
+        // reconfigured.
+        if self.state.load() == ChildState::Open {
```
tiagolobocastro, line 356: do we need to check if a child is in a rebuild state?

jonathan-teh: Also, I think this is clearer as a match statement without a default state so that no states are forgotten if, say, in the future, new states are introduced.

paulyoong: @tiagolobocastro good catch. I've added a check for the out-of-sync state also so that the child doesn't get re-added to the write only channels.

@jonathan-teh good suggestion. I've changed it to a match statement.

jonathan-teh (approved):

gila (approved):

tiagolobocastro (approved):

paulyoong:
bors merge

PR #538, opened 2020-11-25
Introduce init stage for disabling default commit status in the future

jkryl:
Why I'm trying to push this change that does not change anything?  I have spent quite some time yesterday figuring out how to disable default commit status updates. My findings lead me to the conclusion that it could be done with checkout scm parameter but I don't know how exactly and it requires more experiments. Nevertheless, the small changes that I had for that in my branch could be useful and also describe the future work that needs to be done in a comment.

I hope that the code as it is won't break the pipeline. That remains to be tested by bors 🤞

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,13 +1,5 @@
 #!/usr/bin/env groovy
 
-// Will ABORT current job for cases when we don't want to build
-if (currentBuild.getBuildCauses('jenkins.branch.BranchIndexingCause') &&
-    BRANCH_NAME == "develop") {
-    print "INFO: Branch Indexing, aborting job."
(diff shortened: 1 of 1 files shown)


tiagolobocastro (approved):

bobek (approved):
I guess, we need to do `checkout scm` to be able to do `updateGithubCommitStatus(env.GIT_COMMIT`, right? That' s a bit sad, but I guess no big deal.

jkryl:
@bobek : exactly. That's my understanding. That's why I have moved the abort step inside the pipeline after the checkout. AFAIK checkout scm should accept a parameter that can disable the automatic commit status updates.

jkryl:
bors merge

PR #541, opened 2020-11-26
Change install test to use external resources CAS-541

blaisedias:
Change the install test to use defined docker registry,
and yaml files for pools.

Note pool status is not checked for success.

Communication of resources to use is via environment variables.
That appears to be the way for go test.

Previous behaviour is supported, see README.md


chriswldenyer (approved):
LGTM 
Maybe in the description of getTestClusterDetails say that the IP address and port are now returned in the first returned string

jkryl (approved):

blaisedias:
bors merge

PR #542, opened 2020-11-27
child disconnect fixes

gila:
This has been tested by @chriswldenyer as well as myself.


gila:
bors try

gila:
bors try

gila:
Bors retry

gila:
bors retry

gila:
bors retry

mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -348,6 +290,65 @@ impl Bio {
...
+                        "core {} thread {:?}, faulting child {}",
+                        Cores::current(),
+                        Mthread::current(),
+                        child,
+                    );
+
+                    let uri = child.name.clone();
+                    nexus.pause().await.unwrap();
+                    nexus.reconfigure(DREvent::ChildFault).await;
+                    //nexus.remove_child(&uri).await.unwrap();
```
paulyoong, line 339: Should we remove this line?

mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -348,6 +290,65 @@ impl Bio {
...
+                    bdev_destroy(&uri).await.unwrap();
+                    if nexus.status() != NexusStatus::Faulted {
+                        nexus.resume().await.unwrap();
+                    } else {
+                        error!(":{} has no children left... ", nexus);
```
paulyoong, line 344: If the nexus is faulted, this means we stay in the "paused" state. Is this ok? I assume so as at this point the nexus is unrecoverable.

paulyoong (approved):

mtzaurus (approved):

mayastor/src/bdev/nexus/nexus_channel.rs
```diff
@@ -75,14 +75,21 @@ pub enum DREvent {
...
+        if self.readers.is_empty() {
+            None
         } else {
-            self.previous = 0;
+            if self.previous < self.readers.len() - 1 {
+                self.previous += 1;
+            } else {
+                self.previous = 0;
+            }
+            Some(self.previous)
```
mtzaurus, line 91: Just a nit. Looks like this could be replaced by something like this
   self.previous = (self.previous + 1) % self.readers.len();

mayastor/src/bdev/nexus/nexus_io.rs
gila: Yes, for now, we have no children left so we cant turn it back on...

mayastor/src/bdev/nexus/nexus_channel.rs
gila: Yes mike, indeed. There is a snag hidden in the grass here, and that's that this checking for 0 should never happen in the first place. Or at least that's my thinking. Regardless your suggestion indeed is much nicer :)

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -771,6 +765,9 @@ impl Nexus {
     pub(crate) fn readv(&self, io: &Bio, channels: &mut NexusChannelInner) {
         // we use RR to read from the children.
         let child = channels.child_select();
+        if child.is_none() {
+            return;
```
chriswldenyer, line 769: I wonder if the io should be failed here?

chriswldenyer (approved):
LGTM

mayastor/src/bdev/nexus/nexus_channel.rs
chriswldenyer: I think if self.readers.len() happens to be zero, you get a divide-by-zero error

PR #544, opened 2020-11-30
E2e nvmf basic

chriswldenyer:
E2e test which creates a PVC using the mayastor-nvmf storage class, deploys a fio pod, runs fio against the volume, expects it to succeed, and then tears down the pod and the PVC. 
A module common/util.go contains functions available to other later e2e tests.
Not yet run by CI testing. Assumes an existing running cluster with a mayastor deployment.

### DIFF
--- a/mayastor-test/e2e/common/test.go
+++ b/mayastor-test/e2e/common/test.go
@@ -28,7 +28,9 @@ type TestEnvironment struct {
 	DynamicClient dynamic.Interface
 }
 
-func SetupTestEnv() TestEnvironment {
+var gTestEnv TestEnvironment
+
+func SetupTestEnv() {
 
 	By("bootstrapping test environment")
 	useCluster := true
@@ -70,7 +72,7 @@ func SetupTestEnv() TestEnvironment {
 	dynamicClient := dynamic.NewForConfigOrDie(restConfig)
 	Expect(dynamicClient).ToNot(BeNil())
 
-	return TestEnvironment{
+	gTestEnv = TestEnvironment{
 		Cfg:           cfg,
 		K8sClient:     k8sClient,
 		KubeInt:       kubeInt,
@@ -80,7 +82,7 @@ func SetupTestEnv() TestEnvironment {
 	}
 }
 
-func TeardownTestEnv(env *TestEnvironment) {
-	err := env.TestEnv.Stop()
+func TeardownTestEnv() {
+	err := gTestEnv.TestEnv.Stop()
 	Expect(err).ToNot(HaveOccurred())
 }
--- a/mayastor-test/e2e/common/test.go
+++ b/mayastor-test/e2e/common/test.go
@@ -1,0 +1,86 @@
+package common
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"k8s.io/client-go/deprecated/scheme"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+)
+
+type TestEnvironment struct {
+	Cfg           *rest.Config
+	K8sClient     client.Client
+	KubeInt       kubernetes.Interface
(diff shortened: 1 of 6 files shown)


mayastor-test/e2e/nightly/common/util.go
```diff
@@ -0,0 +1,361 @@
...
+	Expect(pvc).ToNot(BeNil())
+	switch pvc.Status.Phase {
+	case
+		corev1.ClaimBound,
+		corev1.ClaimPending,
+		corev1.ClaimLost:
+		return false
+	default:
+		return true
+	}
```
paulyoong, line 149: Could we invert this "switch" statement so that we match on the specific "invalid" status phase and return true? Then the default case would catch all other statuses and return false. This way, if an additional status phase is added, we would, by default, return false rather than true.

If possible, this would also be the case for the PV.

mayastor-test/e2e/nightly/common/util.go
```diff
@@ -0,0 +1,361 @@
...
+		corev1.VolumePending,
+		corev1.VolumeReleased:
+		return false
+	default:
+		return true
+	}
+}
+
+// Retrieve status phase of a Persistent Volume Claim
+func GetPvcClaimStatusPhase(volname string, kubeInt * kubernetes.Interface) (phase corev1.PersistentVolumeClaimPhase) {
```
paulyoong, line 182: We could remove "Claim" from the name as we already use PVC.

mayastor-test/e2e/nightly/common/util.go
```diff
@@ -0,0 +1,361 @@
...
+
+func RmPVC(volName string, scName string, dynamicClient * dynamic.Interface, kubeInt * kubernetes.Interface) {
+	fmt.Printf("removing %s, %s\n", volName, scName)
+
+	PVCApi := (*kubeInt).CoreV1().PersistentVolumeClaims
+
+	// Confirm the PVC has been created.
+	pvc, getPvcErr := PVCApi("default").Get(context.TODO(), volName, metav1.GetOptions{})
+	Expect(getPvcErr).To(BeNil())
+	Expect(pvc).ToNot(BeNil())
```
paulyoong, line 298: Maybe we should move this into its own function as it is used a number of times.

mayastor-test/e2e/nightly/nvmf_vol/nvmf_vol_test.go
```diff
@@ -0,0 +1,124 @@
...
+	go func() {
+		err = k8sManager.Start(ctrl.SetupSignalHandler())
+		Expect(err).ToNot(HaveOccurred())
+	}()
+
+	mgrSyncCtx, mgrSyncCtxCancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer mgrSyncCtxCancel()
+	if synced := k8sManager.GetCache().WaitForCacheSync(mgrSyncCtx.Done()); !synced {
+		fmt.Println("Failed to sync")
+	}
```
paulyoong, line 83: I wonder if the common bootstrapping stuff should also be moved to the common library? We are going to be doing this for every test.

paulyoong (approved):

mayastor-test/e2e/nightly/nvmf_vol/nvmf_vol_test.go
chriswldenyer: Yes, I've been thinking about doing that. It's the obvious next step.

mayastor-test/e2e/nightly/common/util.go
chriswldenyer: Ok, I'll take a look at doing that.

mayastor-test/e2e/nightly/common/util.go
chriswldenyer: Fair enough

mayastor-test/e2e/nightly/common/util.go
chriswldenyer: Ok, will do.

mayastor-test/e2e/nightly/common/util.go
chriswldenyer: Actually, I'm not sure if we gain much at the expense of making the code less clear. I think I'll leave it as-is, at least for now.

mayastor-test/e2e/nightly/nvmf_vol/nvmf_vol_test.go
chriswldenyer: Now refactored

blaisedias (approved):

blaisedias (approved):

chriswldenyer:
bors merge

PR #545, opened 2020-11-30
e2e tests for a disconnected replica in a 2-replica volume

chriswldenyer:
Currently tests are specific to a vagrant deployment  set up according to the READMEs. It is assumed that a cluster is running with mayastor installed with some label changes applied to k8s-1, and using the default IP addresses used by the vagrant deployment.
Isolation of the replica is achieved by modifying the iptables of the affected node.
The test takes care not to isolate a node running the nexus, moac or the kubernetes master node.
A successful fio run is used to indicate a functional volume.


chriswldenyer:
closed for refactoring

Issue #547, opened 2020-12-01
Create mayastor-client docker image


bobek:
It would be helpful to have an image on DockerHub with `mayastor-client` in it. Typical use case is to be able to quickly run in within existing k8s cluster with something like

```
kubectl run debian --rm -i --tty --image mayastor-client -- mayastor-client
```

PR #548, opened 2020-12-01
Build a mayastor-client image (#547)

bobek:
Ideas is to have a published image on dockerhub as desribed at #547 .

image


nix/pkgs/mayastor/default.nix
```diff
@@ -132,4 +132,38 @@ in
       install * $out/bin
     '';
   };
+
+  # this image to simplify troubleshooting at existing clusters
+  client = stdenv.mkDerivation {
+    name = "mayastor-client";
+    inherit version;
+    src = [
+      ../../../target/debug/mayastor-client
```
tiagolobocastro, line 141: not sure this is going to work for CI... if it gets scheduled on a workspace were the debug build was not done

bobek: Thank you for the comment. I hope that this is solved by producing `mayastor-client` as overlay on the top of the `mayastor`.  So, "standard" way current.

bobek:
I will investigate options for even smaller images, but we should be fine for now with basing the image off `mayastor`.

tiagolobocastro (approved):

blaisedias (approved):
LGTM

bobek:
bors merge

PR #550, opened 2020-12-01
feat: add pool and volume services to control plane composer

tiagolobocastro:
- Fix composer container restart and... 

Composer containers were not restarting because the containers were
being created with two networks, the global test and a "fake" one.

Address this by using a network with the compose name with defaults
to the global one but allow using different networks by making sure
that we always clean up networks with our tests labels.

... also add logging from containers on panic, tracing for the cargo
 code and common message bus connect fn to facilitate testing.

- Move the message bus client trait...

... into the common mbus library where it is more accessible.
Make the message bus traits Send

- Add detailed errors to the message bus

This way we're able to know where send/receive went wrong, eg failing
to send or an error received from the remote service.
Reply back with an error when we fail to deserialize a request, etc..
example:
"Reply message came back with an error: Failed to deserialize the
request: 'Failed to deserialize the publish payload: '....' into type
bus_api::v0::GetNodes': missing field `node` at line 1 column x'"

Update the ReceivedMessage type to auto include the reply type
using the base Message trait ::Reply associated type.

Added liveness probe which can be used to probe services for liveness.

- Add initial pool service, extended the rest...

... server and the rest client library o include get/put/delete verbs
for the pool and replicas.

Added simple compose test which makes use of the rest client library
to create/delete pools/replicas.

The pool service makes use of the node service to query existing nodes
and then issues gRPC commands to mayastor instances.

- Add initial volume service and the rest...

... server and client bindings.
The volume service is comprised of the nexus and volume operations where
the nexus is the "dumb" part of the service which creates nexuses where
it's told and the volume is the slightly smarter version which attempts
to decide where to create replicas, place nexuses etc...

The volume will effectively be the brain of the composer testing and
act as the equivalent to moac as csi controller plugin for mozart.

todo: mozart will need a clean and simple way of starting all the
required control plane services. Also, in the test do we want to
"talk" to the services through rest or the message bus?

- feat(jaeger): add tracing for the rest calls

Trace all rest request and responses via a command line option for
the the control plane rest server and its k8s operator counterpart.

The rest client library now has flag to enable/disable tracing.

Added jaegertracing/all-in-one:latest container to the rest client test
which allows us to see all the rest requests/responses @localhost:16686


composer/src/lib.rs
```diff
@@ -55,7 +55,7 @@ impl RpcHandle {
         name: String,
         endpoint: SocketAddr,
     ) -> Result<Self, String> {
-        let mut attempts = 60;
+        let mut attempts = 40;
         loop {
```
cjones1024, line 59: Is this the sort of thing that ought to be tunable via a configuration file or some such?

tiagolobocastro: Well, this is just the test library, I had to add a bigger delay before because of the nats lookup but managed to reduce it now by specifying a port. I can make it a default value part of Self which could be overridden.

mayastor/src/subsys/mbus/mod.rs
```diff
@@ -31,10 +31,15 @@ pub fn mbus_endpoint(endpoint: Option<String>) -> Option<String> {
                 (endpoint.as_str(), 4222)
             };
 
+            debug!("Looking up nats endpoint {}...", address_or_ip);
             if let Ok(ipv4) = address_or_ip.parse::<Ipv4Addr>() {
-                lookup_addr(&IpAddr::V4(ipv4)).expect("Invalid Ipv4 Address");
+                let nats = lookup_addr(&IpAddr::V4(ipv4))
+                    .expect("Invalid Ipv4 Address");
+                debug!("Nats endpoint looked up at {}", nats);
```
paulyoong, line 38: Could maybe word this as "Nats endpoint found at...."?

services/common/src/wrapper/mod.rs
```diff
@@ -0,0 +1,4 @@
+//! Pool Service backend for the message bus and gRPC
```
paulyoong, line 1: This comment looks like it should be removed.

services/common/src/wrapper/v0/mod.rs
```diff
@@ -0,0 +1,806 @@
+//! Implementation of the pool service backend which interacts with
```
paulyoong, line 1: This isn't specific for the pool service.

mayastor/src/subsys/mbus/mod.rs
tiagolobocastro: Done

mbus-api/src/lib.rs
```diff
@@ -13,16 +15,114 @@ pub mod v0;
...
+        options: TimeoutOptions,
+    },
+    #[snafu(display(
+        "Failed to reply back to message through the message bus"
+    ))]
+    Reply { source: io::Error },
+    #[snafu(display("Failed to flush the message bus"))]
+    Flush { source: io::Error },
+    #[snafu(display("Failed to subscribe to channel on the message bus"))]
+    Subscribe { source: io::Error },
```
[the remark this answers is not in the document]
tiagolobocastro: Done

mbus-api/src/message_bus/v0.rs
```diff
@@ -0,0 +1,264 @@
...
+                Ok(obj.clone())
+            }
+        } else {
+            Err(BusError::NotFound)
+        }
+    };
+}
+
+/// Interface used by the rest service to interact with the mayastor
+/// services via the message bus
```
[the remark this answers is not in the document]
tiagolobocastro: Yeah true, done

mbus-api/src/message_bus/v0.rs
```diff
@@ -0,0 +1,264 @@
...
+    async fn get_pools(filter: Filter) -> BusResult<Vec<Pool>> {
+        let pools = GetPools {
+            filter,
+        }
+        .request()
+        .await?;
+        Ok(pools.0)
+    }
+
+    /// Get replica with `id`
```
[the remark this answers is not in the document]
tiagolobocastro: Done

services/common/src/wrapper/mod.rs
tiagolobocastro: Done

services/common/src/wrapper/v0/mod.rs
tiagolobocastro: Done

composer/src/lib.rs
```diff
@@ -289,21 +335,56 @@ impl Builder {
...
+            tracing_subscriber::fmt().with_env_filter(RUST_LOG).init();
+        }
+        self
+    }
+
+    /// setup tracing for the cargo test code with `filter`
+    pub fn with_tracing(self, filter: &str) -> Self {
+        if let Ok(filter) =
+            tracing_subscriber::EnvFilter::try_from_default_env()
+        {
```
chriswldenyer, line 366: These two cases seem to have the same code

mbus-api/src/message_bus/v0.rs
```diff
@@ -0,0 +1,264 @@
...
+    /// Get node `id`
+    #[tracing::instrument(level = "debug", err)]
+    async fn get_node(id: String) -> BusResult<Node> {
+        let nodes = Self::get_nodes().await?;
+        let nodes =
+            nodes.into_iter().filter(|n| n.id == id).collect::<Vec<_>>();
+        only_one!(nodes)
+    }
+
+    /// Get pool with `id`
```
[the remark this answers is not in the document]
tiagolobocastro: Done

mbus-api/src/message_bus/v0.rs
```diff
@@ -0,0 +1,264 @@
...
+#[async_trait]
+pub trait MessageBusTrait: Sized {
+    /// Get all known nodes from the registry
+    #[tracing::instrument(level = "debug", err)]
+    async fn get_nodes() -> BusResult<Vec<Node>> {
+        let nodes = GetNodes {}.request().await?;
+        Ok(nodes.0)
+    }
+
+    /// Get node `id`
```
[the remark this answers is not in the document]
tiagolobocastro: Done

composer/src/lib.rs
tiagolobocastro: oh yeah, I got one to call the other now, thanks

tiagolobocastro:
bors try

tiagolobocastro:
bors retry

tiagolobocastro:
bors try

mbus-api/src/message_bus/v0.rs
```diff
@@ -168,6 +202,113 @@ pub trait MessageBusTrait: Sized {
...
+        .request()
+        .await?;
+        Ok(nexuses.0)
+    }
+
+    /// Get nexus with filter
+    #[tracing::instrument(level = "debug", err)]
+    async fn get_nexus(filter: Filter) -> BusResult<Nexus> {
+        let nexuses = Self::get_nexuses(filter).await?;
+        only_one!(nexuses)
```
gila, line 221: The macro usage seems somewhat confusing. The only_one!() macro does 
```rust
if let Some(obj) = list.first() { 
  if list.len() > 1 { } else {} 
}
```
I don't understand why `list.len()` is called since `list.first()` would return none if `list.len()` would be empty?

gila: Also when there is no element, i.e `nexuses.empty() == true` should we not return None rather then Error?

mbus-api/src/message_bus/v0.rs
```diff
@@ -168,6 +202,113 @@ pub trait MessageBusTrait: Sized {
...
+    /// create nexus
+    #[tracing::instrument(level = "debug", err)]
+    async fn create_nexus(request: CreateNexus) -> BusResult<Nexus> {
+        let nexus = request.request().await?;
+        Ok(nexus)
+    }
+
+    /// destroy nexus
+    #[tracing::instrument(level = "debug", err)]
+    async fn destroy_nexus(request: DestroyNexus) -> BusResult<()> {
```
gila, line 233: For all cases like this, Clippy will start to complain (and so will our build system when upgraded) where return units are used. Also when dealing with multiple requests I would argue that having "what is destroyed" be part of the return message might be useful? So what if we would return Ok(request.xxx) ?

mbus-api/src/message_bus/v0.rs
```diff
@@ -168,6 +202,113 @@ pub trait MessageBusTrait: Sized {
...
+        request.request().await?;
+        Ok(())
+    }
+
+    /// share nexus
+    #[tracing::instrument(level = "debug", err)]
+    async fn share_nexus(request: ShareNexus) -> BusResult<String> {
+        let share = request.request().await?;
+        Ok(share)
```
gila, line 242: This should be written as request.request.await without the explicit `?;` and `Ok(share)` The same goes for most lines where we have this format or Ok(request.wait?) formats.

mbus-api/src/message_bus/v0.rs
```diff
@@ -168,6 +202,113 @@ pub trait MessageBusTrait: Sized {
...
+
+    /// Get nexuses with filter
+    #[tracing::instrument(level = "debug", err)]
+    async fn get_nexuses(filter: Filter) -> BusResult<Vec<Nexus>> {
+        let nexuses = GetNexuses {
+            filter,
+        }
+        .request()
+        .await?;
+        Ok(nexuses.0)
```
gila, line 214: I prefer to use .into_inner() rather then `.0` calls. into_inner() is more verbose but if we do this consistently a lot easier to manage in the future. This is the same approach taken by other crates.

mbus-api/src/v0.rs
```diff
@@ -475,3 +547,274 @@ impl From<i32> for ReplicaState {
...
+    pub filter: Filter,
+}
+
+/// Nexus information
+#[derive(Serialize, Deserialize, Default, Debug, Clone, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct Nexus {
+    /// id of the mayastor instance
+    pub node: String,
+    /// name of the nexus
```
gila, line 566: The name or the UUID? they  *_MAYBE_*  the same, but don't have to be

mbus-api/src/v0.rs
```diff
@@ -475,3 +547,274 @@ impl From<i32> for ReplicaState {
...
+    pub uri: String,
+    /// state of the child
+    pub state: ChildState,
+    /// current rebuild progress (%)
+    pub rebuild_progress: Option<i32>,
+}
+
+/// Child State information
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
+pub enum ChildState {
```
gila, line 595: We need to think about this I guess. What is the single source of truth here for states? What is this state based on? the real state, the mapped state? And what if we changed the real state how will that brake code here? etc..

mbus-api/src/v0.rs
```diff
@@ -475,3 +547,274 @@ impl From<i32> for ReplicaState {
...
+#[derive(Serialize, Deserialize, Default, Debug, Clone)]
+pub struct Volumes(pub Vec<Volume>);
+bus_impl_message_all!(GetVolumes, GetVolumes, Volumes, Volume);
+
+/// Create volume
+#[derive(Serialize, Deserialize, Default, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct CreateVolume {
+    /// name of the volume
+    pub uuid: String,
```
gila, line 772: name != UUID.

rest/service/src/v0/children.rs
```diff
@@ -0,0 +1,185 @@
+use super::*;
+
```
gila, line 2: I see a lot of of fn some_rest_method() -> match xxx.await { Ok(o) => Http.ok().json(o), Err(e) => Err(rest::from(e).into) }

or something like that at least. I cant help but wonder: can this be made more ergonomic?

services/volume/src/service.rs
```diff
@@ -0,0 +1,333 @@
...
+
+    /// Get nexuses according to the filter
+    #[tracing::instrument(level = "debug", err)]
+    pub(super) async fn get_nexuses(
+        &self,
+        request: &GetNexuses,
+    ) -> Result<Nexuses, SvcError> {
+        let filter = request.filter.clone();
+        let nexuses = match filter {
+            Filter::All => self.get_node_nexuses(None).await?,
```
gila, line 45: Filter all or filter None?

mbus-api/src/message_bus/v0.rs
```diff
@@ -1,28 +1,172 @@
...
-    async fn get_node(id: String) -> BusResult<Option<Node>> {
+
+    /// Get node with `id`
+    #[tracing::instrument(level = "debug", err)]
+    async fn get_node(id: &str) -> BusResult<Node> {
         let nodes = Self::get_nodes().await?;
-        Ok(nodes.into_iter().find(|n| n.id == id))
+        let nodes =
+            nodes.into_iter().filter(|n| n.id == id).collect::<Vec<_>>();
+        only_one!(nodes)
```
gila, line 125: IMHO nodes.first() would be nicer where we map none to err

mbus-api/src/v0.rs
```diff
@@ -182,3 +202,276 @@ pub struct Node {
 #[derive(Serialize, Deserialize, Default, Debug, Clone)]
 pub struct Nodes(pub Vec<Node>);
 bus_impl_message_all!(GetNodes, GetNodes, Nodes, Node);
+
+/// Pool Service
+/// Filter query objects
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub enum Filter {
+    /// All objects
+    All,
```
gila, line 241: Filter all means, all objects so in effect the variant should be None?

mbus-api/src/v0.rs
```diff
@@ -182,3 +202,276 @@ pub struct Node {
...
+    /// Pool filters
+    ///
+    /// Filter by Pool id
+    Pool(String),
+    /// Filter by Node and Pool id
+    NodePool(String, String),
+    /// Filter by Node and Replica id
+    NodeReplica(String, String),
+    /// Filter by Node, Pool and Replica id
+    NodePoolReplica(String, String, String),
```
gila, line 253: These filters are cool, but I just wonder how easy it is to for example swap to strings? In a different commit I see we do lots of type definitions (good) but why don't we use it? i.e

struct NodeReplica { NodeName, PoolName, ReplicaName }, ?

mbus-api/src/message_bus/v0.rs
tiagolobocastro: So the reason was in case there are multiple objects with the same id, and in that case the error "NotUnique" is returned. I was torn between catering for this case or not so I'm happy to remove it.

tiagolobocastro: For the empty case, the error was to avoid having an option within the Result.

mbus-api/src/message_bus/v0.rs
tiagolobocastro: Ah, yeah then we can use that way and return the UUID of the object being destroyed for example.

mbus-api/src/v0.rs
tiagolobocastro: Yeah should just be the UUID indeed. I think the only one where we use an actual name is the pool, which is the name the user assigns to the pool.

mbus-api/src/v0.rs
tiagolobocastro: Yeah... another option for these services is for us to use the actual state (not mapped one) from the get-go.

rest/service/src/v0/children.rs
tiagolobocastro: Yeah, I know... The `?` probably doesn't help here because we return a responder trait, not an error. I'll see what I can do...

mbus-api/src/v0.rs
tiagolobocastro: Hmm good point :+1:

mbus-api/src/v0.rs
tiagolobocastro: You mean so it'd become like:
```
let replicas = get_replicas(Filter::NodePoolReplica {
node_name: "node".into(),
pool_name: "pool".into(),
replica_id: "1234...".into()
}).await?;
```

mbus-api/src/message_bus/v0.rs
gila: I see. But why is it a Result<T, E> to begin with rather than an Option<T>? Its a bit of bikeshedding perhaps, OTOH, if I have a list get_xx() method having it return what I look for is either there -- or it is not. i.e Some or None. But it seems we at this level need to account for transport errors and that's why it is a Result?

gila: > So the reason was in case there are multiple objects with the same id, and in that case the error "NotUnique" is returned. I was torn between catering for this case or not so I'm happy to remove it.

Im not sure what that means... A nexus (should) always be unique?

mbus-api/src/v0.rs
gila: Yes its an "oops" that got in and it started to lead its own life. Let us make sure we correct in the new API.

mbus-api/src/v0.rs
gila: Kinda sort yes. This way, when writing the code, and due to the rich nature of the rust tooling, you can by accident replace the node_name string, with a let's say, pool_name. 

```rust
let node_name  = NodeName::from(arg1);
let pool_name = PoolName::from(arg2);
let replica_id = ReplicaId::from(arg3);

let replica = get_replicas( Filter::NoneRepica { node_name, pool_name, replica_id }).await?;
```

This also will allow for, in the future nicer builder patterns? 

```rust
let replicas = Filter.with_nodes(vec![arg1]).with_pool("pool").build().await?;
```

This is a lot of personal taste of course and I'm not suggesting that the above is better or is nicer or whatever, but:

 - as arguments grow, builder patterns are nice
 - we have strongly typed system lets try to make use of it!
 - bikeshedding now, about these horrible small details, will prevent redoing things later

mbus-api/src/message_bus/v0.rs
tiagolobocastro: It should be yes, and in k8s it definitely will, was not sure if that would be the case for testing or for other types of objects like pools which have non uuid names. Also if we have ANA I guess we will have >1 nexus with the same UUID? So to get a nexus from a specific node you'd have to use the Filter by Node I guess.

tiagolobocastro: > I see. But why is it a Result<T, E> to begin with rather than an Option? Its a bit of bikeshedding perhaps, OTOH, if I have a list get_xx() method having it return what I look for is either there -- or it is not. i.e Some or None. But it seems we at this level need to account for transport errors and that's why it is a Result?

Yes that's right, the Result is to account for transport errors.

mbus-api/src/v0.rs
tiagolobocastro: I like the builder patter there! Maybe the `T` can be inferred from usage, so you could have all nexus from the node, or all pools from the node, like so:
```
let nexus: Vec<Nexus> = Filter.with_node(MY_NODE).build().await?;
let pools: Vec<Pool> = Filter.with_node(MY_NODE).build().await?;
```
But I'll probably start with the type changes first and leave the builder for latter.

services/volume/src/service.rs
tiagolobocastro: I wonder if Filter is wrong here. We're not filtering, it's more like querying so maybe it should be Query?

mbus-api/src/v0.rs
tiagolobocastro: I've raised CAS-608 for this.

mbus-api/src/v0.rs
tiagolobocastro: @gila coming back to this, would adding a specific type to each (NodeName/ReplicaId/etc) suffice? And keep the tuple since now you cannot swap the types?:
`// this makes it very explicit
let node_name  = NodeName::from(arg1);
let pool_name = PoolName::from(arg2);
let replica_id = ReplicaId::from(arg3);

// type checker will catch the error here
let replica = get_replicas(node_name, pool_name, replica_id).await?;`

gila: Yes I think so! A tuple (roughly to my understanding) just is a struct with nameless members so that would work just fine!

rest/service/src/v0/children.rs
tiagolobocastro: Yeah this works, I'll change all of the handlers from impl Responder to Result<...>

tiagolobocastro: Hmm the return type doesn't quite work after all... back to the drawing board...

gila (approved):
Nice work

tiagolobocastro:
bors try

services/common/src/wrapper/v0/pool.rs
```diff
@@ -0,0 +1,373 @@
...
+    request: &UnshareReplica,
+) -> rpc::mayastor::ShareReplicaRequest {
+    let request = request.clone();
+    rpc::mayastor::ShareReplicaRequest {
+        uuid: request.uuid.into(),
+        share: Protocol::Off as i32,
+    }
+}
+
+/// convert a message bus replica share to an rpc replica share
```
paulyoong, line 348: replica -> pool

services/common/src/wrapper/v0/registry.rs
```diff
@@ -0,0 +1,533 @@
...
+        nodes.values().cloned().collect()
+    }
+
+    /// List all cached nodes
+    pub async fn list_nodes(&self) -> Vec<Node> {
+        let nodes = self.list_nodes_wrapper().await;
+        nodes.iter().map(|n| n.node()).collect()
+    }
+
+    /// List all cached pools
```
paulyoong, line 67: Should "pools" be "pool wrapper" similar to line 55?

services/common/src/wrapper/v0/registry.rs
```diff
@@ -0,0 +1,533 @@
...
+                    "Failed to fetch the latest node information, '{}'",
+                    error
+                );
+            }
+        };
+
+        self.list_pools_wrapper().await
+    }
+
+    /// List all cached pool wrappers
```
paulyoong, line 96: Should "pool wrappers" be "pools"

services/common/src/wrapper/v0/registry.rs
```diff
@@ -0,0 +1,533 @@
...
+        &self,
+        request: &DestroyReplica,
+    ) -> Result<(), SvcError> {
+        let node = self.get_node(&request.node).await?;
+        node.destroy_replica(request).await?;
+        self.on_replica_removed(request).await;
+        Ok(())
+    }
+
+    /// Create replica and update registry
```
paulyoong, line 265: Create -> Share

services/common/src/wrapper/v0/registry.rs
```diff
@@ -0,0 +1,533 @@
...
+            &request.pool,
+            &request.uuid,
+            &request.protocol,
+            &share,
+        )
+        .await;
+        Ok(share)
+    }
+
+    /// Create replica and update registry
```
paulyoong, line 283: Create -> Unshare

services/common/src/wrapper/v0/registry.rs
```diff
@@ -0,0 +1,533 @@
...
+        &self,
+        request: &DestroyNexus,
+    ) -> Result<(), SvcError> {
+        let node = self.get_node(&request.node).await?;
+        node.destroy_nexus(request).await?;
+        self.on_destroy_nexus(request).await;
+        Ok(())
+    }
+
+    /// Create nexus
```
paulyoong, line 385: Create -> Share

services/common/src/wrapper/v0/registry.rs
```diff
@@ -0,0 +1,533 @@
...
+        request: &ShareNexus,
+    ) -> Result<String, SvcError> {
+        let node = self.get_node(&request.node).await?;
+        let share = node.share_nexus(request).await?;
+        self.on_update_nexus(&request.node, &request.uuid, &share)
+            .await;
+        Ok(share)
+    }
+
+    /// Destroy nexus
```
paulyoong, line 397: Destroy -> Unshare

services/pool/src/service.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// Destroy replica
+    #[tracing::instrument(level = "debug", err)]
+    pub(super) async fn destroy_replica(
+        &self,
+        request: &DestroyReplica,
+    ) -> Result<(), SvcError> {
+        self.registry.destroy_replica(&request).await
+    }
+
+    /// Create replica
```
paulyoong, line 169: Create -> Share

services/pool/src/service.rs
```diff
@@ -0,0 +1,204 @@
...
+    #[tracing::instrument(level = "debug", err)]
+    pub(super) async fn share_replica(
+        &self,
+        request: &ShareReplica,
+    ) -> Result<String, SvcError> {
+        self.registry.share_replica(&request).await
+    }
+
+    /// Create replica
```
paulyoong, line 178: Create -> Unshare

services/volume/src/service.rs
```diff
@@ -0,0 +1,338 @@
...
+        };
+        obj.start();
+        obj
+    }
+    /// Start registry poller
+    fn start(&self) {
+        self.registry.start();
+    }
+
+    /// Get all pools from node or from all nodes
```
paulyoong, line 26: pools -> nexuses

services/volume/src/service.rs
```diff
@@ -0,0 +1,338 @@
...
+    /// Destroy nexus
+    #[tracing::instrument(level = "debug", err)]
+    pub(super) async fn destroy_nexus(
+        &self,
+        request: &DestroyNexus,
+    ) -> Result<(), SvcError> {
+        self.registry.destroy_nexus(request).await
+    }
+
+    /// Create nexus
```
paulyoong, line 92: Create -> Share

services/volume/src/service.rs
```diff
@@ -0,0 +1,338 @@
...
+    #[tracing::instrument(level = "debug", err)]
+    pub(super) async fn share_nexus(
+        &self,
+        request: &ShareNexus,
+    ) -> Result<String, SvcError> {
+        self.registry.share_nexus(request).await
+    }
+
+    /// Destroy nexus
```
paulyoong, line 101: Destroy -> Unshare

paulyoong (approved):
Nice work @tiagolobocastro :+1: 

Just some minor comments.

composer/src/lib.rs
```diff
@@ -132,7 +137,33 @@ impl Binary {
...
+        }
+        self
+    }
+    /// pick up the nats argument name for a particular binary from nats_arg
+    /// and fill up the nats server endpoint using the network name
+    fn setup_nats(&mut self, network: &str) {
+        if !self.nats_arg.is_empty() {
+            self.arguments.push(self.nats_arg.clone());
+            self.arguments.push(format!("nats.{}:4222", network));
+            self.nats_arg = String::new();
```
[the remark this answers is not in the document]
tiagolobocastro: Done

mbus-api/src/v0.rs
```diff
@@ -172,14 +221,650 @@ impl Default for NodeState {
...
+}
+
+/// Pool information
+#[derive(Serialize, Deserialize, Default, Debug, Clone, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct Pool {
+    /// id of the mayastor instance
+    pub node: NodeId,
+    /// name of the pool
+    pub name: PoolId,
```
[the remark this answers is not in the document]
tiagolobocastro: Sort of both, but I've changed it to an ID as it's used as an ID anyway.

mbus-api/src/v0.rs
```diff
@@ -172,14 +221,650 @@ impl Default for NodeState {
...
+bus_impl_vector_request!(Nexuses, Nexus);
+bus_impl_message_all!(GetNexuses, GetNexuses, Nexuses, Nexus);
+
+/// Create Nexus Request
+#[derive(Serialize, Deserialize, Default, Debug, Clone, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct CreateNexus {
+    /// id of the mayastor instance
+    pub node: NodeId,
+    /// this UUID will be set in as the UUID
```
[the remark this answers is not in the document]
tiagolobocastro: It's copied verbatim from mayastor, but maybe we can improve it:
/// the nexus uuid will be set to this ?

rest/tests/v0_test.rs
```diff
@@ -1,101 +1,268 @@
...
+                "jaegertracing/all-in-one:latest",
+            )
+            .with_portmap("16686", "16686")
+            .with_portmap("6831/udp", "6831/udp")
+            .with_portmap("6832/udp", "6832/udp"),
+        )
+        //.with_base_image("alpine:latest".to_string())
+        .with_default_tracing()
+        .autorun(false)
+        //.with_clean(false)
```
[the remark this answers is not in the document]
tiagolobocastro: What about leaving it like this? [see code]

services/common/src/lib.rs
```diff
@@ -100,9 +102,17 @@ impl<'a> Context<'a> {
...
-            .expect("Requested data type not shared via with_shared_data!")
+        match self.state.try_get() {
+            Some(state) => state,
+            None => {
+                let type_name = std::any::type_name::<T>();
+                let error = format!(
+                    "Requested data type '{}' not shared via with_shared_data",
+                    type_name
+                );
+                panic!(error);
```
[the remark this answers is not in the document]
tiagolobocastro: We have to, otherwise what do we return?

services/common/src/wrapper/v0/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+    #[snafu(display("Failed to get nodes from the node service"))]
+    BusGetNodes { source: BusError },
+    #[snafu(display("Failed to get pools from the pool service"))]
+    BusGetPools { source: mbus_api::Error },
+    #[snafu(display("Failed to create pool from the pool service"))]
+    BusCreatePool { source: mbus_api::Error },
+    #[snafu(display("Failed to destroy pool from the pool service"))]
+    BusDestroyPool { source: mbus_api::Error },
+    #[snafu(display("Failed to destroy pool from the pool service"))]
+    BusGetReplicas { source: mbus_api::Error },
```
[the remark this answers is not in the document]
tiagolobocastro: Yeah thanks fixed, tbh all of those errors need scrubbing and adding more details, but I'll leave that for another time.

tiagolobocastro:
bors try

services/common/src/lib.rs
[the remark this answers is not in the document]
paulyoong: Fair enough.

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

Issue #508, opened 2020-11-12
Bringing down a node with nexus brings down the whole volume


jkryl:
It is a kind of known issue in mayastor that we plan to solve in long term by providing multiple instances of nexus for each volume running on different nodes. Then when one nexus goes down, nvme initiator in Linux kernel can switch to another nexus that is healthy. Though there is something more we could do in the meantime to recover a volume and although not as fast it could work for all protocols.

What happens if a node crashes or is willingly removed from the cluster is that k8s CSI calls unpublish on the volume and then it publishes the volume on a different node. We could recreate the nexus in control plane if we get a publish request for a volume with missing nexus (currently we fail such request) and that would make the volume available again. That is relatively a small change. We could take that even further and create nexus only if the volume is published. That has been proposed in the past and we should evaluate how much work that would be to implement and what are the cons (if any).

Note: This is a pure control plane issue that does not affect the data plane (mayastor).

jkryl:
It has been fixed and will be part of the next release - that is v0.7.

Just to clarify we cannot recover from nexus that has disappeared but what we do now is that we recreate the nexus on the same node when it comes back online. The best we can do now without having nvmf multipath feature.

PR #553, opened 2020-12-02
Defer nexus creation until the volume is published.

jkryl:
This addresses github issue:
Bringing down a node with nexus brings down the whole volume #508

Previously nexus was created when volume was created and it could
not be moved between nodes until the volume was destroyed. With
these changes nexus is not created until the volume is actually
published and it is destroyed when unpublished. Why?

* nexus with unused volumes was occasionally causing problems in
  data plane and unnecessary traffic (keep alive requests) between
  nexus and replicas.
* we could not recover a volume with nexus and app running on the
  node that was offline because nexus could not move between nodes
  even if volume was unpublished.
* we could not recover nexus of a published volume that came online
  after being offline for some time. Once the nexus was lost, it was
  lost forever.

So it is a significant improvement over existing situation, though we
still cannot recover a volume with crashed nexus that had been published.
For that we need to support multipathing in nvmf initiator which will be
done at some point in the future.


csi/moac/volume.ts
```diff
@@ -0,0 +1,877 @@
...
+      const r = Object.values(this.replicas).find((r) => r.uri === ch.uri);
+      return {ch, r};
+    });
+    // add newly found replicas to the nexus (one by one)
+    const newReplicas = Object.values(this.replicas).filter((r) => {
+      return (!r.isOffline() &&
+        !childReplicaPairs.find((pair) => pair.r === r) &&
+        !this.nodeBlackList[r.pool!.node!.name]);
+    });
+    for (let i = 0; i < newReplicas.length; i++) {
```
tiagolobocastro, line 342: do we need to check if we've got the required number of replicas already? and if so we don't need to add these to the nexus? Also won't this cause multiple rebuilds at the same time? (which is fine but I'm not if we want that or not)

csi/moac/volume.ts
```diff
@@ -0,0 +1,877 @@
...
+      try {
+        await this.nexus.addReplica(newReplicas[i]);
+        return;
+      } catch (err) {
+        // XXX what should we do with the replica? Destroy it?
+        this.nodeBlackList[newReplicas[i].pool!.node!.name] = true;
```
tiagolobocastro, line 348: I guess this can happen if the node where the replica lives has just gone offline, but it can also happen if there is something wrong with the replica itself. Probably not much else we can do at the moment...

tiagolobocastro (approved):

csi/moac/volume.ts
```diff
@@ -0,0 +1,877 @@
...
+    this.publishedOn = nexus.node.name;
+    log.info(`Published "${this}" at ${uri}`);
+    this.emitEvent('mod');
+    return uri;
+  }
+
+  // Undo publish operation on the volume.
+  async unpublish() {
+    if (this.publishedOn) {
+      this.publishedOn = undefined;
```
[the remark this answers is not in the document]
jkryl: some of the code here will need to change for sure. Currently we destroy nexus no matter what upon unpublish. The problem with unsetting publishedOn only when the nexus is destroyed is that then moac does not know that it should destroy the nexus in FSA. So we would have to introduce another field to distinguish published state from still-published-but-should-not-be and that to me sounds like an overkill. The nexus will be removed at first opportunity when it's possible so unsetting publishedOn and returning OK for the operation in spite of the fact that the nexus might still be there seems ok to me. Also in respect to what might happen next - that the volume may be published again - we just use the nexus that is there, so should not cause any harm.

csi/moac/volume.ts
jkryl: yea. I have introduced nodeBlackList to avoid the most obvious problems of endless loops when trying to do something with replica, it does not work, but next time we are in FSA we do the same thing again and cannot move forward. So it's important feature, however it did not receive any testing (both manual and unit-tests) so later I will probably have to come back to it and make the feature complete if it does not work according to expectations.

csi/moac/volume.ts
jkryl: I can think of just one case when this would be a problem (adding only as many replicas as we need) and that is when moving replicas. How that works is that we create a new replica according to new scheduling criteria and when it's added we remove the least preferred which is the old one. Now if I refused to add a child because there are already enough of them, that would break this feature.

tiagolobocastro: Ah yes that makes sense... However I still don't think we should be adding more than one at once. A while back we had a discussion about rebuild and it was pointed out that we never rebuild more than one at once so we didn't need to have more rebuild test cases for this scenario, so I'm not sure if it's a good idea to start doing that now without revisiting the tests first.

csi/moac/volume.ts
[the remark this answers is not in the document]
tiagolobocastro: hmm, what if the destroy fails for some other reason (bug on mayastor or something). Once the user pod runs again is there a possibility that we would try to create the nexus on a different node? (at that point we'd have 2 of the same nexus on different nodes)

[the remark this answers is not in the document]
jkryl: this goes a bit into the deferred nexus destruction feature that I'm going to implement in the future, but to me it seems it is not possible. If there is a next request to publish a volume then the nexus we failed to destroy is still there, so we will use it. If the nexus is unusable for any reason then application will probably notice that as it will receive IO errors.

csi/moac/volume.ts
jkryl: well, if I cannot have more than one replica under rebuild then this is indeed a problem. I haven't expected that though so I will need to clarify this with you and Paul. Thanks for pointing it out.

jkryl: Paul confirmed that adding multiple children, all being in rebuild state, is ok for mayastor. Let's go with that then and we can change the code later if we see any problem with it.

csi/moac/crds/mayastorvolume.yaml
```diff
@@ -70,13 +70,16 @@ spec:
                 reason:
                   description: Further explanation of the state if applicable.
                   type: string
-                node:
-                  description: Name of the k8s node with the nexus.
+                publishedOn:
```
paulyoong, line 73: I know there was a discussion about the name of this. My suggestion would be "location" or "nexusLocation"?

paulyoong (approved):

csi/moac/crds/mayastorvolume.yaml
jkryl: Hi Paul, there was actually a discussion about that this morning on slack. The conclusion was to go with `targetNodes` . It is plural because in future we would like to support multiple nexuses for a volume.

paulyoong: :+1:

jkryl:
bors merge

PR #554, opened 2020-12-03
Fix for a recent regression to develop branch testing

jkryl:
The current code works but in case that develop should not be
tested and error() is called in init stage, the final result
is ERROR instead of SUCCESS. This fix hasn't been tested but
according to stack-overflow it should work ;-)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -59,6 +59,10 @@ String cron_schedule = BRANCH_NAME == "develop" ? "0 2 * * *" : ""
 
 pipeline {
   agent none
+  options {
+    timeout(time: 2, unit: 'HOURS')
+    parallelsAlwaysFailFast()
+  }
   triggers {
     cron(cron_schedule)
   }
@@ -66,6 +70,7 @@ pipeline {
   stages {
     stage('init') {
       agent { label 'nixos-mayastor' }
+      options { skipDefaultCheckout true }
       steps {
         // TODO: We want to disable built-in github commit notifications.
         // skip-notifications-trait plugin in combination with checkout scm step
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -59,13 +59,18 @@ String cron_schedule = BRANCH_NAME == "develop" ? "0 2 * * *" : ""
 
 pipeline {
   agent none
+  options {
+    timeout(time: 2, unit: 'HOURS')
```
bobek, line 63: :+1:

Jenkinsfile
```diff
@@ -76,8 +81,9 @@ pipeline {
           if (currentBuild.getBuildCauses('jenkins.branch.BranchIndexingCause') &&
               BRANCH_NAME == "develop") {
               print "INFO: Branch Indexing, aborting job."
-              currentBuild.result = 'ABORTED'
-              error('Stopping early')
+              currentBuild.getRawBuild().getExecutor().interrupt(Result.SUCCESS)
+              sleep(3)   // Interrupt is not blocking and does not take effect immediately.
```
bobek, line 85: The magic sleep ...  the glue of the Internet. How safe is 3 (I assume seconds)? The whole pipeline takes long time, so we can also do 10s ;)

bobek (approved):

Jenkinsfile
jkryl: yea :-) The example I took this from had 1 second so I thought that 3 should be enough, but who knows :-)

tiagolobocastro (approved):

jkryl:
bors merge

PR #558, opened 2020-12-04
The fix for commit status updates.

jkryl:
The fix is a combination of changes to jenkinsfile and to the jenkins configuration done through UI.

Tested on test-jenkins branch.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,25 +1,5 @@
 #!/usr/bin/env groovy
 
-// Update status of a commit in github
-def updateGithubCommitStatus(commit, msg, state) {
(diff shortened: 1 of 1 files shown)


chriswldenyer (approved):
Not familiar with groovy, but the change seems to make sense and you say you've tested it. The new code no longer has repoSource and commitShaSource, or the step checkout scm. I assume that is all intentional?

tiagolobocastro (approved):

jkryl:
Right. The default values work actually better than explicit (not sure why)

jkryl:
bors merge

PR #559, opened 2020-12-04
CAS-516: Convert rebuild tests to Docker Compose tests

paulyoong:
Rebuild test cases in nexus_rebuild.rs have been converted into Docker
Compose tests in rebuild.rs. The reason for this change is to make the
tests more representative of real world applications. As an example, the
new tests share replicas over NVMf.

One test (rebuild_lookup) remains in nexus_rebuild.rs. This test cannot
be converted to a Compose test as the functionality is not exposed over
RPC. However, the test is still useful and hence remains.


mayastor/tests/rebuild.rs
```diff
@@ -0,0 +1,1051 @@
...
+
+/// Test faulting the destination of a rebuild.
+#[tokio::test]
+async fn rebuild_fault_dst() {
+    let test = start_infrastructure("rebuild_fault_dst").await;
+    let (mut ms1, _, mut ms3) = setup_test(&test, 1).await;
+    let nexus_hdl = &mut ms1;
+    let child = &get_share_uri(&ms3);
+
+    // Check a rebuild is started for the added child.
```
tiagolobocastro, line 271: this pattern seems to repeat quite a bit, maybe could be made into a single fn

mayastor/tests/rebuild.rs
```diff
@@ -0,0 +1,1051 @@
...
+        NexusState::NexusDegraded as i32
+    );
+    assert_eq!(get_num_rebuilds(nexus_hdl).await, 0);
+}
+
+/// Test rebuild with different sizes of source and destination children.
+#[tokio::test]
+async fn rebuild_sizes() {
+    struct TestCase {
+        child1_size: Option<u64>,
```
tiagolobocastro, line 308: Why are these options? You can do bdev_create_and_share(ms2, **Some(**test.child1_size**)**...

mayastor/tests/rebuild.rs
```diff
@@ -0,0 +1,1051 @@
...
+            share_uri,
+        });
+    }
+
+    // Create a nexus with a single healthy child.
+    let nexus_hdl = &mut test.grpc_handle("ms1").await.unwrap();
+    create_nexus(nexus_hdl, vec![children[1].share_uri.clone()]).await;
+
+    let degraded_children = children[2 ..= 4].to_vec();
+    assert_eq!(degraded_children.len(), 3);
```
tiagolobocastro, line 567: what is this checking? seems to just be checking that [ 2 ..= 4 ] = 3 which will be true?

mayastor/tests/rebuild.rs
```diff
@@ -0,0 +1,1051 @@
...
+    // Connect to nexus over NVMf.
+    let nexus_uri = publish_nexus(nexus_hdl).await;
+    let nexus_tgt = nvmf_connect(nexus_uri.clone());
+
+    // Run fio against nexus.
+    let (s, r) = unbounded::<i32>();
+    let nvmf_tgt = nexus_tgt.clone();
+    std::thread::spawn(move || {
+        s.send(common::fio_verify_size(&nvmf_tgt, NEXUS_SIZE))
+            .unwrap();
```
tiagolobocastro, line 642: probably don't want to unwrap here and it can block the test, should unwrap in the test thread.

mayastor/tests/rebuild.rs
```diff
@@ -0,0 +1,1051 @@
...
+    // Disconnect and destroy nexus
+    nvmf_disconnect(nexus_uri);
+    destroy_nexus(nexus_hdl).await;
+
+    // Check children are in-sync.
+    let child1_tgt = nvmf_connect(child1.clone());
+    let child2_tgt = nvmf_connect(child2.clone());
+    common::compare_devices(&child1_tgt, &child2_tgt, CHILD_SIZE_MB, true);
+    nvmf_disconnect(child1);
+    nvmf_disconnect(child2);
```
tiagolobocastro, line 673: @gila now that we're using nvmf more throughout the cargo tests, should NvmeTarget have a drop where it disconnects to prevent leaking devices on CI?

mayastor/tests/rebuild.rs
```diff
@@ -0,0 +1,1051 @@
...
+            key: "".to_string(),
+            share: ShareProtocolNexus::NexusNvmf as i32,
+        })
+        .await
+        .unwrap()
+        .into_inner();
+    reply.device_uri
+}
+
+/// Create and share a bdev and return.
```
tiagolobocastro, line 724: return the share uri?

tiagolobocastro (approved):

mayastor/tests/rebuild.rs
tiagolobocastro: ah but nvmeadm is also used by the csi so probably not a good idea then...

mayastor/tests/rebuild.rs
```diff
@@ -0,0 +1,1050 @@
...
+        .await
+        .unwrap()
+}
+
+/// Set up the prerequisites for the tests.
+/// Create a nexus on ms1 and create NVMf shares from ms2 & ms3.
+/// The number of children to be added to the nexus is passed in as a parameter.
+async fn setup_test(
+    test: &ComposeTest,
+    num_nexus_children: usize,
```
chriswldenyer, line 695: maybe sanity-check num_nexus_children here. What happens if it is > 2?

mayastor/tests/rebuild.rs
```diff
@@ -0,0 +1,1050 @@
...
+    pause_rebuild(nexus_hdl, child).await;
+    assert!(wait_for_rebuild_state(
+        nexus_hdl,
+        child,
+        "paused",
+        Duration::from_secs(1),
+    )
+    .await
+    .unwrap());
+    assert_eq!(
```
chriswldenyer, line 174: You might want to put this check into a function with the expected state as a parameter

mayastor/tests/rebuild.rs
paulyoong: Yeah I thought about that but there were some tests where additional behaviour was required between adding a child and rebuilding (i.e. pausing) which is why I didn't separate it out.

I'll leave it like this for now and revisit it if needed.

mayastor/tests/rebuild.rs
paulyoong: Good point - changed it.

mayastor/tests/rebuild.rs
paulyoong: I used it to check I was creating the slice correctly and then forgot to remove it. It's removed now.

mayastor/tests/rebuild.rs
paulyoong: Thanks - I've added an error message instead.

mayastor/tests/rebuild.rs
paulyoong: :+1: I've modified the comment.

mayastor/tests/rebuild.rs
paulyoong: I've added an assert to check for <3 children, as that is all that is currently required for the tests.

mayastor/tests/rebuild.rs
paulyoong: Done :+1:

chriswldenyer:
LGTM

chriswldenyer (approved):

paulyoong:
bors merge

PR #560, opened 2020-12-07
Move pipeline abort to global scope again.

jkryl:
Now when default github notifications are disabled in jenkins
configuration we can abort the job at global level. Another reason
is that we don't have to enable execution of unsafe code in the
pipeline then. We have already enabled raw job object access but we
would need more to get rid of the latest error message:

 "Scripts not permitted to use method hudson.model.Run getExecutor"

The init stage remains there because it avoids duplicating the code
for setting pending state on a github commit.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -34,6 +34,14 @@ def notifySlackUponStateChange(build) {
   }
 }
 
+// Will ABORT current job for cases when we don't want to build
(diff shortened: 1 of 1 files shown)


bobek (approved):

tiagolobocastro (approved):

jkryl:
bors merge

PR #567, opened 2020-12-08
Templatize deploy yamls; use in e2e

arne-rusek:
I started templates in shell and quickly run into problems and complicated solutions. And actually with helm I was done faster than with shell and it actually worked.

I've tried to run the tests. I didn't have correct images & registry installed, but it installed mayastor and start of containers failed on pull, obviously. Uninstall test was OK.

Integration to Jenkins will take me a bit (IDK how to commit, and even if I can?). I'll create another PR to automatize generating of deploy yamls per branch. For now one can run `./scripts/generate-deploy-yamls.sh latest` or `./scripts/generate-deploy-yamls.sh develop` to re-create deploy yamls and commit manually if needed.

E2E tests will require `helm` binary on Jenkins machine. I tried... but... wow... Could any nix-er suggest how to do correctly please?

Now it lives in `/etc/profiles/per-user/jenkins/bin`... It can be run via docker if needed... Because we need it to generate files I did not implement it as it would require some volume mounting obscure magic.

Please take a look if acceptable.


jkryl (approved):
Amazing! 👍 

helm can be installed on jenkins slaves by putting the package name to /etc/nixos/configuration.nix and then running nixos-rebuild switch. doc should be updated too (jenkins.md). But before we start to taint the configuration of jenkins slaves I think we should have some idea of how we are going to commit changes back the repo from jenkins. I don't see any easy way to do that. So perhaps we will have to resort to manual action any time anything in the helm charts is changed. We could add some kind of a check to pre-commit hook and even test it on behalf of ci/cd. I mean that the generated yamls are up-to-date. If not simply return an error and that will prevent the commit from getting merged.

bobek:
>  we should have some idea of how we are going to commit changes back the repo from jenkins.

We can do commit and push to remote repo, right? We need to provision a rw user for that.

BTW how much do we need to have everything in Jenkins? I am just thinking that it is super simple with github actions. Something like:

```
      - name: Commit changes
        uses: EndBug/add-and-commit@master
        with:
          author_name: ssh_tf_bot
          author_email: ssh_tf_bot@mayadata.io
          message: "Updating Terraform SSH keys"
          add: "terraform_ssh.json"
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

from other project of ours.

arne-rusek:
> > we should have some idea of how we are going to commit changes back the repo from jenkins.
> 
> We can do commit and push to remote repo, right? We need to provision a rw user for that.

There's a token, but I'm not sure if it is r/w (guessing not :))

> BTW how much do we need to have everything in Jenkins? I am just thinking that it is super simple with github actions. Something like:
> 
> ```
>       - name: Commit changes
>         uses: EndBug/add-and-commit@master
>         with:
>           author_name: ssh_tf_bot
>           author_email: ssh_tf_bot@mayadata.io
>           message: "Updating Terraform SSH keys"
>           add: "terraform_ssh.json"
>         env:
>           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
> ```
> 
> from other project of ours.

I'm not sure we do need it in jenkins. I'm all for GH Action. Thinking about it running it in jenkins with the time our current job takes there's gonna be a relatively large time window in which yamls might be out-of-sync, and that's not nice (i.e. on merging to master images' tags should be changed compared to those in e.g. develop). Separate GH action should be at least faster.

arne-rusek:
bors merge

paulyoong (approved):

arne-rusek:
bors merge

PR #574, opened 2020-12-10
Integrate install & uninstall e2e tests to jenkins

jkryl:
It is just a starting point. There are many things to improve.
For example:

 * reworking test directory structure,
 * dynamic spawn of k8s cluster,
 * making it possible to run multiple e2e tests in parallel,
 * eliminate pool template file in install test
 * generate deployment yaml files to a different dir than deploy/
 * adding more e2e tests

### DIFF
--- a/mayastor-test/e2e/install/install_test.go
+++ b/mayastor-test/e2e/install/install_test.go
@@ -113,8 +113,8 @@ func getDeployYamlDir() string {
 func applyDeployYaml(filename string) {
 	cmd := exec.Command("kubectl", "apply", "-f", filename)
 	cmd.Dir = getDeployYamlDir()
-	_, err := cmd.CombinedOutput()
-	Expect(err).ToNot(HaveOccurred())
+	out, err := cmd.CombinedOutput()
+	Expect(err).ToNot(HaveOccurred(), "%s", out)
 }
 
 // Encapsulate the logic to find where the templated yamls are
@@ -130,8 +130,8 @@ func makeImageName(registryAddress string, imagename string, imageversion string
 func generateYamls(registryAddress string) {
 	bashcmd := "../../../scripts/generate-deploy-yamls.sh  ci " + registryAddress
 	cmd := exec.Command("bash", "-c", bashcmd)
-	_, err := cmd.CombinedOutput()
-	Expect(err).ToNot(HaveOccurred())
+	out, err := cmd.CombinedOutput()
+	Expect(err).ToNot(HaveOccurred(), "%s", out)
 }
 
 func applyTemplatedYaml(filename string, imagename string, registryAddress string) {
@@ -139,8 +139,8 @@ func applyTemplatedYaml(filename string, imagename string, registryAddress strin
 	bashcmd := "IMAGE_NAME=" + fullimagename + " envsubst < " + filename + " | kubectl apply -f -"
 	cmd := exec.Command("bash", "-c", bashcmd)
 	cmd.Dir = getTemplateYamlDir()
-	_, err := cmd.CombinedOutput()
-	Expect(err).ToNot(HaveOccurred())
+	out, err := cmd.CombinedOutput()
+	Expect(err).ToNot(HaveOccurred(), "%s", out)
 }
 
 // We expect this to fail a few times before it succeeds,
@@ -164,6 +164,9 @@ func moacReadyPodCount() int {
 }
 
 // create pools for the cluster
+//
+// TODO: Ideally there should be one way how to create pools without using
+// two env variables to do a similar thing.
 func createPools(mayastorNodes []string) {
 	envPoolYamls := os.Getenv("e2e_pool_yaml_files")
 	poolDevice := os.Getenv("e2e_pool_device")
@@ -186,11 +189,11 @@ func createPools(mayastorNodes []string) {
 			bashcmd := "NODE_NAME=" + mayastorNode + " POOL_DEVICE=" + poolDevice + " envsubst < " + "pool.yaml.template" + " | kubectl apply -f -"
 			cmd := exec.Command("bash", "-c", bashcmd)
 			cmd.Dir = getTemplateYamlDir()
-			_, err := cmd.CombinedOutput()
-			Expect(err).ToNot(HaveOccurred())
+			out, err := cmd.CombinedOutput()
+			Expect(err).ToNot(HaveOccurred(), "%s", out)
 		}
 	} else {
-		// No pools created
+		Expect(false).To(BeTrue(), "Neither e2e_pool_yaml_files nor e2e_pool_device specified")
 	}
 }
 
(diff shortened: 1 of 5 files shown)


mayastor-test/e2e/install/install_test.go
```diff
@@ -165,32 +164,16 @@ func moacReadyPodCount() int {
 
 // create pools for the cluster
 func createPools(mayastorNodes []string) {
-	envPoolYamls := os.Getenv("e2e_pool_yaml_files")
 	poolDevice := os.Getenv("e2e_pool_device")
-	if len(envPoolYamls) != 0 {
```
blaisedias, line 170: By removing this we lose the capability to use arbitrary pool definitions, which is useful when all the nodes in the cluster are not the same.  This was added for precisely that case. The tool deploying the cluster "knows" the shape of the cluster and can generate appropriate pool definitions and pass them to the e2e install test.

mayastor-test/e2e/install/install_test.go
```diff
@@ -165,32 +164,16 @@ func moacReadyPodCount() int {
...
-			fmt.Println("creating pool on:", mayastorNode, " using device:", poolDevice)
-			bashcmd := "NODE_NAME=" + mayastorNode + " POOL_DEVICE=" + poolDevice + " envsubst < " + "pool.yaml.template" + " | kubectl apply -f -"
-			cmd := exec.Command("bash", "-c", bashcmd)
-			cmd.Dir = getTemplateYamlDir()
-			_, err := cmd.CombinedOutput()
-			Expect(err).ToNot(HaveOccurred())
-		}
-	} else {
-		// No pools created
+	// Use the template file to create pools as per the devices
```
blaisedias, line 168: Perhaps we should add a check that e2e_pool_device was defined, and fail if not.
Nothing will work if no pools are defined so kind of pointless, proceeding further.

blaisedias (changes requested):

mayastor-test/e2e/install/install_test.go
jkryl: Oh, you mean when the device has a different path on each node. When one cannot say it's for example /dev/nvme1n1 everywhere. I haven't thought of this use case. I will insert it back unless I come up with a better way how to configure pools. So what I don't like very much about the current code is that we have two entirely different ways how to create pools. Instead, I would like to unite the pool configuration and generate pool CRs based on a cluster description file that would contain mapping between worker nodes and pool devices and possibly we could add more things to it in the future. Now when helm is part of the environment we could perhaps use it to render pool yaml files based on the cluster description file. But it sounds like a non-trivial amount of work, so for now I will just put the code back 🤔

mayastor-test/e2e/install/install_test.go
jkryl: will fix 👍

mayastor-test/e2e/install/install_test.go
blaisedias: agreed, and I would say that the most flexible way would be this way, i.e. the tool standing up the cluster defines the pools, rather than install test itself defining the pools. The second approach made sense to get things running quickly for developers.

blaisedias: Also it is not just the fact that the devices may be different, but I think we should also be testing in clusters which have nodes on which only maystor runs and not csi-mayastor.

scripts/e2e-test.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env sh
+
```
arne-rusek, line 2:
```suggestion
set -e
```
I would recommend adding this stanza. And later run tests with if, e.g.:
```
if ! go test; then
  cleanup_fail=true
fi
```

It seems to me that this is a wrapper that will have tests added later one by one and people tend to expect failures not to be silent. And without `set -e` failures will be silent and will produce hard-to-debug time consuming problems when changes are added to the script. Even now the `cd` commands might fail if tests are moved but the error will be reported for `go test` line with error that will probably look like:

```
$ go test
go: cannot find main module; see 'go help modules'                                                                                                                                                                                        
```
which is confusing and doesn't point to the root cause.

You might wanna take a look at https://www.shellcheck.net/ that will help with quoting and similar issues.

jkryl: fixed 👍

mayastor-test/e2e/install/install_test.go
```diff
@@ -113,8 +113,8 @@ func getDeployYamlDir() string {
 func applyDeployYaml(filename string) {
```
chriswldenyer, line 113: At some point we should move these functions to the e2e/common/ directory.

scripts/e2e-test.sh
```diff
@@ -0,0 +1,36 @@
...
+  echo "Missing parameter registry"
+  exit 1
+fi
+
+test_failed=
+export e2e_docker_registry="$REGISTRY"
+export e2e_pool_device=/dev/nvme1n1
+
+for dir in $TESTS; do
+  cd $SCRIPTDIR/../mayastor-test/e2e/$dir
```
arne-rusek, line 20: Just a nitpick of shellcheck (not mine :smile:):

```suggestion
  cd "$SCRIPTDIR/../mayastor-test/e2e/$dir"
```

scripts/e2e-test.sh
```diff
@@ -0,0 +1,36 @@
...
+  if ! go test; then
+    test_failed=1
+    break
+  fi
+done
+
+# must always run uninstall test in order to clean up the cluster
+cd $SCRIPTDIR/../mayastor-test/e2e/uninstall
```
arne-rusek, line 28:
```suggestion
cd "$SCRIPTDIR/../mayastor-test/e2e/uninstall"
```

mayastor-test/e2e/install/install_test.go
jkryl: agree. It should be done at some point (in a future PR).

arne-rusek (approved):
LGTM - just two suggestions from shellcheck

chriswldenyer (approved):

scripts/e2e-test.sh
jkryl: fixed. thanks!

scripts/e2e-test.sh
jkryl: fixed. thanks!

jkryl:
bors merge

jkryl:
bors retry

jkryl:
bors retry

jkryl:
bors retry

scripts/e2e-test.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env sh
+
+set -e
+
+SCRIPTDIR=$(dirname "$0")
```
blaisedias, line 5: Change to 
`SCRIPTDIR="$(dirname "$(realpath "$0")")"`

blaisedias (changes requested):

scripts/e2e-test.sh
arne-rusek:
```suggestion
SCRIPTDIR="$(realpath "$(dirname "$0")")"
```

jkryl: fixed. thanks 👍

jkryl:
bors retry

blaisedias (approved):

jkryl:
bors retry

PR #576, opened 2020-12-11
cas-561 Provide access to udev databse for mayastor container via additional volume mount.

cjones1024:
mayastor requires access to the udev database that resides in /run/udev in order for the ListBlockDevices() gRPC request to function correctly. This access must be provided via an explicit volume mount when mayastor is run from within a container.
Without such access, the gRPC call returns partial and (highly misleading) incorrect information.

### DIFF
--- a/deploy/mayastor-daemonset.yaml
+++ b/deploy/mayastor-daemonset.yaml
@@ -67,6 +67,8 @@ spec:
         volumeMounts:
         - name: device
           mountPath: /dev
+        - name: udev
+          mountPath: /run/udev
         - name: dshm
           mountPath: /dev/shm
         - name: configlocation
@@ -93,6 +95,10 @@ spec:
       - name: device
         hostPath:
           path: /dev
+          type: Directory
+      - name: udev
+        hostPath:
+          path: /run/udev
           type: Directory
       - name: dshm
         emptyDir:


deploy/mayastor-daemonset.yaml
```diff
@@ -67,6 +67,8 @@ spec:
         volumeMounts:
         - name: device
           mountPath: /dev
+        - name: udev
+          mountPath: /run/udev
```
mittachaitu, line 71: Q: Do we need to mount as Read-Only?

mittachaitu (approved):
lgtm

deploy/mayastor-daemonset.yaml
cjones1024: I believe that a readonly mount is most likely sufficient, but without completely thorough testing it is actually safer to leave as is.

cjones1024:
bors try

paulyoong (changes requested):
@jkryl has highlighted the fact that the deploy YAML files are now constructed from templates. Therefore these changes should be included in the template files. I will work on this.

paulyoong:
Closing this PR as the issue has been addressed by commit b751efca90b6e9572c90daf67b984c05d25c03a1

PR #577, opened 2020-12-14
Add support for disconnection testing on volumes with more than 3 mayastor nodes.

chriswldenyer:
Ensure node to isolate is one hosting a replica for the volume.
Disable tests by using "test -c" to ensure they are at least compiled.
More detailed test descriptions.

### DIFF
--- a/mayastor-test/e2e/common/util.go
+++ b/mayastor-test/e2e/common/util.go
@@ -46,8 +46,9 @@ func DeleteDeployYaml(filename string) {
 
 // Status part of the mayastor volume CRD
 type MayastorVolStatus struct {
-	State string
-	Node  string
+	State    string
+	Node     string
+	Replicas []string
 }
 
 func GetMSV(uuid string) *MayastorVolStatus {
@@ -74,6 +75,9 @@ func GetMSV(uuid string) *MayastorVolStatus {
 		return nil
 	}
 	msVol := MayastorVolStatus{}
+
+	msVol.Replicas = make([]string, 0, 4)
+
 	v := reflect.ValueOf(status)
 	if v.Kind() == reflect.Map {
 		for _, key := range v.MapKeys() {
@@ -82,18 +86,33 @@ func GetMSV(uuid string) *MayastorVolStatus {
 			switch sKey {
 			case "state":
 				msVol.State = val.Interface().(string)
-				break
 			case "nexus":
 				nexusInt := val.Interface().(map[string]interface{})
 				if node, ok := nexusInt["node"].(string); ok {
 					msVol.Node = node
 				}
-				Expect(msVol.Node != "")
-				break
+			case "replicas":
+				replicas := val.Interface().([]interface{})
+				for _, replica := range replicas {
+					replicaMap := reflect.ValueOf(replica)
+					if replicaMap.Kind() == reflect.Map {
+						for _, field := range replicaMap.MapKeys() {
+							switch field.Interface().(string) {
+							case "node":
+								value := replicaMap.MapIndex(field)
+								msVol.Replicas = append(msVol.Replicas, value.Interface().(string))
+							}
+						}
+					}
+				}
 			}
 		}
+		// Note: msVol.Node can be unassigned here if the volume is not mounted
+		Expect(msVol.State).NotTo(Equal(""))
+		Expect(len(msVol.Replicas)).To(BeNumerically(">", 0))
+		return &msVol
 	}
-	return &msVol
+	return nil
 }
 
 // Check for a deleted Mayastor Volume,
@@ -197,11 +216,12 @@ func GetMsvState(uuid string) string {
 	return fmt.Sprintf("%s", msv.State)
 }
 
-// Retrieve the nexus node hosting the Mayastor Volume
-func GetMsvNode(uuid string) string {
+// Retrieve the nexus node hosting the Mayastor Volume,
+// and the names of the replica nodes
+func GetMsvNodes(uuid string) (string, []string) {
 	msv := GetMSV(uuid)
 	Expect(msv).ToNot(BeNil())
-	return fmt.Sprintf("%s", msv.Node)
+	return msv.Node, msv.Replicas
 }
 
 // Create a PVC and verify that
--- a/mayastor-test/e2e/node_disconnect/iscsi_drop/node_disconnect_iscsi_drop_test.go
+++ b/mayastor-test/e2e/node_disconnect/iscsi_drop/node_disconnect_iscsi_drop_test.go
@@ -30,7 +30,7 @@ func TestNodeLoss(t *testing.T) {
 }
 
 var _ = Describe("Mayastor node loss test", func() {
-	It("should verify behaviour when a node becomes inaccessible", func() {
+	It("should verify iscsi nexus behaviour when a node becomes inaccessible (iptables DROP)", func() {
 		lossTest()
 	})
 })
--- a/mayastor-test/e2e/node_disconnect/iscsi_reject/node_disconnect_iscsi_reject_test.go
+++ b/mayastor-test/e2e/node_disconnect/iscsi_reject/node_disconnect_iscsi_reject_test.go
@@ -30,7 +30,7 @@ func TestNodeLoss(t *testing.T) {
 }
 
 var _ = Describe("Mayastor node loss test", func() {
-	It("should verify behaviour when a node becomes inaccessible", func() {
+	It("should verify iscsi nexus behaviour when a node becomes inaccessible (iptables REJECT)", func() {
 		lossTest()
 	})
 })
--- a/mayastor-test/e2e/node_disconnect/iscsi_reject_idle/node_disconnect_iscsi_reject_idle_test.go
+++ b/mayastor-test/e2e/node_disconnect/iscsi_reject_idle/node_disconnect_iscsi_reject_idle_test.go
@@ -31,7 +31,7 @@ func TestNodeLoss(t *testing.T) {
 }
 
 var _ = Describe("Mayastor node loss test", func() {
-	It("should verify behaviour when a node becomes inaccessible when no IO is received", func() {
+	It("should verify iscsi nexus behaviour when a node becomes inaccessible when no IO is received (iptables REJECT)", func() {
 		lossTest()
 	})
 })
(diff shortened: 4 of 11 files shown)


paulyoong (approved):

mayastor-test/e2e/common/util.go
```diff
@@ -90,6 +94,22 @@ func GetMSV(uuid string) *MayastorVolStatus {
...
+								msVol.Replicas = append(msVol.Replicas, value.Interface().(string))
+								break
+							}
+						}
+					}
+				}
+				break
 			}
 		}
 	}
```
arne-rusek, line 115: The way I understand we count replicas is everything is replica so 1 replica is minimum (the one that nexus is running on and is attached as `bdev://`), isn't it? Shouldn't we test that we have at least one replica otherwise something is wrong? I.e. `Expect(len(msVol.Replicas)>0)`

mayastor-test/e2e/node_disconnect/iscsi_reject/node_disconnect_iscsi_reject_test.go
```diff
@@ -30,7 +30,7 @@ func TestNodeLoss(t *testing.T) {
 }
 
 var _ = Describe("Mayastor node loss test", func() {
-	It("should verify behaviour when a node becomes inaccessible", func() {
+	It("should verify iscsi nexus behaviour when a node becomes inaccessible", func() {
```
arne-rusek, line 33: I love how you rewritten this (and above) to be more specific. I with devops/sre hat on regard dropped connection as "default" behavior - something went wrong on the network - but sure REJECT happens when something is misconfigured - which I'm guessing we're testing here (also filename suggest this). 

Thus I'm suggesting being explicit here too:
```suggestion
	It("should verify iscsi nexus behaviour when a node becomes inaccessible (iptables REJECT)", func() {
```

mayastor-test/e2e/node_disconnect/iscsi_reject_idle/node_disconnect_iscsi_reject_idle_test.go
```diff
@@ -31,7 +31,7 @@ func TestNodeLoss(t *testing.T) {
 }
 
 var _ = Describe("Mayastor node loss test", func() {
-	It("should verify behaviour when a node becomes inaccessible when no IO is received", func() {
+	It("should verify iscsi nexus behaviour when a node becomes inaccessible when no IO is received", func() {
```
arne-rusek, line 34:
```suggestion
	It("should verify iscsi nexus behaviour when a node becomes inaccessible when no IO is received (iptables REJECT)", func() {
```

mayastor-test/e2e/node_disconnect/iscsi_drop/node_disconnect_iscsi_drop_test.go
```diff
@@ -30,7 +30,7 @@ func TestNodeLoss(t *testing.T) {
 }
 
 var _ = Describe("Mayastor node loss test", func() {
-	It("should verify behaviour when a node becomes inaccessible", func() {
+	It("should verify iscsi nexus behaviour when a node becomes inaccessible (dropped connection)", func() {
```
arne-rusek, line 33: Love how it is now explicit (and I'm expecting on failed test one doesn't have to go through code to get the details). :+1: I'm however a bit perplexed by `dropped connection`. The reason is we're running over TCP so it is just packets that are dropped (not the connection - it might even recover later when e.g. network problem we're testing here is fixed and e.g. packet loss stops). I found out that being explicit about what is being done - rather than the attempted outcome being tested - helps when test starts to fail as it is more explicit about how the environment is setup. What about:

```suggestion
	It("should verify iscsi nexus behaviour when a node becomes inaccessible (iptables DROP)", func() {
```

mayastor-test/e2e/node_disconnect/nvmf_drop/node_disconnect_nvmf_drop_test.go
```diff
@@ -30,7 +30,7 @@ func TestNodeLoss(t *testing.T) {
 }
 
 var _ = Describe("Mayastor node loss test", func() {
-	It("should verify behaviour when a node becomes inaccessible", func() {
+	It("should verify nvmf nexus behaviour when a node becomes inaccessible (dropped connection)", func() {
```
arne-rusek, line 33:
```suggestion
	It("should verify nvmf nexus behaviour when a node becomes inaccessible (iptables DROP)", func() {
```

mayastor-test/e2e/node_disconnect/nvmf_reject/node_disconnect_nvmf_reject_test.go
```diff
@@ -31,7 +31,7 @@ func TestNodeLoss(t *testing.T) {
 }
 
 var _ = Describe("Mayastor node loss test", func() {
-	It("should verify behaviour when a node becomes inaccessible", func() {
+	It("should verify nvmf nexus behaviour when a node becomes inaccessible", func() {
```
arne-rusek, line 34:
```suggestion
	It("should verify nvmf nexus behaviour when a node becomes inaccessible (iptables REJECT)", func() {
```

mayastor-test/e2e/node_disconnect/nvmf_reject_idle/node_disconnect_nvmf_reject_idle_test.go
```diff
@@ -31,7 +31,7 @@ func TestNodeLoss(t *testing.T) {
 }
 
 var _ = Describe("Mayastor node loss test", func() {
-	It("should verify behaviour when a node becomes inaccessible when no IO is received", func() {
+	It("should verify nvmf nexus behaviour when a node becomes inaccessible when no IO is received", func() {
```
arne-rusek, line 34:
```suggestion
	It("should verify nvmf nexus behaviour when a node becomes inaccessible when no IO is received (iptables REJECT)", func() {
```

mayastor-test/e2e/node_disconnect/lib/node_disconnect_lib.go
```diff
@@ -105,8 +105,6 @@ func LossWhenIdleTest(nodeToIsolate string, otherNodes []string, disconnectionMe
 	DisconnectNode(nodeToIsolate, otherNodes, disconnectionMethod)
 
 	fmt.Printf("waiting up to 90s for disconnection to affect the nexus\n")
```
arne-rusek, line 107: A nitpick (not a functional problem), but I always try to address magic numbers. It's not a problem per-se, but I like to help readers of the code. I recommend putting 90 to a constant and using it throughout code. https://en.wikipedia.org/wiki/Magic_number_(programming)

mayastor-test/e2e/node_disconnect/test.sh
```diff
@@ -11,9 +11,9 @@ timeout=200
 (cd nvmf_reject_idle && go test -timeout "${timeout}s")
 (cd iscsi_reject_idle && go test -timeout "${timeout}s")
 
-# These tests currently fail
-# (cd nvmf_drop && go test -timeout "${timeout}s")
-# (cd iscsi_drop && go test -timeout "${timeout}s")
+# These two tests currently fail so are run with -c (compile only)
```
arne-rusek, line 14: Sorry, almost fixed.

arne-rusek (changes requested):

mayastor-test/e2e/common/util.go
```diff
@@ -74,6 +75,9 @@ func GetMSV(uuid string) *MayastorVolStatus {
 		return nil
 	}
 	msVol := MayastorVolStatus{}
+
+	msVol.Replicas = make([]string, 0, 4)
+
 	v := reflect.ValueOf(status)
 	if v.Kind() == reflect.Map {
```
arne-rusek, line 82: I'm not sure if this is expected, but it feels that if for any reason (unless impossible, and if it is impossible I don't understand the if) we return `msVol` filled with default values if `v.Kind != reflect.Map`. And when calling GetMSV we test it for `!= nil` and that test will succeed.

I know this isn't part of the CR, but I'd like to understand if it's a bug.

mayastor-test/e2e/common/util.go
chriswldenyer: Yes, that sounds reasonable. It could also catch a problem if the MSV definition changes so that replicas are no longer found.

mayastor-test/e2e/common/util.go
```diff
@@ -90,6 +94,22 @@ func GetMSV(uuid string) *MayastorVolStatus {
...
+				replicas := val.Interface().([]interface{})
+				for _, replica := range replicas {
+					replicaMap := reflect.ValueOf(replica)
+					if replicaMap.Kind() == reflect.Map {
+						for _, field := range replicaMap.MapKeys() {
+							switch field.Interface().(string) {
+							case "node":
+								value := replicaMap.MapIndex(field)
+								msVol.Replicas = append(msVol.Replicas, value.Interface().(string))
+								break
```
arne-rusek, line 107: Is this break really needed? It breaks out of innermost `switch` which looks like C, but isn't needed in golang AFAIK: https://golang.org/ref/spec#Break_statements

mayastor-test/e2e/node_disconnect/iscsi_reject/node_disconnect_iscsi_reject_test.go
chriswldenyer: Yes, happy to make these changes in the description.

mayastor-test/e2e/common/util.go
chriswldenyer: You are right, I'll fix that.

mayastor-test/e2e/node_disconnect/lib/node_disconnect_lib.go
chriswldenyer: Agreed, I'll make the change.

mayastor-test/e2e/common/util.go
chriswldenyer: Yes I think it is, we might hit it if the MSV changes again. I'll get it to return nil if v is not of the right type.

arne-rusek (approved):
LGTM

chriswldenyer:
bors merge

paulyoong (approved):

arne-rusek (approved):

chriswldenyer:
bors merge

PR #579, opened 2020-12-14
Add replica re-assignment e2e test. 

chriswldenyer:
This test causes a replica node to be faulted by IO packet rejection, and then waits for the control plane
to provision a new replica. The test requires 4 mayastor nodes.

PR #582, opened 2020-12-15
Rework directory structure of the tests.

jkryl:
NodeJS mayastor tests are in test/grpc directory. The name grpc
was chosen because they test gRPC API of mayastor.

E2E tests are in test/e2e directory.

Third-party CSI E2E tests are in test/csi-e2e directory.

I had to fix some lint errors in JS files because moving the file counts as modifying the file.


paulyoong (approved):

arne-rusek:
Please update also `.dockerignore` - i presume it should now contain `test`, doesn't it?

arne-rusek:
Risking broadening PR's scope maybe too much for you (sorry :sweat_smile: ), what about `composer`, would it make sense to move it as well to `test`? And maybe rename to something telling "what it does" compared to "how it does it"?

test/e2e/README.md
```diff
@@ -0,0 +1,50 @@
+# Pre-requisites
+
+The test doesn't yet manage the lifecycle of the cluster being tested,
+therefore the test hosts' kubeconfig must point to a Kubernetes cluster.
+You can verify that the kubeconfig is setup correctly simply with
+`kubectl get nodes`.
+
+The cluster under test must meet the following requirements:
+* Have 3 nodes
```
arne-rusek, line 9: I believe that's no longer true: https://github.com/openebs/Mayastor/pull/579/files#diff-04c6409df4a44eb62116a9030ccf6bdab282e29f0e57bdc0628f5261f002edbeR8 or at least won't be in the near future when the test is really run, what about raising the requirements right away (as I feel we might forget to update doc later):

```suggestion
* Have 4 nodes
```

test/e2e/README.md
```diff
@@ -0,0 +1,50 @@
...
+* Each node must be configured per the quick start:
+  * At least 512 2MiB hugepages available
```
arne-rusek, line 11: There were some discussions regarding this. I'm thinking more generic and maybe more correct would be:
```suggestion
  * At least 1GiB of hugepages available (e.g. 512 2MiB hugepages)
```

test/e2e/README.md
```diff
@@ -0,0 +1,50 @@
...
+  * Each node must be labelled for use by mayastor (ie "openebs.io/engine=mayastor")
+
+The test host must have the following installed:
+* go (>= v1.15)
+* ginkgo (tested with v1.2)
```
arne-rusek, line 16: :thinking: I did not install it and was able to run install+uninstall e2e test. Is it really required? It might be installed automagically (or I didn't run the test that really needs it).

test/e2e/README.md
```diff
@@ -0,0 +1,50 @@
...
+* helm
+
+# Setting up the cluster
+
+### Create a cluster in public cloud
+
+Use terraform script in `aws-kubeadm/` from
+[terraform repo](https://github.com/mayadata-io/mayastor-terraform-playground) to
+set up a cluster in AWS suitable for running the tests. The configuration file for
+terraform could look like this (replace docker registry value by yours):
```
arne-rusek, line 27:
```suggestion
terraform could look like this (replace docker registry value by yours - must be reachable from the cluster):
```

test/e2e/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+```
+cluster_name    = "my-e2e-test-cluster"
```
arne-rusek, line 30: As a good measure I always create testing resources so they can be easily identified. This is a small nudge to hint that, what do you think?

```suggestion
cluster_name    = "<NICKNAME>-e2e-test-cluster"
```

test/grpc/test_replica.js
```diff
@@ -24,10 +24,10 @@ const BASE_UUID = 'c35fa4dd-d527-4b7b-9cf0-436b8bb0ba7';
 const NVMF_URI = /^nvmf:\/\/(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}):\d{1,5}\/nqn.2019-05.io.openebs:/;
 
 // tunables of the test suite
-var endpoint = process.env.MAYASTOR_ENDPOINT;
-var disks = process.env.MAYASTOR_DISKS;
+let endpoint = process.env.MAYASTOR_ENDPOINT;
+let disks = process.env.MAYASTOR_DISKS;
```
arne-rusek, line 28: Should these be documented somewhere?

arne-rusek (changes requested):

test/e2e/README.md
jkryl: me neither. I think it is just a dependency that is automatically downloaded and installed when building the test code. I will remove that from README 👍

test/e2e/README.md
jkryl: 👍

test/e2e/README.md
jkryl: indeed 👍

test/grpc/test_replica.js
jkryl: that is a long history. Originally I designed mayastor grpc tests so that they can be run on any mayastor instance later on that proved unfeasible and no one runs the tests that way. Instead, the test suite always starts mayastor and provides necessary resources like disks. So it is a sort of a dead code or dead functionality that would be probably better to remove. But that's out of the scope of this PR.

test/e2e/README.md
jkryl: 👍

test/e2e/README.md
jkryl: thanks 👍

jkryl:
I agree it would make sense to put docker compose tests under the new /test directory. I will mention that on slack and we can do that any time later if the original authors of the code agree.

test/grpc/test_replica.js
arne-rusek: Could you please add a comment stating that?

```suggestion
// TODO: sort of a dead code or dead functionality that would be probably better to remove
// originally made to run tests on any mayastor instance, currently test suite always starts mayastor
// and provides necessary resources like disks
let endpoint = process.env.MAYASTOR_ENDPOINT;
let disks = process.env.MAYASTOR_DISKS;
```

arne-rusek (approved):
I've suggested one more change - documenting what you've told me in the comment. Sounds like important enough to put to the code for future generations. Otherwise LGTM

test/grpc/test_replica.js
jkryl: fixed, thanks

jkryl:
bors merge

jkryl:
bors retry

jkryl:
bors retry

PR #585, opened 2020-12-16
CAS-582: Set nexus state to degraded when adding replica

paulyoong:
When adding a replica to a nexus, ensure that the nexus state in the MSV
is set to degraded.


jkryl (approved):
Thanks!

gila (approved):

paulyoong:
bors merge

PR #587, opened 2020-12-17
Volume IO Tests CAS-505 CAS-506

blaisedias:
Remove redundant test in nvmf_vol


chriswldenyer (commented):
There is a problem with leaking PVCs on failure, which Jan also mentioned. My tests also have this issue. It also prevents any subsequent test creating a PVC with the same name if it still exists.

blaisedias:
> There is a problem with leaking PVCs on failure, which Jan also mentioned. My tests also have this issue. It also prevents any subsequent test creating a PVC with the same name if it still exists.

I have fixed this now.

test/e2e/basic_volume_io/basic_volume_io_test.go
```diff
@@ -0,0 +1,93 @@
+// JIRA: CAS-505
+// JIRA: CAS-506
```
paulyoong, line 2: I think we should remove this?

paulyoong (approved):

test/e2e/basic_volume_io/basic_volume_io_test.go
blaisedias: There was some discussion early on in which it was decided that we add this so for ease of tracking test results.

test/e2e/basic_volume_io/basic_volume_io_test.go
```diff
@@ -0,0 +1,93 @@
...
+	Eventually(func() bool {
+		return common.IsPodRunning(fioPodName)
+	},
+		defTimeoutSecs,
+		"1s",
+	).Should(Equal(true))
+
+	// Run the fio test
+	common.RunFio(fioPodName, 20)
+	podNames = podNames[:len(podNames)-1]
```
chriswldenyer, line 54: Shouldn't you do this after DeletePod() has succeeded?

jkryl (approved):

blaisedias:
bors try

chriswldenyer (approved):

blaisedias:
bors merge

PR #588, opened 2020-12-17
test: add basic rebuild E2E test (CAS-285)

paulyoong:
This test checks that a rebuild can be started and completes correctly.

Some helper functions have been added to the common directory and could
be used to refactor some of the existing code in the future.

I have purposely not added the test to the Jenkinsfile yet, as I want to wait for
the discussion on the E2E cleanup procedure to conclude. I will raise a separate
work item for this.

### DIFF
--- a/test/e2e/rebuild/basic_rebuild_test.go
+++ b/test/e2e/rebuild/basic_rebuild_test.go
@@ -1,0 +1,93 @@
+package basic_rebuild_test
+
+import (
+	"e2e-basic/common"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+var (
+	pvcName      = "rebuild-test-pvc"
+	storageClass = "mayastor-nvmf"
+)
+
+const ApplicationPod = "fio.yaml"
+
+func basicRebuildTest() {
+	// Create a PVC
+	common.MkPVC(pvcName, storageClass)
+	pvc, err := common.GetPVC(pvcName)
+	Expect(err).To(BeNil())
+	Expect(pvc).ToNot(BeNil())
+
+	timeout := "90s"
+	pollPeriod := "1s"
+
+	// Create an application pod and wait for the PVC to be bound to it.
+	common.ApplyDeployYaml(ApplicationPod)
+	Eventually(func() bool { return common.IsPvcBound(pvcName) }, timeout, pollPeriod).Should(Equal(true))
+
+	uuid := string(pvc.ObjectMeta.UID)
+	repl, err := common.GetNumReplicas(uuid)
+	Expect(err).To(BeNil())
+	Expect(repl).Should(Equal(int64(1)))
+
+	// Wait for volume to be published before adding a child.
+	// This ensures that a nexus exists when the child is added.
+	Eventually(func() bool { return common.IsVolumePublished(uuid) }, timeout, pollPeriod).Should(Equal(true))
+
+	// Add another child which should kick off a rebuild.
+	common.UpdateNumReplicas(uuid, 2)
+	repl, err = common.GetNumReplicas(uuid)
+	Expect(err).To(BeNil())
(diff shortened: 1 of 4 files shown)


test/e2e/rebuild/basic_rebuild_test.go
```diff
@@ -0,0 +1,99 @@
...
+})
+
+const ApplicationPod = "fio.yaml"
+
+// CreateDummyPod deploys a pod which mounts the PVC but doesn't do anything with it.
+// This is to allow a nexus entry to be created in the MSV (because a nexus is only created on volume publish).
+func CreateDummyPod() {
+	common.ApplyDeployYaml(ApplicationPod)
+	// Allow some time for the PVC to be mounted.
+	time.Sleep(3 * time.Second)
```
jkryl, line 93: Is there a way how we could test that the volume has been mounted without using fixed time intervals? I think we should have a utility function in common library that executes a callback with exponential backoff until successful. Just like we have in mocha tests. No idea how difficult it is to implement in golang. But I suppose it would have many use cases in other tests as well.

blaisedias: @jkryl  I think the idiom is `Eventually` for example in `pvc_stress` test we have
```
	// Wait for the PVC to be bound.
	Eventually(func() coreV1.PersistentVolumeClaimPhase {
		return Cmn.GetPvcStatusPhase(volName)
	},
		defTimeoutSecs, // timeout
		"1s",           // polling interval
	).Should(Equal(coreV1.ClaimBound))
```

test/e2e/rebuild/basic_rebuild_test.go
```diff
@@ -0,0 +1,99 @@
...
+var _ = Describe("Mayastor rebuild test", func() {
+	It("should run a rebuild job to completion", func() {
+		basicRebuildTest()
+	})
+})
+
+var _ = BeforeSuite(func(done Done) {
+	logf.SetLogger(zap.LoggerTo(GinkgoWriter, true))
+	common.SetupTestEnv()
+	common.MkPVC(pvcName, storageClass)
```
chriswldenyer, line 74: You might want to move MkPVC and CreateDummyPod out of BeforeSuite so that any errors they trigger will mean that AfterSuite gets called to clear up.

test/e2e/rebuild/basic_rebuild_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// Wait for the added child to show up.
+	Eventually(func() int { return common.GetNumChildren(uuid) }, timeout, pollPeriod).Should(BeEquivalentTo(2))
+
+	getChildrenFunc := func(uuid string) []common.NexusChild {
+		children, err := common.GetChildren(uuid)
+		if err != nil {
+			panic("Failed to get children")
+		}
+		Expect(len(children)).Should(BeEquivalentTo(2))
```
chriswldenyer, line 46: Could Equal() be used for ints instead of BeEquivalentTo() ?

test/e2e/rebuild/basic_rebuild_test.go
paulyoong: The code has been refactored a bit, but I have replaced the sleep with the check that @blaisedias suggested.

The sleep was required to ensure that the volume was published before adding a child. This is needed to address the race condition detailed [here](https://github.com/openebs/Mayastor/issues/590). However, after speaking with @jkryl I have included an additional function to check if a volume has been published. This checks for the existence of the "Target Nodes" field in the MSV. If this field exists, the volume is published.

test/e2e/rebuild/basic_rebuild_test.go
paulyoong: Done. The functionality for creating the PVC and creating the application pod has been moved to the test itself. The cleanup is still performed by the "AfterSuite" function.

test/e2e/rebuild/basic_rebuild_test.go
paulyoong: Done, thanks.

chriswldenyer (approved):
LGTM

jkryl (approved):

paulyoong:
bors merge

paulyoong:
bors merge

PR #589, opened 2020-12-17
Be able to test a specific tag of images in e2e tests

jkryl:


blaisedias (approved):
LGTM

Jenkinsfile
```diff
@@ -146,7 +152,7 @@ pipeline {
         sh 'nix-store --delete /nix/store/*docker-image*'
         withCredentials([file(credentialsId: 'kubeconfig', variable: 'KUBECONFIG')]) {
           sh 'kubectl get nodes -o wide'
-          sh "nix-shell --run './scripts/e2e-test.sh ${env.REGISTRY}'"
+          sh "nix-shell --run './scripts/e2e-test.sh -d /dev/nvme1n1 -t ${env.GIT_COMMIT_SHORT} -r ${env.REGISTRY}'"
```
arne-rusek, line 155:
```suggestion
          sh "nix-shell --run './scripts/e2e-test.sh --device /dev/nvme1n1 --tag \"${env.GIT_COMMIT_SHORT}\" --registry \"${env.REGISTRY}\"'"
```

Makes it easier for future readers. Not sure about quoting, do you think I did it right?

scripts/e2e-test.sh
```diff
@@ -3,18 +3,65 @@
...
+  --tag <name>              Docker image tag of mayastor images (default "ci")
+
+Examples:
+  $(basename $0) --registry 127.0.0.1:5000 --tag a80ce0c
+EOF
+}
+
+# Parse arguments
+while [ "$#" -gt 0 ]; do
+  case $1 in
```
arne-rusek, line 27: Please _always_ (there are exceptions but IMHO it's not worth using them, I'd just quote everything) quote shell variables. Operator typos and/or spaces can have serious unpredictable consequences: https://github.com/MrMEEE/bumblebee-Old-and-abbandoned/issues/123 (similar bug was in some ubuntu pkg, I cannot find it now, but it was because of space in variable) We need the shellcheck ci :smile: but please run it for the time being manually.

scripts/e2e-test.sh
```diff
@@ -3,18 +3,65 @@
...
+  cat <<EOF
+Usage: $(basename $0) [OPTIONS]
+
+Options:
+  --device <path>           Device path to use for storage pools.
+  --registry <host[:port]>  Registry to pull the mayastor images from.
+  --tag <name>              Docker image tag of mayastor images (default "ci")
+
+Examples:
+  $(basename $0) --registry 127.0.0.1:5000 --tag a80ce0c
```
arne-rusek, line 21: FYI I tend to use "$0" as command then can be copy-pasted and will work. YMMW.

Nice to have help btw :+1:

scripts/e2e-test.sh
```diff
@@ -3,18 +3,65 @@
...
+      shift
+      REGISTRY=$1
+      shift
+      ;;
+    -t|--tag)
+      shift
+      TAG=$1
+      shift
+      ;;
+    *)
```
arne-rusek, line 43:
```suggestion
    -h|--help)
      help
      exit 0
    *)
```

For the sake of completeness...

scripts/e2e-test.sh
```diff
@@ -3,18 +3,65 @@
...
+      echo "Unknown option: $1"
+      help
+      exit 1
+      ;;
+  esac
+done
+
+if [ -z "$DEVICE" ]; then
```
arne-rusek, line 51: Readability/maintainability note. I would not rename (i.e. keep the _spelling_) variables throughout the code. I.e. why not use `e2e_pool_device` here? It helps while `grep`-ing through sources as one can see the "path" the value takes throughout system and lowers the cognitive load a little. Please consider not renaming variables between "layers".

test/e2e/install/install_test.go
```diff
@@ -86,9 +86,13 @@ func getTestClusterDetails() (string, int, []string, error) {
 	// Redundant check, but keep it anyway, we are writing a test after all.
 	// We should have found at least one node!
 	if len(nodeIPs) == 0 {
-		return "", 0, nil, errors.New("no usable nodes found")
+		return "", "", 0, nil, errors.New("no usable nodes found")
 	}
 
+	tag := os.Getenv("e2e_image_tag")
```
arne-rusek, line 92: hmm.. follow up to readability/maintainability above -- I see ... well... Also golang's CamelCase doesn't help. well... do as you see fit.

test/e2e/install/install_test.go
```diff
@@ -39,12 +39,12 @@ var testEnv *envtest.Environment
 /// The assumption is that the test-registry is accessible via the IP addr of the master,
 /// or any node in the cluster if the master noe does not exist
 /// TODO Refine how we workout the address of the test-registry
-func getTestClusterDetails() (string, int, []string, error) {
+func getTestClusterDetails() (string, string, int, []string, error) {
```
arne-rusek, line 42: This is getting a bit ridiculous. I would write please consider using struct, but it would probably require doing that all around the code :frowning_face: I think it makes sense to do it at once.

test/e2e/install/install_test.go
```diff
@@ -127,22 +131,13 @@ func makeImageName(registryAddress string, imagename string, imageversion string
 	return registryAddress + "/mayadata/" + imagename + ":" + imageversion
 }
 
-func generateYamls(registryAddress string) {
-	bashcmd := "../../../scripts/generate-deploy-yamls.sh  ci " + registryAddress
+func generateYamls(imageTag string, registryAddress string) {
+	bashcmd := fmt.Sprintf("../../../scripts/generate-deploy-yamls.sh %s %s", imageTag, registryAddress)
```
arne-rusek, line 135: thank you

test/e2e/install/install_test.go
```diff
@@ -127,22 +131,13 @@ func makeImageName(registryAddress string, imagename string, imageversion string
...
 	cmd := exec.Command("bash", "-c", bashcmd)
 	out, err := cmd.CombinedOutput()
 	Expect(err).ToNot(HaveOccurred(), "%s", out)
 }
 
-func applyTemplatedYaml(filename string, imagename string, registryAddress string) {
```
arne-rusek, line 137: :yaay:

Jenkinsfile
```diff
@@ -137,6 +137,12 @@ pipeline {
     }
     stage('e2e tests') {
       agent { label 'nixos-mayastor' }
+      environment {
+        GIT_COMMIT_SHORT = sh(
+          script: "printf \$(git rev-parse --short ${GIT_COMMIT})",
```
arne-rusek, line 142: I was going to ask why the `printf` but that's to get rid of whitespace right? What about comment for dummies like me?
```suggestion
          // using printf to get rid of trailing newline
          script: "printf \$(git rev-parse --short ${GIT_COMMIT})",
```

arne-rusek (changes requested):

test/e2e/install/install_test.go
jkryl: there is also ongoing discussion about using some kind of configuration file instead of env variables. Then we would parse yaml/json into a structure and use it internally in the code. But I'm golang noob hence doing only minimal changes and leaving bigger refactorings to others :-)

scripts/e2e-test.sh
jkryl: agree. this was taken from another script of ours.

scripts/e2e-test.sh
jkryl: 👍

scripts/e2e-test.sh
jkryl: 👍

scripts/e2e-test.sh
jkryl: I'm not sure. variables used by go code are rather long and the current names correspond with the name of shell script arguments which is also good for the reader who tries to understand the code. If we rename them then we should convert e2e variable to uppercase and make sure that empty string is handled equally as undefined variable which I felt is a bit out of scope of this change.

test/e2e/install/install_test.go
jkryl: and more importantly it should be uppercase. But then we should change all e2e variables.

scripts/e2e-test.sh
arne-rusek: TL;DR - I'm resolving the comment as keeping code style isn't really solvable in this one comment :-) Feel free to ignore. I will still provide some thoughts I have around this mainly for the case if you see an error in reasoning and would be willing to provide me with some feedback.

> variables used by go code are rather long; correspond with the name of shell script

I've noticed that envvars are renamed once again inside the go code... back to tag, registry etc go variables. So I would say the whole thing is a problem of env-vars or rather using env-vars instead of cmdline arguments in the _go code_ in the same way as the script does it :+1: (or config). But given the current state if if one searches through env it is nice to see that these random variables are connected to e2e using the prefix that's right (similar to e.g. `MAKE_FLAGS`). Thus I'd say correct solution is to use cmdline args not env vars. Out-of scope of this MR, I agree. And a total nitpick.

>  If we rename them then we should convert e2e variable to uppercase

Well... They should have been uppercase from the start, shouldn't they? But it's kinda random (mostly using uppercase for variables, not constants in shell is common). FYI when I'm in doubt, I tend to fall back to nicely written google style guides: https://google.github.io/styleguide/shellguide.html#s7.2-variable-names https://google.github.io/styleguide/shellguide.html#s7.5-read-only-variables ... And that's how I understand it - constants - which here they are - and should have been uppercase from the start. And variables should be lowercase. But I don't see much value here, unless they're really declared with `readonly <VAR_NAME>` Anyway...

> and make sure that empty string is handled equally as undefined variable which I felt is a bit out of scope of this change

I don't think I follow. Unless you specify `set -u` undefined variable behave as empty ones AFAIK...

test/e2e/install/install_test.go
arne-rusek: Yeah, I don't think this is about config file (you should parse e.g. json to pre-defined structures in golang otherwise you'll get into stuff like `config.(map[string]interface{})["key1"].(map[string]interface{})["key2"].(string)` which you do not want to, believe me :-) But I agree with the reasoning, actually Chris was kind and did some refactoring in different mr merging this to a struct. But I think I now understand your motivation behind the config file (which I did not really).

test/e2e/install/install_test.go
arne-rusek: Indeed. As discussed above constants should be uppercase (which these are/should be), variables (in the shell) should not. But it's and agreement/style, not really meaning too much and one can use any case for shell variables (For me consistency across layers is much more important and that can be solved by `-i` to grep and code-search tools often have "case insensitive" search).

arne-rusek (approved):
LGTM

scripts/e2e-test.sh
jkryl: agree. This whole problem of passing arguments between a caller and the test suites will need to be probably revisited at some point in the future. In fact, currently, it's just the install test suite that requires arguments. But that might change as we add more e2e tests in the future.

re undefined vs empty string: I forgot that there is no difference between that in os.getenv() in go. Please, ignore that comment.

arne-rusek: Oh right. In golang too (if you dig into it there's`LookupEnv` that can distinguish between unset & empty however building any non-trivial functionality on that distinction is evil :smile:). BTW I really meant in shell, there's no real difference between empty and undefined in scripts (without `set -u`, or unless you again dig deeper with e.g. `env`).

jkryl:
bors merge

PR #591, opened 2020-12-17
Do not overwrite versioned files in e2e tests CAS-595

arne-rusek:
Deployment yaml files are now generated to `./test-yamls` which is `.gitignore`-d to avoid polluting developer's git.

### DIFF
--- a/test/e2e/install/install_test.go
+++ b/test/e2e/install/install_test.go
@@ -128,7 +128,7 @@ func makeImageName(registryAddress string, imagename string, imageversion string
 }
 
 func generateYamls(registryAddress string) {
-	bashcmd := "../../../scripts/generate-deploy-yamls.sh  ci " + registryAddress
+	bashcmd := "../../../scripts/generate-deploy-yamls.sh -t ../../../test-yamls ci " + registryAddress
 	cmd := exec.Command("bash", "-c", bashcmd)
 	out, err := cmd.CombinedOutput()
 	Expect(err).ToNot(HaveOccurred(), "%s", out)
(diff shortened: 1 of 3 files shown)


jkryl (approved):

test/e2e/install/install_test.go
```diff
@@ -207,15 +207,16 @@ func installMayastor() {
 
 	fmt.Printf("registry address %v, number of mayastor instances=%v\n", registryAddress, numMayastorInstances)
 
+	// FIXME use absolute paths, do not depend on CWD
```
bobek, line 210: Yes, please, I would suggest doing it even now, with something like

```
MAIN_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```

at the beginning and referencing everything from that.

bobek (changes requested):

blaisedias (approved):
There is a simpler way to do this which is to do `kubectl apply -f - ` and then pipe in the contents.
That way there is no need to mess around with temporary directories.
But this is good enough.

bobek:
> There is a simpler way to do this which is to do `kubectl apply -f - ` and then pipe in the contents.
> That way there is no need to mess around with temporary directories.

Good point, I thought, that this is on purpose -- so you have the artifacts generated in case you need to inspect them. Maybe reapply etc.

bobek (approved):
We have agreed, that requested changes are going to happen later. @arne-rusek will create an appropriate ticket to keep track of it.

test/e2e/install/install_test.go
arne-rusek: We've agreed to address this throughout whole tests in a later PR.

arne-rusek:
> > There is a simpler way to do this which is to do `kubectl apply -f - ` and then pipe in the contents.
> > That way there is no need to mess around with temporary directories.
> 
> Good point, I thought, that this is on purpose -- so you have the artifacts generated in case you need to inspect them. Maybe reapply etc.

I do like artifacts :smile: Often makes debugging much easier. I don't see a way to tell helm to generate just single yaml and pipe it :frowning_face:

arne-rusek:
bors merge

arne-rusek:
bors retry

arne-rusek:
bors retry

PR #594, opened 2020-12-18
Run CSI e2e tests in CI/CD pipeline CAS-577

blaisedias:
Refactor scripts/e2e-test.sh to
 * Run the CSI tests
 * Be more developer friendly

1. Add csi to the set of tests run by default
2. The set of tests to be run can now be specified
   using env var e2e_tests
3. Run uninstall test only if specified,
    and always after all other tests have
    run successfully or not.

### DIFF
--- a/scripts/e2e-test.sh
+++ b/scripts/e2e-test.sh
@@ -3,10 +3,11 @@
 set -eux
 
 SCRIPTDIR=$(dirname "$(realpath "$0")")
-TESTS="install basic_volume_io"
+TESTS=${e2e_tests:-"install basic_volume_io csi uninstall"}
 DEVICE=
 REGISTRY=
 TAG=
+TESTDIR=$(realpath "$SCRIPTDIR/../test/e2e")
 
 help() {
   cat <<EOF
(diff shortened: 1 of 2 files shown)


arne-rusek:
This seem to be conflicting with https://github.com/openebs/Mayastor/pull/597/

Let's discuss on slack what to do as it seems they're doing both very similar things.

arne-rusek (changes requested):
This seem to be conflicting with https://github.com/openebs/Mayastor/pull/597/

Let's discuss on slack what to do as it seems they're doing both very similar things.

jkryl:
I like:
* being able to specify what tests to run
* having shell script function to run a test so we don't need to repeat the same parameters everywhere
what I don't like that much is:
* introducing new env variables instead of command line options - confusing if the tool is configured using both ways
* quite complicated logic iterating over tests skipping uninstall and then iterating again running only uninstall - we could have simply just a list of tests to run

but as Arne suggested it's probably better to discuss that in the chat after the e2e nightly PR is merged.

blaisedias:
bors try

blaisedias:
Superseded by https://github.com/openebs/Mayastor/pull/595

PR #596, opened 2020-12-21
jenkins worker is running out of space

jkryl:
Jenkins worker was running out of space because docker images were not
pruned. I decided to put all space reclaiming into a single script that
is run before each e2e stage. So nix GC is no longer run as a service
on worker nodes. The advantage is that we can synchronize GC with running
jobs that way.

### DIFF
--- a/scripts/reclaim-space.sh
+++ b/scripts/reclaim-space.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env sh
+
+#
+# The script tries to free as much space as possible by removing nix packages
+# and docker images that aren't used.
+#
+
+set -e
+
+MIN_FREE_GIB=$1
+
+get_avail_gib() {
+    echo $(( $(df --output=avail / | awk 'NR == 2 { print $1 }' ) / 1024 / 1024 ))
+}
+
+free=$(get_avail_gib)
+echo "Available space in root partition: $free GiB"
+
+if [ -n "$MIN_FREE_GIB" ]; then
+    if [ "$free" -gt "$MIN_FREE_GIB" ]; then
+        exit 0
+    fi
+fi
+
+set -x
+nix-collect-garbage
+docker prune --force --all
+set +x
+
+echo "Available space after cleanup: $(get_avail_gib) GiB"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -145,15 +145,25 @@ pipeline {
         )
       }
       steps {
+        // e2e tests are the most demanding step for space on the disk so we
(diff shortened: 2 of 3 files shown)


scripts/reclaim-space.sh
```diff
@@ -0,0 +1,30 @@
...
+# The script tries to free as much space as possible by removing nix packages
+# and docker images that aren't used.
+#
+
+set -e
+
+MIN_FREE_GB=$1
+
+get_avail_gb() {
+    echo $(( $(df --output=avail / | awk 'NR == 2 { print $1 }' ) / 1024 / 1024 ))
```
bobek, line 13:
```suggestion
    echo $(( $(df --output=avail / | awk 'NR == 2 { print $1 }' ) / 1000 / 1000 ))
```

You are using `GB` throughout the script. So either this or change `GB` to `GiB` everywhere (ideally in the name of the method as well).

bobek (changes requested):

scripts/reclaim-space.sh
arne-rusek: Please use GiB and 1024.

scripts/reclaim-space.sh
```diff
@@ -0,0 +1,30 @@
...
+echo "Available space in root partition: $free GB"
+
+if [ -n "$MIN_FREE_GB" ]; then
+    if [ "$free" -gt "$MIN_FREE_GB" ]; then
+        exit 0
+    fi
+fi
+
+set -x
+docker prune --force --all
```
arne-rusek, line 26: Consider adding e.g. `--filter until=24h` that would remove stuff older than e.g. 24h. I'm totally oddballing here, but there might be images like busybox, base images etc... that would be re-downloaded right away anyway? So this would do some speedup costing some disk space that would most probably be used right away anyway? Maybe even different filter (longer?shorter?)

jkryl: the dockerd is used in a different way here. Basically it could be completely avoided if we used for example skopeo to upload images to the registry. The only purpose of docker is to load the archive created by nix and then push the images to registry. It does not contain any other images than mayastor images.

arne-rusek: I see.

arne-rusek: BTW, isn't better to run `docker pune --force --all` *after* the image upload then?

jkryl: good point

bobek (approved):

arne-rusek (approved):

jkryl:
bors merge

PR #597, opened 2020-12-21
ci: support for running nightly e2e tests

jkryl:
New variable in e2e-test.sh script is introduced that is expected
to contain names of all e2e tests that should run nightly in the
future.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -54,6 +54,8 @@ if (currentBuild.getBuildCauses('jenkins.branch.BranchIndexingCause') &&
 
 // Only schedule regular builds on develop branch, so we don't need to guard against it
 String cron_schedule = BRANCH_NAME == "develop" ? "0 2 * * *" : ""
+// Some long e2e tests are not suitable to be run for each PR
+boolean run_extended_e2e_tests = (env.BRANCH_NAME != 'staging' && env.BRANCH_NAME != 'trying') ? true : false
 
 pipeline {
   agent none
@@ -206,7 +208,13 @@ pipeline {
                     fingerprintArtifacts: true
                 )
                 sh 'kubectl get nodes -o wide'
-                sh "nix-shell --run './scripts/e2e-test.sh --device /dev/sdb --tag \"${env.GIT_COMMIT_SHORT}\" --registry \"${env.REGISTRY}\"'"
+                script {
+                  def cmd = "./scripts/e2e-test.sh --device /dev/sdb --tag \"${env.GIT_COMMIT_SHORT}\" --registry \"${env.REGISTRY}\""
+                  if (run_extended_e2e_tests) {
+                    cmd = cmd + " --extended"
+                  }
+                  sh "nix-shell --run '${cmd}'"
+                }
               }
               post {
                 failure {
--- a/scripts/e2e-test.sh
+++ b/scripts/e2e-test.sh
@@ -9,12 +9,17 @@ REPORTSDIR=$(realpath "$SCRIPTDIR/..")
 # List and Sequence of tests.
 #tests="install basic_volume_io csi replica rebuild node_disconnect/replica_pod_remove uninstall"
 # Restrictions:
-#   1. resource_check MUST follow csi 
+#   1. resource_check MUST follow csi
 #       resource_check is a follow up check for the 3rd party CSI test suite.
 #   2. replicas_pod_remove SHOULD be the last test before uninstall
 #       this is a disruptive test.
-tests="install basic_volume_io csi resource_check uninstall"
+#TESTS="install basic_volume_io csi replica rebuild node_disconnect/replica_pod_remove uninstall"
+TESTS="install basic_volume_io csi resource_check uninstall"
+EXTENDED_TESTS=""
 
+# Global state variables
+tests=""
+run_extended_tests=
 device=
 registry=
 tag="ci"
@@ -34,6 +39,7 @@ Options:
   --tests <list of tests>   Lists of tests to run, delimited by spaces (default: "$tests")
         Note: the last 2 tests should be (if they are to be run)
              node_disconnect/replica_pod_remove uninstall
+  --extended                Run long running tests also.
   --reportsdir <path>       Path to use for junit xml test reports (default: repo root)
   --logs                    Generate logs and cluster state dump at the end of successful test run,
                             prior to uninstall.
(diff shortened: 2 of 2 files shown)


Jenkinsfile
```diff
@@ -155,7 +155,13 @@ pipeline {
         sh "./scripts/release.sh --alias-tag ci --registry ${env.REGISTRY}"
         withCredentials([file(credentialsId: 'kubeconfig', variable: 'KUBECONFIG')]) {
           sh 'kubectl get nodes -o wide'
-          sh "nix-shell --run './scripts/e2e-test.sh --device /dev/nvme1n1 --tag \"${env.GIT_COMMIT_SHORT}\" --registry \"${env.REGISTRY}\"'"
+          script {
+            def cmd = "./scripts/e2e-test.sh --device /dev/nvme1n1 --tag \"${env.GIT_COMMIT_SHORT}\" --registry \"${env.REGISTRY}\""
```
gila, line 159: Are you sure this will always be the device you think it is?

scripts/e2e-test.sh
```diff
@@ -75,6 +82,16 @@ for dir in $TESTS; do
   fi
 done
 
+if [ -z "$test_failed" ] && [ -n "$RUN_LONG_TESTS" ]; then
+  for dir in $LONG_TESTS; do
+    cd "$SCRIPTDIR/../test/e2e/$dir"
+    if ! go test -v . -ginkgo.v -ginkgo.progress -timeout 0 ; then
```
arne-rusek, line 88: Does `ginko.progress` play well with redirected stdout & jenkins? Does it make sense here?

arne-rusek (changes requested):

Jenkinsfile
jkryl: Right. For now it is a safe choice. When we test multiple clouds and different devices all of this will need to get changed so we will figure that out when we get there 👍

scripts/e2e-test.sh
jkryl: Good point. I have looked into the jenkins log from previous runs and the output seems to look ok. So I think it can stay if it makes the output better for interactive use case.

Jenkinsfile
```diff
@@ -155,7 +155,13 @@ pipeline {
...
+            if (env.BRANCH_NAME != 'staging' && env.BRANCH_NAME != 'trying') {
```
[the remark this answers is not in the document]
jkryl: will do. thanks 👍

arne-rusek (approved):

blaisedias (approved):

jkryl:
bors merge

jkryl:
bors retry

jkryl:
bors retry

jkryl:
bors merge

jkryl:
bors retry

jkryl:
bors merge

jkryl:
bors merge-

jkryl:
bors retry

jkryl:
bors merge

scripts/e2e-test.sh
```diff
@@ -4,8 +4,11 @@ set -eux
 
 SCRIPTDIR=$(dirname "$(realpath "$0")")
 # new tests should be added before the replica_pod_remove test
-#TESTS="install basic_volume_io csi replica rebuild node_disconnect/replica_pod_remove uninstall"
-TESTS="install basic_volume_io csi uninstall"
+#DEFAULT_TESTS="install basic_volume_io csi replica rebuild node_disconnect/replica_pod_remove uninstall"
+DEFAULT_TESTS="install basic_volume_io csi uninstall"
```
arne-rusek, line 8: I would rename the var to something like basic and extended tests. It sorta feels wrong, but I'm not a native speaker.
```suggestion
# DEFAULT_TESTS are run for each PR, LONG_TESTS only daily (see --long option, check Jenkinsfile for current schedule)
DEFAULT_TESTS="install basic_volume_io csi uninstall"
```

scripts/e2e-test.sh
```diff
@@ -4,8 +4,11 @@ set -eux
...
+LONG_TESTS=""
+TESTS=""
```
arne-rusek, line 10: Could you please rename below variables (lowercase) as they're not constants but rather global state and it's confusing to have them together with configuration constants like `DEFAULT_TESTS`. You can also use `readonly VAR=VAL` keyword to enforce. This block of vars (for an outsider) looks confusing.

I would also suggest:
```suggestion
# Global state variables
TESTS=""
```

or something. These glue scripts can get pretty confusing so I would keep them especially readable :-)

arne-rusek (approved):
I've suggested some stylistic changes, but LGTM

scripts/e2e-test.sh
jkryl: I would not make any assumption about how often the tests run or mention anything about PR testing here. That's a matter of jenkins configuration. This script provides a way to run e2e tests regardless if they run in Jenkins, on local computer or somewhere else. As for the name of the variable itself, if you agree, since this concern is more about how to name variables, I will defer to @blaisedias 's opinion if things are understandable enough as they are named now. That can be certainly changed and will do that if there is agreement on what should be the new names. @blaisedias  any opinion on this?

scripts/e2e-test.sh
jkryl: good point. I will convert all global variables used in this script that are neither exported nor constant to lower-case.

blaisedias: Whilst you are at it, probably best to name the variable 'test_list'

scripts/e2e-test.sh
blaisedias: I kind of prefer extended to long, but both names convey something useful about the set of tests being run.
If long... is used then it is natural to expect that short... is defined.
Whereas extension implies that normal set + some more, which is the case.
It just so happens that the extended set now has long running tests. 
There is a case to rename the directory test/e2e/nightly to test/e2e/long or test/e2e/extended for consistency.

blaisedias (approved):

scripts/e2e-test.sh
jkryl: I think we arrived to conclusion then. Let's use "extended" word then. I will update the PR 👍

jkryl: done

jkryl:
bors merge

Hoverbear (approved):

jkryl:
bors retry

jkryl:
bors retry

jkryl:
bors retry

jkryl:
bors retry

jkryl:
bors retry

jkryl:
bors retry

PR #598, opened 2020-12-23
 CAS-599: Use alpine image to run fio

arne-rusek:
nixery.dev isn't recommended for production use and alpine image is smaller

### DIFF
--- a/test/e2e/common/util.go
+++ b/test/e2e/common/util.go
@@ -294,8 +294,8 @@ func MkPVC(volName string, scName string) string {
 	Eventually(func() *MayastorVolStatus {
 		return GetMSV(string(pvc.ObjectMeta.UID))
 	},
-	defTimeoutSecs,
-	"1s",
+		defTimeoutSecs,
+		"1s",
 	).Should(Not(BeNil()))
 
 	return string(pvc.ObjectMeta.UID)
@@ -433,9 +433,8 @@ func CreateFioPodDef(podName string, volName string) *corev1.Pod {
 			Containers: []corev1.Container{
 				{
 					Name:    podName,
-					Image:   "nixery.dev/shell/fio/tini",
-					Command: []string{"tini", "--"},
-					Args:    []string{"sleep", "1000000"},
+					Image:   "alpine",
+					Command: []string{"sh", "-c", "apk add fio && sleep 1000000"},
 					VolumeMounts: []corev1.VolumeMount{
 						{
 							Name:      "ms-volume",
--- a/deploy/fio.yaml
+++ b/deploy/fio.yaml
@@ -9,11 +9,8 @@ spec:
        claimName: ms-volume-claim
   containers:
     - name: fio
-      image: nixery.dev/shell/fio/tini
-      command: [ "tini", "--" ]
-      args:
-        - sleep
-        - "1000000"
+      image: alpine
+      command: [ "sh", "-c", "apk add fio && sleep 1000000" ]
       volumeMounts:
         - mountPath: "/volume"
           name: ms-volume


gila:
This is not meant for production. Do you know why we use nixery to begin with?

test/e2e/common/util.go
```diff
@@ -433,9 +433,8 @@ func CreateFioPodDef(podName string, volName string) *corev1.Pod {
 			Containers: []corev1.Container{
 				{
 					Name:    podName,
-					Image:   "nixery.dev/shell/fio/tini",
-					Command: []string{"tini", "--"},
-					Args:    []string{"sleep", "1000000"},
+					Image:   "alpine",
```
gila, line 436: This looks horrible. Code changes required to make changes to containers? Why do we not read this from YAML?

gila (changes requested):

test/e2e/common/util.go
arne-rusek: Barking at the wrong tree here :-) I would tend to agree. What changes would you imagine? Also please understand scope of this (fixing currently existing problem) and if you want some refactoring (which I would definitely recommend and it already started on some places) please create a ticket with requirements. Not sure, if I understand correctly, but I cannot see much of a difference, if the configuration YAML would be hosted in the repository as well.

arne-rusek:
> This is not meant for production. Do you know why we use nixery to begin with?

Would you enlighten me please?

gila:
> > This is not meant for production. Do you know why we use nixery to begin with?
> 
> Would you enlighten me please?

It allows us to create any container content by changing the URI itself. So if we want to have fio and, I don't know let us say, sysstat we can change it to nixery/bash/fio/sysstat. 

With your change if have to (always) do apk update && apk install what-i-need and would always be stuck with older SW. With nixery we follow the most latest packages, which we need in a lot of cases because we use it during testing (which the YAML file is for)

test/e2e/common/util.go
gila: So you change fio.yaml in the repo, but you don't use that file in the test. That's basically what I was alluding too. So to me the change to fio.yaml not needed at all?

jkryl:
nixery is really cool. The only downside is that it is unstable. Outages are frequent and long (a couple of hours is not an exception). So if you do an interactive experiment or testing it is a great choice. If you rely on nixery in ci/cd pipeline then it is not so great. IMHO the change as it is now introduces two problems (correct me if I'm wrong):

* fio is installed after the container is created so if I exec fio before the installation is completed I will get an error
* having to install fio every time container is started makes it more sensitive to network problems
* when trying to terminate container with sleep it will take a very long time because sleep does not have a handler for sigterm (that's why `tiny` was used in previous version of the container).

Clearly we could create our own image with preinstalled fio and use it in ci/cd. We don't get the latest version of the fio then, but actually for the purpose of e2e tests I would argue that it is even desireable if fio version is not changing unexpectedly. But for the yaml file in deploy directory, I would probably continue to use the nixery as it's used interactively and the benefits outweigh the instability of nixery.

gila:
Yes, exactly. We should leave the YAML file as-is. The changes to the go code are fine with me. Also, if the deploy directory is assumed to be "part of ci-cd" then I'm not sure if that is a good idea because they were never intended for e2e tests. If we start doing that, then that's fine, but it somewhat changes the initial purpose of these files.

test/e2e/common/util.go
arne-rusek: I understood `fio.yaml` to be a file for the random user to test deploy fio when playing with mayastor. That what I've used when installing mayastor the first few times. It doesn't have connection to the tests but to the fact that when a random user from the internet wants to test it. She has a high chance that it will fails because of nixery and I regard it as a quality issue with the SW. Anyway YMMW.

arne-rusek:
@jkryl:
> nixery is really cool. The only downside is that it is unstable. Outages are frequent and long (a couple of hours is not an exception). So if you do an interactive experiment or testing it is a great choice. If you rely on nixery in ci/cd pipeline then it is not so great. IMHO the change as it is now introduces two problems (correct me if I'm wrong):

>  fio is installed after the container is created so if I exec fio before the installation is completed I will get an error
having to install fio every time container is started makes it more sensitive to network problems when trying to terminate container with sleep it will take a very long time because sleep does not have a handler for sigterm (that's why tiny was used in previous version of the container).

There's indeed a race which I didn't notice. CI-CD pipeline waits in a loop of 1s (IIRC) for the availability of the container so it will probably hit sooner or later. Even though alpine pkg install is really fast, networking issues might help the race to hit here. Good point. As for network problems in general, they can prevent pulling images as well, depends. But that's not something we could prevent and given the generic case I don't think it's clear what's better (pulling larger image vs installing small package - yes, images are often cached, but still I would dare to say which is better generally). As for `tini` I haven't realized that as well, good point, thanks.

> Clearly we could create our own image with preinstalled fio and use it in ci/cd. We don't get the latest version of the fio then, but actually for the purpose of e2e tests I would argue that it is even desireable if fio version is not changing unexpectedly. But for the yaml file in deploy directory, I would probably continue to use the nixery as it's used interactively and the benefits outweigh the instability of nixery.

I will update the ticket. We might build our own image (we have own registry anyway), or just cache nixery's one there. As for versioning, I know too little about NixOS branches and nixery to comment on that but Alpine is pretty fresh (and also this bleeding-edge vs. stable is not a debate for this PR, but I don't agree that latest is always greatest and I agree that version stability and avoiding unintended upgrades makes stuff generally more stable and dependable).

@gila:
> Yes, exactly. We should leave the YAML file as-is. The changes to the go code are fine with me. Also, if the deploy directory is assumed to be "part of ci-cd" then I'm not sure if that is a good idea because they were never intended for e2e tests. If we start doing that, then that's fine, but it somewhat changes the initial purpose of these files.

Again just re-iterating - `fio.yaml` is in our repo and it has a high chance to be used by users. And also high chance to fail due to nixery (clearly stated) not so good availability. Anyway I'm closing this issue due to @jkryl reported race condition which as all race conditions will hit sooner or later.

PR #601, opened 2020-12-30
Unschedule Mayastor pod to test replica loss

chriswldenyer:
Unschedule Mayastor pod to test replica loss

In order to avoid low-level network manipulation to fault a child,
the new test unschedules the corresponding mayastor replica pod,
so removing it from the IO path, to achieve the same effect.
This allows greater flexibility in where the test can be run
(e.g. CI testing) as the kubernetes API is all that is needed.
Slight refactoring of node-labelling code.


chriswldenyer:
bors try

chriswldenyer:
bors try

chriswldenyer:
Closing to try and fix commit message issues

PR #607, opened 2021-01-06
feat(poller): want poller abstraction

gila:
This change adds a safe abstraction around the poller interface. We
did not use pollers much (if not at all) but now that we need to
construct, among others our own qpairs we need to create them often
and frequently.

A small trampoline function (which is always inlined) is used to avoid
passing around raw pointers. The poller will execute a closure and allows
us to capture variables so there is no need for boxing arguments.

### DIFF
--- a/mayastor/src/core/poller.rs
+++ b/mayastor/src/core/poller.rs
@@ -1,0 +1,148 @@
+use std::{
+    ffi::{c_void, CString},
+    ptr::NonNull,
+    time::Duration,
+};
+
+use spdk_sys::{
+    spdk_poller,
+    spdk_poller_pause,
+    spdk_poller_register,
+    spdk_poller_register_named,
+    spdk_poller_resume,
+    spdk_poller_unregister,
+};
+
+/// structure holding our function and context
+struct PollCtx<'a>(Box<dyn FnMut() -> i32 + 'a>);
+
+/// indirection to avoid raw pointers at upper layers
+#[inline(always)]
+extern "C" fn _cb(ctx: *mut c_void) -> i32 {
+    let poll = unsafe { &mut *(ctx as *mut PollCtx) };
+    (poll.0)()
+}
+
+/// Poller structure that allows us to pause, stop, resume periodic tasks
+pub struct Poller<'a> {
+    inner: NonNull<spdk_poller>,
+    ctx: NonNull<PollCtx<'a>>,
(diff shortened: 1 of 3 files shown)


mayastor/src/core/poller.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub fn stop(mut self) {
+        unsafe {
+            spdk_poller_unregister(&mut self.inner.as_ptr());
+            Box::from_raw(self.ctx.as_ptr());
+            self.stopped = true;
+        }
+    }
+
+    /// pause the given poller
+    pub fn pause(&mut self) {
```
mtzaurus, line 45: Shall we disallow calling spdk_poller_pause() in case poller is stopped when 'self.stopped == true' ?
According to the code of spdk_poller_unregister(), it NULL's the poller (see https://github.com/spdk/spdk/blob/3c2190c214514d7b1d2fb2dd9ae8214dcb36c8c5/lib/thread/thread.c#L1144),
 
but 'spdk_poller_pause()' does not check the pointer against being non-NULL and just blindly dereferences it, which will cause NULL pointer dereference:
https://github.com/spdk/spdk/blob/3c2190c214514d7b1d2fb2dd9ae8214dcb36c8c5/lib/thread/thread.c#L1180

The same suggestion applies to resume(), which uses spdk_poller_resume() with the same behaviour.

gila: Good thinking, I was looking into that myself as well, but I do not think it is needed. Perhaps I'm overlooking so please correct me if I'm wrong. `stop(mut self)` will consume self so when I have this:
```rust
let p = poller::Builder::new().with_whatever().build();
// stop takes mut self after this call, and I can't use "p" any more
p.stop();
```

Right?

mtzaurus: Ah, yes. You are right. After stopping the controller it dissolves.

mtzaurus (approved):

mayastor/src/core/poller.rs
gila: I've updated the comments mike, to add the consuming explicitly.

tiagolobocastro (approved):
Might be worth having a test for a poller with a non 0 interval to make sure the time conversion is working

jkryl (approved):

PR #608, opened 2021-01-07
feat(js): use new version of grpc-uds with cached binary

jkryl:
0.1.6 has a cached prebuilt binary on github so it is downloaded if
available instead of having to build grpc library each time the module
is installed.

Unfortunately, with nix that does not work. More over mayastor tests
still depend on official node-grpc package so the benefit of this change
is not as anticipated. That said the following cases were improved:

1. when running moac unit tests, npm install will be much faster
2. when running grpc api tests, we have to build grpc just once (the
   official one) instead of twice as before.

### DIFF
--- a/csi/moac/node-packages.nix
+++ b/csi/moac/node-packages.nix
@@ -2020,13 +2020,13 @@ let
         sha512 = "4BBXHXb5OjjBh7luylu8vFqL6H6aPn/LeqpQaSBeRzO/Xv95wHW/WkU9TJRqaCTMZ5wq9jTSvlJWp0vRJy1pVA==";
       };
     };
-    "grpc-uds-0.1.4" = {
+    "grpc-uds-0.1.6" = {
       name = "grpc-uds";
       packageName = "grpc-uds";
-      version = "0.1.4";
+      version = "0.1.6";
       src = fetchurl {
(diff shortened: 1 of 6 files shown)


paulyoong (approved):

arne-rusek:
Not sure if it applies, but npm install seems to be running by default on one core. I have greatly sped it up on my machine by `export npm_config_jobs=$(nproc)` added to `scripts/moac-test.sh` and `scripts/grpc-test.sh`. Not sure it fits into this PR, but it can always be renamed to "speedup stuff" :-)

jkryl:
Good point Arne. The change is made by you in a separate PR so I think we are good 👍

gila (approved):

jkryl:
bors merge

arne-rusek:
> Not sure if it applies, but npm install seems to be running by default on one core. I have greatly sped it up on my machine by `export npm_config_jobs=$(nproc)` added to `scripts/moac-test.sh` and `scripts/grpc-test.sh`. Not sure it fits into this PR, but it can always be renamed to "speedup stuff" :-)

Done in https://github.com/openebs/Mayastor/pull/609

jkryl:
bors retry

jkryl:
bors merge-

jkryl:
bors merge

PR #609, opened 2021-01-07
perf(tests): set npm jobs to cpu count

arne-rusek:
Speed up installing npm packages by utilizing all available cores on a machine. I opted to put it to script rather than jenkins so development on local machine can do that too (nixos, rust, go compilation does it by default anyway, so it should produce no surprises).

If I missed some npm install please comment.

### DIFF
--- a/scripts/grpc-test.sh
+++ b/scripts/grpc-test.sh
@@ -1,11 +1,12 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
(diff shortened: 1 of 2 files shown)


jkryl (approved):
thanks!

gila (approved):

arne-rusek:
bors merge

arne-rusek:
bors merge

PR #610, opened 2021-01-07
fix(replica): delay adding replica (CAS-585)

paulyoong:
Adding a replica to an unpublished nexus was causing the child to be
added as online rather than degraded when the volume was subsequently
published. Because of this a rebuild was not run and data corruption
could occur when reading from the newly added child.

This change delays the addition of a replica until a volume is
published; ensuring the child is added as degraded and rebuilt
accordingly.

Also added a new "replica_test" E2E test.

### DIFF
--- a/scripts/e2e-test.sh
+++ b/scripts/e2e-test.sh
@@ -6,7 +6,8 @@ exit 0
 set -eux
 
 SCRIPTDIR=$(dirname "$(realpath "$0")")
-TESTS="install basic_volume_io node_disconnect/replica_pod_remove"
+# new tests should be added before the replica_pod_remove test
+TESTS="install basic_volume_io replica node_disconnect/replica_pod_remove"
 DEVICE=
 REGISTRY=
 TAG=
--- a/test/e2e/replica/replica_test.go
+++ b/test/e2e/replica/replica_test.go
@@ -1,0 +1,82 @@
+package replica_test
+
+import (
+	"e2e-basic/common"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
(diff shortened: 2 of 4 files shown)


scripts/e2e-test.sh
```diff
@@ -3,7 +3,7 @@
 set -eux
 
 SCRIPTDIR=$(dirname "$(realpath "$0")")
-TESTS="install basic_volume_io node_disconnect/replica_pod_remove"
```
chriswldenyer, line 6: Could you move the replica test to before the replica_pod_remove (which should be last, being the most disruptive)? 
I should have added a comment to that effect, but I've asked Blaise to add a comment to his PR.

chriswldenyer: Ah, he didn't. Could you add a comment along the lines of "add new tests before replica_pod_remove" ?

test/e2e/replica/replica_test.go
```diff
@@ -0,0 +1,82 @@
+package replica_test
```
chriswldenyer, line 1: Please mention this additional e2e test in the description.

scripts/e2e-test.sh
paulyoong: Done.

test/e2e/replica/replica_test.go
paulyoong: Done. I've also added an additional basic README.

tiagolobocastro (approved):

chriswldenyer (approved):
LGTM

paulyoong:
bors merge

paulyoong:
bors retry

paulyoong:
bors cancel

paulyoong:
bors merge

paulyoong:
bors merge

PR #620, opened 2021-01-11
ci: enable github actions for nvmx pieces

gila:
The nvmex tests are very isolated and so we really don't need
to run full CI/CD every time we make a change right now.


.github/workflows/nvmx.yml
```diff
@@ -0,0 +1,19 @@
+name: NVMx testing
+on: push
```
bobek, line 2: Do you want to run it on push to every branch. What about something like

```suggestion
on: 
  push:
    branches:
      - cas-534-nvmx
```

gila: Can't hurt to filter somewhat :+1:

gila:
merged manually with 8512d5a0

PR #621, opened 2021-01-12
fix(linter): fix linter git hook

tiagolobocastro:
@commitlint/config-conventional is not installed by npx so we need to
install it before running @commitlint/cli.
since we're doing this, install @commitlint/cli which speeds up npx as
it no longer needs to pull it.


paulyoong (approved):

blaisedias (approved):

tiagolobocastro:
bors merge

PR #625, opened 2021-01-13
fix(moac): import vols after nodes are registered

jkryl:
Give storage nodes a chance to register with control plane before
we start to import volumes from crds. Otherwise moac tries to fix
volumes that are ok, but it does not know about it.

### DIFF
--- a/mayastor/src/subsys/mbus/registration.rs
+++ b/mayastor/src/subsys/mbus/registration.rs
@@ -14,7 +14,7 @@ use snafu::Snafu;
 use std::{env, time::Duration};
 
 /// Mayastor sends registration messages in this interval (kind of heart-beat)
-const HB_INTERVAL: Duration = Duration::from_secs(10);
+const HB_INTERVAL: Duration = Duration::from_secs(5);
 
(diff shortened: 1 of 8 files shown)


mayastor/src/subsys/mbus/registration.rs
```diff
@@ -14,7 +14,7 @@ use snafu::Snafu;
 use std::{env, time::Duration};
 
 /// Mayastor sends registration messages in this interval (kind of heart-beat)
-const HB_INTERVAL: Duration = Duration::from_secs(10);
+const HB_INTERVAL: Duration = Duration::from_secs(5);
```
tiagolobocastro, line 17: if you're reducing the period can you also please reduce the default deadline in services/node/src/server.rs:19?

tiagolobocastro (approved):

mayastor/src/subsys/mbus/registration.rs
jkryl: thanks. I will 👍

gila (approved):

PR #627, opened 2021-01-14
test: remove unused nix tests (CAS-496)

paulyoong:
The nix tests have not been run for some time. Most of the test cases
are now covered by either the Docker Compose tests or the E2E tests -
making the nix tests redundant.

Some of the rebuild test cases that weren't covered have been rewritten
as Docker Compose tests.


composer/src/lib.rs
```diff
@@ -939,6 +943,24 @@ impl ComposeTest {
         Ok(())
     }
 
+    /// disconnect container from the network
+    pub async fn disconnect(
+        &self,
+        container_name: &str,
+        network_name: &str,
```
tiagolobocastro, line 950: you already have the network name in self so you don't need to pass it as an argument

tiagolobocastro (approved):

composer/src/lib.rs
paulyoong: Good point. I have replaced the use of _network_name_ for _self.network_id_.

mayastor/tests/rebuild.rs
```diff
@@ -636,6 +636,191 @@ async fn rebuild_with_load() {
...
+        rebuild_dst,
+        "running",
+        Duration::from_secs(1),
+    )
+    .await
+    .unwrap());
+
+    // Disconnect the rebuild source container from the network and check that
+    // the rebuild terminates. This requires a large timeout because it takes
+    // some time for the NVMf subsystem to report the error up.
```
jonathan-teh, line 751: Could address this by reducing the timeouts in `nvme_bdev_opts`, that would reduce test runtime.

paulyoong: Thanks for the suggestion. I gave it a go, changing the various times, but it didn't seem to make any noticeable difference in run time when I tried it locally. Perhaps I'm missing something. I think for now, I'll leave it as it is, but I'll raise a ticket for it. I think it would be worth revisiting.

jonathan-teh (approved):

paulyoong:
bors merge

paulyoong:
bors try

paulyoong:
bors merge

PR #630, opened 2021-01-14
refactor(nexus): add nexus create log

tiagolobocastro:
Add log message with the nexus create arguments.

### DIFF
--- a/mayastor/src/grpc/mayastor_grpc.rs
+++ b/mayastor/src/grpc/mayastor_grpc.rs
@@ -117,7 +117,7 @@ impl mayastor_server::Mayastor for MayastorSvc {
         sync_config(pool_grpc::share_replica(args)).await
     }
 
-    #[instrument(level = "debug", err)]
+    #[instrument(level = "info", err)]
     async fn create_nexus(
         &self,
         request: Request<CreateNexusRequest>,
@@ -128,8 +128,7 @@ impl mayastor_server::Mayastor for MayastorSvc {
             let name = uuid_to_name(&args.uuid)?;
             locally! { async move {
                 nexus_create(&name, args.size, Some(&args.uuid), &args.children).await
-            }}
-            ;
+            }};
             let nexus = nexus_lookup(&uuid)?;
             info!("Created nexus {}", uuid);
             Ok(Response::new(nexus.to_grpc()))


mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -126,6 +126,7 @@ impl mayastor_server::Mayastor for MayastorSvc {
             let args = request.into_inner();
             let uuid = args.uuid.clone();
             let name = uuid_to_name(&args.uuid)?;
+            info!("Creating nexus '{}' with arguments: '{:?}'", uuid, args);
```
paulyoong, line 129: Would it be better to use the name rather than the uuid? The uuid is already in the args whereas the name is constructed from the uuid.

paulyoong (approved):

mayastor/src/grpc/mayastor_grpc.rs
gila: @tiagolobocastro what about we try:

```
#[instrument(err)]
```
We should see the call with arguments -- all the time (and during debug) so we don't need to add `info!()` calls at all?

tiagolobocastro: But that would be only in case of error?

tiagolobocastro: Yeah that works, I've removed the log and added that instrument with info.
Though tbh I tend not to like to inspect those instrument thingies, too confusing to visually inspect without any tooling but it's all we have atm..

gila: Since we are being honest -- I don't like it either not one bit! The only thing I like is that can add "log in and out". 

The reason for this is its very easy to have args change or returns for that matter and the logging messages are forgotten or not updated. 

OTOH, all things considered it's (to me) one of those "o will look into that later"

I see you have mastered proc-macro powers so we can make look into it later unless it bugs you so much you want to kill it with fire right now :)

gila (approved):

mayastor/src/grpc/mayastor_grpc.rs
tiagolobocastro: I actually had a look at this before, we can configure what the instrument macro does, I think mayastor has it "configured for tracing" with all the events showing. On the control plane services, you only get the one log file with the function arguments for example.

I also got the instrument enter/exit thingies to actually show up on the jaeger container at some point before Christmas, might raise a story for that.

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #631, opened 2021-01-14
fix: in child_retire, use close not bdev_destroy

chriswldenyer:
This caters for the simultaneous deletion of a bdev within
nexus code (due to io-error) with an attempt via the control plane,
and so avoid a race condition where the nexus attempts to 
delete a bdev that no longer exists. child.close() includes
checks for the case where the bdev has already been deleted.


chriswldenyer:
bors try

PR #633, opened 2021-01-15
fix: avoid panic with missing bdev

chriswldenyer:
When calling child_retire() and bdev_destroy returns
an error, log the error instead of panicking. The bdev
may be missing due to a simultaneous bdev removal via
the control plane. A further change will introduce
better synchronization to avoid this condition.


paulyoong (approved):

chriswldenyer:
bors try

gila (approved):
Thank you @chriswldenyer

jonathan-teh (approved):

chriswldenyer:
bors merge

PR #636, opened 2021-01-18
test(rebuild): disable test

paulyoong:
Disable the rebuild_dst_disconnect test. This test is occasionally
failing on CI/CD even with a suitably high timeout time. When this
occurs an error message is output indicating that we failed to get IO
channel because we were probably low on memory.

A ticket will be raised to investigate this fully and re-enable the
test.

### DIFF
--- a/mayastor/tests/rebuild.rs
+++ b/mayastor/tests/rebuild.rs
@@ -776,6 +776,7 @@ async fn rebuild_src_disconnect() {
 /// Test rebuild when disconnecting the destination container from the
 /// network.
 #[tokio::test]
+#[ignore]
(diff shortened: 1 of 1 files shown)


chriswldenyer (approved):
Maybe comment why it is disabled, in the source code?

blaisedias (approved):

paulyoong:
> Maybe comment why it is disabled, in the source code?

I thought it would be sufficient to include the reasoning in the commit message as we would probably be looking at that anyway when we come to re-enable it.

paulyoong:
bors merge

PR #638, opened 2021-01-18
fix(moac): distribute nexuses more evenly

jkryl:
We still keep the rule that one replica is always local to the
nexus, but when deciding which replica is chosen as local we try
to choose the node with smallest number of nexuses.


gila (approved):

tiagolobocastro (approved):

mtzaurus (approved):

jkryl:
bors merge

PR #640, opened 2021-01-19
test(cargo): verify unmap operation for aio module

cjones1024:
Test is currently disabled until spdk is updated.

### DIFF
--- a/mayastor/tests/aio_unmap.rs
+++ b/mayastor/tests/aio_unmap.rs
@@ -1,0 +1,569 @@
+use common::MayastorTest;
+
+use mayastor::{
+    core::{Bdev, BdevHandle, MayastorCliArgs, Share},
+    ffihelper::cb_arg,
+    lvs::{Lvol, Lvs},
+    nexus_uri::{bdev_create, bdev_destroy},
+};
+
+use rpc::mayastor::CreatePoolRequest;
+
+use futures::channel::oneshot;
+use once_cell::sync::Lazy;
+use std::{
+    ffi::{c_void, CString},
+    io::{Error, ErrorKind},
+    mem::MaybeUninit,
+};
+
+use spdk_sys::{
+    spdk_bdev_free_io,
+    spdk_bdev_io,
+    spdk_bdev_unmap,
+    spdk_bdev_write_zeroes,
+};
+
+use nvmeadm::NvmeTarget;
+use std::convert::TryFrom;
+
+use tracing::info;
+
+pub mod common;
+
+static BDEVNAME1: &str = "aio:///tmp/disk1.img";
+static DISKNAME1: &str = "/tmp/disk1.img";
+
+// Get the I/O blocksize for the filesystem upon which the given file resides.
+fn get_fs_blocksize(path: &str) -> Result<u64, Error> {
+    Ok(stat(path)?.st_blksize as u64)
+}
+
+// Return the number of (512-byte) blocks allocated for a given file.
+// Note that the result is ALWAYS in terms of 512-byte blocks,
+// regardless of the underlying filesystem I/O blocksize.
+fn get_allocated_blocks(path: &str) -> Result<u64, Error> {
+    Ok(stat(path)?.st_blocks as u64)
+}
+
+// Obtain stat information for given file
+fn stat(path: &str) -> Result<libc::stat64, Error> {
+    let mut data: MaybeUninit<libc::stat64> = MaybeUninit::uninit();
+    let cpath = CString::new(path).unwrap();
+
+    if unsafe { libc::stat64(cpath.as_ptr(), data.as_mut_ptr()) } < 0 {
+        return Err(Error::last_os_error());
+    }
+
+    Ok(unsafe { data.assume_init() })
+}
+
+extern "C" fn io_completion_cb(
+    io: *mut spdk_bdev_io,
+    success: bool,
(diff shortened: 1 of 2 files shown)


mayastor/tests/aio_unmap.rs
```diff
@@ -0,0 +1,118 @@
...
+                .unwrap();
+        }
+
+        let pool = Lvs::lookup("tpool").unwrap();
+        assert_eq!(pool.lvols().unwrap().count(), 4);
+    })
+    .await;
+
+    // verify that some blocks have been allocated
+    assert_ne!(get_allocated_blocks(DISKNAME1).unwrap(), 0);
```
tiagolobocastro, line 76: Looks like these lvols only have metadata, should we also write some "user data" to the lvol?
Would it be worth sending explicit unmap to each lvol or even via a nexus? (to cover the case for unmaps from the pod host)?

mayastor/tests/aio_unmap.rs
```diff
@@ -0,0 +1,118 @@
...
+    .await;
+
+    // Check that we're able to find our new LVS.
+    ms.spawn(async {
+        assert_eq!(Lvs::iter().count(), 1);
+        let pool = Lvs::lookup("tpool").unwrap();
+        assert_eq!(pool.name(), "tpool");
+        assert_eq!(pool.used(), 0);
+        dbg!(pool.uuid());
+        assert_eq!(pool.base_bdev().name(), "/tmp/disk1.img");
```
jonathan-teh, line 57: Use DISKNAME here.

mayastor/tests/aio_unmap.rs
```diff
@@ -0,0 +1,118 @@
...
+        assert_eq!(Lvs::iter().count(), 0);
+
+        // no bdevs
+        assert_eq!(Bdev::bdev_first().into_iter().count(), 0);
+    })
+    .await;
+
+    // Verify that all used blocks have been discarded by confirming
+    // that no blocks are currently allocated for the backing store.
+    assert_eq!(get_allocated_blocks(DISKNAME1).unwrap(), 0);
```
jonathan-teh, line 115: Use a custom message with the assert so it's clear if it fails and the assert itself becomes self-documenting.

mayastor/tests/aio_unmap.rs
cjones1024: Yes we could write user data to the lvols. However this was not supposed to be a test to validate that the upper layers do the right thing when the lvol or pool is destroyed. This was intended to be a simple test to validate that when the upper layers want to zero a region, the blocks comprising that region are in fact unallocated. The fact that everything is zeroed means I don't need to complicate things by having to verify precisely where a hole is located in a file.

cjones1024: After discussing with tiago, will create 2 tests:
- low level tests that verifies calling unmap on a bdev actually unallocates the blocks form the underlying file.
- a higher level test that verifies that destroying an lvol or pool does in fact zero the data (without being concerned as to how the zeroes are written).

mayastor/tests/aio_unmap.rs
```diff
@@ -0,0 +1,118 @@
...
+        assert_eq!(pool.used(), 0);
+        dbg!(pool.uuid());
+        assert_eq!(pool.base_bdev().name(), "/tmp/disk1.img");
+    })
+    .await;
+
+    // Create 4 lvols on this pool.
+    ms.spawn(async {
+        let pool = Lvs::lookup("tpool").unwrap();
+        for i in 0 .. 4 {
```
mtzaurus, line 64: Could you please use a constant to define the amount of volumes to be created/tested ?

mtzaurus:
I have a generic question: is it worth adding a testcase to read unmapped range and make sure we read zeroes from unmapped range ?

cjones1024:
> 
> 
> I have a generic question: is it worth adding a testcase to read unmapped range and make sure we read zeroes from unmapped range ?

Good Idea, Will add

mayastor/tests/aio_unmap.rs
cjones1024: Done

mayastor/tests/aio_unmap.rs
cjones1024: Done

tiagolobocastro (approved):

cjones1024:
bors try

cjones1024:
bors try

mayastor/tests/aio_unmap.rs
```diff
@@ -0,0 +1,570 @@
...
+        .await
+        .expect("failed awaiting write_zeroes completion")
+    {
+        return Ok(());
+    }
+
+    Err(Error::new(ErrorKind::Other, "write_zeroes failed"))
+}
+
+fn setup() -> &'static MayastorTest<'static> {
```
gila, line 146: Can we use Compose:::MaystorTest for this?

cjones1024:
bors check

cjones1024:
bors try

jonathan-teh:
This only tests the "file" codepath in the aio bdev module, are there plans to test the "block" codepath too? Perhaps not here as it's tricky to set up a temporary block device and clean up afterwards.

jonathan-teh (approved):

cjones1024:
bors merge

PR #642, opened 2021-01-19
chore: more fine grained mayastor test permissions

Hoverbear:
This reduces our usage of `sudo` in Mayastor tests.

Currently, they run as root with full privileges
(via a cargo alias), this isn't awesome.

This PR still uses `sudo`, but only uses `sudo`
long enough to gain the required capabilities
(which @gila told me, thank you). I think the
ambient capabilities are optional, but I suspect
the way it plays into inherited binaries will be
relevant to our use case.

I haven't used `capsh` much, so I'd appreciate
any pointers! :blush: 

## Testing

From `mayastor/`, as a normal user:

```bash
nix-shell ../shell.nix
> cargo test
```

From `mayastor/`, as root:

```bash
sudo nix-shell ../shell.nix
> cargo test
```

### DIFF
--- a/mayastor/.cargo/config
+++ b/mayastor/.cargo/config
@@ -1,4 +1,4 @@
 # we need elevated privileges to run mayastor related tests
 # cargo will ask you nicely to type your password
 [target.x86_64-unknown-linux-gnu]
-runner = 'sudo -E'
+runner = ".cargo/runner.sh"
--- a/mayastor/.cargo/runner.sh
+++ b/mayastor/.cargo/runner.sh
@@ -1,15 +1,20 @@
 #! /usr/bin/env bash
-set -x
 
 # Grab the arguments passed to the runner.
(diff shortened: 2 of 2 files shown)


jkryl (commented):
awesome! Just a nit. If it already runs as root then we should skip the trick with sudo in runner.sh. Some users might not have sudo installed, but they would still like to run everything as root.

paulyoong (approved):

tiagolobocastro (approved):

gila (approved):

Hoverbear:
@jkryl Yeah!!! Let me get that!

jkryl (approved):

gila:
bors merge

Hoverbear:
I think this is also an unrelated failure.

bors merge

Hoverbear:
bors merge

PR #644, opened 2021-01-19
fix(lint): sanitize PR messages

tiagolobocastro:
Sanitize PR title and body when doing the lint pull request stage.
Also explicitly set type-enum values so we add "example" to the list.

PR #646, opened 2021-01-20
chore: permit default conventional commit lengths

Hoverbear:
Previously we overrode the defaults set by
`@commitlint/config-conventional`. Now we do not.

We flagged this during a discussion on Slack about
#644 as we noted the fix still left #642 invalid
due to line length issues.

I'm not sure why we overrode this, this might cause
breakage in scripts which depend on the presumed
line length.

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,8 +1,6 @@
 module.exports = {
   extends: ['@commitlint/config-conventional'],
   rules: {
-    "header-max-length": async () => [2, "always", 50],
-    "body-max-line-length": async () => [2, "always", 72],
     'type-enum': [2, 'always', ['build', 'chore', 'ci', 'docs', 'feat', 'fix', 'perf', 'refactor', 'revert', 'style', 'test', 'example']],
   },
   defaultIgnores: false,


tiagolobocastro:
The golden ration 50/72 was once suggested during a meeting before we had the commitlint, which is why I've set it as such, though there was no discussion over it now that we're actually enforcing it.
Does the changelog generator work well with 50/72?
Another point is also, do we really want to lint the body from the PR, since it will only be part of the Merge commit by bors?

Hoverbear:
Ahh! I mean, if there is a reason to be doing this, then great!

I've been using 100 in other projects since it aligns with the [`config-conventional`](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional) as well as [Rust's guidelines](https://github.com/rust-dev-tools/fmt-rfcs/blob/b208df842cbfa013fe2d3123a96ef2780a4a9950/guide/guide.md#indentation-and-line-width).

I've previously used https://convco.github.io/ which is a rust crate and didn't experience these lints.

jkryl (approved):

Hoverbear:
I noted from @tiagolobocastro that we plan to use https://github.com/clog-tool/clog-cli to generate our changelogs. So I'll ensure compatability.

gila (approved):

tiagolobocastro (approved):

Hoverbear:
bors merge

Hoverbear:
bors merge

PR #647, opened 2021-01-21
feat(nvmx): controller shutdown/removal

mtzaurus:
This fix introduces shutdown logic for NVMe controllers, which kicks in
upon device removal and properly cleans up controller' I/O resources.

I/O device now uses a wrapper around SPDK I/O device handle, which
allows better I/O channel iteration and hides low-level SPDK functions
inside iterator.

Implements CAS-627


mtzaurus:
Closing PR due to lint-unfriendly PR header.

PR #648, opened 2021-01-21
feat(nvmx): controller shutdown/removal

mtzaurus:
This fix introduces shutdown logic for NVMe controllers, which kicks in
upon device removal and properly cleans up controller I/O resources.

I/O device now uses a wrapper around SPDK I/O device handle, which
allows better I/O channel iteration and hides low-level SPDK functions
inside iterator.

Implements CAS-627

### DIFF
--- a/mayastor/src/bdev/dev/nvmx/channel.rs
+++ b/mayastor/src/bdev/dev/nvmx/channel.rs
@@ -1,17 +1,19 @@
 /* I/O channel for NVMe controller, one per core. */
 
 use crate::{
-    bdev::dev::nvmx::NVME_CONTROLLERS,
+    bdev::dev::nvmx::{NvmeControllerState, NVME_CONTROLLERS},
     core::poller,
     subsys::NvmeBdevOpts,
 };
 use std::{cmp::max, mem::size_of, os::raw::c_void, ptr::NonNull};
 
 use spdk_sys::{
+    nvme_qpair_abort_reqs,
     spdk_io_channel,
     spdk_nvme_ctrlr,
     spdk_nvme_ctrlr_alloc_io_qpair,
     spdk_nvme_ctrlr_connect_io_qpair,
+    spdk_nvme_ctrlr_disconnect_io_qpair,
     spdk_nvme_ctrlr_free_io_qpair,
     spdk_nvme_ctrlr_get_default_io_qpair_opts,
     spdk_nvme_ctrlr_reconnect_io_qpair,
@@ -63,24 +65,75 @@ pub struct NvmeIoChannelInner<'a> {
     pub qpair: *mut spdk_nvme_qpair,
     poll_group: NonNull<spdk_nvme_poll_group>,
     poller: poller::Poller<'a>,
+    // Flag to indicate the shutdown state of the channel.
+    // We need such a flag to differentiate between channel reset and shutdown.
+    // Channel reset is a reversible operation, which is followed by
+    // reinitialize(), which 'resurrects' channel (i.e. recreates all its
+    // I/O resources): such behaviour is observed during controller reset.
+    // Shutdown, in contrary, means 'one-way' ticket for the channel, which
+    // doesn't assume any further resurrections: such behaviour is seen
+    // upon controller shutdown. Being able to differentiate between these
+    // 2 states allows controller reset to behave properly in parallel with
+    // shutdown (if case reset is initiated before shutdown), and
+    // not to reinitialize channels already processed by shutdown logic.
+    is_shutdown: bool,
 }
 
 impl NvmeIoChannelInner<'_> {
-    /// Resets channel, making it unusable till reinitialize() is called.
+    /// Reset channel, making it unusable till reinitialize() is called.
     pub fn reset(&mut self) -> i32 {
         if self.qpair.is_null() {
             return 0;
         }
 
-        0
+        self._abort_queue();
+
+        let q = self.qpair;
+        self.qpair = std::ptr::null_mut();
+
+        unsafe {
+            spdk_nvme_ctrlr_disconnect_io_qpair(q);
+            spdk_nvme_ctrlr_free_io_qpair(q)
+        }
     }
 
-    /// Reinitializes channel after reset.
+    /// Checks whether the I/O channel is shutdown.
+    pub fn is_shutdown(&self) -> bool {
+        self.is_shutdown
+    }
+
+    /// Shutdown I/O channel and make it completely unusable for I/O.
+    pub fn shutdown(&mut self) -> i32 {
+        if self.is_shutdown {
+            return 0;
+        }
+
+        let rc = self.reset();
+        if rc == 0 {
+            self.is_shutdown = true;
+        }
+        rc
+    }
+
+    /// Aborts all the requests in the qpair.
+    fn _abort_queue(&self) {
+        unsafe { nvme_qpair_abort_reqs(self.qpair, 1) };
+    }
+
+    /// Reinitializes channel after reset unless the channel is shutdown.
     pub fn reinitialize(
         &mut self,
         ctrlr_name: &str,
         ctrlr_handle: *mut spdk_nvme_ctrlr,
     ) -> i32 {
+        if self.is_shutdown {
+            error!(
+                "{} I/O channel is shutdown, channel reinitialization not possible",
+                ctrlr_name
+            );
+            return -libc::ENODEV;
+        }
+
         // Create qpair for target controller.
         let mut opts = spdk_nvme_io_qpair_opts::default();
         let default_opts = NvmeBdevOpts::default();
@@ -182,6 +235,17 @@ impl NvmeControllerIoChannel {
         };
 
         let controller = controller.lock().expect("lock error");
+
+        // Make sure controller is available.
+        if controller.get_state() != NvmeControllerState::Running {
+            error!(
+                "{} controller is in {:?} state, I/O channel creation not possible",
+                controller.get_name(),
+                controller.get_state()
+            );
+            return 1;
+        }
+
         let nvme_channel = NvmeIoChannel::from_raw(ctx);
         let mut opts = spdk_nvme_io_qpair_opts::default();
         let default_opts = NvmeBdevOpts::default();
@@ -234,6 +298,7 @@ impl NvmeControllerIoChannel {
             qpair,
             poll_group: NonNull::new(poll_group).unwrap(),
             poller,
+            is_shutdown: false,
         });
 
         nvme_channel.inner = Box::into_raw(inner);
--- a/mayastor/src/bdev/dev/nvmx/controller.rs
+++ b/mayastor/src/bdev/dev/nvmx/controller.rs
@@ -11,7 +11,6 @@ use spdk_sys::{
     spdk_io_channel_iter,
     spdk_io_channel_iter_get_channel,
     spdk_io_channel_iter_get_ctx,
-    spdk_io_channel_iter_get_io_device,
     spdk_io_device_register,
     spdk_io_device_unregister,
     spdk_nvme_ctrlr,
@@ -23,13 +22,19 @@ use spdk_sys::{
 
 use crate::{
     bdev::dev::nvmx::{
-        channel::{NvmeControllerIoChannel, NvmeIoChannel},
+        channel::{NvmeControllerIoChannel, NvmeIoChannel, NvmeIoChannelInner},
         nvme_bdev_running_config,
         uri::NvmeControllerContext,
         NvmeNamespace,
         NVME_CONTROLLERS,
     },
-    core::{mempool::MemoryPool, poller, CoreError, IoCompletionCallback},
+    core::{
+        mempool::MemoryPool,
+        poller,
+        CoreError,
+        IoCompletionCallback,
+        IoCompletionCallbackArg,
+    },
 };
 
 const RESET_CTX_POOL_SIZE: u64 = 1024 - 1;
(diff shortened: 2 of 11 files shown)


mayastor/src/bdev/dev/nvmx/channel.rs
```diff
@@ -63,24 +65,75 @@ pub struct NvmeIoChannelInner<'a> {
...
+    // Channel reset is a reversible operation, which is followed by
+    // reinitialize(), which 'resurrects' channel (i.e. recreates all its
+    // I/O resources): such behaviour is observed during controller reset.
+    // Shutdown, in contrary, means 'one-way' ticket for the channel, which
+    // doesn't assume any further resurrections: such behaviour is seen
+    // upon controller shutdown. Being able to differentiate between these
+    // 2 states allows controller reset to behave properly in parallel with
+    // shutdown (if case reset is initiated before shutdown), and
+    // not to reinitialize channels already processed by shutdown logic.
+    is_shutdown: bool,
```
tiagolobocastro, line 79: If it's false, do we need differentiate between reset and not reset? If so should this be an enum?

tiagolobocastro (commented):
Any reasons for the fn prefixed with _? Seems to be private/pub, though that should not be needed in rust.

mayastor/src/bdev/dev/nvmx/controller.rs
```diff
@@ -236,155 +242,224 @@ impl<'a> NvmeController<'a> {
...
 
-        unsafe {
-            spdk_for_each_channel(
-                self.id as *mut c_void,
-                Some(NvmeController::reset_destroy_channels),
-                reset_ctx as *mut c_void,
-                Some(NvmeController::reset_destroy_channels_done),
-            );
-        }
+        // Iterate over all I/O channels and rrest/econfigure them one by one.
```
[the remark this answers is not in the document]
mtzaurus: Yup, should be 'reset', thanks.

mayastor/src/bdev/dev/nvmx/uri.rs
```diff
@@ -286,6 +285,43 @@ impl CreateDestroy for NvmfDeviceTemplate {
 
     // nvme_bdev_ctrlr_create
     async fn destroy(self: Box<Self>) -> Result<(), Self::Error> {
+        // 1. Initiate controller shutdown, which shuts down all I/O resources
+        // of the controller.
+        let controller = NVME_CONTROLLERS
+            .lookup_by_name(self.get_name())
```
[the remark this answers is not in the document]
mtzaurus: Not sure it's an issue.
We access NVME_CONTROLLERS in similar manner in lots of our other async functions, all works fine.
Just a side note: lint mostly complains about ".await'ing with a mutex lock held", which is not the case here.

mayastor/src/bdev/dev/nvmx/channel.rs
mtzaurus: There is no need to differentiate between reset/shutdown when it's false. We just need a flag telling whether recover for the channel is possible or not. There is no complex logic around it.

mayastor/src/bdev/dev/nvmx/uri.rs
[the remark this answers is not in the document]
tiagolobocastro: It wouldn't be a lint issue, but what would happen if one async thread grabs the lock, then another async thread tries to grab the lock. At this point this is a deadlock because the other async fn will "block" the reactor and the initial thread won't run?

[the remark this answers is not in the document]
mtzaurus: If the first thread doesn't block forever while holding the lock, then it's not an issue, as the lock will be released upon leaving the lock scope and the second thread will proceed. We don't have any sleep/rescheduling with the lock held, we just grab the lock and do some 'atomic work' (I mean, not interrupted by other jobs).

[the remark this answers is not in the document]
tiagolobocastro: Yeah, I guess it's fine as long as there's no async code called while the lock is held.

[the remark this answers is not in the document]
mtzaurus: Yes.
And that's exactly what Lint blames about when it sees '.await' with a lock held.

mayastor/src/bdev/dev/nvmx/channel.rs
```diff
@@ -63,24 +65,75 @@ pub struct NvmeIoChannelInner<'a> {
...
+    /// Reset channel, making it unusable till reinitialize() is called.
     pub fn reset(&mut self) -> i32 {
         if self.qpair.is_null() {
             return 0;
         }
 
-        0
+        self._abort_queue();
+
+        let q = self.qpair;
```
gila, line 91: nit; we could use `std::ptr::swap()`  but likely its better that we abstract all of this in `qp.rs` if you will just as we did with the poller (i can work on this)

mayastor/src/bdev/dev/nvmx/channel.rs
```diff
@@ -63,24 +65,75 @@ pub struct NvmeIoChannelInner<'a> {
...
+        let rc = self.reset();
+        if rc == 0 {
+            self.is_shutdown = true;
+        }
+        rc
     }
 
-    /// Reinitializes channel after reset.
+    /// Aborts all the requests in the qpair.
+    fn _abort_queue(&self) {
```
gila, line 119: we dont need to have _ prefixed I think, the fn already is private.

mayastor/src/bdev/dev/nvmx/controller.rs
```diff
@@ -101,6 +117,7 @@ pub struct NvmeController<'a> {
     prchk_flags: u32,
     pub(crate) state: NvmeControllerState,
     inner: Option<NvmeControllerInner<'a>>,
+    reset_active: bool,
```
gila, line 120: Would reset_active not be more at home as a variant of NvmeControllerState? i.e ::Resetting ?

mayastor/src/bdev/dev/nvmx/controller.rs
```diff
@@ -201,33 +201,39 @@ impl<'a> NvmeController<'a> {
...
+                self.name
             );
         }
 
         self.inner.as_mut().unwrap().namespaces =
             vec![Arc::new(NvmeNamespace::from_ptr(ns))]
     }
 
+    /// Get controller state.
+    pub fn get_state(&self) -> NvmeControllerState {
```
gila, line 213: nit; would just `state()` not be nicer?

mayastor/src/bdev/dev/nvmx/controller.rs
```diff
@@ -236,155 +242,224 @@ impl<'a> NvmeController<'a> {
...
+            name: self.name.clone(),
+            cb,
+            cb_arg,
+            spdk_handle: self.ctrlr_as_ptr(),
+            io_device,
+        };
 
         // Mark controller as being under reset and schedule asynchronous reset.
-        self.set_state(NvmeControllerState::Resetting);
+        self.reset_active = true;
```
gila, line 269: ah! any reason for this?

gila (approved):
LGTM, I suggest though we abstract qpairs to hide ffi interfaces somewhat more even. I can work on that if it does not interfere with your other planning.

mayastor/src/bdev/dev/nvmx/controller.rs
mtzaurus: Yes.
I decided to get rid of Resetting as a separate state to simplify checking against controller availability in other code that makes judgements about availability of the controller.
Its much easier to check against ::Running as a valid state then ::Running && ::Resetting, because ::Resetting means 'Controller is OK and just being reset'.
So reset feels to me like an action inside ::Running state.

mayastor/src/bdev/dev/nvmx/channel.rs
```diff
@@ -63,24 +65,75 @@ pub struct NvmeIoChannelInner<'a> {
     pub qpair: *mut spdk_nvme_qpair,
     poll_group: NonNull<spdk_nvme_poll_group>,
     poller: poller::Poller<'a>,
+    // Flag to indicate the shutdown state of the channel.
```
Hoverbear, line 68: I suggest using `///` based doc comments here so IDE tooling! (It's not public, so it wouldn't show up in generated docs, but IDE tools can helpful and it's nearly for free)

mayastor/src/bdev/dev/nvmx/channel.rs
```diff
@@ -63,24 +65,75 @@ pub struct NvmeIoChannelInner<'a> {
...
+        if rc == 0 {
+            self.is_shutdown = true;
+        }
+        rc
     }
 
-    /// Reinitializes channel after reset.
+    /// Aborts all the requests in the qpair.
+    fn _abort_queue(&self) {
+        unsafe { nvme_qpair_abort_reqs(self.qpair, 1) };
```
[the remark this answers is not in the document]
Hoverbear: Would making this a const with a name, or a comment adding context, be a worthy interm mitigation step?

mayastor/src/bdev/dev/nvmx/controller.rs
```diff
@@ -60,25 +74,26 @@ impl<'a> NvmeControllerInner<'a> {
...
     fn to_string(&self) -> String {
         match *self {
             NvmeControllerState::Initializing => "Initializing",
             NvmeControllerState::Running => "Running",
-            NvmeControllerState::Resetting => "Resetting",
-            NvmeControllerState::Destroying => "Destroying",
+            NvmeControllerState::Unconfiguring => "Unconfiguring",
+            NvmeControllerState::Unconfigured => "Unconfigured",
         }
         .parse()
```
Hoverbear, line 98: I think you can call `"Unconfigured".to_string()` and avoid the unwrap here.

mayastor/src/bdev/dev/nvmx/controller.rs
```diff
@@ -236,155 +242,224 @@ impl<'a> NvmeController<'a> {
...
         Ok(())
     }
 
-    fn complete_reset(reset_ctx: &ResetCtx, status: i32) {
-        // Set controller state to Running and invoke completion callback.
-        let c = NVME_CONTROLLERS
-            .lookup_by_name(&reset_ctx.name)
-            .expect("Controller was removed while reset is in progress");
-        let mut controller = c.lock().expect("lock poisoned");
+    fn _shutdown_channels(
```
Hoverbear, line 280: Is there a significance to these `_`'d functions?

mtzaurus: Nope.
I just wanted to denote callbacks. I will remove underscores.

mayastor/src/bdev/dev/nvmx/controller.rs
mtzaurus: Good point, thanks.

mayastor/src/bdev/dev/nvmx/channel.rs
mtzaurus: Indeed it's useful for IDE/tools. Thanks !

mayastor/src/bdev/dev/nvmx/controller.rs
gila: Yes, I understand your thinking, and it makes sense. But at the same time, it also seems to lead to situations that function like `.state()` or `.set_state()` would not really tell you the whole picture when using these functions and their respective return types :thinking: 

So when we are resetting, we have the controller's global state (which is guarded by the mutex) and we keep `self.rest_active == true` until we have iterated all cores with an io channel to the device, right? 

If so we can perhaps introduce an extra variant of the enum type for example:

```rust
//to make it more clear for ourselves
type resetting: bool; 

pub enum NvmeControllerState {
 /// the old states we had and change Running too:
 Running(resetting)
}
```

This would  allow when calling `state()`  for the callee to really understand whats going on i.e:
```rust
match self.state() {
   Running(restting) if !resetting => {  // we are running and not resetting }, 
   Running(restting) if resetting => {  // guess what...   },
}
```

For cases where we don't care we can do:

```rust 
 match self.state() {
    Running(_) => {  // we are running but we don't care what state  we are in }, 
}
``` 

As this is somewhat verbose we can make use of the `matches!()` macro that is available in our version of rust: 
```rust
if matches!(self.state(), Running(_)) { } 
```

So this sounds like an awful lot like bikeshedding, but there is reason to the madness. We have a pretty messy state assessment within the current code base, and I would like to avoid that if at all possible. For example, we set a state from A to B but forgot to set a bool to false, making the code go poof.  

So what do you think? Is it worth trying it out such that we can differentiate internally and have just a single variable (and method) to consult to get the fully defined state rather than state + some other variable to consult? Then again, I might be pedantic :v:

mtzaurus: Interesting idea, I like it.
So we basically can hide sub-states for Running state inside the state itself.
We should definitely give it a try, so I will update the PR to reflect that.
Thanks !

mayastor/src/bdev/dev/nvmx/controller.rs
Hoverbear: Ahhh! I think idiomatically `cb` is usually used for callbacks, so maybe `fn_name_cb`? https://stackoverflow.com/a/41081702

mayastor/src/bdev/dev/nvmx/controller.rs
Hoverbear: That would be idiomatic: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

mayastor/src/bdev/dev/nvmx/channel.rs
Hoverbear: This is touched in https://github.com/openebs/Mayastor/pull/648#discussion_r562084675, it's to signify callbacks!

PR #654, opened 2021-01-25
fix(moac): sync replicas more often

jkryl:
fix(moac): sync replicas more often

One of the previous fixes for moac introduced a significant change to
lifecycle of the volumes and that was that a nexus for a volume was
created only if the volume was used (published in k8s speak). That
creates a number of occasions when we don't attempt to synchronize
data on replicas although we should. To solve that we create a nexus
each time we think that data need to be synchronized. We don't publish
it so that no one can access the volume. The new cases are:

* when volume is created: nexus writes labels to replicas and goes
  away again.
* when volume is imported: nexus checks replicas and sets their
  status (not implemented in data plane yet).
* when replicas are scaled up: new replica gets added to the new
  nexus and when the synchronization is finished, the nexus is
  removed again.
* when replicas are scaled down: replica gets removed from the new
  nexus, replica is destroyed and nexus is removed again.

By performing these type of operations while nexus is there we improve
the stability of the data throughout volume's lifecycle. There are
number of cases that aren't still solved and that require other
changes in data and control plane (i.e. persisten generation numbers
stored on replicas, storing gen numbers in control plane), though
this is a step forward.

### DIFF
--- a/csi/moac/test/volumes_test.js
+++ b/csi/moac/test/volumes_test.js
@@ -33,6 +33,7 @@ module.exports = function () {
   let nexus, replica1, replica2;
   let volume;
   let volEvents;
+  let isSynced1, isSynced2, isSynced3;
 
   // Create pristine test env with 3 pools on 3 nodes
   function createTestEnv () {
@@ -41,6 +42,13 @@ module.exports = function () {
     node1 = new Node('node1');
     node2 = new Node('node2');
     node3 = new Node('node3');
+    isSynced1 = sinon.stub(node1, 'isSynced');
+    isSynced1.returns(true);
+    isSynced2 = sinon.stub(node2, 'isSynced');
+    isSynced2.returns(true);
+    isSynced3 = sinon.stub(node3, 'isSynced');
+    isSynced3.returns(true);
+
     // pools sorted from the most to the least preferred
     pool1 = new Pool({
       name: 'pool1',
@@ -196,6 +204,18 @@ module.exports = function () {
         share: 'REPLICA_NONE',
         uri: 'bdev:///' + UUID
       });
+      stub1.onCall(1).resolves({
+        uuid: UUID,
+        size: 90,
+        state: 'NEXUS_ONLINE',
+        children: [
+          {
+            uri: 'bdev:///' + UUID,
+            state: 'CHILD_ONLINE',
+            rebuildProgress: 0
+          }
+        ]
+      });
 
       volumes.start();
       volume = await volumes.createVolume(UUID, {
@@ -206,8 +226,8 @@ module.exports = function () {
         limitBytes: 0,
         protocol: 'nbd'
       });
+      await waitUntil(() => volume.state === 'healthy', 'healthy volume');
       expect(volume.size).to.equal(90);
-      expect(volume.state).to.equal('healthy');
       sinon.assert.calledWithMatch(stub1.firstCall, 'createReplica', {
         uuid: UUID,
         pool: 'pool1',
@@ -215,8 +235,7 @@ module.exports = function () {
         thin: false,
         share: 'REPLICA_NONE'
       });
-      // 1 new + 2 mods
-      expect(volEvents).to.have.lengthOf(3);
+      expect(volEvents).to.have.lengthOf(5);
     });
 
     it('should limit the size of created volume', async () => {
@@ -229,6 +248,18 @@ module.exports = function () {
         share: 'REPLICA_NONE',
         uri: 'bdev:///' + UUID
       });
+      stub1.onCall(1).resolves({
+        uuid: UUID,
+        size: 50,
+        state: 'NEXUS_ONLINE',
+        children: [
+          {
+            uri: 'bdev:///' + UUID,
+            state: 'CHILD_ONLINE',
+            rebuildProgress: 0
+          }
+        ]
+      });
 
       volumes.start();
       volume = await volumes.createVolume(UUID, {
@@ -239,8 +270,8 @@ module.exports = function () {
         limitBytes: 50,
         protocol: 'nbd'
       });
+      await waitUntil(() => volume.state === 'healthy', 'healthy volume');
       expect(volume.size).to.equal(50);
-      expect(volume.state).to.equal('healthy');
       sinon.assert.calledWithMatch(stub1.firstCall, 'createReplica', {
         uuid: UUID,
         pool: 'pool1',
@@ -248,8 +279,7 @@ module.exports = function () {
         thin: false,
         share: 'REPLICA_NONE'
       });
-      // 1 new + 2 mods
-      expect(volEvents).to.have.lengthOf(3);
+      expect(volEvents).to.have.lengthOf(5);
     });
 
     it('should fail if the size is zero', async () => {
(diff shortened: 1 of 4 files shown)


csi/moac/test/volumes_test.js
```diff
@@ -302,13 +332,18 @@ module.exports = function () {
...
+        children: [`bdev:///${UUID}`]
+      });
       expect(Object.keys(volume.replicas)).to.have.lengthOf(1);
       expect(Object.values(volume.replicas)[0]).to.equal(replica);
-      expect(volume.state).to.equal('healthy');
-      expect(volEvents).to.have.lengthOf(3);
+      expect(volEvents).to.have.lengthOf(4);
       expect(volEvents[0].eventType).to.equal('new');
       expect(volEvents[1].eventType).to.equal('mod');
       expect(volEvents[2].eventType).to.equal('mod');
```
paulyoong, line 349: Given that we have an additional event, should we be adding another check for the additional event i.e. volEvents[3]?

csi/moac/test/volumes_test.js
```diff
@@ -1501,7 +1671,7 @@ module.exports = function () {
       expect(volume.state).to.equal('healthy');
       expect(Object.keys(volume.replicas)).to.have.length(3);
       // 2 nexus events
```
paulyoong, line 1673: Comment needs updating to "3 nexus events"

csi/moac/volume.ts
```diff
@@ -490,14 +454,41 @@ export class Volume {
...
+
     // Ensure there is sufficient number of replicas for the volume.
     const newReplicaCount = this.replicaCount - Object.keys(this.replicas).length;
     if (newReplicaCount > 0) {
       // create more replicas if higher replication factor is desired
       await this._createReplicas(newReplicaCount);
     }
-    this._setState(VolumeState.Healthy);
+    const [nexusNode, replicaSet] = await this._ensureReplicaShareProtocols();
+    // If the nexus poped up while we were creating replicas pick it up now.
```
paulyoong, line 466: Typo: "poped" -> "popped"

csi/moac/volume.ts
```diff
@@ -490,14 +454,41 @@ export class Volume {
...
+    // Though it's an unsual situation so we log a warning if it happens.
+    if (!this.nexus) {
+      const nexus = this.registry.getNexus(this.uuid);
+      if (nexus) {
+        log.warn(
+          `The nexus "${nexus}" appeared while creating replicas - using it`
+        );
+        this.newNexus(nexus);
```
paulyoong, line 474: Is it possible that the nexus that we are now using has a different number of replicas and therefore we don't have the same configuration as if we were to create a new one?

paulyoong (approved):

csi/moac/test/volumes_test.js
```diff
@@ -336,7 +371,7 @@ module.exports = function () {
         state: 'NEXUS_ONLINE',
         children: [
           {
-            uri: `nvmf:///blabla/${UUID}`,
+            uri: `nvmf://blabla/${UUID}`,
```
tiagolobocastro, line 374: i think the original /// were correct

csi/moac/volume.ts
```diff
@@ -247,72 +234,37 @@ export class Volume {
...
-      assert(this.publishedOn);
-      let nexusNode = this.registry.getNode(this.publishedOn);
-      if (nexusNode && nexusNode.isSynced()) {
-        let replicas = [];
-        for (let nodeName in this.replicas) {
-          if (!this.replicas[nodeName].isOffline() && !this.nodeBlackList[nodeName]) {
-            replicas.push(this.replicas[nodeName]);
+      if (
+        this.publishedOn ||
+        replicaSet.length !== this.replicaCount
```
tiagolobocastro, line 252: If the user reduces the number of replicas, we don't need to actually bring up a nexus?

tiagolobocastro (approved):

csi/moac/volume.ts
jkryl: It's true that it's not strictly necessary. It's more kind of a double check that we don't do anything stupid. The case is as follows. All that moac sees is that there are n replicas and there should be n - 1. It does not know if all replicas are sound and healthy. They should be but perhaps not in all cases. Then creating a nexus to tell us what the nexus thinks about those replicas (i.e. is replica reachable from nexus) before we try to destroy any of them, can be seen as another safety check. It cannot tell us if any replica is out of sync because we haven't implemented generation numbers still, but that would be also possible in the future. But I admit that I'm being maybe too much paranoid. So I will defer to reviewers opinion on this. Although that the code is simpler as it is now, adding a new block of code for removing replica when there is no nexus would not be probably too much work to do.

csi/moac/test/volumes_test.js
jkryl: I always confuse these. The general scheme is `protocol://host:port/path`, right? So if there is no host part then you get 3 slashes in row. With nvmf there is the host part or isn't? It's not really important as long as we use the same string here and in the assert that is testing the value, but would be nice to get it right.

csi/moac/volume.ts
jkryl: I think that might be possible. We have already created replicas and ensured the share protocols on them and then suddenly a strange nexus pops up. What is the chance that this will work if we don't know anything about the nexus? FSA would fix any inconsistencies however it is non-deterministic as we don't know when and IO could fail in the meantime if volume is used. Maybe this part of the code is a bit over-thought and we should just go straight to the nexus creation without checking if it exists. AFAIK the create will fail in that case and we simply return an error to the CSI method caller. If you think so, I will remove this part of code to make it simpler.

csi/moac/volume.ts
jkryl: will fix 👍

csi/moac/test/volumes_test.js
jkryl: thanks 👍

csi/moac/test/volumes_test.js
jkryl: yup, you are right 👍

csi/moac/test/volumes_test.js
tiagolobocastro: Right, with nvmf you need the host part I believe. Ah is blabla the host path?

csi/moac/volume.ts
tiagolobocastro: Yeah I suppose it doesn't hurt to do it, and we can check if the replicas are still "alive". Although if any replica is not "alive" then the nexus creation will fail IIRC, but that's a separate issue altogether.

csi/moac/test/volumes_test.js
jkryl: yea, very inaccurate approximation of what you would normally find there :-) But there are places where I got it even worse without host part just nvmf://${UUID} ... :-)

csi/moac/volume.ts
jkryl: yea, that's true and I did not think about it. It's like stepping through a mine field now  :-D  So we might end up in a situation when we simply cannot remove a replica from volume, because the replica that appears to be ok is not reachable from nexus, that's why nexus creation fails, that's why we can't remove any replica 🤔  not a nice case. On the other hand, moac is not the only piece of puzzle in this game and it is a general problem that can bite us at many places (i.e. also when we try to scale up the volume).

csi/moac/volume.ts
paulyoong: I think I would prefer what you suggested - to attempt to create the nexus and return a failure if it fails. At least that way we know what we're getting. Plus, from what I understand, the chances of hitting this particular case is small, so we shouldn't be seeing nexus create failures often.

csi/moac/volume.ts
tiagolobocastro: Yeah, so if we fix that issue, then bring up the nexus to reduce a replica should not pose any issue. Meanwhile if we wish to decrease the replicas when 1 replica is not "alive", then we could always manually remove it from the list of children?

csi/moac/test/volumes_test.js
tiagolobocastro: hehe then it's now fixed :-)

csi/moac/volume.ts
jkryl: I have never tried that (editing replicas in msv). In general it is the part of crd that users shouldn't touch. Though IMHO it is probably acceptable to live with this corner case until nexus create is fixed.

csi/moac/volume.ts
jkryl: Agree. I will make this change 👍

jkryl:
bors merge

PR #655, opened 2021-01-25
ci(moac): scale up/down test fails sometimes

jkryl:
The problem was that after we add a new replica to the volume it
does not immediately transition to degraded state (the new replica
usually starts in degraded state because it is not synced) but the
volume state change happens shortly after it. So the test code was
wrong that immediately after adding a child to nexus the state should
be degraded. It has to wait for it.

### DIFF
--- a/csi/moac/test/volumes_test.js
+++ b/csi/moac/test/volumes_test.js
@@ -1045,8 +1045,10 @@ module.exports = function () {
         protocol: 'nbd'
       });
 
-      await waitUntil(() => nexus.children.length === 3, 'new replica');
-      expect(volume.state).to.equal('degraded');
+      await waitUntil(
+        () => nexus.children.length === 3 && volume.state === 'degraded',
+        'degraded volume with new replica'
+      );
 
       const newChild = volume.nexus.children.find(
         (ch) => ch.state === 'CHILD_DEGRADED'


csi/moac/test/volumes_test.js
```diff
@@ -1045,8 +1045,10 @@ module.exports = function () {
         protocol: 'nbd'
       });
 
-      await waitUntil(() => nexus.children.length === 3, 'new replica');
-      expect(volume.state).to.equal('degraded');
+      await waitUntil(
+        () => nexus.children.length === 3 && volume.state === 'degraded',
```
tiagolobocastro, line 1049: hmm, I don't know how exactly waitUntil works, but would this not trip up the CI failure in the same way? Because the test complained that healthy != degraded, which means when moac looked, the child was already rebuild/healthy?

jkryl: Note that no state of replica can change unless we explicitly change it from the test code as we are using mocks and not real mayastor instances. So it can't happen that the state of replica would go from degraded to healthy without us doing that change. waitUntil() waits for the condition to become true. So if replica is added to the nexus, the addChild gRPC method returns state of the nexus including its children. We mock the reply and set the state of newly added child to degraded to mimic real behaviour and it stays so until the test is over.

paulyoong (approved):

tiagolobocastro (approved):

jkryl:
bors merge

jkryl:
bors retry

jkryl:
bors retry

jkryl:
bors retry

PR #661, opened 2021-01-26
fix(nexus): allow re-adding child removed by AEN

jonathan-teh:
Add a destroying flag to NexusChild and only set it in destroy() so
a nexus child destroyed via the normal close() path actually removes
the bdev.

This allows a remove via e.g. an NVMe async event notification to only
mark the child as faulted, which does remove it from the I/O path, and
allows a subsequent normal destroy of a child to remove the bdev.
Previously it would have set NexusChild::bdev to None without
destroying the bdev, which leaks it.

This requires that the shutdown process destroys nexuses first so that
it then destroys each child and its bdev, in that order, as the remove
event from SPDK as part of shutdown would not destroy the bdev, which
would result in the process hanging.

Fixes CAS-624

### DIFF
--- a/mayastor/src/bdev/nexus/mod.rs
+++ b/mayastor/src/bdev/nexus/mod.rs
@@ -59,3 +59,17 @@ pub fn nexus_instance_new(name: String, size: u64, children: Vec<String>) {
     let list = instances();
     list.push(Nexus::new(&name, size, None, Some(&children)));
 }
+
+/// called during shutdown so that the bdevs for each child is destroyed first
+pub async fn nexus_destroy_all() {
+    info!("destroying all nexuses...");
+    for nexus in instances() {
+        if let Err(e) = nexus.destroy().await {
+            error!(
+                "failed to destroy nexus {} during shutdown: {}",
+                nexus.name, e
+            );
+        }
+    }
+    info!("destroyed all nexuses");
+}
--- a/mayastor/src/bdev/nexus/nexus_child.rs
+++ b/mayastor/src/bdev/nexus/nexus_child.rs
@@ -132,6 +132,9 @@ pub struct NexusChild {
     /// current state of the child
     #[serde(skip_serializing)]
     pub state: AtomicCell<ChildState>,
+    /// destruction started, differentiates remove events
+    #[serde(skip_serializing)]
+    destroying: AtomicCell<bool>,
     /// record of most-recent IO errors
     #[serde(skip_serializing)]
     pub(crate) err_store: Option<NexusErrStore>,
@@ -229,6 +232,7 @@ impl NexusChild {
             },
         )?);
 
+        self.destroying.store(false);
         self.desc = Some(desc);
 
         let cfg = Config::get();
@@ -318,6 +322,11 @@ impl NexusChild {
         self.state.load()
     }
 
+    /// returns whether destruction has started
+    pub fn destroying(&self) -> bool {
+        self.destroying.load()
+    }
+
     pub(crate) fn rebuilding(&self) -> bool {
         match RebuildJob::lookup(&self.name) {
             Ok(_) => self.state() == ChildState::Faulted(Reason::OutOfSync),
@@ -369,8 +378,14 @@ impl NexusChild {
     pub(crate) fn remove(&mut self) {
         info!("Removing child {}", self.name);
 
-        // The bdev is being removed, so ensure we don't use it again.
-        self.bdev = None;
+        // Only remove the bdev if the child is being destroyed instead of a
+        // hot remove event e.g. NVMe AEN (asynchronous event notification)
+        let destroying = self.destroying();
+        info!("Destroying child: {}", destroying);
+        if destroying {
+            // The bdev is being removed, so ensure we don't use it again.
+            self.bdev = None;
+        }
 
         let state = self.state();
 
@@ -385,7 +400,7 @@ impl NexusChild {
         }
 
         // Remove the child from the I/O path. If we had an IO error the bdev,
-        // the channels where already reconfigured so we dont have to do
+        // the channels were already reconfigured so we don't have to do
         // that twice.
         if state != ChildState::Faulted(Reason::IoError) {
             let nexus_name = self.parent.clone();
@@ -397,10 +412,12 @@ impl NexusChild {
             });
         }
 
-        // Dropping the last descriptor results in the bdev being removed.
-        // This must be performed in this function.
-        let desc = self.desc.take();
-        drop(desc);
+        if destroying {
+            // Dropping the last descriptor results in the bdev being removed.
+            // This must be performed in this function.
+            let desc = self.desc.take();
+            drop(desc);
+        }
 
         self.remove_complete();
         info!("Child {} removed", self.name);
@@ -428,6 +445,7 @@ impl NexusChild {
             parent,
             desc: None,
             state: AtomicCell::new(ChildState::Init),
+            destroying: AtomicCell::new(false),
             err_store: None,
             remove_channel: mpsc::channel(0),
         }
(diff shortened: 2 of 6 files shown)


jonathan-teh:
bors try

mayastor/src/bdev/nexus/mod.rs
```diff
@@ -59,3 +59,17 @@ pub fn nexus_instance_new(name: String, size: u64, children: Vec<String>) {
     let list = instances();
     list.push(Nexus::new(&name, size, None, Some(&children)));
 }
+
+/// called during shutdown so that the bdevs for each child is destroyed first
+pub async fn nexus_destroy_all() {
+    debug!("destroying all nexuses...");
+    for nexus in instances() {
```
Hoverbear, line 66: Could we use [`futures::join_all`](https://docs.rs/futures/0.3.12/futures/future/fn.join_all.html) here? So we could start them in parallel instead of waiting for each?

jonathan-teh: I guess that's possible as a future enhancement but it appears that the existing SPDK behaviour (based on the log messages) is to destroy them sequentially so I'm mostly preserving existing behaviour here, just re-ordering which subsystems to destroy first.

mayastor/src/core/bdev.rs
```diff
@@ -171,9 +171,9 @@ impl Bdev {
                 }
             }
             spdk_sys::SPDK_BDEV_EVENT_RESIZE => {
-                info!("Received resize event for bdev {}", bdev.name())
+                warn!("Received resize event for bdev {}", bdev.name())
```
Hoverbear, line 174: According to [Tracing docs](https://tracing.rs/tracing/struct.level) `warn` is conventionally used only for hazardous situations. This seems to be a normal system oepration, so perhaps `info` was indeed appropriate?

mayastor/src/core/bdev.rs
```diff
@@ -171,9 +171,9 @@ impl Bdev {
...
             }
-            spdk_sys::SPDK_BDEV_EVENT_MEDIA_MANAGEMENT => info!(
+            spdk_sys::SPDK_BDEV_EVENT_MEDIA_MANAGEMENT => warn!(
```
Hoverbear, line 176: Perhaps this log's level was also appropriate at `info`... Would operators need to be aware of management plane events in their honeycomb/datadog instances?

mayastor/src/core/bdev.rs
jonathan-teh: I forgot to note that in the commit message, I changed it to a warning because it's unimplemented. I'll add the author as a reviewer.

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -132,6 +132,9 @@ pub struct NexusChild {
     /// current state of the child
     #[serde(skip_serializing)]
     pub state: AtomicCell<ChildState>,
+    /// destruction started, differentiates remove events
+    #[serde(skip_serializing)]
+    destroying: AtomicCell<bool>,
```
gila, line 137: can we add this state into the child state as a variant?

gila (dismissed):

mayastor/src/bdev/nexus/nexus_child.rs
jonathan-teh: I used a separate "Destroying" state initially but the handling of different states in the remove handler itself seemed to mean that we'd end up with some of those states logical-ORed with the "Destroying" so I thought it was cleaner to just use a separate flag.

test/grpc/test_nexus.js
```diff
@@ -836,6 +846,25 @@ describe('nexus', function () {
       common.jsonrpcCommand('/tmp/target.sock', 'nvmf_subsystem_remove_ns', args, done);
     });
 
+    it('should still have bdev of removed child after remove event', (done) => {
+      common.jsonrpcCommand(null, 'bdev_get_bdevs', (err, out) => {
```
mtzaurus, line 850: Should we also check that nexus child has also changed its state after removing the underlaying namespace ?

mayastor/src/bdev/nexus/mod.rs
```diff
@@ -59,3 +59,17 @@ pub fn nexus_instance_new(name: String, size: u64, children: Vec<String>) {
     let list = instances();
     list.push(Nexus::new(&name, size, None, Some(&children)));
 }
+
+/// called during shutdown so that the bdevs for each child is destroyed first
+pub async fn nexus_destroy_all() {
+    debug!("destroying all nexuses...");
```
tiagolobocastro, line 65: should these be info? will only happen on shutdown and might be useful to see them if for some reason we crash while shutting down

jonathan-teh: I thought I was trying to be consistent with some of the other log entries but at least "NVMe-oF target shutdown completed" and "Unloading Nexus CAS Module" are at info level. I'll change these.

test/grpc/test_nexus.js
jonathan-teh: This is sort of covered by the cargo test replica_stop_cont but that's actually more a data plane test. I guess it won't hurt to add another test here.

jonathan-teh:
bors try

jonathan-teh:
bors retry

mayastor/src/bdev/nexus/nexus_child.rs
gila: Yeah, I can see its easier, and considering the short term goals appropriate, however; we can now open/share a nexus, that is being destroyed? (as we look at the state of the bdev to determine if we want to share?) We also now do `if destroying() { } else { }` which seems a whole lot less ergonomic to me than using [match](https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html) statements.

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -369,8 +378,14 @@ impl NexusChild {
     pub(crate) fn remove(&mut self) {
         info!("Removing child {}", self.name);
 
-        // The bdev is being removed, so ensure we don't use it again.
-        self.bdev = None;
+        // Only remove the bdev if the child is being destroyed instead of a
+        // hot remove event e.g. NVMe AEN (asynchronous event notification)
```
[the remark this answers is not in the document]
jonathan-teh: I may have misheard something you mentioned. I'll remove that part of the comment, certainly there are other reasons for a hot remove event.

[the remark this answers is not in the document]
gila: Sorry I meant to say, these are not NVMe AEN events they *could* be but as the bdev layer, they are generalized.

mayastor/src/bdev/nexus/nexus_child.rs
jonathan-teh: > we can now open/share a nexus, that is being destroyed?
Good point, I'll try to work in a separate "Destroying" state.

jonathan-teh: I've changed the flag to a new ChildState. I can squash the separate commit later.

jonathan-teh:
bors try

jonathan-teh:
bors try

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -397,10 +432,12 @@ impl NexusChild {
...
-        // Dropping the last descriptor results in the bdev being removed.
-        // This must be performed in this function.
-        let desc = self.desc.take();
-        drop(desc);
+        if destroying {
+            // Dropping the last descriptor results in the bdev being removed.
+            // This must be performed in this function.
+            let desc = self.desc.take();
+            drop(desc);
+        }
```
cjones1024, line 440: I suspect the drop is now redundant as desc is about to go out of scope.

cjones1024 (approved):
Looks reasonable to me.

mayastor/src/bdev/nexus/nexus_child.rs
jonathan-teh: Well spotted, I've removed the drop.

jonathan-teh:
bors try

tiagolobocastro (approved):

jonathan-teh:
bors merge

jonathan-teh:
bors try

jonathan-teh:
bors merge

PR #664, opened 2021-01-27
refactor(moac): drop support for NBD in CSI

jkryl:
NBD does not work well with new nexus feature when nexus is created
only if it is used (mounted). For NBD we need to know which node the
nexus is located and the location must not change throughout the
lifetime of the volume. Apparently that does not work now when nexus
is created on whatever node is available and the location can change
between mounts of the volume.

And as part of that convert csi.js to typescript.

Resolves: CAS-672

### DIFF
--- a/csi/moac/.gitignore
+++ b/csi/moac/.gitignore
@@ -1,7 +1,7 @@
 /node_modules/
 /proto/
 /result
-/watcher.js
+/csi.js
 /nexus.js
 /node.js
 /node_operator.js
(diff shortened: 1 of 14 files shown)


gila (approved):
@Jan, oddly enough this song popped into my head when seeing this PR. It clearly shows I need to step away a bit :joy: 

https://www.youtube.com/watch?v=4jAyZ4njHsc

tiagolobocastro (approved):
> And Now His Watch Is Ended

jkryl:
bors merge

jkryl:
bors retry

jkryl:
bors retry

PR #669, opened 2021-01-29
feat(control plane): allow images to be built

paulyoong:
Provide the ability to build container images for the control plane
services and operators.
    
To build the node service, for example:
    nix-build -A images.node-service
or to build a debug version, add the suffix "-dev":
    nix-build -A images.node-service-dev
    
To build the node operator, for example:
    nix-build -A images.node-operator
or to build a debug version, add the suffix "-dev":
    nix-build -A images.node-operator-dev
    
Additional change:
    
Changed the existing image names to remove the "-image" suffix. So the
command to build the mayastor image now looks like:
    nix-build -A images.mayastor
rather than:
    nix-build -A images.mayastor-image

### DIFF
--- a/nix/pkgs/mayastor/default.nix
+++ b/nix/pkgs/mayastor/default.nix
@@ -107,6 +107,10 @@ in
       ../../../target/debug/mayastor-client
       ../../../target/debug/jsonrpc
       ../../../target/debug/kiiss
+      ../../../target/debug/node
+      ../../../target/debug/volume
+      ../../../target/debug/pool
+      ../../../target/debug/rest
     ];
 
     buildInputs = [
--- a/doc/build.md
+++ b/doc/build.md
@@ -217,7 +217,7 @@ short commit hash of the top-most commit or a tag name of the commit if
 present. Example of building a moac package:
 
 ```bash
-nix-build -A images.moac-image
+nix-build -A images.moac
 ```
 
 At the end of the build is printed path to docker image tar archive. Load the
@@ -229,9 +229,9 @@ docker run --rm -it image-hash /bin/bash
 ```
 
 Mayastor and csi plugin images can have multiple flavours. Production image
-name does not contain the flavour name (i.e. `mayastor-image`). Debug image
-contains the `dev` in its name (i.e. `mayastor-dev-image`). Mayastor package
-has additional flavour called `adhoc` (`mayastor-adhoc-image`), that is handy
+name does not contain the flavour name (i.e. `mayastor`). Debug image
+contains the `dev` in its name (i.e. `mayastor-dev`). Mayastor package
+has additional flavour called `adhoc` (`mayastor-adhoc`), that is handy
 for testing because it is not based on mayastor package but rather on whatever
 binaries are present in `target/debug` directory.
 
--- a/nix/pkgs/images/default.nix
+++ b/nix/pkgs/images/default.nix
@@ -51,7 +51,7 @@ let
       mkdir -p var/tmp
     '';
   };
-  servicesImageProps = {
+  clientImageProps = {
     tag = version;
     created = "now";
     config = {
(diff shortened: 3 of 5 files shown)


nix/pkgs/images/operators/default.nix
```diff
@@ -0,0 +1,50 @@
...
+    tag = version;
+    created = "now";
+    config = {
+      Env = [ "PATH=${env}" ];
+    };
+  };
+in
+{
+  # Release image of node operator.
+  node-image = dockerTools.buildLayeredImage (operatorImageProps // {
```
Hoverbear, line 36: I was under the impression from the PR description we were renaming this to `node`? Or is that only in `images.*`? Will we see a non-`image` variant of `node`?

nix/pkgs/images/services/default.nix
```diff
@@ -0,0 +1,109 @@
...
+  serviceImageProps = {
+    tag = version;
+    created = "now";
+    config = {
+      Env = [ "PATH=${env}" ];
+    };
+  };
+in
+{
+  # Release image of kiiss service.
```
Hoverbear, line 35: Similar question about the `-image` postfix, will we see a non `image` variant?

nix/pkgs/mayastor/default.nix
```diff
@@ -107,6 +107,10 @@ in
       ../../../target/debug/mayastor-client
       ../../../target/debug/jsonrpc
       ../../../target/debug/kiiss
+      ../../../target/debug/node
```
Hoverbear, line 110: I am concerned about naming conflicts around these.

Could these be `mayastor-node`?

nix/mayastor-overlay.nix
```diff
@@ -3,11 +3,13 @@ self: super: {
   nvmet-cli = super.callPackage ./pkgs/nvmet-cli { };
   libspdk = (super.callPackage ./pkgs/libspdk { }).release;
   libspdk-dev = (super.callPackage ./pkgs/libspdk { }).debug;
-  mayastor = (super.callPackage ./pkgs/mayastor { }).release;
-  mayastor-dev = (super.callPackage ./pkgs/mayastor { }).debug;
-  mayastor-adhoc = (super.callPackage ./pkgs/mayastor { }).adhoc;
+  mayastor-release = (super.callPackage ./pkgs/mayastor { }).release;
```
Hoverbear, line 6: I think keeping `mayastor` as the canonical name is worthwhile. 

Based on my current understanding of Nix, It is not idiomatic to have `-dev` postfixes in Nix. It seems like this is configured via `--argstr` usually? I feel like this is for future work, though.

nix/mayastor-overlay.nix
```diff
@@ -3,11 +3,13 @@ self: super: {
...
+  mayastor-debug = (super.callPackage ./pkgs/mayastor { }).debug;
+  mayastor-ad-hoc = (super.callPackage ./pkgs/mayastor { }).adhoc;
   moac = (import ./../csi/moac { pkgs = super; }).package;
   images = super.callPackage ./pkgs/images { };
+  services = super.callPackage ./pkgs/images/services { };
```
Hoverbear, line 11: It's unclear to me what the difference between `images` and `services` is? Are the services not images? Do the images (like `mayastor` -- which is a daemon/service) not operate as services?

nix/pkgs/images/operators/default.nix
paulyoong: I was debating what to do here. I removed the `image` suffix from those that were prefixed by `images.*` only as they seemed redundant.

The reason I kept it for the others was that they weren't prefixed with `images.*` so I thought that I should be more explicit here. Really this is a result of me trying to reorganise the images directory. Perhaps I should just put them all into the same place and that would be clearer.

nix/pkgs/images/services/default.nix
paulyoong: Again, I think maybe I should just put all the images into the same place and drop the postfix.

nix/pkgs/mayastor/default.nix
paulyoong: I think maybe they could be renamed to something like `node-service`. A `mayastor-node` tends to indicate a node that mayastor is running on. These are control plane services.

nix/mayastor-overlay.nix
paulyoong: I wanted to keep `mayastor` as the name here but because the name in `images/default.nix` was changed from `mayastor-image` to `mayastor` when you try to build it using `nix-build -A images.mayastor` you get an error message indicating `infinite recursion encountered`

nix/mayastor-overlay.nix
paulyoong: The distinction was meant to be between the new control plane services and the existing images, but maybe that is confusing. I'll keep them with the other images and then I can remove this.

nix/pkgs/images/operators/default.nix
paulyoong: I have removed the `-image` now, but I do make a distinction between `-service` and `-operator`.

nix/pkgs/images/services/default.nix
paulyoong: Done - I have removed the `-image` postfix now.

nix/pkgs/mayastor/default.nix
paulyoong: I would like to leave them as they are for now. We will probably be looking to move all these separate services into a single binary anyway, so this will probably change.

nix/mayastor-overlay.nix
paulyoong: I figured out how to keep what we already had, so it is back to `mayastor` as the canonical name.

nix/mayastor-overlay.nix
paulyoong: I have removed the `services` definition now. Only `images` remain. Apologies for the confusion.

tiagolobocastro (approved):

nix/pkgs/mayastor/default.nix
Hoverbear: Okay!

Hoverbear (approved):

paulyoong:
bors merge

paulyoong:
bors merge

paulyoong:
bors merge

paulyoong:
bors merge

tiagolobocastro:
bors help

paulyoong:
bors merge

paulyoong:
bors merge

paulyoong:
bors merge

paulyoong:
bors merge

PR #672, opened 2021-02-01
fix(moac): serialize volume creation

jkryl:
As part of that, request cache has been implemented, that is used to
detect duplicate CSI requests. All of this improves stability of the
moac if it gets under the load.

e2e stress tests are out of scope and will be implemented on behalf
of a different ticket.

Minor improvement losely related to the changes is workq getting
converted to typescript to take advantage of type checking.

Resolves: CAS-673


csi/moac/node_operator.ts
```diff
@@ -92,7 +92,7 @@ export class NodeOperator {
   ) {
     assert(registry);
     this.namespace = namespace;
-    this.workq = new Workq();
+    this.workq = new Workq('mayastornode');
```
paulyoong, line 95: For the pool operator you have a commenting indicating the workq is for serializing pool operations. Do you want a similar comment here?

paulyoong (approved):

tiagolobocastro (approved):

csi/moac/node_operator.ts
jkryl: Good point. Will add it 👍

jkryl: I just realised that there is already such comment at member definition. So that solves it.

jkryl:
bors merge

PR #673, opened 2021-02-01
ci(helm): clean up deploy yamls and improve helm templates

jkryl:
List of changes:

* use defined storage class from storage-class.yaml in pvc.yaml
* use develop tags on develop branch
* generate pool.yaml from flexible template
* add cpu-count, hugepages and image-pull-policy chart parameters
* avoid hardcoded namespace in yamls
* remove unused mayastor-daemonset-config.yaml
* remove grafana monitor files that aren't maintained anymore
* remove namespace.yaml and use "kubectl create namespace mayastor"
  instead

Future directions:

1. We can use the new helm template to generate pool yaml according
   to our needs in e2e install test avoiding duplicate pool yaml.
2. scripts/generate-deploy-yamls.sh should be a highlevel tool with
   profile name argument (i.e. release, develop, test, ...) to make
   it really useful. For test profile we can generate config with
   two cpus per mayastor, for release we can use just one, etc.
   The point being that this is the recommended config for given env.


deploy/pool.yaml
```diff
@@ -1,10 +1,13 @@
...
+  # Or use the following to generate unique albeit less descriptive pool names
+  #generateName: pool-
   namespace: mayastor
 spec:
-  node: MYHOSTNAME
-  disks: ["MYBLOCKDEVICE"]
-
+  node: worker-0
+  # Just one disk device is supported at this moment
+  disks: ["/dev/nvme1n1"]
```
GlennBullingham, line 13: As the /deploy yaml configuration files also serve as the "minimal viable configuration" example for end users to raise a mayastor cluster, we previously have values for the node and disk fields which should make it clear that they are required to make a substitution against non-viable, example values.

I therefore wonder whether we should be setting them to a value (eg. the /dev/nvme1n1) where it is no longer immediately clear that this not immediately clear and which a user might also take to be a mandatory value which should not be changed?

jkryl: I was wondering whether this is the right change too. There are some pros and cons. Seeing a real value here makes it more clear than anything else what is the expected value. On the other hand it's not easy to see what is supposed to be replaced in this file. What people would think if we used helm patterns here instead of real values because that's what many people in k8s community are familiar with and give an example of real value in a comment above? So it would look like:

```yaml
apiVersion: "openebs.io/v1alpha1"
kind: MayastorPool
metadata:
  # Name can be anything as long as it is unique
  name: pool-on-{{ NODE_NAME }}
  # or let k8s to generate a unique pool name
  #generateName: pool-
  namespace: mayastor
spec:
  node: {{ NODE_NAME }}
  # Just one disk device is supported at this moment (i.e. /dev/nvme1n1)
  disks: ["{{ DEVICE }}"]
```

How does that look? Or are there any other suggestions?

GlennBullingham: In general, I like that.  One very minor (potentially pedantic) point and that is, to me, the use of i.e. ("that is") rather e.g. ("for example)  on the comment line still suggests that this is the only valid value, rather than just an example.

tiagolobocastro: What if the user does not know about helm patterns, might they get confused with the `{{ }}` ?

jkryl: That could happen. On the other hand just as it can happen with any other substitution patterns `<NAME>`, `$NAME`,  etc. The fact that many k8s users (unlike me a while ago ;-) ) have probably some experience with helm templates speaks in favour of using it. But I'm fine with anything that people prefer. I will raise this question in slack to get some feedback on it.

jkryl: Please, ignore the change.  Helm is unable to work with a value that looks like a replacement pattern. All my attempts to escape the value failed. Falling back to uppercase names as we had before. The new comments will stay.

tiagolobocastro (approved):

GlennBullingham (approved):

jkryl:
bors merge

jkryl:
bors retry

jkryl:
bors retry

jkryl:
bors retry

jkryl:
Closing PR because bors does not want to merge it. I will open a new one.

PR #675, opened 2021-02-04
feat(nvmx): i/o timeouts for nvme controller

mtzaurus:
This fix introduces tuneable actions for I/O timeouts for NVMe
controllers (Abort, Reset and Ignore), which are now configurable
in place for running controllers and allow setting different timeout
actions for different controllers.

State transitions for NVMe controllers are now handled by a dedicated
state machine entity, which is now part of the global state of every
controller. Such an approach provides better state encapsulation and
validation and is less error prone than direct state manipulations.

Implements CAS-659

### DIFF
--- a/mayastor/src/bdev/dev.rs
+++ b/mayastor/src/bdev/dev.rs
@@ -31,6 +31,7 @@ use crate::{
     nexus_uri::{self, NexusBdevError},
 };
 
+mod malloc;
 mod nvmx;
 
 impl Uri {
@@ -47,6 +48,7 @@ impl Uri {
         match url.scheme() {
             // backend NVMF target - fairly unstable (as of Linux 5.2)
             "nvmf" => Ok(Box::new(nvmx::NvmfDeviceTemplate::try_from(&url)?)),
+            "malloc" => Ok(Box::new(malloc::Malloc::try_from(&url)?)),
 
             scheme => Err(NexusBdevError::UriSchemeUnsupported {
                 scheme: scheme.to_string(),
--- a/mayastor/src/bdev/dev/nvmx/channel.rs
+++ b/mayastor/src/bdev/dev/nvmx/channel.rs
@@ -191,7 +191,7 @@ extern "C" fn disconnected_qpair_cb(
     qpair: *mut spdk_nvme_qpair,
     _ctx: *mut c_void,
 ) {
-    warn!("NVMe qpair disconnected !");
+    warn!(?qpair, "NVMe qpair disconnected");
     /*
      * Currently, just try to reconnect indefinitely. If we are doing a
      * reset, the reset will reconnect a qpair and we will stop getting a
@@ -226,7 +226,7 @@ impl NvmeControllerIoChannel {
 
         debug!("Creating IO channel for controller ID 0x{:X}", id);
 
-        let controller = match NVME_CONTROLLERS.lookup_by_name(id.to_string()) {
+        let carc = match NVME_CONTROLLERS.lookup_by_name(id.to_string()) {
             None => {
                 error!("No NVMe controller found for ID 0x{:X}", id);
                 return 1;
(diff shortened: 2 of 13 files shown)


mayastor/src/bdev/dev/nvmx/controller_inner.rs
```diff
@@ -0,0 +1,333 @@
...
+                        qpair,
+                        cid,
+                        Some(NvmeController::command_abort_handler),
+                        cb_arg,
+                    );
+                    if rc == 0 {
+                        info!(
+                            "{}: initiated abort for CID {}",
+                            timeout_cfg.name, cid
+                        );
```
jonathan-teh, line 264: I think it's slightly cleaner to return early here, then the next test for `timeout_action == Reset` or `rc != 0` can be removed altogether.

mayastor/src/bdev/dev/nvmx/controller.rs
```diff
@@ -194,7 +178,7 @@ impl<'a> NvmeController<'a> {
         })
     }
 
-    /// populate name spaces, at current we only populate the first namespace
+    /// populate name spaces, current we only populate the first namespace
```
jonathan-teh, line 181: currently?

mayastor/src/bdev/dev.rs
```diff
@@ -47,6 +48,7 @@ impl Uri {
         match url.scheme() {
             // backend NVMF target - fairly unstable (as of Linux 5.2)
             "nvmf" => Ok(Box::new(nvmx::NvmfDeviceTemplate::try_from(&url)?)),
+            "malloc" => Ok(Box::new(malloc::Malloc::try_from(&url)?)),
```
jonathan-teh, line 51: This file looks quite different from the upstream version. If it's going to replace it, the comment for the malloc URI should be copied across.

jonathan-teh (approved):
I think this looks OK

mayastor/src/bdev/dev/nvmx/channel.rs
```diff
@@ -191,7 +191,7 @@ extern "C" fn disconnected_qpair_cb(
     qpair: *mut spdk_nvme_qpair,
     _ctx: *mut c_void,
 ) {
-    warn!("NVMe qpair disconnected !");
+    warn!("NVMe qpair disconnected, qpair={:p}", qpair);
```
Hoverbear, line 194: We use [`tracing`](https://github.com/tokio-rs/tracing) so you could use this in a structured manner!

https://github.com/openebs/Mayastor/blob/ce2ca62c841b5715e6c2b482f3d614023025f618/mayastor/Cargo.toml#L75-L79

You can see ["Recording fields"](https://tracing-rs.netlify.app/tracing/index.html#recording-fields), maybe like:

```rust
tracing::warn!(
    ?qpair,
    "NVMe qpair disconnected",
);
```

mayastor/src/bdev/dev/nvmx/channel.rs
```diff
@@ -226,24 +226,36 @@ impl NvmeControllerIoChannel {
...
-            error!(
-                "{} controller is in {:?} state, I/O channel creation not possible",
-                controller.get_name(),
-                controller.get_state()
-            );
-            return 1;
+        let cname;
+        let spdk_handle;
+
+        {
```
Hoverbear, line 240: Not an issue, but I'd note Rust is fine with multiple returns via a tuple:

```suggestion
        let (cname, spdk_handle) = {
```

Then you can later return:

```rust
(controller.get_name(), controller.ctrlr_as_ptr())
```

mayastor/src/bdev/dev.rs
mtzaurus: We will restore all other required URI schemes when merging with the upstream. Now its a WIP.

mayastor/src/bdev/dev/nvmx/channel.rs
mtzaurus: Thanks, fixed.

mayastor/src/bdev/dev/nvmx/channel.rs
mtzaurus: Thanks for the suggestion, the code looks nicer. Fixed.

mayastor/src/bdev/dev/nvmx/controller_inner.rs
mtzaurus: Indeed the code looks shorter than. Thanks.

PR #682, opened 2021-02-08
ci: generate logs as jenkins artefacts CAS-692

blaisedias:


blaisedias:
bors try

blaisedias:
bors try

blaisedias:
bors try

blaisedias:
bors try

blaisedias:
bors try

blaisedias:
bors try

blaisedias:
bors try

Jenkinsfile
```diff
@@ -206,7 +206,12 @@ pipeline {
                     fingerprintArtifacts: true
                 )
                 sh 'kubectl get nodes -o wide'
-                sh "nix-shell --run './scripts/e2e-test.sh --device /dev/sdb --tag \"${env.GIT_COMMIT_SHORT}\" --registry \"${env.REGISTRY}\"'"
+                sh "nix-shell --run './scripts/e2e-test.sh --device /dev/sdb --tag \"${env.GIT_COMMIT_SHORT}\" --registry \"${env.REGISTRY}\" --logsdir ./logs/mayastor --logs '"
```
jkryl, line 209: This is a comment about logging in general. From usability perspective it seems that having both options, logsdir and logs, does not make sense. Most of the time I want either the logs to be generated in which case I specify logsdir or I don't want the logs in which case I run it without the option. I mean, who would like to see all logs printed to stdout if we can save them to files now. Is there such use case?

scripts/e2e-test.sh
```diff
@@ -192,6 +192,7 @@ done
 
 if [ "$generate_logs" -ne 0 ]; then
     if [ -n "$logsdir" ]; then
+        mkdir -p "$logsdir"
```
jkryl, line 195: not directly related but on the line below: `"$SCRIPTDIR"/e2e-cluster-dump.sh` - I find it very unusual to use quoting this way and looks unnatural to me. The whole string should be quoted either like this:
```
"${SCRIPTDIR}/something"
```
or like this
```
"$SCRIPTDIR/something"
```
does not matter in this case because the end of variable name is clearly delimited by the slash.

But if I'm the only one who finds this quoting pattern weird then feel free to ignore me (cc @arne-rusek ) :-)

Jenkinsfile
```diff
@@ -206,7 +206,12 @@ pipeline {
...
+              }
+              post {
+                always {
```
jkryl, line 212: I think we want to save logs only in case of an error and not always.

jkryl (commented):
The change looks good! Just a couple of related comments.

scripts/e2e-test.sh
arne-rusek: Personally I tend to quote whole string (as it looks nicer/more readable when syntax-highlighted to me), except when I don't e.g. when using glob: `"$SCRIPTDIR/"*` I am not aware of any repercussions doing it either way as long as variables are quoted and Having said that I would keep this consistent throughout the project, but just for the sake of being consistent (avoiding confusion in reader that is not bash-expert).

blaisedias:
bors try

Jenkinsfile
blaisedias: Capture  logs always and dispose regularly is more useful approach.

Jenkinsfile
blaisedias: The semantics of --logs is `always generate logs` , if not specified logs are only generated on test failure.
--logsdir specifies where the log files (if any) are the be generated.

arne-rusek (approved):
LGTM

scripts/e2e-test.sh
blaisedias: fixed

blaisedias:
bors try

jkryl (approved):

blaisedias:
bors try

blaisedias:
bors merge

blaisedias:
bors merge

PR #685, opened 2021-02-09
Feature/cas 628

mtzaurus:
refactor(nvmx): more abstractions for spdk objects

This fix introduces better API abstractions for such SPDK objects as
i/o qpair and poll group. All low-level SPDK handles are hidden
inside high-level Rust wrappers, which incapsulates low-level SPDK
library functions.
As a plus, high-level wrappers allow using Options instead of raw
pointers, which makes code look a lot better.

    Implements CAS-628


mtzaurus:
Will submit another PR.

PR #686, opened 2021-02-09
ci: add post 3rd party test resource check CAS-691

blaisedias:
The CSI test suite is a 3rd party test suite.
We would like to apply stringent resource leakage checks for
each test.
We have test code which does this, it makes sense to reuse this.
However it is not straightforward to integrate this with the
3rd party source code.
Doing so makes it more difficult to integrate future changes in
the 3rd party source code.

Create a simple resource check test which is sequenced to run after
each 3rd party test suite.

Rename CheckNo* functions to the more apt CheckFor*

### DIFF
--- a/scripts/e2e-test.sh
+++ b/scripts/e2e-test.sh
@@ -5,9 +5,16 @@ set -eu
 SCRIPTDIR=$(dirname "$(realpath "$0")")
 TESTDIR=$(realpath "$SCRIPTDIR/../test/e2e")
 REPORTSDIR=$(realpath "$SCRIPTDIR/..")
-# new tests should be added before the replica_pod_remove test
+
+# List and Sequence of tests.
 #tests="install basic_volume_io csi replica rebuild node_disconnect/replica_pod_remove uninstall"
-tests="install basic_volume_io csi uninstall"
+# Restrictions:
+#   1. resource_check MUST follow csi 
+#       resource_check is a follow up check for the 3rd party CSI test suite.
+#   2. replicas_pod_remove SHOULD be the last test before uninstall
+#       this is a disruptive test.
+tests="install basic_volume_io csi resource_check uninstall"
+
 device=
 registry=
 tag="ci"
--- a/test/e2e/resource_check/resource_check_test.go
+++ b/test/e2e/resource_check/resource_check_test.go
@@ -1,0 +1,72 @@
+package basic_test
+
+import (
+	"e2e-basic/common"
+	"os"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/reporters"
+	. "github.com/onsi/gomega"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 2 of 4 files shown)


scripts/e2e-test.sh
```diff
@@ -5,9 +5,16 @@ set -eu
...
-# new tests should be added before the replica_pod_remove test
+
+# List and Sequence of tests.
 #tests="install basic_volume_io csi replica rebuild node_disconnect/replica_pod_remove uninstall"
-tests="install basic_volume_io csi uninstall"
+# Restrictions:
+#   1. resource_check MUST follow csi 
+#       resource_check is a follow up check for the 3rd party CSI test suite.
+#   2. replicas_pod_remove SHOULD be the last test before uninstall
+#       this is a disruptive test.
```
paulyoong, line 15: This comment is important but the _replicas_pod_remove_ test isn't in the list of tests. I wonder if this should be moved above line 10 where it is commented out? Then, when it is included, the comment can be updated.

test/e2e/resource_check/resource_check_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+// Check that there are no artefacts left over from
+// the previous 3rd party test.
+func resourceCheck() {
+
+	found, err := common.CheckNoTestPods()
+	if err != nil {
+		logf.Log.Error(err, "Failed to checking for test pods.")
```
paulyoong, line 21: checking -> check

test/e2e/resource_check/resource_check_test.go
```diff
@@ -0,0 +1,72 @@
...
+	} else {
+		Expect(found).To(BeFalse())
+	}
+
+	found, err = common.CheckNoPVCs()
```
paulyoong, line 26: Nit: I assumed that if we were calling "CheckNoPVCs" that it would return true if there were no PVCs; because we are checking that there are no PVCs. Maybe the function name could be changed to something like "HasPVCs"?

paulyoong (approved):

tiagolobocastro (approved):

scripts/e2e-test.sh
chriswldenyer: It should be back in, e.g. in back-to-back testing fairly soon.

test/e2e/resource_check/resource_check_test.go
blaisedias: Done.

test/e2e/resource_check/resource_check_test.go
blaisedias: There are a bunch of functions prefixed with CheckNo, I will change all the prefixes to CheckFor

jkryl (approved):

blaisedias:
bors merge

blaisedias:
bors merge

blaisedias:
bors merge

PR #687, opened 2021-02-09
refactor(nvmx): more abstractions for spdk objects

mtzaurus:
This fix introduces better API abstractions for such SPDK objects as
i/o qpair and poll group. All low-level SPDK handles are hidden
inside high-level Rust wrappers, which incapsulates low-level SPDK
library functions.
As a plus, high-level wrappers allow using Options instead of raw
pointers, which makes code look a lot better.

Implements CAS-628

### DIFF
--- a/mayastor/src/bdev/dev/nvmx/channel.rs
+++ b/mayastor/src/bdev/dev/nvmx/channel.rs
@@ -1,9 +1,12 @@
 /* I/O channel for NVMe controller, one per core. */
 
 use crate::{
-    bdev::dev::nvmx::{NvmeControllerState, NVME_CONTROLLERS},
-    core::poller,
-    subsys::NvmeBdevOpts,
+    bdev::dev::nvmx::{
+        nvme_bdev_running_config,
+        NvmeControllerState,
+        NVME_CONTROLLERS,
+    },
+    core::{poller, CoreError},
 };
 use std::{cmp::max, mem::size_of, os::raw::c_void, ptr::NonNull};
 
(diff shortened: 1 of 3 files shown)


mayastor/src/bdev/dev/nvmx/channel.rs
```diff
@@ -58,13 +61,139 @@ impl<'a> NvmeIoChannel<'a> {
...
+
+        let qpair: *mut spdk_nvme_qpair = unsafe {
+            spdk_nvme_ctrlr_alloc_io_qpair(
+                ctrlr_handle,
+                &qpair_opts,
+                size_of::<spdk_nvme_io_qpair_opts>() as u64,
+            )
+        };
+
+        if qpair.is_null() {
```
tiagolobocastro, line 106: might be able to avoid 1 unwrap by reversing the logic to: if let Some(qpair) = NonNull::new(qpair) { ... Self { qpair, ... }}

mtzaurus: Yup, indeed. Thanks !

mayastor/src/bdev/dev/nvmx/channel.rs
```diff
@@ -58,13 +61,139 @@ impl<'a> NvmeIoChannel<'a> {
...
+            error!(?ctrlr_name, "Failed to allocate I/O qpair for controller",);
+            Err(CoreError::GetIoChannel {
+                name: ctrlr_name.to_string(),
+            })
+        } else {
+            debug!(?qpair, ?ctrlr_name, "qpair created for controller");
+            Ok(Self {
+                qpair: NonNull::new(qpair).unwrap(),
+                ctrlr_handle: NonNull::new(ctrlr_handle).unwrap(),
```
tiagolobocastro, line 115: do we need to return an error if ctrlr_handle is null?

mtzaurus: If it's null, qpair allocation will fail and we end up in 'if qpair.is_null() {' branch.

mtzaurus: But I can add an assertion against null, as null must not be passed.

mayastor/src/bdev/dev/nvmx/channel.rs
```diff
@@ -134,54 +250,43 @@ impl NvmeIoChannelInner<'_> {
...
-                &opts,
-                size_of::<spdk_nvme_io_qpair_opts>() as u64,
+        // We assume that channel is reinitialized after being reset, so we
+        // expect to see no I/O qpair.
+        if self.qpair.is_some() {
+            warn!(
+                ?ctrlr_name,
+                "I/O channel has active I/O qpair while being reinitialized, clearing"
             );
+            self.qpair.take().unwrap();
```
tiagolobocastro, line 260: hmm could we avoid unwrap by using like: if let Some(q) = q.take() ?

mtzaurus: Good point.

tiagolobocastro (approved):

Hoverbear (approved):

mayastor/src/bdev/dev/nvmx/channel.rs
```diff
@@ -58,13 +61,139 @@ impl<'a> NvmeIoChannel<'a> {
...
+    fn add_qpair(&mut self, qpair: &IoQpair) -> i32 {
+        unsafe { spdk_nvme_poll_group_add(self.0.as_ptr(), qpair.as_ptr()) }
+    }
+
+    /// Remove I/O qpair to poll group.
+    fn remove_qpair(&mut self, qpair: &IoQpair) -> i32 {
+        unsafe { spdk_nvme_poll_group_remove(self.0.as_ptr(), qpair.as_ptr()) }
+    }
+
+    /// Get SPDK handle for poll group.
```
cjones1024, line 163: is it worth adding an #[inline] here

mtzaurus: Good idea, thanks !

PR #688, opened 2021-02-09
fix(nexus): set children to destroying state on shutdown

jonathan-teh:
Set all children to Destroying state instead to address possible
segfault on shutdown when destroying nexuses.

Fixes CAS-697

### DIFF
--- a/mayastor/src/bdev/nexus/mod.rs
+++ b/mayastor/src/bdev/nexus/mod.rs
@@ -60,16 +60,14 @@ pub fn nexus_instance_new(name: String, size: u64, children: Vec<String>) {
     list.push(Nexus::new(&name, size, None, Some(&children)));
 }
 
-/// called during shutdown so that the bdevs for each child is destroyed first
-pub async fn nexus_destroy_all() {
-    info!("destroying all nexuses...");
+/// called during shutdown so that all nexus children are in Destroying state
+/// so that a possible remove event from SPDK also results in bdev removal
+pub async fn nexus_children_to_destroying_state() {
+    info!("setting all nexus children to destroying state...");
     for nexus in instances() {
-        if let Err(e) = nexus.destroy().await {
-            error!(
-                "failed to destroy nexus {} during shutdown: {}",
-                nexus.name, e
-            );
+        for child in nexus.children.iter() {
+            child.set_state(nexus_child::ChildState::Destroying);
         }
     }
-    info!("destroyed all nexuses");
+    info!("set all nexus children to destroying state");
 }
--- a/mayastor/src/core/env.rs
+++ b/mayastor/src/core/env.rs
@@ -279,7 +279,7 @@ async fn do_shutdown(arg: *mut c_void) {
     }
 
     iscsi::fini();
-    nexus::nexus_destroy_all().await;
+    nexus::nexus_children_to_destroying_state().await;
     unsafe {
         spdk_rpc_finish();
         spdk_subsystem_fini(Some(reactors_stop), arg);


jonathan-teh:
bors try

paulyoong (approved):

jkryl:
Hi Jon, could you explain more about the problem? What you do is clear, but what is unclear to me is what was the problem before and how setting state of children instead of destroying nexus is going to fix the problem. Thanks!

jonathan-teh:
This fixes an issue that was introduced by the fix to #661, which introduced a new Destroying state for the nexus child which is set when calling `NexusChild::destroy()` so that the subsequent remove event from SPDK removes the reference to the bdev only the child was in Destroying state.
It turns out that remove events can occur during shutdown, such as with the `fault_child` test, but it doesn't remove the reference to the bdev in that case and the mayastor process hangs. To address that, it was decided to destroy all nexuses during shutdown but it introduced the issue in CAS-697 instead.

mayastor/src/bdev/nexus/mod.rs
```diff
@@ -60,16 +60,14 @@ pub fn nexus_instance_new(name: String, size: u64, children: Vec<String>) {
...
 
-/// called during shutdown so that the bdevs for each child is destroyed first
-pub async fn nexus_destroy_all() {
-    info!("destroying all nexuses...");
+/// called during shutdown so that all nexus children are in Destroying state
+/// so that a possible remove event from SPDK also results in bdev removal
+pub async fn nexus_children_to_destroying_state() {
+    info!("setting all nexus children to destroying state...");
     for nexus in instances() {
-        if let Err(e) = nexus.destroy().await {
```
mtzaurus, line 67: Can you please clarify where the nexus is destroyed, since nexus.destroy() is removed by this fix ?

jonathan-teh: It's not explicitly destroyed, this essentially restores the original behaviour prior to #661. The nexus instances themselves are cleared as part of shutdown when `NexusModule::nexus_mod_fini()` is called.

mtzaurus: Thanks for clarification.

mtzaurus:
> This fixes an issue that was introduced by the fix to #661, which introduced a new Destroying state for the nexus child which is set when calling `NexusChild::destroy()` so that the subsequent remove event from SPDK removes the reference to the bdev only the child was in Destroying state.
> It turns out that remove events can occur during shutdown, such as with the `fault_child` test, but it doesn't remove the reference to the bdev in that case and the mayastor process hangs. To address that, it was decided to destroy all nexuses during shutdown but it introduced the issue in CAS-697 instead.

Shouldn't all the nexus children be destroyed as part of the shutdown too, as part of Nexus removal ? NexusChild::destroy() will be called then (I presume so, as all Nexus devices should be removed upon Nexus removal), so children will transition into Destroying state, hence the device reference will be removed properly in response to SPDK events.

jonathan-teh:
> Shouldn't all the nexus children be destroyed as part of the shutdown too, as part of Nexus removal ?
That doesn't seem to be necessary, it didn't do that prior to #661. Destroying nexus children also records the state in the config file, which is an unintended consequence.

mtzaurus:
> That doesn't seem to be necessary, it didn't do that prior to #661. Destroying nexus children also records the state in the config file, which is an unintended consequence.

Right, it's shutdown, not removal. :-)

mtzaurus (approved):

jonathan-teh:
bors merge

cjones1024 (approved):
Looks reasonable

jonathan-teh:
bors merge

jonathan-teh:
bors merge

jonathan-teh:
bors merge

PR #691, opened 2021-02-09
test: send xray reports

chriswldenyer:
test: send xray reports

Send reports to Jira xray. Specify which test spec in the
Jenkinsfile. Currently implemented to on-demand E2E tests.
Updated one readme.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,6 +9,8 @@ def e2e_destroy_cluster_job='k8s-destroy-cluster' // Jenkins job to destroy clus
 def e2e_environment="hcloud-kubeadm"
 // Global variable to pass current k8s job between stages
 def k8s_job=""
+def xray_projectkey='MQ'
+def xray_e2e_testplan='MQ-1'
 
 // Searches previous builds to find first non aborted one
 def getLastNonAbortedBuild(build) {
@@ -237,6 +239,19 @@ pipeline {
                         "Note: you need to click `proceed` and will get an empty page when using destroy link. " +
                         "(<https://mayadata.atlassian.net/wiki/spaces/MS/pages/247332965/Test+infrastructure#On-Demand-E2E-K8S-Clusters|doc>)"
                     )
+                  }
+                }
+                always { // always send the junit results back to Xray and Jenkins
+                  junit 'e2e.*.xml'
+                  script {
+                    step([
+                      $class: 'XrayImportBuilder',
+                      endpointName: '/junit',
+                      importFilePath: 'e2e.*.xml',
+                      importToSameExecution: 'true',
+                      projectKey: "${xray_projectkey}",
+                      testPlanKey: "${xray_e2e_testplan}",
+                      serverInstance: "${env.JIRASERVERUUID}"])
                   }
                 }
               }
--- a/test/e2e/node_disconnect/README.md
+++ b/test/e2e/node_disconnect/README.md
@@ -1,25 +1,9 @@
-## Note
-The tests in directories replica_disconnect and replica_reassign
-are not currently deployable by the CI system
-as those tests assume a vagrant installation.
-
 ## Pre-requisites for replica_pod_remove
 
-* A Kubernetes cluster with 3 nodes, with mayastor installed.
-
-## Pre-requisites for the other directories
-
-* A Kubernetes cluster with at least 3 nodes, with mayastor installed.
-* The replica_reassign test requires at least 4 nodes.
-* The cluster is deployed using vagrant and KUBESPRAY_REPO is correctly
-  defined in ``` ./lib/io_connect_node.sh ```
+* A Kubernetes cluster with at least 2 nodes, with mayastor installed.
 
 ## Overview
-The tests verify the behaviour of the cluster under fault conditions
-affecting the availability of resources in the cluster, for example
-a missing mayastor pod or a disconnected node.
+This directory is for tests verify the behaviour of the cluster under
+fault conditions affecting the availability of resources in the
+cluster, for example a missing mayastor pod or a disconnected node.
 
-To run, cd to the test directory then:
-```bash
-go test
-```
--- a/test/e2e/basic_volume_io/basic_volume_io_test.go
+++ b/test/e2e/basic_volume_io/basic_volume_io_test.go
@@ -4,7 +4,7 @@ package basic_volume_io_test
 
 import (
 	"e2e-basic/common"
-	"os"
+	"e2e-basic/common/junit"
 	"testing"
 
 	. "github.com/onsi/ginkgo"
@@ -26,8 +26,8 @@ var volNames []volSc
 
 func TestBasicVolumeIO(t *testing.T) {
 	RegisterFailHandler(Fail)
-	reportDir := os.Getenv("e2e_reports_dir")
-	junitReporter := reporters.NewJUnitReporter(reportDir + "/basic-volume-io-junit.xml")
+	junitReporter := reporters.NewJUnitReporter(junit.ConstructJunitFileName("basic-volume-io-junit.xml"))
+
 	RunSpecsWithDefaultAndCustomReporters(t, "Basic volume IO tests, NVMe-oF TCP and iSCSI", []Reporter{junitReporter})
 }
 
--- a/test/e2e/common/junit/junit.go
+++ b/test/e2e/common/junit/junit.go
@@ -1,0 +1,11 @@
+package junit
+
+import (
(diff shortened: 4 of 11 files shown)


chriswldenyer:
bors try

chriswldenyer:
bors try

blaisedias (approved):
LGTM

Jenkinsfile
```diff
@@ -192,21 +193,43 @@ pipeline {
...
+                  script {
+                    step([
+                      $class: 'XrayImportBuilder',
+                      endpointName: '/junit',
+                      importFilePath: 'e2e_junit/*.xml',
+                      importToSameExecution: 'true',
+                      projectKey: "${projectkey}",
+                      testPlanKey: "${env.TESTPLAN}",
+                      serverInstance: "${env.JIRASERVERUUID}"])
+                    sh 'rm -f e2e_junit/*.xml'
```
arne-rusek, line 230: Please do not remove it. It will be automatically removed when new job is run in the same workspace and can be useful for debugging later. Or is it required?
```suggestion
```

test/e2e/node_disconnect/README.md
```diff
@@ -1,25 +1,9 @@
...
 
-* A Kubernetes cluster with 3 nodes, with mayastor installed.
-
-## Pre-requisites for the other directories
-
-* A Kubernetes cluster with at least 3 nodes, with mayastor installed.
-* The replica_reassign test requires at least 4 nodes.
-* The cluster is deployed using vagrant and KUBESPRAY_REPO is correctly
-  defined in ``` ./lib/io_connect_node.sh ```
+* A Kubernetes cluster with at least 2 nodes, with mayastor installed.
```
arne-rusek, line 3: E2E clusters always have 3 nodes. Is it really true? (We might as well lower the number of nodes on e2e clusters)

Jenkinsfile
```diff
@@ -192,21 +193,43 @@ pipeline {
...
+                always { // always send the junit results back to Xray and Jenkins
+                  junit 'e2e_junit/*.xml'
+                  script {
+                    step([
+                      $class: 'XrayImportBuilder',
+                      endpointName: '/junit',
+                      importFilePath: 'e2e_junit/*.xml',
+                      importToSameExecution: 'true',
+                      projectKey: "${projectkey}",
+                      testPlanKey: "${env.TESTPLAN}",
```
arne-rusek, line 228:
```suggestion
                      testPlanKey: "${xray_testplan}",
```

Jenkinsfile
```diff
@@ -192,21 +193,43 @@ pipeline {
...
+                      selector: specific("${k8s_job.getNumber()}"),
+                      filter: "${e2e_environment}/modules/k8s/secrets/admin.conf",
+                      target: "",
+                      fingerprintArtifacts: true
+                  )
+                  sh 'kubectl get nodes -o wide'
+                  sh 'mkdir -p ./e2e_junit'
+                  def pd = sh(script: 'pwd', returnStdout:true).trim()
+                  def junitdir = "${pd}/e2e_junit"
+                  sh "nix-shell --run './scripts/e2e-test.sh --device /dev/sdb --tag \"${env.GIT_COMMIT_SHORT}\" --registry \"${env.REGISTRY}\" --reportsdir \"${junitdir}\"'"
```
arne-rusek, line 215:
```suggestion
                  sh """
                    mkdir -p e2e_junit
                    nix-shell --run './scripts/e2e-test.sh --device /dev/sdb --tag "${env.GIT_COMMIT_SHORT}" --registry "${env.REGISTRY}" --reportsdir e2e_junit'
                  """
```

This way you could also remove the additional `script {` nesting level as IMHO is not needed anymore.

If `e2e-test.sh` requires absolute path, than that's a bug in the script as I imagine one can run it manually and it should work with any user specified path. If that is indeed the case, please call `realpath` on `--reportsdir` value in the script itself. This looks like hot-fixing the `e2e-script.sh` in a wrapper.

arne-rusek (changes requested):

Jenkinsfile
```diff
@@ -9,6 +9,7 @@ def e2e_destroy_cluster_job='k8s-destroy-cluster' // Jenkins job to destroy clus
 def e2e_environment="hcloud-kubeadm"
 // Global variable to pass current k8s job between stages
 def k8s_job=""
+def projectkey="MQ"
```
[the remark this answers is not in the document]
chriswldenyer: Ok, will do.

Jenkinsfile
```diff
@@ -192,21 +193,43 @@ pipeline {
...
+              post {
+                always { // always send the junit results back to Xray and Jenkins
+                  junit 'e2e_junit/*.xml'
+                  script {
+                    step([
+                      $class: 'XrayImportBuilder',
+                      endpointName: '/junit',
+                      importFilePath: 'e2e_junit/*.xml',
+                      importToSameExecution: 'true',
+                      projectKey: "${projectkey}",
```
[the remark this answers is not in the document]
chriswldenyer: Will do.

Jenkinsfile
```diff
@@ -192,21 +193,43 @@ pipeline {
                   returnStdout: true
                 )
                 KUBECONFIG = "${env.WORKSPACE}/${e2e_environment}/modules/k8s/secrets/admin.conf"
+                TESTPLAN = "MQ-1"
```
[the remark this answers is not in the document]
chriswldenyer: It is likely that different parts of the testing, e.g. grpc could get sent to a different test plan. That is Glenn's suggestion, so I wanted to make the variable local to the stage. The project, on the other hand will always be "MQ", which is why I had that at the top.
I will add the xray refix.

Jenkinsfile
chriswldenyer: In my test pipeline they seem to hang around between runs. Maybe that is different in the Mayastor pipeline. I was just concerned that if we remove a test, old copies of the junit files will still be sent.
If you say they do get cleaned up, I will remove the line.

test/e2e/node_disconnect/README.md
chriswldenyer: They don't have to have 3 nodes, (but my test needs at least 2 replicas). For some reason we have settled on using 3 in our setups. Maybe this is worth a wider discussion?

Jenkinsfile
chriswldenyer: Will change.

Jenkinsfile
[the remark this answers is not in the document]
chriswldenyer: Changed my mind. Moved it to the top. I'll deal with multiple test plans when the need arises.

Jenkinsfile
chriswldenyer: I'll move the rm to before the test run, so we get to examine the files and they won't corrupt later runs.

Jenkinsfile
chriswldenyer: Talking to Blaise about this, I think I'll avoid the use of a subdirectory altogether as it sounds preferable to keep the xml files in the default location, which is the repository root.

Jenkinsfile
arne-rusek: They **cannot** corrupt later runs as jenkins cleans up all untracked files before running the next job (`git clean -fdx` FYI we even tested it with @jkryl and it works) you do not need to clean up anything.

Jenkinsfile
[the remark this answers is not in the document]
arne-rusek: FYI groovy supports maps, so you can have configuration per stage nicely set up at one spot and e.g. index by stage name. https://stackoverflow.com/questions/42770775/how-to-define-and-iterate-over-map-in-jenkinsfile
https://stackoverflow.com/questions/36502945/access-stage-name-during-the-build-in-jenkins-pipeline/47268363

test/e2e/node_disconnect/README.md
arne-rusek: It was to be able to accommodate the disconnect test IIRC. @blaisedias might remember as in some PR he changed it from 2 to 3 :) That's why I'm surprised we're changing it back to 2. Anyway. I would suggest that if we change it to 2, we also should change node count for e2e clusters so we actually make sure it's true and 2 is enough?

Jenkinsfile
chriswldenyer: Ok fair enough, I didn't know about the git clean action.

chriswldenyer:
bors try

chriswldenyer:
Close pending further changes

PR #693, opened 2021-02-10
ci(helm): clean up deploy yamls and improve helm templates

jkryl:
List of changes:

* use defined storage class from storage-class.yaml in pvc.yaml
* use develop tags on develop branch
* generate pool.yaml from flexible template
* add cpu-count, hugepages and image-pull-policy chart parameters
* avoid hardcoded namespace in yamls
* remove unused mayastor-daemonset-config.yaml
* remove grafana monitor files that aren't maintained anymore
* remove namespace.yaml and use "kubectl create namespace mayastor"
  instead

Future directions:

1. We can use the new helm template to generate pool yaml according
   to our needs in e2e install test avoiding duplicate pool yaml.
2. scripts/generate-deploy-yamls.sh should be a highlevel tool with
   profile name argument (i.e. release, develop, test, ...) to make
   it really useful. For test profile we can generate config with
   two cpus per mayastor, for release we can use just one, etc.
   The point being that this is the recommended config for given env.


GlennBullingham (approved):

paulyoong (approved):

jkryl:
bors merge

jkryl:
bors retry

PR #694, opened 2021-02-10
chore: do profile-based yaml generation

jkryl:
The purpose of profiles is to provide reasonable defaults for yaml
templates for each of the major use cases:

* develop: used by developers to run mayastor
* release: used for stable releases of mayastor
* test: used by mayastor e2e tests

In addition to that I have:

* increased debug level of moac when running in develop env
* fixed wrong usage of -l parameter (indexing starts at 0)
* introduced rule of thumb for cpu count and hugepage mem (1:1)

Resolves: CAS-699

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -1,8 +1,12 @@
-mayastorImagesTag: develop
+mayastorImagesTag: latest
 mayastorImagePullPolicy: Always
-mayastorCpuCount: "2"
+mayastorCpuCount: "1"
 mayastorHugePagesGiB: "1"
-mayastorImagesRepo: ""
+mayastorImagesRegistry: ""
 mayastorPools:
   - node: "NODE_NAME"
     device: "DEVICE"
+# This option is intended for development yamls and motivated by the problem of
+# moac that does not update status of msp resource in some cases. Feel free to
+# remove when no longer needed.
+moacDebug: false
--- a/deploy/mayastor-daemonset.yaml
+++ b/deploy/mayastor-daemonset.yaml
@@ -47,21 +47,15 @@ spec:
         - name: IMPORT_NEXUSES
           value: "false"
         args:
-        # In order to select what cores mayastor should be running on, a mask or a list can be specified.
-        # For example: -m 0x1 will tell mayastor to only use one core which is equivalent to -l 1
-        # Using a mask of 0x3 will use the first 2 cores, which is equivalent to -l 1-2
-        #
-        # The -l argument supports ranges to be able to do the same as passing a mask for example:
-        # -l 1,2,10-20 means use core 1, 2, 10 to 20
-        #
-        # Note:
-        # 1. When both -m and -l are specified the -l argument is takes precedence.
-        # 2. Ensure that the CPU resources are updated accordingly. If you use 2 CPUs, the CPU: field should also read 2.
+        # The -l argument accepts cpu-list. Indexing starts at zero.
+        # For example -l 1,2,10-20 means use core 1, 2, 10 to 20.
+        # Note: Ensure that the CPU resources are updated accordingly.
+        #       If you use 2 CPUs, the CPU: field should also read 2.
         - "-N$(MY_NODE_NAME)"
         - "-g$(MY_POD_IP)"
         - "-nnats"
         - "-y/var/local/mayastor/config.yaml"
-        - "-l1,2"
+        - "-l0"
         securityContext:
           privileged: true
         volumeMounts:
@@ -80,11 +74,11 @@ spec:
           # belong to Guaranteed QoS class, hence can never get evicted in case of
           # pressure unless they exceed those limits. limits and requests must be the same.
           limits:
-            cpu: "2"
+            cpu: "1"
             memory: "512Mi"
             hugepages-2Mi: "1Gi"
           requests:
-            cpu: "2"
+            cpu: "1"
             memory: "512Mi"
             hugepages-2Mi: "1Gi"
         ports:
--- a/scripts/generate-deploy-yamls.sh
+++ b/scripts/generate-deploy-yamls.sh
@@ -1,52 +1,158 @@
-#! /bin/sh
+#!/bin/sh
+
+# This is a wrapper script for helm to generate yaml files for deploying
+# mayastor. It provides reasonable defaults for helm values based on
+# selected profile. Easy to use and minimizing risk of error.
+# Keep the script as simple as possible - ad-hoc use cases can be addressed
+# by running helm directly.
 
 set -e
 
-if [ "x$1" = x ]; then
-cat <<EOF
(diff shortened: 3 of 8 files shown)


jkryl:
@arne-rusek : We will have to bump up hugepage memory for e2e-test nodes from 1G to 2G. Not sure what should be changed where. Could you advise me? thx

scripts/generate-deploy-yamls.sh
```diff
@@ -1,52 +1,158 @@
-#! /bin/sh
```
arne-rusek, line 1: Oh you. It's there on purpose, to show it's possible  :trollface:

arne-rusek:
> @arne-rusek : We will have to bump up hugepage memory for e2e-test nodes from 1G to 2G. Not sure what should be changed where. Could you advise me? thx

I don't see hugepages being templated. Did you forget to template it?

Anyway I'd lower default instance on AWS/HCloud to a smaller one as 2 CPUs should now be enough (if anybody uses https://github.com/mayadata-io/mayastor-terraform-playground/ it will be cheaper) and just set variables during building e2e-test clusters. How and where? That would probably depend on how/where would be the `scripts/generate-deploy-yamls.sh` called. I presume in a separate PR?

scripts/generate-deploy-yamls.sh
```diff
@@ -1,52 +1,158 @@
...
+if ! which helm >/dev/null 2>&1; then
+	echo "Install helm (>v3.4.1) to PATH"
 	exit 1
 fi
 
+if [ ! -d "$output_dir" ]; then
+	mkdir -p "$output_dir"
+fi
+
 tmpd=$(mktemp -d /tmp/generate-deploy.sh.XXXXXXXX)
```
arne-rusek, line 131:
```suggestion
tmpd=$(mktemp -d /tmp/generate-deploy-yamls.sh.XXXXXXXX)
```

please fix my inconsistency

scripts/generate-deploy-yamls.sh
```diff
@@ -1,52 +1,158 @@
...
+template_params="mayastorImagesTag=$tag"
+template_params="$template_params,mayastorImagePullPolicy=$pull_policy"
+template_params="$template_params,mayastorCpuCount=$cores"
+template_params="$template_params,mayastorHugePagesGiB=$cores"
+template_params="$template_params,moacDebug=$moac_debug"
+if [ -n "$registry" ]; then
+	template_params="$template_params,mayastorImagesRepo=$registry"
 fi
+if [ -n "$pools" ]; then
+  i=0
```
arne-rusek, line 145: strange indentation, is it intended?

scripts/generate-deploy-yamls.sh
```diff
@@ -1,52 +1,158 @@
...
+		template_params="$template_params,mayastorPools[$i].device=$device"
+		i=$((i + 1))
+	done
+fi
+
 helm template --set "$template_params" mayastor "$SCRIPTDIR/../chart" --output-dir="$tmpd" --namespace mayastor
 
-mv "$tmpd"/mayastor/templates/*.yaml "$TARGET_DIR"
+cd "$tmpd/mayastor/templates"
+mv ./*.yaml "$output_dir/"
```
arne-rusek, line 158: Please don't change working directory in scripts. Working directory should be there for the user, not scripts. Moreover it's like using global variables/mutating global state. I don't even understand the need here. Sooner or later it always causes problems to someone (e.g. continuing this script all relative paths are now broken). If you really need to do this (which I don't understand and don't see a reason) please use at least a subshell.
`( cd /whatever && do_stuff )`

scripts/generate-deploy-yamls.sh
jkryl: some indentation is wrong. will fix it. thanks

scripts/generate-deploy-yamls.sh
jkryl: will fix 👍

scripts/generate-deploy-yamls.sh
jkryl: The reason for this was basically because I wanted to avoid string quotation like this:
```
mv "$tmpd"/mayastor/templates/*.yaml ...
```
Note that I cannot quote the whole string as usually because that will not expand to multiple files:
```
mv "$tmpd/mayastor/templates/*.yaml" ...   <--- does not work
```
So either I have to quote just the variable and not the whole string, or I have to change the directory first and then do the copy (and I can do that in subshell - no problem). I'm fine with either option. Let me know what you would prefer.

arne-rusek: I prefer leaving it as it is as is perfectly correct shell code :smile:. Second best is the subshell.

When at it (just brainstorm, you made me think :smile:) `find "$tmpd/mayastor/templates" -name '*.yaml' | xargs mv -t "$output_dir"` is also valid and doesn't contain partial path quoting.

FYI just to get the feel for it I did a small research over my `/usr/bin` and partial quoting is ubiquitous "technique":
```
file /usr/bin/* | grep shell | cut -f1 -d: | xargs grep --color=always '"/' 
```

arne-rusek (approved):
LGTM

chart/values.yaml
```diff
@@ -1,8 +1,12 @@
-mayastorImagesTag: develop
+mayastorImagesTag: latest
 mayastorImagePullPolicy: Always
-mayastorCpuCount: "2"
+mayastorCpuCount: "1"
```
gila, line 3: I don't know exactly how this is supposed to work but.. how does this get filled in the daemon set values?

deploy/mayastor-daemonset.yaml
```diff
@@ -80,11 +74,11 @@ spec:
           # belong to Guaranteed QoS class, hence can never get evicted in case of
           # pressure unless they exceed those limits. limits and requests must be the same.
           limits:
-            cpu: "2"
+            cpu: "1"
             memory: "512Mi"
             hugepages-2Mi: "1Gi"
           requests:
-            cpu: "2"
+            cpu: "1"
```
gila, line 81: Can this me {{cpus}}.len() or something?

chart/values.yaml
jkryl: these values replace placeholders in chart/templates and the result is saved to the deploy directory. In a nutshell -  just a simplified explanation.

deploy/mayastor-daemonset.yaml
jkryl: yup. It is more or less like you have described in template directory of the mayastor chart. This file is the result of templating process so that's why we see hardcoded value here.

blaisedias (approved):
I think we should use the symbol `mayastorImagesRegistry` instead of `mayastorImagesRepo` to avoid confusion in the future.
But otherwise LGTM

jkryl:
@blaisedias : I will change that. thanks

jkryl:
bors merge

jkryl:
bors retry

jkryl:
bors retry

PR #701, opened 2021-02-11
refactor(nexus): clean up label parsing code

cjones1024:
- dispensed with NexusChildLabel struct entirely
  (which maintained a child reference and seemed to be causing deadlocks)
- added extra sanity checking for valid labels
- create labels correctly for disks of varying size
- allow changing of disk GUID if creating nexus with different UUID

Resolves CAS-720

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_bdev.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev.rs
@@ -6,6 +6,7 @@
 
 use std::{
     convert::TryFrom,
+    env,
     fmt::{Display, Formatter},
     os::raw::c_void,
 };
@@ -51,7 +52,7 @@ use crate::{
             nexus_nbd::{NbdDisk, NbdError},
         },
     },
-    core::{Bdev, CoreError, DmaError, Protocol, Reactor, Share},
+    core::{Bdev, CoreError, Protocol, Reactor, Share},
     ffihelper::errno_result_from_i32,
     lvs::Lvol,
     nexus_uri::{bdev_destroy, NexusBdevError},
@@ -113,19 +114,25 @@ pub enum Error {
     ShareNvmfNexus { source: CoreError, name: String },
     #[snafu(display("Failed to unshare nexus {}", name))]
     UnshareNexus { source: CoreError, name: String },
-    #[snafu(display("Failed to allocate label of nexus {}", name))]
-    AllocLabel { source: DmaError, name: String },
-    #[snafu(display("Failed to write label of nexus {}", name))]
+    #[snafu(display(
+        "Failed to read child label of nexus {}: {}",
+        name,
+        source
+    ))]
+    ReadLabel { source: LabelError, name: String },
+    #[snafu(display(
+        "Failed to write child label of nexus {}: {}",
+        name,
+        source
+    ))]
     WriteLabel { source: LabelError, name: String },
-    #[snafu(display("Failed to read label from a child of nexus {}", name))]
-    ReadLabel { source: ChildError, name: String },
-    #[snafu(display("Labels of the nexus {} are not the same", name))]
-    CheckLabels { name: String },
-    #[snafu(display("Failed to write protective MBR of nexus {}", name))]
-    WritePmbr { source: LabelError, name: String },
-    #[snafu(display("Failed to register IO device nexus {}", name))]
+    #[snafu(display(
+        "Failed to register IO device nexus {}: {}",
+        name,
+        source
+    ))]
     RegisterNexus { source: Errno, name: String },
-    #[snafu(display("Failed to create child of nexus {}", name))]
+    #[snafu(display("Failed to create child of nexus {}: {}", name, source))]
     CreateChild {
         source: NexusBdevError,
         name: String,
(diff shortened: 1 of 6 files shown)


cjones1024:
bors try

cjones1024:
bors try

mayastor/src/bdev/nexus/nexus_label.rs
```diff
@@ -126,340 +141,221 @@ pub enum LabelError {
...
-                .context(WriteAlloc {
-                    name: String::from("secondary"),
-                })?;
-        let mut writer = Cursor::new(buf.as_mut_slice());
+        let data = header.lba_start + metadata_size;
 
-        // Secondary partition table
-        for entry in &label.partitions {
-            serialize_into(&mut writer, &entry).context(SerializeError {})?;
+        if data > header.lba_end {
```
tiagolobocastro, line 248: I imagine this block size is the block size of this particular child and not the first one in the list?
Though IIRC there's a validation somewhere that they all need to have the same block size?

tiagolobocastro (approved):

mayastor/src/bdev/nexus/nexus_label.rs
```diff
@@ -653,14 +591,76 @@ pub struct NexusLabel {
...
+    /// locate a partition by name
+    fn get_partition(&self, name: &str) -> Option<&GptEntry> {
+        self.partitions
+            .iter()
+            .find(|entry| entry.ent_name.name == name)
+    }
+
+    #[allow(dead_code)]
+    /// returns the offset of the first metadata block
+    pub(crate) fn metadata_offset(&self) -> Result<u64, ProbeError> {
```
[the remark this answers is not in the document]
cjones1024: Yes, and in this particular case very soon. Think it's justified to add them here now, and remove the annotation when it's no longer required.

mayastor/src/bdev/nexus/nexus_label.rs
cjones1024: The block size is now always obtained from the child that it pertains to. You are correct in that there is a (necessary) check to ensure that all children have the same block size, and currently lots of things will break if this is not the case. However if we are ever in a position to relax that requirement, this code will do the right thing. (Well mostly - there are still places in the labelling code that requires all children to have the same block_size - but this is no longer one of them).

cjones1024:
bors try

mayastor/src/bdev/nexus/nexus_label.rs
tiagolobocastro: That's good. In theory it could be possible to have the nexus always configured as a 4k block device, and then that would allow having children with mixed 512 and 4k block sizes without having to do ugly things like RMW.

cjones1024:
bors try

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -481,21 +488,26 @@ impl Nexus {
     }
 
     pub async fn sync_labels(&mut self) -> Result<(), Error> {
-        let label = self.update_child_labels().await.context(WriteLabel {
+        if env::var("NEXUS_DONT_READ_LABELS").is_ok() {
```
Hoverbear, line 491: I wonder if there is a better way to do this?

mayastor/src/bdev/nexus/nexus_label.rs
```diff
@@ -126,340 +141,221 @@ pub enum LabelError {
...
 
-    fn get_secondary_data(
-        &self,
-        label: &NexusLabel,
-    ) -> Result<LabelData, LabelError> {
-        let block_size = self.bdev.block_len() as u64;
-        let blocks = Aligned::get_blocks(
-            GPTHeader::PARTITION_TABLE_SIZE as u64,
-            block_size,
+    // Create partition table entries for the MayaMeta and
```
Hoverbear, line 234: This is probably better served as a `///` comment for IDE tooling.

mayastor/src/bdev/nexus/nexus_label.rs
```diff
@@ -126,340 +141,221 @@ pub enum LabelError {
...
                 }
-                NexusLabelStatus::Neither => {
-                    futures.push(
-                        label.child.write_at(primary.offset, &primary.buf),
-                    );
-                    futures.push(
-                        label.child.write_at(secondary.offset, &secondary.buf),
-                    );
+                Err(error) => {
+                    return Err(LabelError::Generic {
```
Hoverbear, line 295: Since this one has a source, I wonder if a `GenericSourced` error variant with a `source` would be more appropriate?

gila (approved):
This looks nicer! Thanks, @cjones1024

mayastor/src/bdev/nexus/nexus_label.rs
cjones1024: In this case the "source" is a string rather than an error. snafu treats a "source" field automagically and so using that term here would only add to the confusion.

mayastor/src/bdev/nexus/nexus_label.rs
cjones1024: Done

mayastor/src/bdev/nexus/nexus_bdev.rs
cjones1024: As per the comments, this really only exists for one particular use case - where we want to use NULL bdevs purely for the purpose of performance testing, and as such is only ever needed in testing. I don't really want to see this exposed elsewhere, but am always open to suggestions about how to do it better. :)

cjones1024:
bors try

mayastor/src/bdev/nexus/nexus_bdev.rs
jonathan-teh: The existing `NEXUS_LABEL_IGNORE_ERRORS` environment variable is also used in the `nvmet` binary. It should either be kept in sync or the environment variable name changed.

mtzaurus:
For the sake of testability and taking into account the future changes in Nexus/BlockDevices approach,  can we make the code that performs actual I/O (i.e. reads/writes GPT sectors) wore abstract over some generic interfaces, like read / write or so ?
This gives us possibility to smoothly switch to a new device model by just implementing a thin I/O adapter, rather than changing all label code + we can easily write tests that validate GPT using files as backend storage.

cjones1024:
Yes I think this is definitely the way to proceed, but I think it would also be better to this as under a separate Jira (which could also encompass reading and writing to the MetaData partition)

cjones1024:
bors try

mayastor/src/bdev/nexus/nexus_label.rs
cjones1024: Now removed entirely

cjones1024:
bors try

cjones1024:
bors merge

cjones1024:
bors merge

cjones1024:
bors merge

cjones1024:
bors merge

PR #703, opened 2021-02-11
feat(rest-test): new rest test project 

tiagolobocastro:
This is the initial bits of a new test project which leverages the new
control plane to test mayastor.
It has already revealed a few issues.

Make replica share protocol conversion from gRPC fallible as it was
causing mayastor to crash on the REST tests.

Split the deployer into a library and a binary allowing for the library
 to be used by the coming rest test project.

Allow for a JSON Value to traverse the stack allowing swagger to
 display it as JSON and with colours.
Also make sure the rest client returns the correct errors.

### DIFF
--- a/composer/src/lib.rs
+++ b/composer/src/lib.rs
@@ -113,11 +113,11 @@ impl Binary {
         let path = std::path::PathBuf::from(std::env!("CARGO_MANIFEST_DIR"));
         let srcdir = path.parent().unwrap().to_string_lossy();
 
-        Self::new(format!("{}/target/debug/{}", srcdir, name), vec![])
+        Self::new(&format!("{}/target/debug/{}", srcdir, name), vec![])
     }
     /// Setup nix shell binary from path and arguments
     pub fn from_nix(name: &str) -> Self {
-        Self::new(Self::which(name).expect("binary should exist"), vec![])
+        Self::new(name, vec![])
     }
     /// Add single argument
     /// Only one argument can be passed per use. So instead of:
@@ -171,11 +171,17 @@ impl Binary {
     }
     fn which(name: &str) -> std::io::Result<String> {
         let output = std::process::Command::new("which").arg(name).output()?;
+        if !output.status.success() {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::NotFound,
+                name,
+            ));
+        }
         Ok(String::from_utf8_lossy(&output.stdout).trim().into())
     }
-    fn new(path: String, args: Vec<String>) -> Self {
+    fn new(path: &str, args: Vec<String>) -> Self {
         Self {
-            path,
+            path: Self::which(path).expect("Binary path should exist!"),
             arguments: args,
             ..Default::default()
         }
@@ -310,6 +316,10 @@ pub struct Builder {
     containers: Vec<ContainerSpec>,
     /// the network for the tests used
     network: String,
+    /// reuse existing containers
+    reuse: bool,
+    /// allow cleaning up on a panic (if clean is true)
+    allow_clean_on_panic: bool,
     /// delete the container and network when dropped
     clean: bool,
     /// destroy existing containers if any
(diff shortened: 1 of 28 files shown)


tiagolobocastro:
bors try

control-plane/tests/tests/common/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+                    pool: pool.id(),
+                    size,
+                    thin: false,
+                    share: share.clone(),
+                });
+            }
+        }
+        self
+    }
+    /// Build into the resulting Cluster
```
paulyoong, line 210: I think this comment needs removing. It is already on line 221.

control-plane/tests/tests/pools.rs
```diff
@@ -0,0 +1,218 @@
...
+        .create_pool(v0::CreatePool {
+            node: "mayastor".into(),
+            id: "pooloop".into(),
+            disks: vec!["malloc:///disk?size_mb=100".into()],
+        })
+        .await
+        .unwrap();
+}
+
+/// FIXME: CAS-710
```
paulyoong, line 104: Maybe add to the ticket that the `#[allow_fail]` should be removed once fixed, so that someone less familiar with it knows what to do.

control-plane/tests/tests/replicas.rs
```diff
@@ -0,0 +1,131 @@
+#![feature(allow_fail)]
+
+pub mod common;
+use common::*;
+
+// FIXME: CAS-721, CAS-722
```
paulyoong, line 6: Again, maybe just a note in the tickets about removing the `#[allow_fail]`.

control-plane/tests/tests/replicas.rs
```diff
@@ -0,0 +1,131 @@
...
+                thin: true,
+                share: protocol.clone(),
+            }),
+        )
+        .await
+        .unwrap();
+    }
+}
+
+/// todo: RAISE CAS!
```
paulyoong, line 74: Can we just raise this ticket and reference it here? We don't want to forget about it.

paulyoong (approved):

composer/src/lib.rs
```diff
@@ -430,6 +438,13 @@ impl Builder {
         self.add_container_spec(ContainerSpec::from_binary(name, image))
     }
 
+    /// attempt to reuse and restart containers instead of starting new ones
+    pub fn with_reuse(mut self, reuse: bool) -> Builder {
+        self.reuse = reuse;
+        self.prune = false;
+        self
+    }
```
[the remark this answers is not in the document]
tiagolobocastro: oh good catch :+1:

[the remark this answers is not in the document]
tiagolobocastro: The clean only runs at the drop, so we're ok there I think.

control-plane/rest/src/lib.rs
```diff
@@ -155,6 +169,49 @@ impl ActixRestClient {
...
+#[derive(Debug, Clone, Apiv2Schema)]
+pub struct JsonGeneric {
+    inner: serde_json::Value,
+}
+impl Serialize for JsonGeneric {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        self.inner.serialize(serializer)
```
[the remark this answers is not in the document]
tiagolobocastro: I think you're reading it wrong. The serializer is the `serializer` which is being called through a trait. Inner is the `value`.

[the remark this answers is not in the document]
paulyoong: Ah, doh!

control-plane/tests/tests/replicas.rs
tiagolobocastro: dohh, I've raised it now.

tiagolobocastro:
bors try

gila (approved):
👍

tiagolobocastro:
bors merge

PR #705, opened 2021-02-11
feat(nvmx): initial support of device events

mtzaurus:
BlockDevice now supports registration of callbacks to notify about
device events (as of now only device removal event is supported).
In addition, nvmf tests now use dynamically launched Mayastor
instances.

Implements CAS-670


Hoverbear (approved):

mayastor/src/bdev/dev/nvmx/controller.rs
```diff
@@ -96,22 +106,35 @@ pub struct NvmeControllerInner<'a> {
     adminq_poller: poller::Poller<'a>,
     io_device: Arc<IoDevice>,
 }
+
+type EventCallbackList = Vec<fn(DeviceEventType, &str)>;
```
gila, line 110: Should we limit the number of entries we can insert? (or maybe a different data structure) or are the number of events, not a concern? If they *might* be, I think we should add an issue for it to look into it later.

mayastor/src/bdev/dev/nvmx/controller.rs
```diff
@@ -502,6 +529,38 @@ impl<'a> NvmeController<'a> {
...
+            (*l)(event, &self.name);
+        }
+        listeners.len()
+    }
+
+    /// Register listener to monitor device events related to this controller.
+    pub fn add_event_listener(
+        &self,
+        listener: fn(DeviceEventType, &str),
+    ) -> Result<(), CoreError> {
```
gila, line 554: nit; we never generate an error here.

mayastor/src/bdev/dev/nvmx/device.rs
```diff
@@ -119,6 +120,17 @@ impl NvmeBlockDevice {
             name: String::from(name),
         }
     }
+
+    fn get_controller(
+        &self,
+    ) -> Result<Arc<Mutex<NvmeController<'static>>>, CoreError> {
```
gila, line 126: 'static? 🤔

gila (commented):
few minor questions, thanks @mtzaurus 👍

mayastor/src/bdev/dev/nvmx/device.rs
Hoverbear: Do we want the returned NVMe controller to last possibly the entire life of the binary run?

mayastor/src/bdev/dev/nvmx/controller.rs
mtzaurus: Sure thing. I will create a ticket to consider improvement in device events storing.

mayastor/src/bdev/dev/nvmx/controller.rs
mtzaurus: I wanted to keep the API more flexible. Later on we might add checks against the maximum limit of event listeners, etc.

mayastor/src/bdev/dev/nvmx/device.rs
mtzaurus: Indeed we can move controller lookup close to the caller and get rid of get_controller(), as its used only once.

mtzaurus:
Merged manually after rebasing on top of develop.

PR #718, opened 2021-02-15
fix(nvmx): controller i/o statistics

mtzaurus:
This fix introduces I/O statistics for read, write and unmap
operations for NVMe block devices.

Implements CAS-722

### DIFF
--- a/mayastor/src/bdev/dev/nvmx/channel.rs
+++ b/mayastor/src/bdev/dev/nvmx/channel.rs
@@ -1,12 +1,15 @@
 /* I/O channel for NVMe controller, one per core. */
 
 use crate::{
-    bdev::dev::nvmx::{
-        nvme_bdev_running_config,
-        NvmeControllerState,
-        NVME_CONTROLLERS,
+    bdev::{
+        dev::nvmx::{
+            nvme_bdev_running_config,
+            NvmeControllerState,
+            NVME_CONTROLLERS,
+        },
+        nexus::nexus_io::IoType,
     },
-    core::{poller, CoreError},
+    core::{poller, BlockDeviceIoStats, CoreError},
 };
 use std::{cmp::max, mem::size_of, os::raw::c_void, ptr::NonNull};
 
@@ -197,6 +200,7 @@ pub struct NvmeIoChannelInner<'a> {
     poll_group: PollGroup,
     poller: poller::Poller<'a>,
     pub qpair: Option<IoQpair>,
+    io_stats_controller: IoStatsController,
     // Flag to indicate the shutdown state of the channel.
     // We need such a flag to differentiate between channel reset and shutdown.
     // Channel reset is a reversible operation, which is followed by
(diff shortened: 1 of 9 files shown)


gila (approved):
👍

mayastor/src/bdev/dev/nvmx/controller.rs
```diff
@@ -328,6 +330,70 @@ impl<'a> NvmeController<'a> {
...
     }
 
+    /// Get I/O statistics for all I/O channels of the controller.
+    pub fn get_io_stats<T: 'static + Sized, F>(
+        &self,
+        cb: F,
+        cb_arg: T,
+    ) -> Result<(), CoreError>
+    where
+        F: Fn(Result<BlockDeviceIoStats, CoreError>, T) + 'static,
```
jkryl, line 340: just an idea. By pushing the oneshot channel from io_stats() to this routine, you could keep larger part of the rust code clean (async method without without callbacks). But maybe it cannot be done for some reason.

mayastor/src/bdev/dev/nvmx/channel.rs
```diff
@@ -291,6 +295,61 @@ impl NvmeIoChannelInner<'_> {
...
+
+    #[inline]
+    /// Account amount of blocks and I/O operations.
+    pub fn account_block_io(
+        &mut self,
+        op: IoType,
+        num_ops: u64,
+        num_blocks: u64,
+    ) {
+        let num_bytes = num_blocks * self.block_size;
```
jkryl, line 327: Not sure how much raw the stats should be. But I personally would prefer as raw numbers as possible and I can do the remaining math as a consumer of the stats. Also the code in IO path that is called often, could be simpler then. So if we recorded just number of reads and writes we could save a couple of instructions and a couple of bytes of memory. Assuming the consumer knows the block size of the device, she can calculate the rest easily. Just an idea and it's really a micro optimisation.

jkryl (approved):
👍

mayastor/src/bdev/dev/nvmx/channel.rs
mtzaurus: Actually, libSPDK does the same as we do, they calculate the real stats in place:

bdev_io->internal.ch->stat.bytes_read += bdev_io->u.bdev.num_blocks * bdev_io->bdev->blocklen;

Looks like they don't care much about the speed. :-)

But not a problem, I can move multiplication step to 'get_stats()' and return updated stats when explicitly requested. I would prefer to keep translation as close to the stats itself as possible and not to put any burden to the caller.

jonathan-teh: It appears that all callers pass 1 to `num_ops` so perhaps that can be dropped.

mayastor/src/bdev/dev/nvmx/controller.rs
mtzaurus: Unfortunately, we can't use this method as async. We invoke controller's methods under protection of the mutex, and await()'ing under mutex is not always a good idea. In fact, our linter complains about such a thing. That's why we initiate an async operation and then wait with the controller freed.

mayastor/src/bdev/dev/nvmx/device.rs
```diff
@@ -173,8 +176,28 @@ impl BlockDevice for NvmeBlockDevice {
...
+            CoreError::BdevNotFound {
+                name: self.name.to_string(),
+            },
+        )?;
+
+        let (s, r) =
+            oneshot::channel::<Result<BlockDeviceIoStats, CoreError>>();
+        // Schedule async I/O stats collection and wait for the result.
+        {
+            let controller = carc.lock().expect("lock poisened");
```
jonathan-teh, line 190: Typo: poisoned.

jonathan-teh (approved):

mayastor/src/bdev/dev/nvmx/device.rs
mtzaurus: Typo fixed, thanks.

PR #730, opened 2021-02-18
ci: remove failing tests

gila:
The tests keep failing where the culprit seems that we run multiple
instances. These tests are simply to heavy/complex to run in CI this way
and we have to come up with something else.

The tests are removed instead of disabled it does not make sense to keep
them around as they due get compiled.


blaisedias (approved):
LGTM

jkryl (approved):
🔥

gila:
bors merge

PR #733, opened 2021-02-19
ci(e2e): run e2e on shared worker

arne-rusek:
Because of new beefier worker nodes for Jenkins we can run e2e tests in parallel
with other jobs thus keeping more CI capacity for other jobs


paulyoong (approved):

chriswldenyer (approved):

arne-rusek:
bors merge

PR #734, opened 2021-02-19
chore(rest): remove volume operations

paulyoong:
The Volume operations were included as a convenience to the consumer of
the REST API. For example, this allowed a volume to be easily created
without having to worry about creating pools, replicas etc. However,
the consumer of the REST API will typically be an Operator that does
have domain knowledge and should therefore be responsible for
constructing a volume. This makes the existing volume operations
redundant and so they have been removed.


tiagolobocastro (approved):

gila (commented):
So I'm not sure about the reasoning for the removal here.  How does one create volume through the rest API once this is removed?

tiagolobocastro:
> So I'm not sure about the reasoning for the removal here. How does one create volume through the rest API once this is removed?

This service was created at the time where the services would be the "brain" and decide where pools and nexus get placed. Now that the "brain" is moving higher up then we don't need the volume endpoints as all can be done via the nexus endpoints. That will allow the control plane to choose exactly where to place the nexus children and where to place the N nexuses (in case of ANA).

paulyoong:
Closing PR - after speaking with @gila these higher level REST calls will be useful and should be kept.

PR #735, opened 2021-02-19
test(rest): add more rests tests and remove rust operator

tiagolobocastro:
Remove the rust node operator from the tree as it will be rewritten in
 go and might not even live in this repo.

Anyhow requires errors to be Send which is not the case with the rest
 return Http types and errors so use snafu instead.
Enhanced errors to include failure to send errors, http header etc..

Add more rest tests around replica and the nexus.
Get the correct error from the rest to avoid false negatives.
Deferring the "replica outage" tests as those kind of tests were often
 failing on mayastor already.


control-plane/rest/src/lib.rs
```diff
@@ -148,51 +161,160 @@ impl ActixRestClient {
...
-                Ok(result) => Ok(result),
-                Err(_) => Ok(vec![serde_json::from_slice::<R>(&rest_body)?]),
+        let status = rest_response.status();
+        let headers = rest_response.headers().clone();
+        let head = || {
+            let mut head = ResponseHead::new(status);
+            head.headers = headers.clone();
+            head
+        };
+        let body = rest_response.body().await.map_err(|_| {
```
paulyoong, line 185: Should we include the error being mapped?

control-plane/rest/src/lib.rs
```diff
@@ -148,51 +161,160 @@ impl ActixRestClient {
...
+        /// http Header
+        head: ResponseHead,
+    },
+    /// Invalid Body
+    #[snafu(display("Invalid body, header: {:?}", head))]
+    InvalidBody {
+        /// http Header
+        head: ResponseHead,
+        /// http Body
+        body: Bytes,
```
paulyoong, line 296: This doesn't seem to be used in the error message.

control-plane/rest/src/lib.rs
```diff
@@ -148,51 +161,160 @@ impl ActixRestClient {
...
+    },
+    /// No Body
+    #[snafu(display("No body, header: {:?}", head))]
+    Header {
+        /// http Header
+        head: ResponseHead,
+    },
+    /// Body in JSON format
+    #[snafu(display("Http status: {}, error: {}", head.status, error.to_string()))]
+    Valid {
```
paulyoong, line 306: Could we rename `Valid` to something else? It seems a little strange to have a valid error.

control-plane/tests/tests/common/mod.rs
```diff
@@ -169,19 +197,40 @@ impl ClusterBuilder {
...
+        }
+        self
+    }
+    /// Add pool with `disk` to each node
+    pub fn with_pool(mut self, disk: &str) -> Self {
+        self.pools.push(PoolDisk::Uri(disk.to_string()));
         self
     }
-    /// Add `count` replicas to each node per pool
+    /// Specify `count` replicas to each node per pool
```
paulyoong, line 217: "...to **add to** each node per pool"?

control-plane/tests/tests/common/mod.rs
```diff
@@ -211,7 +260,8 @@ impl ClusterBuilder {
             .with_base_image(self.opts.base_image.clone())
             .autorun(false)
             .with_default_tracing()
-            .with_clean(true)
+            // todo: remove before committing
```
paulyoong, line 263: Do you need to action this todo?

control-plane/tests/tests/nexus.rs
```diff
@@ -4,22 +4,200 @@ pub mod common;
...
+        .with_rest_timeout(std::time::Duration::from_secs(1))
+        // don't log whilst we have the allow_fail
+        .compose_build(|c| c.with_logs(false))
+        .await
+        .unwrap();
+
+    for size_mb in &vec![6, 10, 100] {
+        let size = size_mb * 1024 * 1024;
+        let disk =
+            || format!("malloc:///disk?size_mb={}", size / (1024 * 1024));
```
paulyoong, line 36: You could just use `size_mb` here.

control-plane/tests/tests/nexus.rs
```diff
@@ -4,22 +4,200 @@ pub mod common;
...
+            })
+            .await
+            .unwrap();
+        }
+    }
+
+    for size_mb in &vec![1, 2, 4] {
+        let size = size_mb * 1024 * 1024;
+        let disk =
+            || format!("malloc:///disk?size_mb={}", size / (1024 * 1024));
```
paulyoong, line 70: Again, you could use `size_mb` here.

control-plane/tests/tests/nexus.rs
```diff
@@ -4,22 +4,200 @@ pub mod common;
...
+    cluster
+        .rest_v0()
+        .create_nexus(v0::CreateNexus {
+            node: cluster.node(0),
+            uuid: v0::NexusId::new(),
+            size,
+            children: vec![local.into(), remote.into()],
+        })
+        .await
+        .expect_err("One replica is not present so nexus can't be created");
```
paulyoong, line 202: I think `can't` should be replaced with `shouldn't` as this error message will only be shown if the nexus is erroneously created.

paulyoong (approved):

control-plane/rest/src/lib.rs
tiagolobocastro: yeah, I've changed all the map's to context now :+1:

control-plane/rest/src/lib.rs
tiagolobocastro: No, it's still on the error, just not on the default to_string(). Yeah, maybe all fields should be on it by default then.

control-plane/rest/src/lib.rs
tiagolobocastro: Yeah, it was valid in the sense that it's a valid json format error. Any suggestions?

control-plane/tests/tests/common/mod.rs
tiagolobocastro: that was just a reminder for myself to change the clean back to true again before commiting, which I clearly forgot. I wonder, should we have an ENV variable to tell the composer not to cleanup during local testing? Would avoid having to change those flags all the time while debugging.

control-plane/tests/tests/nexus.rs
tiagolobocastro: dohhh!

control-plane/rest/src/lib.rs
tiagolobocastro: Renamed to `RestServer`

control-plane/tests/tests/common/mod.rs
tiagolobocastro: Done as a separate commit.

tiagolobocastro:
bors try

tiagolobocastro:
bors try

control-plane/rest/src/lib.rs
paulyoong: :+1:

control-plane/tests/tests/nexus.rs
```diff
@@ -4,22 +4,198 @@ pub mod common;
...
+    let size = 10 * 1024 * 1024;
     let cluster = ClusterBuilder::builder()
         .with_pools(1)
-        .with_replicas(2, 5 * 1024 * 1024, v0::Protocol::Off)
+        .with_replicas(1, size, v0::Protocol::Off)
         .build()
         .await
         .unwrap();
 
+    let replica = format!("bdev:///{}", Cluster::replica(0, 0));
```
jonathan-teh, line 110: `bdev://` is being kept around for backwards compatibility, new code should use `loopback://`.

jonathan-teh (approved):

control-plane/tests/tests/nexus.rs
tiagolobocastro: Ah thanks, should probably also change the volume agent - I'm changing the agents for another pr so I'll do it there.

tiagolobocastro:
bors merge

PR #738, opened 2021-02-22
feat(nexus): provide mechanism for persisting child generation data

cjones1024:
- refactor old nexus_metadata module
- provide means to create a "metadata index" on the MayaMeta partition of each and any child
- provide means to add and retrieve child generation data from the "metadata index"

resolves CAS-746


cjones1024:
bors try

cjones1024:
bors try

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -5,821 +5,745 @@
...
-//!  - The first usable "data" block is the first block following the index
-//!    (whose size is aligned to the blocksize of the disk).
+//!  - The second block contains the MetaDataIndex (currently 80 bytes) while
+//!    the remainder of the block is padded with zeros.
+//!  - The data in the "index" starts at the third block.
+//!  - This and each subsequent block contains a single MetaDataObject entry.
+//!  - The size of a MetaDataObject entry therefore cannot exceed the block_size
+//!    of the device.
+//!  - Once the index is "full", when a new entry is added, the oldest entry is
+//!    (silently) removed in order to make space for the new entry.
```
jkryl, line 15: Looks like a ring buffer to me, right? And the metadataindex has a pointer to latest entry into this ring buffer. So when we fill the whole ring, we start overwriting entries from 0 again. That also means that for each change to metadata we have to update two blocks: the index and the entry. Is my understanding right?

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -5,821 +5,745 @@
...
 
-impl MetaDataHeader {
-    pub const METADATA_HEADER_SIZE: u32 = 72;
-    pub const MAX_INDEX_ENTRIES: u32 = 32;
-    pub const INDEX_ENTRY_SIZE: u32 = 44;
+impl MetaDataObject {
+    /// Observe that sizeof(MetaDataHeader) + 26 * sizeof(MetaDataChildEntry) is
+    /// exactly 512. While it may be smaller, the value of this expression
+    /// must NEVER exceed 512.
+    pub const MAX_CHILD_ENTRIES: u32 = 26;
```
jkryl, line 161: Would be good to have common definition of this constant and use it also in add-nexus-child method in order to be consistent. I assume that nothing prevents grpc client to add more than 26 children at this point. But maybe we want to enforce much lower limit at the grpc level. Not sure. In any case the count should never go beyond 26.

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -5,821 +5,745 @@
...
+    #[snafu(display("Incorrect MetaDataIndex size"))]
+    IndexSize {},
+    #[snafu(display("Incorrect MetaDataIndex checksum"))]
+    IndexChecksum {},
+}
+
+#[derive(Debug, Default, Deserialize, PartialEq, Serialize)]
+pub struct MetaDataChildEntry {
+    pub guid: Guid,
+    pub state: u16,
```
jkryl, line 107: just now we store only the state of the child, so we can tell which replica was online and which was degraded or even faulted. I'm not sure if generation number isn't a more valuable information when it comes to the rebuild. I wasn't able to imagine a case when we could not detect wrong replica (after nexus is recreated) however increasing generation number is simpler to do, does not rely on representation of state by a number that can potentially change. What do you think Colin and @gila ? Some time ago we discussed RPC method for obtaining the generation number that would be used by the operator to decide what to do when importing a volume. So that sounds like a more useful information.

cjones1024: Each "index" entry represents a snapshot of the nexus, and contains a timestap, generation number, and the current state of each child (identified by it's uuid).  So each child in effect inherits the same generation number. The idea is that a new index entry is created whenever the status of any child changes. Also, the intention is that the state of each child is represented as a bitmap - where the meaning of each bit is yet to be determined. However the code here at the persistence layer does not need to know what the "state" represents - it's just a value that needs to be stored and retrieved.

And, yes there will be gRPC methods to retrieve the latest index entry from one or all children.

mayastor/src/bdev/nexus/nexus_metadata.rs
cjones1024: Correct. Each change will involve writing out a new entry, and also updating the block containing the index.

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -5,821 +5,745 @@
...
+}
+
+impl MetaDataChildEntry {
+    /// Calculate checksum over the child entries
+    pub fn checksum(children: &[MetaDataChildEntry]) -> u32 {
```
Hoverbear, line 112: It does appear this function has some reasonable error conditions due to serialization. Are all of these situations ones in which we should panic?

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -5,821 +5,745 @@
...
-        let data_start = index_start
-            + Aligned::get_blocks(
-                MetaDataHeader::MAX_INDEX_ENTRIES
-                    * MetaDataHeader::INDEX_ENTRY_SIZE,
-                block_size,
-            );
+    /// Perform basic checks to ensure object is valid,
+    /// and then calculate checksums.
+    pub fn validate(&mut self) -> Result<(), MetaDataError> {
+        if self.signature != [0x4d, 0x61, 0x79, 0x61, 0x44, 0x61, 0x74, 0x61] {
```
Hoverbear, line 218: It strikes me that this is a magic value which we should probably const so we could use it and find it later.

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -5,821 +5,745 @@
...
-        self.sync_metadata(metadata).await
+        Ok(list)
     }
 
-    /// Append a new config object to "MetaData" partition.
-    pub async fn append_config_object(
-        &mut self,
-        metadata: &mut NexusMetaData,
-        config: &NexusConfig,
+    /// Purge oldest entries from index.
```
Hoverbear, line 708: This doc comment might be improved with details about what the return value is, such as:

```rust
/// Purge oldest entries from index.
///
/// Returns the number of entries removed.
```

mayastor/src/bdev/nexus/nexus_metadata.rs
jkryl: thanks for confirmation 👍

mayastor/src/bdev/nexus/nexus_metadata.rs
jkryl: You are right. The generation number in the index is what I was looking for. Looks good 👍

mayastor/src/bdev/nexus/nexus_metadata.rs
cjones1024: As far as I can tell, most of the variants of the relevant error type pertain to deserialization rather than serialization. The only possible errors here would result from an I/O error on the internal writer (Cursor) which, shouldn't ever fail as long as we don't run out of memory. But I take your point, it is still good practice to test for all (theoretically) possible error conditions.

mayastor/src/bdev/nexus/nexus_metadata.rs
cjones1024: Agreed. Will expand the description for the public functions.

mayastor/src/bdev/nexus/nexus_metadata.rs
tiagolobocastro: When we create the nexus, we should add the first child states, so we should get a validation error which could then fail the creation altogether?

mayastor/src/bdev/nexus/nexus_metadata.rs
tiagolobocastro: what's the idea behind the ring buffer?

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -5,821 +5,745 @@
...
-    },
-    #[snafu(display("Incorrect MetaData header signature"))]
-    HeaderSignature {},
-    #[snafu(display("Incorrect MetaData header checksum"))]
-    HeaderChecksum {},
-    #[snafu(display("Incorrect MetaData index checksum"))]
-    IndexChecksum {},
-    #[snafu(display("Incorrect MetaData configuration object checksum"))]
+    HandleError { source: CoreError, name: String },
+    #[snafu(display("Incorrect MetaDataObject signature"))]
```
tiagolobocastro, line 88: what it be useful ti display what all of these failed things are, what what the correct ones should have been?

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -5,821 +5,745 @@
...
+        let object = NexusMetaData::read_object(&buf)?;
+
+        Ok(Some(object))
     }
 
-    /// Retrieve header + index from "MetaData" partition.
-    pub async fn get_metadata(&self) -> Result<NexusMetaData, MetaDataError> {
-        if let Some(partition) = self
-            .probe_label()
+    /// Return list of the most recent entries but do not modify the index.
```
tiagolobocastro, line 636: If all entries are being used, do we need to make sure we don't return the oldest entry? Eg, we could've previously attempted to write to it and failed or failed update the index so we can't trust it?

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -5,821 +5,745 @@
...
+    IndexSignature {},
+    #[snafu(display("Incorrect MetaDataIndex size"))]
+    IndexSize {},
+    #[snafu(display("Incorrect MetaDataIndex checksum"))]
+    IndexChecksum {},
+}
+
+#[derive(Debug, Default, Deserialize, PartialEq, Serialize)]
+pub struct MetaDataChildEntry {
+    pub guid: Guid,
```
tiagolobocastro, line 106: which guid is this? Currently the children are added by URI. Spdk probably does indeed assign a uuid to each bdev, though not sure if that would be persistent on the children side of things.

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -5,821 +5,745 @@
...
-        .context(ReadError {
-            name: String::from("object"),
-        })?;
+        index.generation += 1;
+        index.timestamp = NexusMetaData::timestamp(now);
+        index.checksum();
+        serialize_into(&mut writer, index).context(SerializeError {})
+    }
+
+    fn write_object(
```
gila, line 355: Im not sure if we have ever discussed this I do recall is talking about this subject but to be sure, how do we handle atomicity? 

What if the buffer we need to write exceeds the blk_size?

mayastor/src/bdev/nexus/nexus_metadata.rs
mtzaurus: From the consistency perspective, it would be nice for this constant to be explicitly expressed via sizes of all the participating structures: that would allow facing errors when we add some new fields to them. Also, would be nice to have some defensive logic (assertions, etc) to guarantee that the total size never goes beyond 512 (if such defensive checks already exist in the code, sorry for overlooking).

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -5,821 +5,745 @@
...
+    pub total_entries: u64,
 }
 
-impl NexusMetaData {
-    /// Construct a MetaDataHeader from raw data
-    fn read_header(buf: &DmaBuf) -> Result<MetaDataHeader, MetaDataError> {
-        MetaDataHeader::from_slice(buf.as_slice())
-    }
+impl MetaDataIndex {
+    pub const METADATA_INDEX_SIZE: u32 = 80;
```
mtzaurus, line 278: Would't it be better to express this constant via the size of the structure ?

cjones1024: Yes. Done.

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -5,821 +5,745 @@
...
-//!    remainder of the block is padded with zeros.
-//!  - The "index" starts at the third block and contains a fixed number of
-//!    MetaDataIndexEntry entries, each of which contains the address of an
-//!    object that has been written to the partition.
-//!  - The first usable "data" block is the first block following the index
-//!    (whose size is aligned to the blocksize of the disk).
+//!  - The second block contains the MetaDataIndex (currently 80 bytes) while
+//!    the remainder of the block is padded with zeros.
+//!  - The data in the "index" starts at the third block.
+//!  - This and each subsequent block contains a single MetaDataObject entry.
```
mtzaurus, line 11: If I understand this correctly, writing a new entry requires update 2 different disk sectors: MetaDataIndex and the MetaDataObject sector. How do we handle scenarios when crash occurs between these 2 writes ?

cjones1024:
bors try

mayastor/src/bdev/nexus/nexus_metadata.rs
cjones1024: Have updated to return a Result rather than trigger a panic even though the error shouldn't really ever occur in practice

mayastor/src/bdev/nexus/nexus_metadata.rs
cjones1024: Done

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -418,7 +418,17 @@ impl Nexus {
...
 
-        n.bdev.set_uuid(uuid.map(String::from));
+        if let Some(s) = uuid {
+            match uuid::Uuid::parse_str(s) {
+                Ok(u) => {
+                    info!("nexus {}: setting bdev UUID: {}", name, s);
+                    n.bdev.set_uuid(u);
+                }
+                Err(error) => {
+                    warn!("nexus {}: invalid UUID: {}: {}", name, s, error);
```
Hoverbear, line 428: Seems this should reutrn the error and destruct the nexus.

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -459,6 +464,9 @@ impl NexusChild {
             state: AtomicCell::new(ChildState::Init),
             prev_state: AtomicCell::new(ChildState::Init),
             remove_channel: mpsc::channel(0),
+            guid: Guid::from_str("00000000-0000-0000-0000-000000000000")
```
Hoverbear, line 467: If we're just setting itt to the same thing every time can we not print it? (Why do we have this?)

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -487,6 +495,7 @@ impl NexusChild {
 
     /// return reference to child's bdev and a new BdevHandle
     /// both must be present - otherwise it is considered an error
+    #[allow(dead_code)]
```
Hoverbear, line 498: Just delete it!

mayastor/src/bdev/nexus/nexus_label.rs
```diff
@@ -566,77 +566,44 @@ pub struct NexusLabel {
 }
 
 impl NexusLabel {
+    #[allow(dead_code)]
```
Hoverbear, line 569: Can we just delete it?

mayastor/src/bdev/nexus/nexus_label.rs
```diff
@@ -1181,81 +1172,125 @@ impl NexusChild {
...
-    /// Create a new label on this child
-    async fn create_label(
-        &mut self,
-        config: &LabelConfig,
-        block_size: u32,
-        data_blocks: u64,
-        total_blocks: u64,
-    ) -> Result<NexusLabel, LabelError> {
-        info!("creating new label for child {}", self.name);
+    // Helper function to generate a new label and index for this child
```
Hoverbear, line 1186: This should be a `///` comment.

mayastor/src/bdev/nexus/nexus_label.rs
```diff
@@ -1269,121 +1304,82 @@ impl NexusChild {
...
-
-        // Generate "reference" partition table entries
-        let header = GptHeader::reference(block_size, nexus_blocks, guid);
-        let reference = Nexus::create_maya_partitions(
-            &config,
-            &header,
-            block_size,
-            nexus_blocks,
-        )?;
+        let now = SystemTime::now();
```
Hoverbear, line 1378: I am a bit worried how using SystemTime might interact over rebuilds with remotes. Do we need to worry about cases where one machine has a very different time than the other?

mayastor/src/bdev/nexus/nexus_label.rs
```diff
@@ -1611,3 +1616,13 @@ impl Aligned for u64 {
         }
     }
 }
+
+fn unique(offsets: &[u64]) -> Option<u64> {
```
Hoverbear, line 1620: Could you document this function? It's not clear what the intent is. It's called unique but it returns an optional single integer. I would typically expect a `unique()` function to return a bool or a number of unique valuesin the vector? Or even a collection of only the unique values?

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -5,821 +5,929 @@
...
-    pub header: MetaDataHeader,
-    pub index: Vec<MetaDataIndexEntry>,
-}
+pub struct NexusMetaData;
 
 impl NexusMetaData {
-    /// Construct a MetaDataHeader from raw data
-    fn read_header(buf: &DmaBuf) -> Result<MetaDataHeader, MetaDataError> {
-        MetaDataHeader::from_slice(buf.as_slice())
+    fn timestamp(time: &SystemTime) -> u128 {
```
Hoverbear, line 361: This function does not interact with `NexusMetaData` at all, it should probably be removed from this trait into a bare namespace.

mayastor/src/bdev/nexus/nexus_metadata.rs
```diff
@@ -5,821 +5,929 @@
...
+    pub table_crc: u32,
+    pub children: Vec<MetaDataChildEntry>,
 }
 
-impl MetaDataHeader {
-    pub const METADATA_HEADER_SIZE: u32 = 72;
-    pub const MAX_INDEX_ENTRIES: u32 = 32;
-    pub const INDEX_ENTRY_SIZE: u32 = 44;
+impl MetaDataObject {
+    /// Observe that sizeof(MetaDataHeader) + 26 * sizeof(MetaDataChildEntry) is
```
Hoverbear, line 174: You might be able to guarantee this with https://docs.rs/static_assertions/1.1.0/static_assertions/macro.const_assert.html

mayastor/src/bdev/nexus/nexus_bdev.rs
cjones1024: This is consistent with the previous behaviour which was to generate a new random UUID. This is unnecessary as the bdev already has a valid UUID (randomly generated at the point of creation).

mayastor/src/bdev/nexus/nexus_child.rs
cjones1024: This is needed to identify the child in a manner that is persistent across restarts. The proper value is "initialised" from the metatadata in the MayaMeta partition, but it is too early to do that here. It just needs to be initialised to something. However have now changed to initialise it to uuid::Uuid::nil()

mayastor/src/bdev/nexus/nexus_label.rs
cjones1024: Done

mayastor/src/bdev/nexus/nexus_label.rs
cjones1024: Have added a self documenting comment. The function is used to check if all elements of an array (slice) have the same value, and returns "that" value wrapped as an Option - and None is returned if the criteria is not satisfied (ie. empty or not all the same). I'm open to suggestions for a better name.

cjones1024:
bors try

mayastor/src/bdev/nexus/nexus_bdev.rs
tiagolobocastro: This might be a bug. Otherwise it'd mean the nexus gets shared via nvmf on the wrong nqn? The csi will not be able to connect to it?

mayastor/src/bdev/nexus/nexus_label.rs
```diff
@@ -367,61 +389,70 @@ pub struct GptHeader {
 
 impl GptHeader {
     pub const PARTITION_TABLE_SIZE: u64 = 128 * 128;
+    pub const DATA_OFFSET: u64 = 1024 * 1024;
```
tiagolobocastro, line 392: how is this offset and size calculated?

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -459,6 +462,8 @@ impl NexusChild {
             state: AtomicCell::new(ChildState::Init),
             prev_state: AtomicCell::new(ChildState::Init),
             remove_channel: mpsc::channel(0),
+            guid: Guid::from(uuid::Uuid::nil()),
```
tiagolobocastro, line 465: Is this nil "here" because it will be overridden later? If so should this be an Option?

mayastor/src/bdev/nexus/nexus_label.rs
cjones1024: The GPT spec definitely does not require this, but most partitioning tools will align partitions to 1 MiB boundaries. It is certainly the case that fdisk does this, and I suspect that is the main reason why we do also.
This means that the start of the first partition is always at an offset of 1024 * 1024 bytes (note this is regardless of the underlying block size).

cjones1024: Also, the partition table size of 128 x 128 represents the "maximum" possible partition table size, although it could conceivably be smaller. However, we always reserve the maximum amount of space so that we can add partitions to the partition table without having to move existing partition boundaries. Again this is entirely consistent with how fdisk behaves.

mayastor/src/bdev/nexus/nexus_bdev.rs
cjones1024: I'm coming around to the view that we probably ought to fail outright in this case.

mayastor/src/bdev/nexus/nexus_child.rs
tiagolobocastro: ok although I'd suggest adding some kind of test or assert in the code that makes sure that we do set the GUID.

mayastor/src/bdev/nexus/nexus_label.rs
tiagolobocastro: :+1:

mayastor/src/bdev/nexus/nexus_bdev.rs
cjones1024: Passing the error up would require more changes that I'd rather not make at this time. Have now changed code to effectively replicate the old behaviour: in the case that no uuid is specified, or if the specified uuid cannot be parsed, then auto-generate a new one.

tiagolobocastro (approved):

cjones1024:
bors try

mayastor/src/bdev/nexus/nexus_metadata.rs
cjones1024: In my view the actual values of these signatures/checksums aren't all that useful. They are just to ensure that the bytes read from the MayaMeta partition actually represent a valid object

mayastor/src/bdev/nexus/nexus_metadata.rs
cjones1024: We don't handle the error explicitly. The only thing that can happen is that (assuming the index is currently full) the oldest entry will be overwritten. This would not normally be a problem if we only retrieve the most recent entry. The only issue is if we retrieve the entire contents of the index, and iven in that case we could detect this case because the generation number would be out of order. This is a minor issue for now as this functionality (retrieving all entries at once) is not used at present.

mayastor/src/bdev/nexus/nexus_metadata.rs
cjones1024: There is a checks insofar as any attempt to add a MetaDataObject that had more than 26 child entries will fail. However as far as I am aware, there are no other checks elsewhere to enforce any limit on the number of children that a Nexus may have.

mayastor/src/bdev/nexus/nexus_metadata.rs
cjones1024: We can detect this case if the generation numbers are out of order, but at this stage it would be up to the caller to do that.

paulyoong (approved):
Approved on the basis that there are known limitations that will be subsequently fixed up.

mayastor/src/bdev/nexus/nexus_label.rs
```diff
@@ -102,16 +107,39 @@ pub enum LabelError {
     ))]
     HandleError { source: CoreError, name: String },
     #[snafu(display(
-        "Device is too small to accomodate Metadata partition: blocks={}",
-        blocks
+        "Device is too small to accomodate Metadata partition: size = {} x {}",
```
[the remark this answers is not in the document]
cjones1024: Fixed

cjones1024:
bors merge

cjones1024:
bors merge

PR #739, opened 2021-02-22
ci: jenkins should fail the pipeline if there are new coredumps

jkryl:
Not only it should fail, but also print more information about
the core dump including the stack.

I have tried to update doc/jenkins.md and correct information that
got stale. However we might remove it in the future or at least
reduce the amount of details. It's difficult to keep up with the
changes to the infrastructure and we have a repo dedicated to
jenkins infrastructure now.

Resolves: CAS-733


tiagolobocastro:
Some of the cargo tests are currently testing existing bugs on the mayastor, and one of them actually causes mayastor to crash (although according to @cjones1024 now it doesn't crash, it just hangs) so not sure if that would trip up this new check.

jkryl:
Hi @tiagolobocastro , if it is a rust panic then that does not trigger the check I have added. From system point of view rust panic is a clean exit of a process. But what it hopefully will catch are crashes caused by segfault, sigbus, sigabort, etc. Those could have been missed until now. That's one thing. The other thing is that even if the test fails because something crashes, then we don't have enough information to debug the issue because we don't know anything about the crash itself. Hopefully this change should improve things on both fronts.

tiagolobocastro:
> Hi @tiagolobocastro , if it is a rust panic then that does not trigger the check I have added. From system point of view rust panic is a clean exit of a process. But what it hopefully will catch are crashes caused by segfault, sigbus, sigabort, etc. Those could have been missed until now. That's one thing. The other thing is that even if the test fails because something crashes, then we don't have enough information to debug the issue because we don't know anything about the crash itself. Hopefully this change should improve things on both fronts.

By default that's true, but our Cargo.toml has overrides that I think:
`[profile.dev]
panic = "abort"`

I'm not sure why that was added, maybe when the tests were failing all over is no longer required?

jkryl:
I recall something about it. It was added because the errors from child threads were not properly propagated to and reported by the parent. So we had unit test that were reported as ok but in fact one of the threads panic'd :-( Not sure if we need to keep that behaviour or not. Maybe @gila would know.

Back to your original question, I don't understand the reason behind such test. We test a known failure in SPDK, let it dump a core and if the core is dumped we report ok. I'm missing a sense of such test. IMHO if we know that the test crashes mayastor then we should either fix mayastor/spdk or we should not run it. cc @cjones1024  in case he can provide more info about the test that you have mentioned Tiago. Thanks for bringing this up!

gila:
> I recall something about it. It was added because the errors from child threads were not properly propagated to and reported by the parent. So we had unit test that were reported as ok but in fact one of the threads panic'd :-( Not sure if we need to keep that behaviour or not. Maybe @gila would know.
> 
> Back to your original question, I don't understand the reason behind such test. We test a known failure in SPDK, let it dump a core and if the core is dumped we report ok. I'm missing a sense of such test. IMHO if we know that the test crashes mayastor then we should either fix mayastor/spdk or we should not run it. cc @cjones1024 in case he can provide more info about the test that you have mentioned Tiago. Thanks for bringing this up!

IIRC the setting only indicates if we unwind and then panic, or right away, panic. When unwinding, we try to walk back up the stack. According to the docs unwinding with FFI is undefined. Panic behaviour does not change what Rust does. If threads panic, it is caught. So a panic in thread A does not impact thread B, but only if the code itself is UnwindSafe (which FFI is not). So in the example of the tests, if we call `unwrap()`, which induces a panic, the thread is caught and the test is failed. 

It is possible to override this behaviour and call `core::intrinsics::abort`, which would result (I think) in the more "traditional" abort behaviour.

tiagolobocastro:
> I recall something about it. It was added because the errors from child threads were not properly propagated to and reported by the parent. So we had unit test that were reported as ok but in fact one of the threads panic'd :-( Not sure if we need to keep that behaviour or not. Maybe @gila would know.
> 
> Back to your original question, I don't understand the reason behind such test. We test a known failure in SPDK, let it dump a core and if the core is dumped we report ok. I'm missing a sense of such test. IMHO if we know that the test crashes mayastor then we should either fix mayastor/spdk or we should not run it. cc @cjones1024 in case he can provide more info about the test that you have mentioned Tiago. Thanks for bringing this up!

Yes we have a jira for it and @cjones1024 is working on it. Also I remember now it's not cargo that panics, it's the mayastor running in a container that panics. If this fails this PR as expected then please remove the second for loop from the test `create_nexus_sizes`. Once Colin fixes the issue we can add it back.

scripts/check-coredumps.sh
```diff
@@ -0,0 +1,70 @@
...
+# Globals
+since="2000-01-01 00:00:00"
+
+help() {
+  cat <<EOF
+Usage: $0 [--since DATE]
+
+Options:
+  --since DATE   Report only coredumps newer than provided date.
+                 (DATE format is YYYY-MM-DD HH:MM:SS)
```
arne-rusek, line 17:
```suggestion
                 (DATE format is YYYY-MM-DD HH:MM:SS - due to space needs quotes!)
```

arne-rusek (approved):
LGTM; help update suggested

tiagolobocastro (approved):
I haven't looked at the CI to see whether we get coredumps on a regular basis but if we do I'd recommend we introduce this without actually failing the pipeline for now.

arne-rusek:
> I haven't looked at the CI to see whether we get coredumps on a regular basis but if we do I'd recommend we introduce this without actually failing the pipeline for now.

50 since 2021-02-23 :face_with_head_bandage:

jkryl:
let's find out. I will merge it and if we get many coredumps that we don't know how to fix, we can always disable the check as a last resort.

jkryl:
bors merge

PR #742, opened 2021-02-24
test(e2e): extend e2e continuous tests

chriswldenyer:
Define the continuous tests in the test script rather
than the Jenkinsfile. Add script option --profile
to specify this. 
Add rebuild and replica e2e tests to continuous runs.
Ensure commit status is not affected by running continuous tests.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -85,8 +85,6 @@ if (currentBuild.getBuildCauses('jenkins.branch.BranchIndexingCause') &&
 
 // Only schedule regular builds on develop branch, so we don't need to guard against it
 String cron_schedule = BRANCH_NAME == "develop" ? "0 2 * * *" : ""
-// Some long e2e tests are not suitable to be run for each PR
-boolean run_extended_e2e_tests = (env.BRANCH_NAME != 'staging' && env.BRANCH_NAME != 'trying') ? true : false
 
 // Determine which stages to run
 if (params.e2e_continuous == true) {
@@ -95,7 +93,7 @@ if (params.e2e_continuous == true) {
   grpc_test = false
   moac_test = false
   e2e_test = true
-  e2e_test_set = 'install basic_volume_io csi resource_check uninstall'
+  e2e_test_profile = "continuous"
   // use images from dockerhub tagged with e2e_continuous_image_tag instead of building from current source
   e2e_build_images = false
   // do not push images even when running on master/develop/release branches
@@ -106,7 +104,8 @@ if (params.e2e_continuous == true) {
   grpc_test = true
   moac_test = true
   e2e_test = true
-  e2e_test_set = ''
+  // Some long e2e tests are not suitable to be run for each PR
+  e2e_test_profile = (env.BRANCH_NAME != 'staging' && env.BRANCH_NAME != 'trying') ? "extended" : "ondemand"
   e2e_build_images = true
   do_not_push_images = false
 }
@@ -291,17 +290,11 @@ pipeline {
                   } else {
                     tag = e2e_continuous_image_tag
                   }
-                  def cmd = "./scripts/e2e-test.sh --device /dev/sdb --tag \"${tag}\" --logs --logsdir \"./logs/mayastor\" "
+                  def cmd = "./scripts/e2e-test.sh --device /dev/sdb --tag \"${tag}\" --logs --logsdir \"./logs/mayastor\" --profile \"${e2e_test_profile}\" "
 
                   // building images also means using the CI registry
                   if (e2e_build_images == true) {
                     cmd = cmd + " --registry \"" + env.REGISTRY + "\""
-                  }
-                  if (run_extended_e2e_tests) {
-                    cmd = cmd + " --extended"
-                  }
-                  if (e2e_test_set != '') {
-                      cmd = cmd + " --tests \"" + e2e_test_set + "\""
                   }
                   sh "nix-shell --run '${cmd}'"
                 }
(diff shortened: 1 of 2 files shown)


chriswldenyer:
bors try

chriswldenyer:
bors try

chriswldenyer:
bors try-

chriswldenyer:
bors try

chriswldenyer:
bors try

blaisedias (commented):
The e2e-test.sh script is getting complicated, with multiple options to select tests,
some of which are mutually exclusive.
It strikes me that it would be simpler if we should replace those options
with a single option which selects an "test execution profile".
So profiles would be something like "extended", "continuous" and "ondemand"

scripts/e2e-test.sh
```diff
@@ -142,11 +146,32 @@ fi
 
 export e2e_docker_registry="$registry" # can be empty string
 
-if [ -z "$tests" ]; then
-  tests="$TESTS"
-  if [ -n "$run_extended_tests" ]; then
-    tests="$tests $EXTENDED_TESTS"
+if [ -n "$profile" ]; then
```
blaisedias, line 149: I think for clarity we should 
1) check and exit if  `$tests` and `$profile` are both set.
2) Default the value of `$profile` to `ondemand`, if `$profile` is empty.

Then it is eminently clear which profile - if any - is being used,
in later parts of the script (if required)

scripts/e2e-test.sh
```diff
@@ -41,7 +43,8 @@ Options:
   --tests <list of tests>   Lists of tests to run, delimited by spaces (default: "$tests")
         Note: the last 2 tests should be (if they are to be run)
              node_disconnect/replica_pod_remove uninstall
-  --extended                Run long running tests also.
+  --profile <extended|continuous|ondemand>
+                            Run the tests corresponding to the profile.
```
blaisedias, line 47: Add `(default :ondemand)` to help text

blaisedias (changes requested):

scripts/e2e-test.sh
chriswldenyer: Maybe the default profile should be "default" - "ondemand" shouldn't be special.
"--profile default" is the same as --profile being absent, in which case $profile is set to "default"

chriswldenyer: OTOH allowing "--profile default" is maybe going too far.

scripts/e2e-test.sh
chriswldenyer: The default is no longer equivalent to ondemand.

scripts/e2e-test.sh
chriswldenyer: 1. Done,
2. Now has a profile named "default" if profile is not specified.

scripts/e2e-test.sh
blaisedias: Then it is unclear what happens if no profile is set?

blaisedias (approved):
Looks okay to me.

scripts/e2e-test.sh
chriswldenyer: Like the current behaviour of e2e-test.sh ?
I could change the text to "Run the tests corresponding to the profile instead of the default tests."

blaisedias: Yeah that would work.

blaisedias (approved):
LGTM

chriswldenyer:
bors try

jkryl (commented):
The changes look good. I'm just not sure what is the use case for continuous tests. We already have extended tests that should run nightly. I can see that continuous tests have two more tests compared to extended tests. So when one should run continuous and when extended?

chriswldenyer:
> The changes look good. I'm just not sure what is the use case for continuous tests. We already have extended tests that should run nightly. I can see that continuous tests have two more tests compared to extended tests. So when one should run continuous and when extended?

The continuous tests are orientated towards e2e tests rather than component tests. We want them running as often as possible to have the best chance of catching transient errors. We also want to run them on released images rather than code that is still in develop.
We also run a smaller set of e2e tests as part of the git merge on-demand CI. (At one point we were going to remove e2e tests completely from the on-demand tests, in which case the continuous runs would be the only ones doing e2e.) As it stands now, we are doing e2e in both on-demand and continuous, but have significantly fewer of the e2e tests running in on-demand so as not to add too much of a time penalty (or risk of transient failures) to merge testing.

I'm not sure about extended. They are run on non-trying and non-staging branches only because we don't want to add a time penalty to PRs. But they are probably worth doing because they are testing more recent code than published images, and because we have the time to do it in a nightly run.

chriswldenyer:
bors try

Jenkinsfile
```diff
@@ -445,25 +438,27 @@ pipeline {
...
-              contextSource: [
-                $class: 'ManuallyEnteredCommitContextSource',
-                context: 'continuous-integration/jenkins/branch'
-              ],
-              statusResultSource: [
-                $class: 'ConditionalStatusResultSource',
-                results: [
-                  [$class: 'AnyBuildResult', message: 'Pipeline result', state: currentBuild.getResult()]
+            // Do not update the commit status for continuous tests
+            if (params.e2e_continuous == false) {
```
arne-rusek, line 442: I _think_ you still need the `currentBuild.result != null` condition as sometimes all steps are skipped but this will be setting status. Please check "empty" pipelines in https://mayastor-ci.mayadata.io/job/Mayastor/job/develop/

arne-rusek (changes requested):

Jenkinsfile
arne-rusek: I'm colorblind I guess, sorry, the condition is there still :)

arne-rusek (approved):

chriswldenyer:
bors merge

PR #745, opened 2021-02-25
fix(moac): mayastor pool state is shown as blank

jkryl:
Well known problem with a decent workaround now. After experimenting
with tcp keep-alive that did not bring any relief I decided to brush
the dust of my previous mini-feature with restarting watcher connections
every n minutes. By default, as defined in the yaml file now, it is every
10 minutes. Most of the diff is about updating version of javascript
k8s client. That change will be merged to the upstream eventually.
What it does is that the MOD event is not emitted upon watcher restart
if the resource has not changed. Hence it makes the watcher restarts really
smooth. I have tested the change in Google Cloud.

Resolves: CAS-645 and CAS-630


jkryl:
The link to the changes to upstream: https://github.com/kubernetes-client/javascript/pull/597

tiagolobocastro (approved):
:+1:

GlennBullingham (approved):

jkryl:
bors merge

PR #757, opened 2021-03-02
feat: support multiple client output formats

Hoverbear:
Support both tables/plaintext and JSON in outputs.

Users can now specify `-o json` or `-o default` to configure their output options.
`-o default` is implied. This mimics the `nvme` tool.

Normalizes the output of commands to be similar to `docker` or `kubectl` style commands,
returning relevant URIs, names, etc.

Resolves https://github.com/openebs/Mayastor/issues/756

### DIFF
--- a/doc/mayastor-client.md
+++ b/doc/mayastor-client.md
@@ -81,7 +81,7 @@ a URI to the resource and we can start using it.
 
 ```bash
 > mayastor-client nexus create `uuidgen -r` 1GiB aio:///dev/sdb
-Nexus 4db90841-5ee8-4b7d-a4e9-13be1043bcb3 created
+4db90841-5ee8-4b7d-a4e9-13be1043bcb3
 ```
 Tip: To find out what the arguments are, simply append the `-h` flag to any command.
 
@@ -101,7 +101,7 @@ fine writing to it as it were a local disk.
 ```bash
 > fallocate -l 2GiB /data/file.img
 > mayastor-client nexus create `uuidgen -r` 1GiB 'aio:///data/file.img?blk_size=512 aio:///dev/sdb'
-Nexus d0c47a07-d104-48e6-8f36-bfdb47e8e766 created
+d0c47a07-d104-48e6-8f36-bfdb47e8e766
 ```
 
 Notice the added query parameter `blk_size`, required as files do not have block sizes.
@@ -117,7 +117,7 @@ yourself by running fio on top of the NBD device.
 
 ```bash
 > mayastor-client nexus remove d0c47a07-d104-48e6-8f36-bfdb47e8e766 'aio:///data/file.img?blk_size=512'
-Removed aio:///data/file.img?blk_size=512 from children of d0c47a07-d104-48e6-8f36-bfdb47e8e766
+aio:///data/file.img?blk_size=512
 ```
 
 In the logs of mayastor you will see something like:
@@ -134,7 +134,7 @@ Now we can add the device again:
 
 ```bash
 > mayastor-client nexus add d0c47a07-d104-48e6-8f36-bfdb47e8e766 'aio:///data/file.img?blk_size=512'
-Added aio:///data/file.img?blk_size=512 to children of d0c47a07-d104-48e6-8f36-bfdb47e8e766
+aio:///data/file.img?blk_size=512
 ```
 
 Both the nexus and the newly added children are now degraded. The child is degraded because it needs to be rebuilt
@@ -280,9 +280,8 @@ We are maxing out at roughly 90MB, as this a 1 GbE network that is to be expecte
 Now, let's disconnect it and create a Nexus that that consumes one of the NVMe targets and rerun the test:
 
 ```bash
-    nvme disconnect -d {/dev/nvme1,/dev/nvme2}
-    mayastor-client nexus create `uuidgen -r` 64MiB 'nvmf://192.168.1.2/nqn.2019-05.io.openebs:cnode2
-        nvmf://192.168.1.2/nqn.2019-05.io.openebs:cnode1'
+nvme disconnect -d {/dev/nvme1,/dev/nvme2}
+mayastor-client nexus create `uuidgen -r` 64MiB 'nvmf://192.168.1.2/nqn.2019-05.io.openebs:cnode2 nvmf://192.168.1.2/nqn.2019-05.io.openebs:cnode1'
 ```
 
 Ok we now have created a nexus that consists out of 2 replica's:
@@ -300,7 +299,7 @@ and NVMF
 
 ```bash
 > mayastor-client nexus publish 787f82e7-e7d8-4ae1-8a25-5d48ead4f4cd
-Nexus published at file:///dev/nbd0
+file:///dev/nbd0
 ```
 
 And the results:
--- a/mayastor/src/bin/mayastor-client/device_cli.rs
+++ b/mayastor/src/bin/mayastor-client/device_cli.rs
@@ -2,27 +2,23 @@
 //! methods to obtain information about block devices on the current host
 
 use super::context::Context;
+use crate::{context::OutputFormat, GrpcStatus};
 use ::rpc::mayastor as rpc;
 use clap::{App, AppSettings, Arg, ArgMatches, SubCommand};
 use colored_json::ToColoredJson;
+use snafu::ResultExt;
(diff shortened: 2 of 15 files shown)


Hoverbear:
This allows for stuff like:

```bash
SHARE_URIS=(`mayastor-client replica list -o json | jq -r ".replicas[].uri"`)
PUBLISH_URI=`mayastor-client nexus list -o json | jq -r ".nexus_list[].device_uri"`
```

Or

```bash
truncate -s 1G tmp/loopbackfile3.img
sudo losetup --sector-size 4096 /dev/loop3 tmp/loopbackfile3.img
mayastor-client pool create pool-${NUM} "uring:///dev/loop${NUM}?blk_size=4096"
REPLICA_UUID=`uuidgen`
NEW_URI=`mayastor-client replica create pool-${NUM} ${REPLICA_UUID} --size ${SIZE}`
mayastor-client nexus add ${NEXUS_UUID} ${NEW_URI}
```

Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
bors try

doc/mayastor-client.md
```diff
@@ -101,7 +101,7 @@ fine writing to it as it were a local disk.
 ```bash
 > fallocate -l 2GiB /data/file.img
 > mayastor-client nexus create `uuidgen -r` 1GiB 'aio:///data/file.img?blk_size=512 aio:///dev/sdb'
```
cjones1024, line 103: Should change this to reflect that children are now to be passed as separate arguments?

doc/mayastor-client.md
```diff
@@ -280,9 +280,8 @@ We are maxing out at roughly 90MB, as this a 1 GbE network that is to be expecte
 Now, let's disconnect it and create a Nexus that that consumes one of the NVMe targets and rerun the test:
 
 ```bash
-    nvme disconnect -d {/dev/nvme1,/dev/nvme2}
-    mayastor-client nexus create `uuidgen -r` 64MiB 'nvmf://192.168.1.2/nqn.2019-05.io.openebs:cnode2
-        nvmf://192.168.1.2/nqn.2019-05.io.openebs:cnode1'
+nvme disconnect -d {/dev/nvme1,/dev/nvme2}
+mayastor-client nexus create `uuidgen -r` 64MiB 'nvmf://192.168.1.2/nqn.2019-05.io.openebs:cnode2 nvmf://192.168.1.2/nqn.2019-05.io.openebs:cnode1'
```
cjones1024, line 284: Same change as above

cjones1024:
This is more me thinking aloud rather than a critique, but I had the idea that we could give this code more structure and make it more standardised by introducing a trait along the lines of
`
trait OutputDisplay {
    fn display_default(&self, ctx: &Context, args: &ArgMatches) -> Result<(), Status>;
}
`
which we would implement for each gRPC response type.

We could then go further and add a display_json() method as well -- with a default implementation to avoid duplicating much common code.  We would end end up with something like the following:
`
trait OutputDisplay: serde::ser::Serialize {
    fn display(&self, ctx: &Context, args: &ArgMatches) -> Result<(), Status> {
        match ctx.output {
            OutputFormat::Json => self.display_json(ctx, args),
            OutputFormat::Default => self.display_default(ctx, args),
        }
    }
    fn display_json(&self, _ctx: &Context, _args: &ArgMatches) -> Result<(), Status> {
        println!(
            "{}",
            serde_json::to_string_pretty(self).unwrap().to_colored_json_auto().unwrap()
        );
        Ok(())
    }
    fn display_default(&self, ctx: &Context, args: &ArgMatches) -> Result<(), Status>;
}
`
and then the implementation for each method would be, for example:
`
async fn nexus_list(
    mut ctx: Context,
    args: &ArgMatches<'_>,
) -> Result<(), Status> {
    let response = ctx.client.list_nexus(rpc::Null {}).await?;
    response.get_ref().display(&ctx, args)
}

impl OutputDisplay for rpc::ListNexusReply {
    fn display_default(&self, ctx: &Context, args: &ArgMatches) -> Result<(), Status> {
        let nexus = &self.nexus_list;
        ...
    }
}
`

Hoverbear:
@cjones1024 Yesss omg that would be lovely. Do you want that change to be included with this, or a follow up PR? (Happy to do it)

cjones1024:
> 
> 
> @cjones1024 Yesss omg that would be lovely. Do you want that change to be included with this, or a follow up PR? (Happy to do it)

If you want to do it in a separate Jira/PR that's fine, but I think its probably just as easy to do as part of this one.

mayastor/src/bin/mayastor-client/device_cli.rs
```diff
@@ -2,27 +2,16 @@
...
-                .help("List all block devices (ie. also include devices currently in use)"),
-        )
-        .arg(
-            Arg::with_name("raw")
-                .long("raw")
-                .takes_value(false)
-                .help("Display output as raw JSON"),
-        );
+    let list =
+        SubCommand::with_name("list").about("List available block devices");
```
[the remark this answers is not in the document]
Hoverbear: Oh, you're right!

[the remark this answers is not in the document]
Hoverbear: Actually, `--all` was removed. Is there a case to keep it? Admittedly, I've yet to see the command actually output anything.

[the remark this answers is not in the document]
jonathan-teh: It's still there, line 51, the first statement in `list_block_devices()`.

Hoverbear:
@cjones1024 I thought more about this.

We'd have to make these impls either in `rpc` or on wrapper types in `mayastor`. This means `OutputFormat` etc would also need to be created in `rpc`, despite it not relating to `rpc`, if we want to avoid creating dependency cycles. This isn't really ideal.

I'd prefer to address this in a separate ticket which separates this change of user-API from internal API changes.

Perhaps even use something like https://docs.rs/cli-table/0.4.5/cli_table/#derive-macro combined with https://docs.rs/tonic-build/0.4.0/tonic_build/struct.Builder.html#method.type_attribute?

Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
bors try

mayastor/.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
jonathan-teh, line 1: The top-level `.gitignore` already has a `.idea` entry so these shouldn't be committed.

jonathan-teh (approved):
LGTM, the commits should be squashed as they're logically part of the same feature.

mayastor/.idea/.gitignore
Hoverbear: Oh my gosh whoops. :sweat: Sorry.

cjones1024:
> 
> 
> @cjones1024 I thought more about this.
> 
> We'd have to make these impls either in `rpc` or on wrapper types in `mayastor`. This means `OutputFormat` etc would also need to be created in `rpc`, despite it not relating to `rpc`, if we want to avoid creating dependency cycles. This isn't really ideal.
> 
> I'd prefer to address this in a separate ticket which separates this change of user-API from internal API changes.
> 

Yes, perfectly fine to do this under a separate ticket.

doc/mayastor-client.md
cjones1024: Has this been changed? - The two children are still passed as a single argument, and this will no longer work.

doc/mayastor-client.md
cjones1024: Again, not clear if this has been changed. Also better if at all possible to try and prevent line breaks in the middle of an argument

Hoverbear:
@cjones1024 is there anything else blocking you from approving this PR? :)

cjones1024 (approved):

Hoverbear:
bors merge

Hoverbear:
bors try

Hoverbear:
Oh wait... bors merge

Hoverbear:
bors merge

Hoverbear:
bors merge

Hoverbear:
bors merge

PR #764, opened 2021-03-03
ci: fixes for correctness and maintainability MQ-232

blaisedias:
Add support for configuration of test parameters by a selectable
yaml file. Can be JSON or TOML as well.
Add a configuration file for CI/CD
Fix install test to conform with other tests
Replace uses of Log.Error with Log.Info when reporting k8s errors
Check mayastor pods health in common AfterEach function
Generate artifacts under single directory artifacts,
and modify so that jenkins collects artifacts from there
for each test run.
This includes configuration files, install yaml files and log files
Use a symbol for mayastor namespace instead of using explicit strings.

### DIFF
--- a/scripts/e2e-test.sh
+++ b/scripts/e2e-test.sh
@@ -5,6 +5,7 @@ set -eu
 SCRIPTDIR=$(dirname "$(realpath "$0")")
 TESTDIR=$(realpath "$SCRIPTDIR/../test/e2e")
 REPORTSDIR=$(realpath "$SCRIPTDIR/..")
+ARTIFACTSDIR=$(realpath "$SCRIPTDIR/../artifacts")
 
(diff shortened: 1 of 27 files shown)


blaisedias:
bors try

scripts/e2e-test.sh
```diff
@@ -103,15 +105,16 @@ while [ "$#" -gt 0 ]; do
...
                 on_fail=$1
                 ;;
             stop)
                 on_fail=$1
                 ;;
             *)
+                echo "invalid option for --onfail"
                 help
-                exit 2
+                exit 1
```
chriswldenyer, line 117: It would be nice to have symbols for these exit values

blaisedias: Done

PR #765, opened 2021-03-03
fix(nexus): mayastor hangs when creating a nexus if there is an error creating a child

cjones1024:
When creating a new nexus, as a final step, the newly created
nexus is added to a global list of nexus instances.

This list is required when handling bdev removal - specifically
it is used by lookup_child_from_bdev() to determine the nexus child
that is associated with a given bdev.

The problem occurs when there is an error creating a nexus,
and proper cleanup necessitates the removal of some children that may
have already been successfully created. The removal code requires the
owning nexus to be present in the global list in order to successfully
remove the children, but the nexus has not yet been added to the list,
as it has not been successfully created.

The solution is to add the (partially created) nexus to the list of global
instances as early as possible in the creation process. This means that we
also need to ensure that it is removed again if any error is encountered.

resolves CAS-757

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_bdev.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev.rs
@@ -11,7 +11,7 @@ use std::{
     os::raw::c_void,
 };
 
-use futures::channel::oneshot;
+use futures::{channel::oneshot, future::join_all};
 use nix::errno::Errno;
 use serde::Serialize;
 use snafu::{ResultExt, Snafu};
@@ -424,10 +424,7 @@ impl Nexus {
             max_io_attempts: cfg.err_store_opts.max_io_attempts,
         });
 
-        n.bdev.set_uuid(match uuid {
-            Some(uuid) => Some(uuid.to_string()),
-            None => None,
-        });
+        n.bdev.set_uuid(uuid.map(String::from));
 
         if let Some(child_bdevs) = child_bdevs {
             n.register_children(child_bdevs);
@@ -1033,11 +1030,13 @@ impl Nexus {
     }
 }
 
-/// If we fail to create one of the children we will fail the whole operation
-/// destroy any created children and return the error. Once created, and we
-/// bring the nexus online, there still might be a configuration mismatch that
-/// would prevent the nexus to come online. We can only determine this
-/// (currently) when online, so we check the errors twice for now.
+/// Create a new nexus and bring it online.
+/// If we fail to create any of the children, then we fail the whole operation.
+/// On failure, we must cleanup by destroying any children that were
+/// successfully created. Also, once the nexus is created, there still might
+/// be a configuration mismatch that would prevent us from going online.
+/// Currently, we can only determine this once we are already online,
+/// and so we check the errors twice for now.
 #[tracing::instrument(level = "debug")]
 pub async fn nexus_create(
     name: &str,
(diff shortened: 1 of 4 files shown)


cjones1024:
bors try

tiagolobocastro (approved):
:+1: Please also remove the fail attribute from the create_nexus_sizes test

jonathan-teh (approved):
Looks fine. Please include in the commit message why it hangs and a summary of how this fixes it.

cjones1024:
bors try

cjones1024:
bors merge

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -1047,61 +1046,85 @@ pub async fn nexus_create(
...
         return Ok(());
     }
 
-    let mut ni = Nexus::new(name, size, uuid, None);
+    // Create a new Nexus object, and immediately add it to the global list.
+    // This is necessary to ensure proper cleanup, as the code responsible for
+    // closing a child assumes that the nexus to which it belongs will appear
+    // in the global list of nexus instances. We must also ensure that the
+    // nexus instance gets removed from the global list if an error occurs.
+    nexus_list.push(Nexus::new(name, size, uuid, None));
```
Hoverbear, line 1061: Can we pass the children in here, instead of having to enumerate them later?

Hoverbear, line 1061: Could you get the nexus by index after you push it in directly, to avoid having to scan?

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -1047,61 +1046,85 @@ pub async fn nexus_create(
...
-        Ok(_) => nexus_list.push(ni),
+        Ok(_) => Ok(()),
+    }
+}
+
+/// Destroy list of child bdevs
+async fn destroy_child_bdevs(name: &str, list: &[String]) {
+    let futures = list.iter().map(String::as_str).map(bdev_destroy);
+    let results = join_all(futures).await;
+    if results.iter().any(|c| c.is_err()) {
```
Hoverbear, line 1117: This will finish after any failure, it seems we'd want to ensure all destruction commands are run and attempted to be completed?

mayastor/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -345,13 +345,13 @@ impl Nexus {
...
     }
-    /// destroy all children that are part of this nexus closes any child
-    /// that might be open first
-    pub(crate) async fn destroy_children(&mut self) {
+
+    /// Close each child that belongs to this nexus.
+    pub(crate) async fn close_children(&mut self) {
         let futures = self.children.iter_mut().map(|c| c.close());
         let results = join_all(futures).await;
         if results.iter().any(|c| c.is_err()) {
```
Hoverbear, line 353: This will finish after any failure, it seems we'd want to ensure all destruction commands are run and attempted to be completed?

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -1047,61 +1046,85 @@ pub async fn nexus_create(
...
-            error!("failed to create child {}: {}", child, err);
-            ni.destroy_children().await;
-            return Err(err).context(CreateChild {
-                name: ni.name.clone(),
+        if let Err(error) = ni.create_and_register(child).await {
+            error!(
+                "failed to create nexus {}: failed to create child {}: {}",
+                name, child, error
+            );
+            ni.close_children().await;
```
Hoverbear, line 1078: This seems to close all children each iteration? Why?

mayastor/src/bdev/nexus/nexus_bdev.rs
cjones1024: Ouch! Yes you are right. I hade mistakenly believed that with join_all(), all subfutures are driven to completion.

cjones1024: Hmm ... According to the documentation, the *old* behaviour was that futures are cancelled immediately on first error. However, as of version 0.3.0, no mention of this is made in the docs any more - suggesting now that the current behaviour is exactly what we want.

mayastor/src/bdev/nexus/nexus_bdev.rs
cjones1024: (Soon) after close_children() is called we return with an error. It will only ever be called at most once.

mayastor/src/bdev/nexus/nexus_bdev.rs
cjones1024: I don't know for sure if that code path still works as intended. This is how the existing code was and I was not prepared to change it.

cjones1024: Not sure if we can safely assume that the list cannot be updated mean-time. This seemed safer if not quite as efficient.
What I really wanted to do was to push the object on the list (a move) while simultaneously retaining a mutable reference to it, but alas the borrow checker says no.

Hoverbear: Yeah, makes sense!

Hoverbear: :sweat_smile: Oh... kay. I guess this is for a later issue.

cjones1024:
bors merge

PR #771, opened 2021-03-04
feat(nexus): implement set ANA state for an NVMf-published Nexus

jonathan-teh:
Implement set_ana_state in nexus_bdev::Nexus and expose that over
gRPC and the client CLI.

Fixes CAS-756, #758

### DIFF
--- a/doc/run.md
+++ b/doc/run.md
@@ -18,7 +18,7 @@ It will also need at least 512 2 MB Hugepages configured.
 
 > Learn more about hugepages: [parts 1][hugepages-lwn-one], [2][hugepages-lwn-two],
 > [3][hugepages-lwn-three], [4][hugepages-lwn-four], [5][hugepages-lwn-five].
-  
+
 In NixOS:
 
 ```nix
@@ -78,7 +78,7 @@ In order to use the full feature set of Mayastor, some or all of the following c
   + `nvme_loop`: NVMe Loop Device support
 
   To load these on NixOS:
-  
+
   ```nix
   # /etc/nixos/configuration.nix
   boot.kernelModules = [
@@ -91,7 +91,40 @@ In order to use the full feature set of Mayastor, some or all of the following c
   ```bash
   modprobe nbd nvmet nvmet_rdma nvme_fabrics nvme_tcp nvme_rdma nvme_loop
   ```
-* An NVMe device. (Typically via PCI-E through an standard slot or [M.2][m-dot-2] port)
+* For Asymmetric Namespace Access (ANA) support (early preview), the following kernel build configuration enabled:
+  + `CONFIG_NVME_MULTIPATH`: enables support for multipath access to NVMe subsystems
+
+  This is usually already enabled in distributions kernels, at least for RHEL/CentOS 8.2, Ubuntu 20.04 LTS, and SUSE Linux Enterprise 15.2.
+
+  On some distributions such as RHEL 8, the feature must be enabled manually:
+
+  ```sh
+  # /etc/modprobe.d/nvme-multipath
+  options nvme_core multipath=1
+  ```
+
+  followed by reloading the `nvme-core` module or rebooting.
+
+  To build this on NixOS:
+
+  ```nix
+  # /etc/nixos/configuration.nix
+  boot.kernelPackages = pkgs.linuxPackages;
+  boot.kernelPatches = [ {
+    name = "nvme-multipath";
+    patch = null;
+    extraConfig = ''
+      NVME_MULTIPATH y
+      '';
+  } ];
+  ```
+
+  followed by:
+
+  ```sh
+  sudo nixos-rebuild boot
+  ```
+* An NVMe device. (Typically via PCI-E through a standard slot or [M.2][m-dot-2] port)
 * A version of [`nix`][nix-install] configured as in the [build guide.][doc-build]
 
 ## Running binaries directly
@@ -158,7 +191,7 @@ Why these parameters?
 - `--privileged` to allow controlling memory policies.
 
   > **TODO:** We can use [control groups][control-groups] for this!
-- `-v /dev:/dev:rw` is needed to get access to any raw device you might want to consume as local 
+- `-v /dev:/dev:rw` is needed to get access to any raw device you might want to consume as local
   storage and huge pages
 - `-v /dev/shm:/dev/shm:rw` is needed as for a circular buffer that can trace any IO operations
   as they happen
@@ -184,7 +217,7 @@ nixpkgs.overlays = [
 ];
 
 systemd.services.mayastor = {
-  wantedBy = [ "multi-user.target" ]; 
+  wantedBy = [ "multi-user.target" ];
   after = [ "network.target" ];
   description = "A cloud native declarative data plane.";
   serviceConfig = {
@@ -273,4 +306,3 @@ production Mayastor deployment and operation instructions.
 [lxd]: https://linuxcontainers.org/
 [libvirtd]: https://libvirt.org/index.html
 [terraform-readme]: ./terraform/readme.adoc
-[aarch64-branch]: 
--- a/mayastor/src/bdev/nexus/nexus_bdev.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev.rs
@@ -13,6 +13,7 @@ use std::{
 
 use futures::{channel::oneshot, future::join_all};
 use nix::errno::Errno;
+use rpc::mayastor::NvmeAnaState;
 use serde::Serialize;
 use snafu::{ResultExt, Snafu};
 use tonic::{Code, Status};
@@ -106,6 +107,8 @@ pub enum Error {
     AlreadyShared { name: String },
     #[snafu(display("The nexus {} has not been shared", name))]
     NotShared { name: String },
+    #[snafu(display("The nexus {} has not been shared over NVMf", name))]
+    NotSharedNvmf { name: String },
     #[snafu(display("Failed to share nexus over NBD {}", name))]
     ShareNbdNexus { source: NbdError, name: String },
     #[snafu(display("Failed to share iscsi nexus {}", name))]
@@ -233,6 +236,8 @@ pub enum Error {
     },
     #[snafu(display("Invalid ShareProtocol value {}", sp_value))]
     InvalidShareProtocol { sp_value: i32 },
+    #[snafu(display("Invalid NvmeAnaState value {}", ana_value))]
+    InvalidNvmeAnaState { ana_value: i32 },
     #[snafu(display("Failed to create nexus {}", name))]
     NexusCreate { name: String },
     #[snafu(display("Failed to destroy nexus {}", name))]
@@ -252,6 +257,16 @@ pub enum Error {
     FailedGetHandle,
     #[snafu(display("Failed to create snapshot on nexus {}", name))]
     FailedCreateSnapshot { name: String, source: CoreError },
+    #[snafu(display("NVMf subsystem error: {}", e))]
+    SubsysNvmfError { e: String },
+}
+
+impl From<subsys::NvmfError> for Error {
+    fn from(error: subsys::NvmfError) -> Self {
+        Error::SubsysNvmfError {
+            e: error.to_string(),
+        }
+    }
 }
 
 impl From<Error> for tonic::Status {
@@ -270,6 +285,9 @@ impl From<Error> for tonic::Status {
                 ..
             } => Status::invalid_argument(e.to_string()),
             Error::NotShared {
+                ..
+            } => Status::invalid_argument(e.to_string()),
+            Error::NotSharedNvmf {
                 ..
             } => Status::invalid_argument(e.to_string()),
             Error::CreateChild {
@@ -628,6 +646,25 @@ impl Nexus {
         }
 
         Ok(())
+    }
+
+    /// set ANA state of the NVMe subsystem
+    pub async fn set_ana_state(
+        &self,
+        ana_state: NvmeAnaState,
+    ) -> Result<(), Error> {
+        if let Some(Protocol::Nvmf) = self.shared() {
+            if let Some(subsystem) = NvmfSubsystem::nqn_lookup(&self.name) {
+                subsystem.pause().await.unwrap();
+                let res = subsystem.set_ana_state(ana_state as u32).await;
+                subsystem.resume().await.unwrap();
+                return Ok(res?);
+            }
+        }
+
+        Err(Error::NotSharedNvmf {
+            name: self.name.clone(),
+        })
     }
 
     /// register the bdev with SPDK and set the callbacks for io channel
--- a/mayastor/src/bdev/nexus/nexus_bdev.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev.rs
@@ -648,6 +648,24 @@ impl Nexus {
         Ok(())
     }
 
+    /// get ANA state of the NVMe subsystem
+    pub async fn get_ana_state(&self) -> Result<NvmeAnaState, Error> {
+        if let Some(Protocol::Nvmf) = self.shared() {
+            if let Some(subsystem) = NvmfSubsystem::nqn_lookup(&self.name) {
+                let ana_state = subsystem.get_ana_state().await? as i32;
+                return NvmeAnaState::from_i32(ana_state).ok_or({
+                    Error::InvalidNvmeAnaState {
+                        ana_value: ana_state,
+                    }
+                });
+            }
+        }
+
+        Err(Error::NotSharedNvmf {
+            name: self.name.clone(),
+        })
+    }
+
     /// set ANA state of the NVMe subsystem
     pub async fn set_ana_state(
         &self,
@@ -655,9 +673,9 @@ impl Nexus {
     ) -> Result<(), Error> {
         if let Some(Protocol::Nvmf) = self.shared() {
             if let Some(subsystem) = NvmfSubsystem::nqn_lookup(&self.name) {
-                subsystem.pause().await.unwrap();
+                subsystem.pause().await?;
                 let res = subsystem.set_ana_state(ana_state as u32).await;
-                subsystem.resume().await.unwrap();
+                subsystem.resume().await?;
                 return Ok(res?);
             }
         }
--- a/mayastor/src/bin/mayastor-client/nexus_cli.rs
+++ b/mayastor/src/bin/mayastor-client/nexus_cli.rs
@@ -54,12 +54,23 @@ pub fn subcommands<'a, 'b>() -> App<'a, 'b> {
                 .help("uuid for the nexus"),
         );
 
-    let set_ana_state = SubCommand::with_name("set_ana_state")
-        .about("set the NVMe ANA state of the nexus")
-        .arg(Arg::with_name("uuid").required(true).index(1)
-            .help("uuid for the nexus"))
-        .arg(Arg::with_name("ana_state").required(true).index(2)
-            .help("NVMe ANA state (optimized,non_optimized,inaccessible) of the nexus"));
+    let ana_state = SubCommand::with_name("ana_state")
+        .about("get or set the NVMe ANA state of the nexus")
+        .arg(
+            Arg::with_name("uuid")
+                .required(true)
+                .index(1)
+                .help("uuid for the nexus"),
+        )
+        .arg(
+            Arg::with_name("state")
+                .required(false)
+                .index(2)
+                .possible_value("optimized")
+                .possible_value("non_optimized")
+                .possible_value("inaccessible")
+                .help("NVMe ANA state of the nexus"),
+        );
 
     let add = SubCommand::with_name("add")
         .about("add a child")
@@ -129,7 +140,7 @@ pub fn subcommands<'a, 'b>() -> App<'a, 'b> {
         .subcommand(add)
         .subcommand(remove)
         .subcommand(unpublish)
-        .subcommand(set_ana_state)
+        .subcommand(ana_state)
         .subcommand(list)
         .subcommand(children)
         .subcommand(nexus_child_cli::subcommands())
@@ -146,9 +157,7 @@ pub async fn handler(
         ("children", Some(args)) => nexus_children(ctx, &args).await,
         ("publish", Some(args)) => nexus_publish(ctx, &args).await,
         ("unpublish", Some(args)) => nexus_unpublish(ctx, &args).await,
-        ("set_ana_state", Some(args)) => {
-            nexus_set_nvme_ana_state(ctx, &args).await
-        }
+        ("ana_state", Some(args)) => nexus_nvme_ana_state(ctx, &args).await,
         ("add", Some(args)) => nexus_add(ctx, &args).await,
         ("remove", Some(args)) => nexus_remove(ctx, &args).await,
         ("child", Some(args)) => nexus_child_cli::handler(ctx, args).await,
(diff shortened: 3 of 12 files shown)


jonathan-teh:
bors try

mayastor/src/bin/mayastor-client/nexus_cli.rs
```diff
@@ -54,6 +54,13 @@ pub fn subcommands<'a, 'b>() -> App<'a, 'b> {
                 .help("uuid for the nexus"),
         );
 
+    let set_ana_state = SubCommand::with_name("set_ana_state")
```
Hoverbear, line 57:
```suggestion
    let set_ana_state = SubCommand::with_name("ana")
```

mayastor/src/bin/mayastor-client/nexus_cli.rs
```diff
@@ -54,6 +54,13 @@ pub fn subcommands<'a, 'b>() -> App<'a, 'b> {
...
+        .about("set the NVMe ANA state of the nexus")
+        .arg(Arg::with_name("uuid").required(true).index(1)
+            .help("uuid for the nexus"))
+        .arg(Arg::with_name("ana_state").required(true).index(2)
```
Hoverbear, line 61:
```suggestion
        .arg(Arg::with_name("state").required(true).index(2)
```

mayastor/src/bin/mayastor-client/nexus_cli.rs
```diff
@@ -54,6 +54,13 @@ pub fn subcommands<'a, 'b>() -> App<'a, 'b> {
...
+            .help("NVMe ANA state (optimized,non_optimized,inaccessible) of the nexus"));
```
Hoverbear, line 62:
```suggestion
            .help("NVMe ANA state (optimized, non_optimized, inaccessible) of the nexus"));
```

mayastor/src/bin/mayastor-client/nexus_cli.rs
```diff
@@ -321,6 +332,40 @@ async fn nexus_unpublish(
     Ok(())
 }
 
+async fn nexus_set_nvme_ana_state(
+    mut ctx: Context,
+    matches: &ArgMatches<'_>,
+) -> Result<(), Status> {
+    let uuid = matches.value_of("uuid").unwrap().to_string();
+    let ana_state = match matches.value_of("ana_state").unwrap() {
```
Hoverbear, line 340: Could we add a `impl FromStr for rpc::NvmeAnaState` somewhere in the codebase and use `.parse()` instead?

mayastor/tests/nexus_multipath.rs
```diff
@@ -138,6 +140,51 @@ async fn nexus_multipath() {
         );
     }
 
+    let output_list = Command::new("nvme").args(&["list"]).output().unwrap();
```
Hoverbear, line 143: `nvme-cli` is not in our shell.nix right now, could you add it so this test won't fail on systems without `nvme`?

Hoverbear (commented):
Can we also add the ANA state to the output colums of `nexus list`?

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -633,6 +651,25 @@ impl Nexus {
...
     }
 
+    /// set ANA state of the NVMe subsystem
+    pub async fn set_ana_state(
+        &self,
+        ana_state: NvmeAnaState,
+    ) -> Result<(), Error> {
+        if let Some(Protocol::Nvmf) = self.shared() {
+            if let Some(subsystem) = NvmfSubsystem::nqn_lookup(&self.name) {
+                subsystem.pause().await.unwrap();
```
gila, line 661: As guilty as others, but as we move along, we should really stop doing the unwrapping and handle the errors properly. If we can't handle the errors properly, then at least let us add .exect("huh?") (much better this way, but you get the point =))

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -633,6 +651,25 @@ impl Nexus {
         Ok(())
     }
 
+    /// set ANA state of the NVMe subsystem
+    pub async fn set_ana_state(
+        &self,
+        ana_state: NvmeAnaState,
+    ) -> Result<(), Error> {
```
gila, line 658: How about we change this prototype towards Option<NvmeAnaState>? 

1. if we change state from A to B we return A (previous state) 
2. if we change from state A to A (i.e. no change), we return None. 
3. in the case of iSCSI, we return None.

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -289,6 +289,33 @@ impl mayastor_server::Mayastor for MayastorSvc {
...
+    #[instrument(level = "debug", err)]
+    async fn set_nvme_ana_state(
+        &self,
+        request: Request<SetNvmeAnaStateRequest>,
+    ) -> GrpcResult<Null> {
+        let args = request.into_inner();
+        let uuid = args.uuid.clone();
+        debug!("Setting NVMe ANA state for nexus {} ...", uuid);
+
+        let ana_state = match NvmeAnaState::from_i32(args.ana_state) {
```
gila, line 301: This is not wrong or anything but, have you looked at the `matches!()` macro? IMHO it makes the code directly more obvious and comes by default with the system.

mayastor/src/subsys/nvmf/subsystem.rs
```diff
@@ -471,6 +472,35 @@ impl NvmfSubsystem {
         }
     }
 
+    /// set ANA state: optimized, non_optimized, inaccessible
+    /// subsystem must be in paused or inactive state
+    pub async fn set_ana_state(&self, ana_state: u32) -> Result<(), Error> {
```
gila, line 477: nice!

mayastor/tests/nexus_multipath.rs
```diff
@@ -138,6 +140,51 @@ async fn nexus_multipath() {
...
+    assert!(
+        output_list.status.success(),
+        "failed to list nvme devices, {}",
+        output_list.status
+    );
+    let sl = String::from_utf8(output_list.stdout).unwrap();
+    let nvmems: Vec<&str> = sl
```
gila, line 150: nvme-cli does support JSON output BTW... not saying it makes life easier in this case but perhaps for future references.

mayastor/src/subsys/nvmf/subsystem.rs
```diff
@@ -471,6 +472,35 @@ impl NvmfSubsystem {
...
+            nvmf_subsystem_set_ana_state(
+                self.0.as_ptr(),
+                trid_replica.as_ptr(),
+                ana_state,
+                Some(set_ana_state_cb),
+                cb_arg(s),
+            );
+        }
+
+        r.await.unwrap().to_result(|e| Error::Subsystem {
```
mtzaurus, line 497: Can we use '.expect()' instead of direct unwrapping ? That gives us some more visibility when error happens.
```
r.await.expect("<error message>")...
```

mayastor/src/subsys/nvmf/subsystem.rs
```diff
@@ -471,6 +472,35 @@ impl NvmfSubsystem {
...
+            source: Errno::from_i32(e),
```
mtzaurus, line 498: It looks like we should use -e here, as Errno::from_i32() expects a positive errno value when translating it into Errno, whereas nvmf_subsystem_set_ana_state() passes negative integers to the callback.
```
ctx = calloc(1, sizeof(*ctx));                                                                                |{
        if (!ctx) {                                                                                                   |        struct spdk_bdev_channel *channel = spdk_io_channel_get_ctx(ch);
                SPDK_ERRLOG("Unable to allocate context\n");                                                          |
                cb_fn(cb_arg, -ENOMEM);                                                                               |        *stat = channel->stat;
                return;                                                                                               |}
        }
```

test/grpc/test_nexus.js
```diff
@@ -685,6 +685,32 @@ describe('nexus', function () {
       });
     });
 
+    it('should change ANA state to inaccessible', (done) => {
```
mtzaurus, line 688: Is it possible to check that the new state has been actually set ?

test/grpc/test_nexus.js
```diff
@@ -685,6 +685,32 @@ describe('nexus', function () {
...
+      client.setNvmeAnaState(
+        {
+          uuid: UUID,
+          ana_state: enums.NVME_ANA_INACCESSIBLE_STATE
+        },
+        (err, res) => {
```
mtzaurus, line 694: IMHO, you can just use 'done' here directly and pass result/error to it directly:
```
client.setNvmeAnaState(
        {
          uuid: UUID,
          ana_state: enums.NVME_ANA_INACCESSIBLE_STATE
        },
        done
)
```

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -233,6 +236,8 @@ pub enum Error {
     },
     #[snafu(display("Invalid ShareProtocol value {}", sp_value))]
     InvalidShareProtocol { sp_value: i32 },
+    #[snafu(display("Invalid NvmeAnaState value {}", ana_value))]
+    InvalidNvmeAnaState { ana_value: i32 },
```
tiagolobocastro, line 240: this i32 could be the NvmeAnaState

rpc/proto/mayastor.proto
```diff
@@ -51,6 +51,9 @@ service Mayastor {
   rpc PublishNexus (PublishNexusRequest) returns (PublishNexusReply) {}
   rpc UnpublishNexus (UnpublishNexusRequest) returns (Null) {}
 
+  // NVMe ANA state
+  rpc SetNvmeAnaState (SetNvmeAnaStateRequest) returns (Null) {}
```
tiagolobocastro, line 55: what about adding a Getter as well?

mayastor/tests/nexus_multipath.rs
tiagolobocastro: hmm it is in the shell as well as nvmet-cli.

mayastor/tests/nexus_multipath.rs
tiagolobocastro: Could we use the `nvmeadm` library for this?

mayastor/src/grpc/mayastor_grpc.rs
tiagolobocastro: what about using the trait `TryFrom<i32>` ?

mayastor/tests/nexus_multipath.rs
jonathan-teh: It is indeed in `shell.nix` as part of `buildInputs`.

mayastor/tests/nexus_multipath.rs
jonathan-teh: I forgot that it does, the existing text-scraping in that test was for the `disconnect` command, which doesn't support output formats. The `nvmeadm` library doesn't (yet) support `ana_state` or possibly the concept of ANA. I'm leaning towards leaving the code here as-is as it's already written and leave that for future improvement.

mayastor/src/subsys/nvmf/subsystem.rs
jonathan-teh: Good point, I've even fixed a couple of instances of that in this file. I'll fix this instance but I think this affects other users in the codebase as well so I'll raise a ticket for that.

test/grpc/test_nexus.js
jonathan-teh: Thanks, fixed this and another similar existing instance nearby.

jonathan-teh:
> Can we also add the ANA state to the output columns of `nexus list`?

Most have asked for a getter and the current value could either be stored in the Nexus or retrieved from `spdk_nvme_ns_get_ana_state()`, with the latter probably a cleaner solution.

mayastor/src/bin/mayastor-client/nexus_cli.rs
jonathan-teh: I've opted for `ana_state` as `ana` on its own is a little vague.

jonathan-teh:
> ... retrieved from `spdk_nvme_ns_get_ana_state()`

Or not, as that's the nvme subsystem, not nvmf. The state is stored in `spdk_nvmf_subsystem_listener` so look for that instead.

mayastor/src/bdev/nexus/nexus_bdev.rs
Hoverbear: `.expect("huh?")` will impact our error messages. It's better to return the error for a backtrace if we can.

mayastor/src/grpc/mayastor_grpc.rs
Hoverbear: Indeed, `std::convert::TryFrom<i32> for NvmeAnaState` seems reasonable.

mayastor/src/bdev/nexus/nexus_bdev.rs
jonathan-teh: As there is now something that converts from `subsys::NvmfError` to this `Error` I've opted to return the error via `?`.

mayastor/src/bin/mayastor-client/nexus_cli.rs
jonathan-teh: I've opted to use `possible_value`s.

mayastor/src/grpc/mayastor_grpc.rs
jonathan-teh: AIUI the `TryFrom<i32>` has to be in the same crate as `::rpc::mayastor::NvmeAnaState`, otherwise I get "compiler says no". Anyway, this is exactly the same pattern used for `ShareProtocolNexus` several lines earlier so probably `matches!()` would be suitable.

mayastor/tests/nexus_multipath.rs
Hoverbear: Sorry, I looked in the mayastor pkg itself. :sweat:

mayastor/src/bin/mayastor-client/nexus_cli.rs
jonathan-teh: Done.

mayastor/src/bin/mayastor-client/nexus_cli.rs
jonathan-teh: Done, in the rpc crate.

mayastor/src/bdev/nexus/nexus_bdev.rs
jonathan-teh: This makes it impossible to return `Error`s, which it can now do from `pause`/`resume` as well so I've opted not to do this.

mayastor/src/subsys/nvmf/subsystem.rs
jonathan-teh: Done.

test/grpc/test_nexus.js
jonathan-teh: Done with the getter.

mayastor/src/bdev/nexus/nexus_bdev.rs
jonathan-teh: Not really as the user only has an `i32` plus if it's invalid we'd want to know what the value was.

rpc/proto/mayastor.proto
jonathan-teh: Done.

jonathan-teh:
> Can we also add the ANA state to the output columns of `nexus list`?

To answer this directly, yes we can but as there is another change to overhaul the CLI output formats it'd be better to leave this to another PR so I'll raise a ticket for that.

jonathan-teh:
bors try

jonathan-teh:
The CI failures are due to this in `dmesg`:
```
nvme nvme3: Please enable CONFIG_NVME_MULTIPATH for full support of multi-port devices.
```

doc/run.md
```diff
@@ -91,7 +91,40 @@ In order to use the full feature set of Mayastor, some or all of the following c
   ```bash
   modprobe nbd nvmet nvmet_rdma nvme_fabrics nvme_tcp nvme_rdma nvme_loop
   ```
-* An NVMe device. (Typically via PCI-E through an standard slot or [M.2][m-dot-2] port)
+* For Asymmetric Namespace Access (ANA) support (early preview), the following kernel build configuration enabled:
+  + `CONFIG_NVME_MULTIPATH`: enables support for multipath access to NVMe subsystems
+
+  This is usually already enabled in distributions kernels, at least for RHEL/CentOS 8.2, Ubuntu 20.04 LTS, and SUSE Linux Enterprise 15.2.
```
Hoverbear, line 97: To keep things short and match above formatting, maybe this line can be deleted, and the modprobe instructions provided in a consistent format to the above?

```bash
modprobe nvme-core multipath=1
```

etc

doc/run.md
```diff
@@ -273,4 +306,4 @@ production Mayastor deployment and operation instructions.
 [lxd]: https://linuxcontainers.org/
 [libvirtd]: https://libvirt.org/index.html
 [terraform-readme]: ./terraform/readme.adoc
-[aarch64-branch]: 
\ No newline at end of file
+[aarch64-branch]:
```
Hoverbear, line 309: Whoops I think this line is left in by mistake, could you trim it? :sweat:

doc/run.md
jonathan-teh: The kernel module option covers the case where `nvme-core.ko` is already loaded during boot, such as systems which use a NVMe drive. The multipath option cannot be modified at run-time.

doc/run.md
jonathan-teh: Done.

jonathan-teh:
bors try

mayastor/src/bin/mayastor.rs
```diff
@@ -42,12 +42,33 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                }
+            }
+        };
 
     info!("Starting Mayastor ..");
     info!(
         "kernel io_uring support: {}",
         if uring_supported { "yes" } else { "no" }
     );
+    info!("kernel nvme multipath support: {}", nvme_mp);
```
tiagolobocastro, line 71: :+1:

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -289,6 +289,52 @@ impl mayastor_server::Mayastor for MayastorSvc {
...
+            None => {
+                return Err(nexus_bdev::Error::InvalidNvmeAnaState {
+                    ana_value: args.ana_state as i32,
+                }
+                .into());
+            }
+        };
+
+        locally! { async move {
+            nexus_lookup(&args.uuid)?.set_ana_state(ana_state).await
```
tiagolobocastro, line 331: are all state transitions valid?

tiagolobocastro (approved):

mayastor/src/grpc/mayastor_grpc.rs
jonathan-teh: Yes, upstream SPDK doesn't restrict to/from any state and neither does the standard.

mayastor/src/bin/mayastor-client/nexus_cli.rs
```diff
@@ -363,6 +443,17 @@ async fn nexus_remove(
     Ok(())
 }
 
+fn ana_state_to_str(idx: i32) -> &'static str {
```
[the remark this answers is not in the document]
jonathan-teh: Done, squashed with the "get state" commit and rebased on to `develop`.

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -630,6 +648,44 @@ impl Nexus {
...
+    /// get ANA state of the NVMe subsystem
+    pub async fn get_ana_state(&self) -> Result<NvmeAnaState, Error> {
+        if let Some(Protocol::Nvmf) = self.shared() {
+            if let Some(subsystem) = NvmfSubsystem::nqn_lookup(&self.name) {
+                let ana_state = subsystem.get_ana_state().await? as i32;
+                return match NvmeAnaState::from_i32(ana_state) {
+                    Some(state) => Ok(state),
+                    None => Err(Error::InvalidNvmeAnaState {
+                        ana_value: ana_state,
+                    }),
```
cjones1024, line 660: I guess it's a matter of personal preference, but you can use Option::ok_or_else() here to convert an Option to a Result

rpc/proto/mayastor.proto
```diff
@@ -273,6 +277,28 @@ message UnpublishNexusRequest {
...
 }
 
+enum NvmeAnaState {
+  NVME_ANA_INVALID_STATE = 0;           // invalid, do not use
+  NVME_ANA_OPTIMIZED_STATE = 0x1;
+  NVME_ANA_NON_OPTIMIZED_STATE = 0x2;
+  NVME_ANA_INACCESSIBLE_STATE = 0x3;
+  NVME_ANA_PERSISTENT_LOSS_STATE = 0x4; // not yet supported
+  NVME_ANA_CHANGE_STATE = 0xF;          // not yet supported
+}
```
jkryl, line 287: I get what optimized and non-optimized states can be used for. What is the meaning and typical use case for inaccessible, persistent loss and change states?

mayastor/src/bin/mayastor.rs
jkryl, line 71: I'm not sure about this message, because it can be confusing. Storage node (mayastor) has nothing to do with multipath support. Storage node always supports multipathing because it is our code and spdk. What matters is if the multipathing is supported on the nodes where the application is running and that can be a mayastor node but also entirely different node when storage nodes do just the storage and don't run any applications. In this sense the message is confusing and should be rather removed from here and printed in our mayastor-csi plugin which is the right place for this and would be really useful there. Not sure what others think.

jkryl (commented):
Looks good. I'm interested in control plane implications. Should the control plane start setting nexus ANA to the value "optimized"? Because it is always optimized when we have just one nexus. Or is it optimized by default?

jonathan-teh:
> Looks good. I'm interested in control plane implications. Should the control plane start setting nexus ANA to the value "optimized"? Because it is always optimized when we have just one nexus. Or is it optimized by default?

It's optimised by default, in fact it's the only state that must be supported, the others optional.

mayastor/src/bin/mayastor.rs
jonathan-teh: I added it because it probably would have helped to debug the CI failures we had but I agree that it only helps if the application and mayastor are running on the same node. @tiago seemed to approve of it. If it helps I can add the word "initiator" after "nvme" to minimise confusion.

rpc/proto/mayastor.proto
jonathan-teh: - Inaccessible: Commands processed by a controller that reports this state for an ANA Group are not able to access user data of namespaces in that ANA Group. 
- Persistent loss:  Commands processed by a controller that reports this state for an ANA Group are persistently not able to access user data of namespaces in that ANA Group. The relationship between a controller and an ANA Group in this state shall not transition to any other ANA state.
- Change state: The change from one asymmetric namespace access state to another asymmetric namespace access state is called a transition. Transitions may occur in such a way that the ANA Change state is not visible to the
host (i.e., the ANA Change state may or may not be reported in the Asymmetric Namespace Access State field in the Asymmetric Namespace Access log).

cjones1024 (approved):

mayastor/src/bdev/nexus/nexus_bdev.rs
jonathan-teh: Thanks, `rust-lint` came back with "error: unnecessary closure used to substitute value for `Option::None`" and suggested `ok_or` instead so I've used that.

mayastor/src/bin/mayastor.rs
jonathan-teh: I've added "initiator" to the message. I think it can help if all nodes in the k8s cluster use the same kernel and kernel configuration. In any case, it's a separate commit so can be easily reverted in future, if necessary.

jonathan-teh:
bors merge

PR #773, opened 2021-03-05
chore: remove control plane from mayastor

tiagolobocastro:
Remove the new control plane from the mayastor repo.
The mbus-api is left with slight modifications so we don't pull in a
 bunch of crates which are not required. In the future we'll probably
 pull it from the control plane repo with feature gates for example.

Add a nix mayastor-unit derivation to allow the control plane to get
 access to specific parts of mayastor only.


tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors try

control-plane/mbus-api/Cargo.toml
```diff
@@ -7,9 +7,7 @@ edition = "2018"
 [dependencies]
```
Hoverbear, line 7: We can probably move this `mbus-api` crate to top level? It seems to be the only child of `control-plane/`.

nix/pkgs/mayastor/default.nix
```diff
@@ -22,114 +22,14 @@
...
-      liburing
-      numactl
-      openssl
-      utillinux
-    ];
-    verifyCargoDeps = false;
-    doCheck = false;
-    meta = { platforms = stdenv.lib.platforms.linux; };
-  };
+  project-builder = pkgs.callPackage ./cargo-package.nix { inherit version; };
```
Hoverbear, line 29: It would be perhaps easier to navigate if these were consistently named. 

You did some restructuring of these nix files to allow single binaries! I was looking at this too! I'm not sure this is the most idiomatic way, but I have no idea what that way is yet... I think this is a definite improvement though. :)

control-plane/mbus-api/Cargo.toml
tiagolobocastro: Yeah good point, there's no more control-plane.

nix/pkgs/mayastor/default.nix
tiagolobocastro: Yeah that can be done too, my main focus here was to allow building just the mayastor binary only as part of the "units" to speed up the control plane builds as we just want the mayastor binary and nothing else.

To get the single binaries on mayastor it's probably better to build the whole project and then have derivations that expose each binary otherwise you might be rebuilding the common imports more than once.

tiagolobocastro: Also, I'm not sure if we want to keep the mayastor/mayastor-dev as the collection of all binaries, just the one binary or a few select. If we have the answer to that I'm happy to add that change to this PR.
I also don't know what's the most idiomatic in nix :/

Hoverbear: Hmm let's leave it for a future PR? :)

paulyoong (approved):

control-plane/mbus-api/Cargo.toml
tiagolobocastro: Done.

cjones1024 (approved):
Looks reasonable.

tiagolobocastro:
bors merge

PR #774, opened 2021-03-05
ci: i/o soak test - ensure remote replicas

blaisedias:
Ensure i/o soak test runs with some volumes provisioned on a remote node.
Run the i/o soak test in a configuration where at least one
instance of mayastor is running on a node where no application pods
are running, so that some volumes can only be provisioned on a remote
node.

Do this by adding labels to "alternate" nodes, and node selectors to pods.
This relies on the fact that if enough pods and volumes are scheduled to
run simultaneously MOAC will schedule some replicas to run on nodes
where the test apps are not running.

This is deemed good enough, and has been observed to work
on a 3 node cluster with 30 fio pods and volumes.


test/e2e/io_soak/io_soak_test.go
```diff
@@ -185,8 +207,8 @@ var _ = Describe("Mayastor Volume IO test", func() {
 
 	It("should verify an NVMe-oF TCP volume can process IO on multiple volumes simultaneously", func() {
```
tiagolobocastro, line 208: This was already here but, what does this mean, "a volume can process IO on multiple volumes"?

scripts/e2e-test.sh
```diff
@@ -16,8 +16,8 @@ REPORTSDIR=$(realpath "$SCRIPTDIR/..")
 #TESTS="install basic_volume_io csi replica rebuild node_disconnect/replica_pod_remove uninstall"
 ALL_TESTS="install basic_volume_io csi resource_check replica rebuild uninstall"
```
chriswldenyer, line 17: should add io_soak to ALL_TESTS. Or maybe rename ALL_TESTS to DEFAULT_TESTS

chriswldenyer: ...or both

test/e2e/io_soak/io_soak_test.go
blaisedias: Doh. It should be "NVMe-oF TCP can process IO on multiple volumes simultaneously"

scripts/e2e-test.sh
blaisedias: I vacillating on that, it would mean that if you do not specify any tests (most developers would probably do that I think) then you end up with a 30 minute + test run, unless one modifies the config
Should soak tests be part of the default?

test/e2e/io_soak/io_soak_test.go
blaisedias: @chriswldenyer  would changing this string impact the xray collection?

blaisedias: Changed to 
`should verify can process IO on multiple volumes simultaneously using NVMe-oF TCP`

tiagolobocastro (approved):

GlennBullingham (approved):

chriswldenyer (approved):
LGTM

blaisedias:
bors merge

PR #775, opened 2021-03-05
chore: allow Nexus::new to create and register children (WIP)

Hoverbear:
Moves the `bdev::nexus::nexus_bdev_children::Nexus::create_and_register` which
`bdev::nexus::nexus_bdev::nexus_create` invokes during it's execution into
`bdev::nexus::nexus_bdev::Nexus::new`.

Previously it would only call `bdev::nexus::nexus_bdev_children::Nexus::register_children`, not
`bdev::nexus::nexus_bdev_children::Nexus::create_and_register`. This created potential for misuse,
as users could pass a child which was not created, and it would be registered invalidly. :boom: 

This has progress towards https://mayadata.atlassian.net/browse/CAS-771 and possibly https://mayadata.atlassian.net/browse/CAS-677 .

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_bdev.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev.rs
@@ -392,12 +392,12 @@ impl Drop for Nexus {
 impl Nexus {
     /// create a new nexus instance with optionally directly attaching
     /// children to it.
-    pub fn new(
+    pub async fn new(
         name: &str,
         size: u64,
         uuid: Option<&str>,
         child_bdevs: Option<&[String]>,
-    ) -> Box<Self> {
+    ) -> Result<Box<Self>, crate::bdev::nexus::nexus_bdev::Error> {
         let mut b = Box::new(spdk_bdev::default());
 
         b.name = c_str!(name);
@@ -410,7 +410,7 @@ impl Nexus {
 
         let cfg = Config::get();
 
-        let mut n = Box::new(Nexus {
+        let mut nexus = Box::new(Nexus {
             name: name.to_string(),
             child_count: 0,
             children: Vec::new(),
@@ -424,17 +424,30 @@ impl Nexus {
             max_io_attempts: cfg.err_store_opts.max_io_attempts,
         });
 
-        n.bdev.set_uuid(uuid.map(String::from));
+        nexus.bdev.set_uuid(uuid.map(String::from));
 
         if let Some(child_bdevs) = child_bdevs {
-            n.register_children(child_bdevs);
+            for child in child_bdevs {
+                if let Err(error) = nexus.create_and_register(child).await {
+                    error!(
+                        "failed to create nexus {}: failed to create child {}: {}",
+                        name, child, error
+                    );
+                    nexus.close_children().await;
+                    return Err(Error::CreateChild {
+                        source: error,
+                        name: String::from(name),
+                    });
+                }
+            }
         }
 
         // store a reference to the Self in the bdev structure.
         unsafe {
-            (*n.bdev.as_ptr()).ctxt = n.as_ref() as *const _ as *mut c_void;
+            (*nexus.bdev.as_ptr()).ctxt =
+                nexus.as_ref() as *const _ as *mut c_void;
         }
-        n
+        Ok(nexus)
     }
 
     /// set the state of the nexus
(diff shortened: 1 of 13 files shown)


Hoverbear:
bors try

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -1058,61 +1071,40 @@ pub async fn nexus_create(
     // closing a child assumes that the nexus to which it belongs will appear
     // in the global list of nexus instances. We must also ensure that the
     // nexus instance gets removed from the global list if an error occurs.
```
cjones1024, line 1073: This comment still applies.

We can now potentially call Nexus::close_children() from within Nexus::new().
NexusChild::close() calls NexusChild::destroy() which raises a SPDK_BDEV_EVENT_REMOVE event.
This ultimately results in callback core::bdev::event_cb being called, which then calls lookup_child_from_bdev().
This last function only works correctly if the owning nexus is present in the global list of nexus instances.
However, it is now currently not present (it's only added at the end), and the bdev will not be properly closed/destroyed.

We must either change the way lookup_child_from_bdev() works to avoid this issue, or otherwise ensure that the current nexus is present in the global list by the time Nexus::close_children() is called, otherwise mayastor will hang.

cjones1024: Just tested on this branch by running
```
/mayastor-client nexus create $id 4Mib 'malloc:///m0?size_mb=4'
``` 
and the client hangs,

The last thing you see in the log is the following:
```
[2021-03-08T10:26:46.504251499+00:00  WARN mayastor::bdev::nexus::nexus_bdev:nexus_bdev.rs:514] error updating child labels: Device is too small to accomodate Metadata partition: blocks=8192
[2021-03-08T10:26:46.505886123+00:00 ERROR mayastor::bdev::nexus::nexus_bdev:nexus_bdev.rs:1092] failed to open nexus nexus-82520992-f822-47d1-93bc-fd5345334d9d: Failed to read child label of nexus nexus-82520992-f822-47d1-93bc-fd5345334d9d: Label is invalid: Incorrect MBR signature
[2021-03-08T10:26:46.506056576+00:00  INFO mayastor::bdev::nexus::nexus_child:nexus_child.rs:361] Closing child malloc:///m0?size_mb=4
[2021-03-08T10:26:46.506211046+00:00  INFO mayastor::nexus_uri:bdev_destroy{uri="malloc:///m0?size_mb=4"}:nexus_uri.rs:79] new
[2021-03-08T10:26:46.506320282+00:00  INFO mayastor::nexus_uri:bdev_destroy{uri="malloc:///m0?size_mb=4"}:nexus_uri.rs:79] enter
[2021-03-08T10:26:46.506628106+00:00  INFO mayastor::nexus_uri:bdev_destroy{uri="malloc:///m0?size_mb=4"}:nexus_uri.rs:79] exit
[2021-03-08T10:26:46.506755874+00:00  INFO mayastor::core::bdev:bdev.rs:168] Received remove event for bdev m0
```
Basically the remove event is not being handled correctly and spdk is stuck in the destroy code for the bdev.

Hoverbear: Hmm so it seems `Nexus::create` could return an `Rc<Nexus>` instead of a `Box<Nexus>` since it needs to be present in the global state to complete child registration? Otherwise I guess it would have to add then remove itself from the list? :thinking: 

Should `Nexus::new()` populate the global list? If not, then this PR should be closed I guess.

Hoverbear: Oh, I noticed this problem even occurs on `develop`... Hm...

cjones1024: Actually the more I think about this, the more I'm inclined to take the view that the best solution at present is to remove the "child_bdevs" argument from Nexus::new() and dis-allow adding children like this altogether. I would also go as far as saying that Nexus::new() probably should not be public either. What do others think?

mayastor/src/bdev/nexus/nexus_config.rs
```diff
@@ -103,7 +103,12 @@ pub(crate) fn parse_ini_config_file() -> i32 {
             name, lu_size, block_size, &child_bdevs
         );
 
-        nexus_instance_new(name, lu_size, child_bdevs);
+        futures::executor::block_on(nexus_instance_new(
```
tiagolobocastro, line 106: Is this "ini" code still relevant since we have the config file? Otherwise might as well remove this altogether

mayastor/src/bdev/nexus/nexus_bdev.rs
Hoverbear: So how is it declarative if you have to build a nexus by iterating over the children?

How would `Nexus::new` be called then? It seems like this is intended to be public API? It's certainly more idiomatic than `nexus_create`, which has no business being in public API.

cjones1024: As far as I can tell, a proper and correct implementation of a public new() would need to do more or less what nexus_create() does. So perhaps we should retain the existing new() as a (private) utility function appropriately renamed, make the existing nexus_create() our new(),  and nexus_create() becomes a simple wrapper around new(). Also, this would make nexus_instance_new() superflous as the nexus needs to be added to the list of global instances as part of new()

PR #778, opened 2021-03-09
test(e2e): add markers to loki log stream

chriswldenyer:
Send log entries directly to Grafana/Loki from golang tests.
Change Loki credentials to username + password as this form is 
needed for the go library.
Currently just one message is generated from a go test to
verify the mechanism, with the expectation that more markers will
be added as and when needed.


chriswldenyer:
bors try

PR #779, opened 2021-03-09
test(e2e): add markers to loki log stream

chriswldenyer:
Send log entries directly to Grafana/Loki from golang tests, with the app string "marker"
Change Loki credentials to username + password as this form is needed for the go library.
Build number now passed into e2e-test.sh to propagate the build number to the markers.
Currently called when each ginkgo test starts.
Fix typo in soak test and temporarily disable from overnight (extended) runs.

### DIFF
--- a/scripts/e2e-test.sh
+++ b/scripts/e2e-test.sh
@@ -18,7 +18,7 @@ TOPDIR=$(realpath "$SCRIPTDIR/..")
 #TESTS="install basic_volume_io csi replica rebuild node_disconnect/replica_pod_remove uninstall"
 DEFAULT_TESTS="install basic_volume_io csi resource_check replica rebuild uninstall"
 ONDEMAND_TESTS="install basic_volume_io csi resource_check uninstall"
-EXTENDED_TESTS="install basic_volume_io csi resource_check io_soak uninstall"
+EXTENDED_TESTS="install basic_volume_io csi resource_check uninstall"
 CONTINUOUS_TESTS="install basic_volume_io csi resource_check replica rebuild io_soak uninstall"
 
 #exit values
@@ -31,6 +31,7 @@ EXITV_FAILED_CLUSTER_OK=255
 
 # Global state variables
 #  test configuration state variables
+build_number=
 device=
 registry=
 tag="ci"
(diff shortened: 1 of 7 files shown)


chriswldenyer:
bors try

chriswldenyer:
bors try

chriswldenyer:
bors try

scripts/e2e-test.sh
```diff
@@ -18,7 +18,7 @@ TOPDIR=$(realpath "$SCRIPTDIR/..")
 #TESTS="install basic_volume_io csi replica rebuild node_disconnect/replica_pod_remove uninstall"
 DEFAULT_TESTS="install basic_volume_io csi resource_check replica rebuild uninstall"
 ONDEMAND_TESTS="install basic_volume_io csi resource_check uninstall"
-EXTENDED_TESTS="install basic_volume_io csi resource_check io_soak uninstall"
+EXTENDED_TESTS="install basic_volume_io csi resource_check uninstall"
```
blaisedias, line 21: Is removing io_soak from extended tests intentional?

test/e2e/common/loki/loki.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+var g_apiUser string
+var g_apiPw string
+var g_buildNumber string
+var g_once sync.Once
+
+func SendLokiMarker(text string) error {
+	g_once.Do(func() {
+		g_apiUser = os.Getenv("grafana_api_user")
```
blaisedias, line 24: I'd say move retrieval of the image tag inside the Once function, then make the decision whether adding markers is possible in that function and store as boolean.
Next time round just check the boolean
Also report errors from within the Once function? 
Not sure about returning an error, caller should  either log the error message or ignore but proceed.
IMO not much point in seeing the same error message multiple times.

chriswldenyer: > I'd say move retrieval of the image tag inside the Once function
But then the image tag would have to be stored as another static variable, mirroring the one inside the config struct which is itself cached. Not sure if caching the same value twice makes sense.
The failure is going to be a configuration issue which will happen rarely. I'm not sure if saving a few extra entries in the log is worth the extra complexity of caching the failure.

chriswldenyer: OTOH we could end up with a lot of errors when the loki marker function is called a lot. 
Also could save some processing when run outside of CI with loki not enabled.
I'll make the change for the code to decide once whether loki is enabled.

chriswldenyer:
bors try

arne-rusek (approved):
I see no evil.

chriswldenyer:
bors merge

PR #780, opened 2021-03-09
refactor: rationalise e2e test structure further

blaisedias:
All tests register fail handlers and get reporters.
At the moment each test does these individually.
Replace the individual calls with a single call,
to a helper function which does both.
An added advantage is that this then is the best,
place to execute common actions, for example emit log markers.


chriswldenyer (approved):
LGTM

blaisedias:
bors merge

PR #782, opened 2021-03-10
ci(moac): fix timing issues in moac watcher tests

jkryl:
It's more a workaround for behaviour that normally should not happen.
On the other hand it does not really matter because it affects only
the test suite.

Resolves: CAS-760

### DIFF
--- a/csi/moac/test/watcher_test.js
+++ b/csi/moac/test/watcher_test.js
@@ -13,6 +13,10 @@ const RESTART_DELAY_MS = 300;
 const EVENT_TIMEOUT_MS = 200;
 const EVENT_DELAY_MS = 100;
 const EYE_BLINK_MS = 30;
+// Believe it or not but it is possible that timeout callback triggers a bit
+// earlier than it should (although that nodejs documentation says that it is
+// not possible). Accomodate this weird behaviour.
+const TOLERATE_MS = 2;
 
 const fakeConfig = {
   clusters: [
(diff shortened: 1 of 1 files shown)


csi/moac/test/watcher_test.js
```diff
@@ -219,7 +223,7 @@ module.exports = function () {
       timeout = setTimeout(() => watcher.emitKubeEvent('add', apple), EVENT_DELAY_MS);
       await watcher.create(apple);
       const delta = new Date() - startTime;
-      expect(delta).to.be.within(EVENT_DELAY_MS - 2, EVENT_DELAY_MS + EYE_BLINK_MS);
+      expect(delta).to.be.within(EVENT_DELAY_MS - TOLERATE_MS, EVENT_DELAY_MS + EYE_BLINK_MS);
```
paulyoong, line 226: Nit: could we assign this calculation to a "delay" constant that way we only need to change it in one place?

csi/moac/test/watcher_test.js
```diff
@@ -230,7 +234,7 @@ module.exports = function () {
       const startTime = new Date();
       await watcher.create(apple);
       const delta = new Date() - startTime;
-      expect(delta).to.be.within(EVENT_TIMEOUT_MS, EVENT_TIMEOUT_MS + EYE_BLINK_MS);
+      expect(delta).to.be.within(EVENT_TIMEOUT_MS - TOLERATE_MS, EVENT_TIMEOUT_MS + EYE_BLINK_MS);
```
paulyoong, line 237: Nit: could we assign this calculation to a "timeout" constant that way we only need to change it in one place?

paulyoong (approved):

csi/moac/test/watcher_test.js
jkryl: thanks for review, Paul. EVENT_DELAY_MS is used elsewhere in the code so I cannot change its value. I could introduce a new constant and use it instead of EVENT_DELAY_MS, but since I already "fixed" all places I think it's too late :-)

tiagolobocastro (approved):

jkryl:
bors merge

PR #784, opened 2021-03-11
chore: refine sync/send impls

Hoverbear:
Went on a little fishing expedition and found some unnecessarily unsafe
implementations which we don't need.

This doesn't have any impact on functionality or users, just safety 
semantics for later.

Also documents some that are safe.

I could use some help ensuring these are correct. :)

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_module.rs
+++ b/mayastor/src/bdev/nexus/nexus_module.rs
@@ -31,13 +31,26 @@ pub struct NexusInstances {
     inner: UnsafeCell<Vec<Box<Nexus>>>,
 }
 
+/// A wrapper around an SPDK Nexus bdev.
+///
+/// See https://spdk.io/doc/structspdk__bdev__module.html for more information.
+///
+/// # Safety
+///
+/// SPDK manages this and it is equivalent to Send/Sync. This means this type is Send/Sync.
 #[derive(Debug)]
 pub struct NexusModule(*mut spdk_bdev_module);
 
+// Safety: Pointer to SPDK managed structure.
 unsafe impl Sync for NexusModule {}
+
+// TODO: This is not safe.
 unsafe impl Sync for NexusInstances {}
 
+// Safety: Pointer to SPDK managed structure.
 unsafe impl Send for NexusModule {}
+
+// TODO: This is not safe.
 unsafe impl Send for NexusInstances {}
 
 impl Default for NexusModule {
--- a/mayastor/src/core/thread.rs
+++ b/mayastor/src/core/thread.rs
@@ -23,12 +23,25 @@ pub enum Error {
     InvalidThread {},
 }
 
-#[derive(Debug, PartialEq, Clone, Copy)]
-/// struct that wraps an SPDK thread. The name thread is chosen poorly and
-/// should not be confused with an actual thread. Consider it more to be
-/// analogous to a container to which you can submit work and poll it to drive
+/// A wrapper around an SPDK thread.
+///
+/// The name thread is chosen poorly and should not be confused with an actual thread. Consider it
+/// more to be analogous to a container to which you can submit work and poll it to drive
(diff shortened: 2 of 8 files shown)


mayastor/src/core/thread.rs
```diff
@@ -23,13 +23,26 @@ pub enum Error {
...
+///
+/// See: https://github.com/spdk/spdk/blob/fb27c710f2c5ee4e666d2b366237e04c08dda977/include/spdk_internal/thread.h#L95-L136
+#[derive(Debug, PartialEq, Clone, Copy)]
 pub struct Mthread(NonNull<spdk_thread>);
 
+// Safety: SPDK considers the inner `spdk_thread` threadsafe.
+unsafe impl Send for Mthread {}
+
+// Safety: SPDK considers the inner `spdk_thread` threadsafe.
+unsafe impl Sync for Mthread {}
```
gila, line 44: I'm not about the sync part. Although in the future, the ability to move threads from/to cores is desired, nothing AFAIK, will prevent two cores/threads call spdk_thread_poll() on each. Having that said, I don't know if that is a problem right now?
As this code works without the sync marker, can, to be sure, omit it?

mayastor/src/subsys/nvmf/subsystem.rs
```diff
@@ -64,7 +64,21 @@ impl Display for SubType {
...
+/// SPDK considers this structure threadsafe. This means this structure can be Send/Sync.
+///
+/// See more: https://github.com/spdk/spdk/blob/312a9d603dfc7dfb7061f33fa3f1fec766579dc7/lib/nvmf/nvmf_internal.h#L264-L305
 pub struct NvmfSubsystem(pub(crate) NonNull<spdk_nvmf_subsystem>);
+
+// Safety: SPDK considers the inner `spdk_nvmf_subsystem` threadsafe.
+unsafe impl Send for NvmfSubsystem {}
+
+// Safety: SPDK considers the inner `spdk_nvmf_subsystem` threadsafe.
+unsafe impl Sync for NvmfSubsystem {}
```
gila, line 80: Again, don't think the subsystem is sync. Two threads can change the subsystem at any given time.

Hoverbear: I see at least partial consideration being given to threading in things like state changes. 

https://github.com/openebs/spdk/blob/312a9d603dfc7dfb7061f33fa3f1fec766579dc7/lib/nvmf/subsystem.c#L448  

https://github.com/openebs/spdk/blob/312a9d603dfc7dfb7061f33fa3f1fec766579dc7/lib/nvmf/subsystem.c#L728

So it does seem like this type is sync?

mayastor/src/core/thread.rs
Hoverbear: It seems like they are sync though? I see what looks like lockless rings in https://github.com/openebs/spdk/blob/050ecb8c3a277e3a2dc8ccb1e370fccde642bf11/lib/thread/thread.c#L890-L944 and you can see in https://github.com/openebs/spdk/blob/050ecb8c3a277e3a2dc8ccb1e370fccde642bf11/lib/thread/thread.c#L951 there is certainly atomics happening.

mayastor/src/bdev/nexus/nexus_module.rs
```diff
@@ -31,13 +31,26 @@ pub struct NexusInstances {
...
 unsafe impl Sync for NexusModule {}
+
+// Safety: Pointer to SPDK managed structure.
 unsafe impl Sync for NexusInstances {}
 
+// Safety: Pointer to SPDK managed structure.
 unsafe impl Send for NexusModule {}
+
+// Safety: Pointer to SPDK managed structure.
 unsafe impl Send for NexusInstances {}
```
Hoverbear, line 54: Wait wait this isn't safe.

PR #787, opened 2021-03-12
test(e2e): add io_soak to extended tests

chriswldenyer:
Run soak tests, as well as replica and
rebuild tests on overnight builds.
Fixed ginkgo test strings to match Xray test definition
for io_soak test.
Removed iscsi from the config list of soak protocols.  
Reduce load factor in soak tests so they
can be run on Hetzner clusters, until we
have larger nodes available or until the new 
e2e-fio image is in use.

### DIFF
--- a/scripts/e2e-test.sh
+++ b/scripts/e2e-test.sh
@@ -18,7 +18,7 @@ TOPDIR=$(realpath "$SCRIPTDIR/..")
 #TESTS="install basic_volume_io csi replica rebuild node_disconnect/replica_pod_remove uninstall"
 DEFAULT_TESTS="install basic_volume_io csi resource_check replica rebuild uninstall"
 ONDEMAND_TESTS="install basic_volume_io csi resource_check uninstall"
-EXTENDED_TESTS="install basic_volume_io csi resource_check uninstall"
+EXTENDED_TESTS="install basic_volume_io csi resource_check replica rebuild io_soak uninstall"
 CONTINUOUS_TESTS="install basic_volume_io csi resource_check replica rebuild io_soak uninstall"
 
 #exit values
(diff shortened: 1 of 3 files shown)


chriswldenyer:
bors try

test/e2e/configurations/ci_e2e_config_hetzner.yaml
```diff
@@ -0,0 +1,34 @@
...
+  replicas: 1
+  cdCycles: 100
+  crudCycles: 10
+ioSoakTest:
+  replicas: 1
+  duration: 10m
+  loadFactor: 2
+  protocols:
+  - nvmf
+  - iscsi
```
blaisedias, line 12: This was an oversight on my path in the original configuration, for the moment we should only check nvmf so remove iscsi ?

chriswldenyer:
bors try

test/e2e/configurations/ci_e2e_config_hetzner.yaml
chriswldenyer: will do

chriswldenyer:
bors try

blaisedias (commented):
Changes look fine to me, but there isn't a need for a new configuration file.
The existing one is named `ci_e2e_config.yaml` and is intended for use with the CI cluster, and should be modified to suit the CI cluster configuration.
When we have a new cluster configuration for nightly or soak tests we should create a new one and use that for those runs.
The intention is to have a config file for each of the cluster types/configurations we use.

chriswldenyer:
> Changes look fine to me, but there isn't a need for a new configuration file.
> The existing one is named `ci_e2e_config.yaml` and is intended for use with the CI cluster, and should be modified to suit the CI cluster configuration.
> When we have a new cluster configuration for nightly or soak tests we should create a new one and use that for those runs.
> The intention is to have a config file for each of the cluster types/configurations we use.

That was my intention as well, with the new file aimed at the particular Hetzner configuration we have now, and the old one for when we have bigger servers. If you think I should edit the existing file, I'll do that.

blaisedias (approved):
Cool.

GlennBullingham (approved):

chriswldenyer:
bors merge

PR #790, opened 2021-03-15
feat(nexus): make ANA reporting configurable

jonathan-teh:
Leave ANA reporting disabled when publishing a Nexus over NVMf unless
the environment variable NEXUS_NVMF_ANA_ENABLE is defined when
creating a Nexus.

Include a chore to switch away from deprecated spdk_nvmf functions.

Fixes CAS-789, CAS-791.

### DIFF
--- a/mayastor/src/subsys/nvmf/subsystem.rs
+++ b/mayastor/src/subsys/nvmf/subsystem.rs
@@ -247,6 +247,16 @@ impl NvmfSubsystem {
 
     /// enable Asymmetric Namespace Access (ANA) reporting
     pub fn set_ana_reporting(&self, enable: bool) -> Result<(), Error> {
+        match std::env::var("NEXUS_NVMF_ANA_ENABLE") {
+            Ok(s) => {
+                if s != "1" {
+                    return Ok(());
+                }
+            }
+            Err(_) => {
+                return Ok(());
+            }
+        }
         unsafe {
             spdk_nvmf_subsystem_set_ana_reporting(self.0.as_ptr(), enable)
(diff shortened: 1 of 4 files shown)


jonathan-teh:
bors try

gila (approved):

mayastor/src/subsys/nvmf/subsystem.rs
```diff
@@ -247,6 +248,9 @@ impl NvmfSubsystem {
 
     /// enable Asymmetric Namespace Access (ANA) reporting
     pub fn set_ana_reporting(&self, enable: bool) -> Result<(), Error> {
+        if std::env::var("NEXUS_NVMF_ANA_ENABLE").is_err() {
```
mtzaurus, line 251: Looks like this variable doesn't take into account the actual value of the variable, just the presence of the variable itself. Is that by design ?
I mean, whether its 'NEXUS_NVMF_ANA_ENABLE=1' or 'NEXUS_NVMF_ANA_ENABLE=0' - doesn't matter.
Shall we also enable/disable based on the actual value of the variable (0 - disable, 1 - enable) or so ?

jonathan-teh: That was somewhat deliberate, it's only meant to be temporary (the Linux kernel NVMe target, for instance, has ANA reporting always enabled, AFAICT). I've changed it to require the env variable to be defined as "1".

Hoverbear (approved):

jonathan-teh:
bors merge

PR #792, opened 2021-03-16
fix(moac): moac assert when publishing a volume

jkryl:
This happened when putting moac under a bit of stress and issuing
create/publish/unpublish/destroy requests in a quick succession.
It revealed a few race conditions. Basically, publish method was
running in "parallel" (interleaved) with FSA code, so the state of
the volume was changing in two places. That has an easy fix of
putting code that changes the state of the volume from publish
method to FSA itself. fsa() guarding method guarantees that we can
have only one _fsa() running at any given time. So that should fix
the race condition. What publish merely does is set a few volume
state variables that tell FSA what to do and kick off FSA that does
the rest. Notification mechanism for letting the publish method know
when the required changes are done is in place but had to be slightly
revamped to be useable for that.

The same change has been done for unpublish and destroy methods (the
same problem although that it wasn't reproduced so far).

Some of the volume object tests have been moved to volumes test
suite because they require to run the FSA now.

Rename children URIs used in the tests so that they reflect the
node where child is located (as it is in real world).

Use events.EventEmitter with type information where possible.

Resolves: CAS-728

### DIFF
--- a/csi/moac/volume.ts
+++ b/csi/moac/volume.ts
@@ -2,6 +2,7 @@
 // replicas and implements algorithms for volume recovery.
 
 import assert from 'assert';
+import events = require('events');
 import * as _ from 'lodash';
 import { Replica } from './replica';
 import { Child, Nexus, Protocol } from './nexus';
@@ -60,7 +61,8 @@ export class Volume {
   public state: VolumeState;
   private publishedOn: string | undefined;
   // internal properties
-  private emitEvent: (type: string) => void;
+  private emitter: events.EventEmitter;
+  private changeCallbacks: (() => void)[];
   private registry: any;
   private runFsa: number; // number of requests to run FSA
   private nodeBlackList: Record<string, boolean>; // replicas on these nodes should be avoided
@@ -83,7 +85,7 @@ export class Volume {
   constructor(
     uuid: string,
     registry: any,
-    emitEvent: (type: string) => void,
+    emitter: events.EventEmitter,
     spec: any,
     state?: VolumeState,
     size?: number,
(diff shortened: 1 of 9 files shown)


csi/moac/volume.ts
```diff
@@ -143,57 +141,78 @@ export class Volume {
...
-    }
-    let uri = nexus.getUri();
-    if (!uri) {
-      uri = await nexus.publish(protocol);
-    } else {
-      log.debug(`Publishing volume ${this} that has been already published`)
+
+    // FSA will take care of publishing the nexus
+    this.publishedOn = this._desiredNexusNode(this._activeReplicas()).name;
+    this.fsa();
```
tiagolobocastro, line 147: How soon will the real fsa run after calling this fn?

csi/moac/volume.ts
```diff
@@ -236,22 +267,32 @@ export class Volume {
       }
     }
 
-    // check that replicas are shared in the way they should be
-    let nexusNode, replicaSet;
-    try {
-      [nexusNode, replicaSet] = await this._ensureReplicaShareProtocols();
-    } catch (err) {
-      log.warn(err.toString());
+    // Unpublish the nexus if it should not be published
```
tiagolobocastro, line 270: Are we missing the `publish` counterpart?

csi/moac/volume.ts
jkryl: just an eye blink - in next reactor loop iteration of nodejs.

csi/moac/volume.ts
jkryl: it is at line 362 in the new file. It is done later because we need to check a couple of things before we publish the volume.

tiagolobocastro: dohh see it now, thanks

tiagolobocastro (approved):

paulyoong (approved):

jkryl:
bors merge

PR #794, opened 2021-03-16
fix(nexus): destroy child bdev when faulted

jonathan-teh:
When a nexus child is faulted, set its state to Destroying to ensure
that its bdev is indeed destroyed when calling bdev_destroy().

Fixes CAS-793

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_io.rs
+++ b/mayastor/src/bdev/nexus/nexus_io.rs
@@ -27,7 +27,6 @@ use crate::{
         Reason,
     },
     core::{Bdev, Cores, GenericStatusCode, Mthread, NvmeStatus, Reactors},
-    nexus_uri::bdev_destroy,
 };
 
 /// NioCtx provides context on a per IO basis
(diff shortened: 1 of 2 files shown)


jonathan-teh:
bors try

jonathan-teh:
@gila suggested calling `child.destroy()` instead, like:
```
-                    child.set_state(ChildState::Destroying);
-                    if let Err(err) = bdev_destroy(&uri).await {
-                        error!("{} destroying bdev {}", err, uri)
+                    if let Err(err) = child.destroy().await {
+                        error!("{} destroying child {}", err, child);
                     }
```
which work just as well and is more maintainable. We couldn't find a reason to call `bdev_destroy()`, I'd only kept it to minimise the changes.

mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -333,18 +332,16 @@ impl Bio {
...
-                    //nexus.remove_child(&uri).await.unwrap();
 
                     // Note, an error can occur here if a separate task,
                     // e.g. grpc request is also deleting the child,
                     // in which case the bdev may no longer exist at
                     // this point. To be addressed by CAS-632 to
                     // improve synchronization.
-                    if let Err(err) = bdev_destroy(&uri).await {
-                        error!("{} destroying bdev {}", err, uri)
+                    if let Err(err) = child.destroy().await {
```
tiagolobocastro, line 343: are there any other places where `child.set_state(ChildState::Destroying);` and `bdev_destroy` was used that could be replaced with this simple call?

tiagolobocastro (approved):

mayastor/src/bdev/nexus/nexus_io.rs
jonathan-teh: There are 2 calls to `bdev_destroy()` in `Nexus::add_child_only()`, the second of which could be replaced as it's after the `NexusChild` has been created but it's not clear how to reproduce that in a test.

Hoverbear (approved):

jonathan-teh:
bors merge

PR #795, opened 2021-03-16
fix(nexus): handle partial IO submission

gila:


mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -1,471 +1,478 @@
...
-        if let Some(nexus) = nexus_lookup(&nexus) {
-            if let Some(child) = nexus.child_lookup(&child.name()) {
-                let current_state = child.state.compare_and_swap(
-                    ChildState::Open,
-                    ChildState::Faulted(Reason::IoError),
+        // determine what to do now based on the context specific info of the IO
+        match self.disposition() {
+            // the happy path, all is good
+            Disposition::Complete(IoStatus::Success) => self.ok(),
+            // All of IO's have failed but all remaining in lights completed now
```
jonathan-teh, line 218: Typo: lights -> flights

mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -1,471 +1,478 @@
...
+            // as well depending on the error we can attempt to do a
+            // retry.
+            Disposition::Complete(IoStatus::Failed) => self.fail(),
+
+            // IOs where submitted before we bumped into an error. The IO has
```
jonathan-teh, line 223: Typo: where -> were

mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -1,471 +1,478 @@
...
+                            Mthread::current(),
+                            child,
+                        );
+
+                        nexus.pause().await.unwrap();
+                        nexus.reconfigure(DrEvent::ChildFault).await;
+                        // TODO: an error can occur here if a separate task,
+                        // e.g. grpc request is also deleting the child.
+                        if let Err(err) =
+                            bdev_destroy(&child.name.clone()).await
```
jonathan-teh, line 458: Would need reapply the fix for #794, depending on which is merged first.

mayastor/src/core/bio.rs
```diff
@@ -0,0 +1,312 @@
...
+}
+
+impl From<*mut spdk_bdev_io> for Bio {
+    fn from(io: *mut spdk_bdev_io) -> Self {
+        Bio(NonNull::new(io as *mut spdk_bdev_io).expect("null ptr"))
+    }
+}
+
+impl Bio {
+    /// obtain tbe Bdev this IO is associated with
```
jonathan-teh, line 159: Typo: tbe -> the

jonathan-teh (commented):
> All IO related functions are not in nexus_io.rs.

Did you mean **now** in?

mayastor/src/bdev/nexus/nexus_io.rs
gila: Yes, its a minor one so we will see, not to worried about that :)

mayastor/tests/replica_snapshot.rs
```diff
@@ -27,6 +27,7 @@ static NXNAME: &str = "replica_snapshot_test";
 static NXNAME_SNAP: &str = "replica_snapshot_test-snap";
 
 #[tokio::test]
+#[ignore]
```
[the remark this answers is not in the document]
gila: Done as part of the force push :+1:

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -752,299 +728,9 @@ impl Nexus {
...
-                    io.nvme_buf(),
-                    io.nvme_nbytes(),
-                    Some(Self::io_completion),
-                    io.as_ptr() as *mut _,
-                )
-            })
-            .collect::<Vec<_>>();
-
-        self.check_io_submission(&results, &io);
+    pub fn io_completion_local(_success: bool, _parent_io: *mut c_void) {
```
mtzaurus, line 732: Is this function used anywhere ?

mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -1,471 +1,478 @@
...
 }
 
-impl NioCtx {
-    #[inline]
-    pub fn dec(&mut self) {
-        self.in_flight -= 1;
-        debug_assert!(self.in_flight >= 0);
-    }
+#[allow(unused_macros)]
+macro_rules! container_of {
```
mtzaurus, line 55: Is this macro used in this module ? I wasn't able to find any occurrences of it.

mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -1,471 +1,478 @@
...
-            if self.ctx_as_mut_ref().status != IoStatus::Success {
-                debug!("BIO for nexus {} failed", self.nexus_as_ref().name)
-            }
-            // we are marking the IO done but not all child IOs have returned,
-            // regardless of their state at this point
-            if self.ctx_as_mut_ref().in_flight != 0 {
-                debug!("BIO for nexus marked completed but has outstanding")
-            }
+pub(crate) fn nexus_submit_io(mut io: NexusBio) {
+    if let Err(e) = match io.cmd() {
```
mtzaurus, line 112: Do we handle IoType::Reset ? Is it handled somewhere else ?

mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -1,471 +1,478 @@
...
+        // these IOs are submitted to all the underlying children
+        IoType::Write | IoType::WriteZeros | IoType::Unmap => io.submit_all(),
+        IoType::Flush => {
+            io.ok();
+            Ok(())
         }
-        unsafe {
-            spdk_bdev_io_complete(self.0.as_ptr(), IoStatus::Success.into())
+        IoType::NvmeAdmin => {
+            io.fail();
```
mtzaurus, line 121: Can we print error message here in addition to returning error, so we can easily see that happening in logs.

mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -1,471 +1,478 @@
...
+            // retry.
+            Disposition::Complete(IoStatus::Failed) => self.fail(),
+
+            // IOs were submitted before we bumped into an error. The IO has
+            // now completed, so we can finally report back to the
+            // callee that we encountered ENOMEM while processing
+            Disposition::Complete(IoStatus::NoMemory) => self.no_mem(),
+
+            // remove child if any from the nexus
+            Disposition::Flying(IoStatus::Failed) => {
```
mtzaurus, line 229: Do we consider a situation when I/O error occurs exactly on the last replica involved in Nexus I/O ? Let's assume a situation with 3 replicas, so we need to distribute our I/O across these 3 replicas. We can observe the following sequence of events (initial in_flight value is 3):

1. First replica completes successfully,  in_flight = 2, disposition = Disposition::Flying(IoStatus::Success)
2. Second replica completes successfully,  in_flight = 1, disposition = Disposition::Flying(IoStatus::Success)
3. Third replica ,  in_flight = 0, ctx.status = IoStatus::Failed, disposition = Disposition::Complete(IoStatus::Failed)

Shall we retire the third replica too, because currently child removal is done only in response to Disposition::Flying(IoStatus::Failed), which means all replicas but not the last serving the I/O.

mayastor/src/bdev/nexus/nexus_bdev.rs
gila: It called in the lvol layer and I did not want to remove all the code when we enable it again. So i left it this way for now.

mayastor/src/bdev/nexus/nexus_io.rs
gila: I use it during development to validate that I'm holding pointers to what I think I am =) I can remove it if you like but I found it rather useful.

mayastor/src/bdev/nexus/nexus_io.rs
gila: Nope, fixed. thx!

mayastor/src/bdev/nexus/nexus_io.rs
gila: The match returns the Error variant and if its an error, we print it at line 132.

mayastor/src/bdev/nexus/nexus_io.rs
gila: Good question. Im not sure. Currently, the behaviour is such that we always retire the IO, and thus have a target, children. We then keep the controller in the paused state which is only valid for the duration of restarts/reboots of the FW. I don't think it's OK to leave it in that state. With this approach, we always keep the target in a valid state all IO's will error out accordingly.

Put differently, we retire up to N-1 children of an N-way mirror which to me seems the right thing to do for now.

mtzaurus: Such a situation is not related to the number of replicas but rather to the sequence of failures, observed on replicas. I mean, the following 2 sequences of I/O completions (among 3 replicas) will be handled differently, although they seem to be identical to me:

1. Succces, Failure, Success  => Replica 2 will be retired.
2. Success, Success, Failure => Replica 3 will not be retired.

gila: Right ok got it, I misunderstood your question. That indeed is exactly what happens right now and should be handled as well :+1:

mayastor/src/bdev/dev/nvme.rs
```diff
@@ -127,14 +127,9 @@ impl CreateDestroy for NVMe {
...
-                errno_result_from_i32((), errno).context(
-                    nexus_uri::DestroyBdev {
-                        name: self.name.clone(),
-                    },
-                )
-            }
-            .await
+            errno_result_from_i32((), errno).context(nexus_uri::DestroyBdev {
+                name: self.name.clone(),
+            })
```
paulyoong, line 132: Can `async` be removed from the function signature now?

gila: Hey @paulyoong Good question! It can not. The trait defined the async nature of the function, that it is in fact not async is some minor overhead we simply have to pay.

paulyoong: :+1:

spdk-sys/build.sh
```diff
@@ -11,12 +11,11 @@ rm libspdk.so
 
 ./configure --enable-debug \
 	--target-arch=nehalem \
-	--disable-tests \
 	--without-isal \
 	--with-iscsi-initiator \
 	--with-crypto \
 	--with-uring \
-	--disable-unit-tests
+	--enable-unit-tests
```
jkryl, line 18: just curious why would we want to build the tests when building spdk as dependency of mayastor

jkryl (approved):
Nice 🎉

mtzaurus (approved):

mayastor/src/bdev/nexus/nexus_io.rs
gila: @mtzaurus I added the case you mentioned, can you have a look?

mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -1,471 +1,504 @@
...
+    fn readv(&mut self) -> Result<(), Errno> {
+        if let Some(i) = self.inner_channel().child_select() {
+            let hdl = self.read_channel_at_index(i);
+            self.submit_read(hdl).map(|_| {
+                self.ctx_as_mut().in_flight += 1;
+            })
+        } else {
+            self.fail();
+            Err(Errno::ENODEV)
+        }
```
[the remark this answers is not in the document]
gila: Internal retries are done within the initiator Jan, we can rotate towards a different child if we want to in the future and I think we actually should. It is however what we had since the start so for now, to not make this any bigger, i'd suggest we keep it as is and rotate later?

spdk-sys/build.sh
gila: when building it this way, and if I'm not sure if its a bug in the library I run them sometimes. I can disable them no problem :+1:

mayastor/src/bdev/nexus/nexus_io.rs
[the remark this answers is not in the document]
jkryl: absolutely. Thanks for explaining 👍

gila:
bors try

mayastor/src/core/bio.rs
```diff
@@ -0,0 +1,316 @@
...
+}
+
+impl std::fmt::Pointer for Bio {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "{:p}", unsafe { self.0.as_ref() })
+    }
+}
+impl Debug for Bio {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(
```
Hoverbear, line 305: Suggest using https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_struct here

mayastor/src/core/bio.rs
```diff
@@ -0,0 +1,316 @@
...
+
+    /// get the IO context specific information for this IO
+    #[inline]
+    pub(crate) fn specific<T>(&self) -> &T {
+        unsafe { &*(self.0.as_ref().driver_ctx.as_ptr() as *const T) }
+    }
+
+    /// get a raw pointer to the base of the iov
+    #[inline]
+    pub(crate) fn iovs(&self) -> *mut spdk_sys::iovec {
```
Hoverbear, line 209: This function is probably unsafe?

mayastor/src/core/bio.rs
```diff
@@ -0,0 +1,316 @@
...
+    fn from(status: i8) -> Self {
+        (status as i32).into()
+    }
+}
+
+#[derive(Clone)]
+#[repr(transparent)]
+pub struct Bio(NonNull<spdk_bdev_io>);
+
+impl From<*mut c_void> for Bio {
```
Hoverbear, line 146: Use `std::convert::TryFrom` for fail-able converstions.

mayastor/src/core/bio.rs
gila: return a *mut T is not unsafe, dereferencing it is AFAIK?

mayastor/src/core/bio.rs
gila: We really dont want to have the system go one when this happens, so unwrap seems appropriate?

gila:
bors try

gila:
bors try

gila:
bors try

mayastor/src/core/bio.rs
Hoverbear: Sure!

mayastor/src/core/bio.rs
Hoverbear: Hmm according to https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html#dereferencing-a-raw-pointer you are indeed correct.

gila:
bors retry

PR #798, opened 2021-03-18
test(e2e): test multiple volumes in a single test pod

blaisedias:
Verify than within an E2E context and single pod can mount and
use (run fio) on multiple volumes provisioned with
	- a single replica
	-  multiple replicas

1. Create a single test pod
2. mount 6 volumes on the test pod
3. run fio on all volumes

Run these steps for volumes with 1 and 2 replicas


blaisedias:
bors try

scripts/e2e-test.sh
```diff
@@ -16,7 +16,9 @@ TOPDIR=$(realpath "$SCRIPTDIR/..")
 #   2. ms_pod_disruption SHOULD be the last test before uninstall
 #
 DEFAULT_TESTS="install basic_volume_io csi resource_check replica rebuild ms_pod_disruption uninstall"
-ONDEMAND_TESTS="install basic_volume_io csi resource_check uninstall"
+#ONDEMAND_TESTS="install basic_volume_io csi resource_check uninstall"
+#temporary
```
paulyoong, line 20: Can this be removed?

test/e2e/multiple_vols_pod_io/multiple_vols_pod_io_test.go
```diff
@@ -0,0 +1,178 @@
...
+var defTimeoutSecs = "120s"
+
+func TestMultipleVolumeIO(t *testing.T) {
+	// Initialise test and set class and file names for reports
+	common.InitTesting(t, "Multiple volumes single pod IO tests", "multiple-volume-pod-io")
+}
+
+func multipleVolumeIOTest(replicas int, volumeCount int, protocol common.ShareProto, volumeType common.VolumeType, binding storageV1.VolumeBindingMode) {
+	logf.Log.Info("MultipleVolumeIOTest", "replicas", replicas, "volumeCount", volumeCount, "protocol", protocol, "volumeType", volumeType, "binding", binding)
+	//	scName := "multi-vol-io-test-" + string(protocol)
```
paulyoong, line 27: Should this be removed? If it is a comment on the below line it doesn't seem to match the formatting i.e. it is missing the volumeType and binding.

paulyoong (approved):

scripts/e2e-test.sh
blaisedias: That was temporary I wanted to do a CI run so that we can get loki logs that I can point to in the bug I will raise. This file has now been restored.
This test will always fail at the moment so the it will not be included into CI runs.

test/e2e/multiple_vols_pod_io/multiple_vols_pod_io_test.go
blaisedias: Oops, yes removed now.

chriswldenyer (approved):
LGTM

blaisedias:
bors merge

PR #804, opened 2021-03-23
feat(nvmeadm): add support for a few controller options

jkryl:
The behaviour of the controller can be tuned better instead of just
accepting the default values from kernel.

Avoid repeating code for connecting to a target and make the code
more rusty.

This work was originally done to address CAS-499 (Mayastor targets
should set sane io timeouts). As it evolved we decided it's better
to document the behavior and suggest how to avoid it instead of
overriding the defaults. Merging it and improving nvmeadm lib
capabilities does not make any harm.

### DIFF
--- a/nvmeadm/src/nvmf_discovery.rs
+++ b/nvmeadm/src/nvmf_discovery.rs
@@ -1,4 +1,5 @@
 use std::{
+    convert::TryFrom,
     fmt,
     fs::OpenOptions,
     io::{ErrorKind, Read, Write},
@@ -41,12 +42,18 @@ static HOST_ID: once_cell::sync::Lazy<String> =
     });
 
 /// The TrType struct for all known transports types note: we are missing loop
-#[derive(Debug, Primitive)]
+#[derive(Clone, Debug, Primitive)]
 #[allow(non_camel_case_types)]
 pub enum TrType {
     rdma = 1,
     fc = 2,
     tcp = 3,
+}
+
+impl Default for TrType {
+    fn default() -> Self {
+        Self::tcp
+    }
 }
 
 impl fmt::Display for TrType {
@@ -56,7 +63,7 @@ impl fmt::Display for TrType {
 }
 
 /// AddressFamily, in case of TCP and RDMA we use IPv6 or IPc4 only
-#[derive(Debug, Primitive)]
+#[derive(Clone, Debug, Primitive)]
 pub enum AddressFamily {
     Pci = 0,
     Ipv4 = 1,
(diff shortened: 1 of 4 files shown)


nvmeadm/src/nvmf_discovery.rs
```diff
@@ -426,91 +440,129 @@ impl DiscoveryBuilder {
...
-        ));
-        connect_args.push_str(&format!("hostid={},", host_id));
+#[derive(Default, Debug, Builder)]
+#[builder(setter(into))]
+#[builder(build_fn(validate = "Self::validate"))]
+pub struct ConnectArgs {
+    /// IP address of the target
+    traddr: String,
+    /// Port number of the target
+    trsvcid: String,
```
tiagolobocastro, line 458: NvmeTarget uses u16 for the ports, would it make sense to do the same here?

nvmeadm/src/nvmf_discovery.rs
```diff
@@ -426,91 +440,129 @@ impl DiscoveryBuilder {
...
-            "hostnqn={},",
-            format!("nqn.2019-05.io.openebs.mayastor:{}", host_id)
-        ));
-        connect_args.push_str(&format!("hostid={},", host_id));
+#[derive(Default, Debug, Builder)]
+#[builder(setter(into))]
+#[builder(build_fn(validate = "Self::validate"))]
+pub struct ConnectArgs {
+    /// IP address of the target
+    traddr: String,
```
tiagolobocastro, line 456: Should this be an IpAddr?

tiagolobocastro (approved):
:+1:

nvmeadm/src/nvmf_discovery.rs
jkryl: we need to keep this structure transport independent because it can be used for rdma transports as well. Not from our CSI code but in target discovery code that is in nvmeadm lib.

nvmeadm/src/nvmf_discovery.rs
jkryl: The same reply as above applies here.

gila:
@jkryl I don't see where we set/change io_timeout in sys/class/block/X/queue/io_timeout this should be set to less insane values, in particular in AWS. What are your thoughts here? thanks!

jkryl:
@gila :  yea, that's not part of these changes. The fix for setting io timeout would have to be done in CSI agent code, while this fix is about nvme connect parameters in nvmeadm lib. Though there is a PR related to io timeout parameter. Your review of that PR would be much appreciated: https://github.com/openebs/mayastor-docs/pull/15 . thanks!

jkryl:
I had a chat with @gila  yesterday and the io timeout will be addressed by a different PR (out of scope of this PR). No worries that CAS-499 would go away. It will be closed when the io timeout problem is fixed.

gila (approved):

jkryl:
bors merge

jkryl:
bors retry

jkryl:
bors retry

PR #805, opened 2021-03-23
feat(nexus): share over nvmf with controller ID range

jonathan-teh:
Add a nexus_share JSON RPC that allows a nexus bdev to be shared over
nvmf with a specified controller ID range. This allows the same nexus
on different Mayastor instances to use non-overlapping controller ID
ranges thereby allowing the initiator to connect to same nexus over
multiple paths.

Add a mocha test to test_nexus to verify that the specified controller
ID range is used when sharing the nexus over nvmf.

Fixes CAS-779

### DIFF
--- a/mayastor/src/bdev/nexus/mod.rs
+++ b/mayastor/src/bdev/nexus/mod.rs
@@ -2,7 +2,13 @@
 
 use spdk_sys::spdk_bdev_module;
 
-use crate::bdev::nexus::{nexus_bdev::Nexus, nexus_fn_table::NexusFnTable};
+use crate::{
+    bdev::nexus::{nexus_bdev::Nexus, nexus_fn_table::NexusFnTable},
+    core::{Bdev, Share},
+    jsonrpc::{jsonrpc_register, Code, JsonRpcError, Result},
+};
+use futures::{future::Future, FutureExt};
+use std::pin::Pin;
 
 /// Allocate C string and return pointer to it.
 /// NOTE: The resulting string must be freed explicitly after use!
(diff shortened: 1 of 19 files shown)


mayastor/src/subsys/nvmf/subsystem.rs
```diff
@@ -269,6 +270,30 @@ impl NvmfSubsystem {
...
     }
 
+    /// set controller ID range
+    pub fn set_cntlid_range(
+        &self,
+        cntlid_min: u16,
+        cntlid_max: u16,
+    ) -> Result<(), Error> {
+        unsafe {
+            spdk_nvmf_subsystem_set_cntlid_range(
```
mtzaurus, line 280: Is this function part of standard SPDK API ? Where it resides ? (I wasn't able to find it via https://spdk.io/doc/)

mayastor/src/bdev/nexus/mod.rs
```diff
@@ -29,9 +35,79 @@ pub mod nexus_module;
...
+                if proto != "iscsi" && proto != "nvmf" {
+                    return Err(JsonRpcError {
+                        code: Code::InvalidParams,
+                        message: "invalid protocol".to_string(),
+                    });
+                }
+                if let Some(bdev) = Bdev::lookup_by_name(&args.name) {
+                    match proto.as_str() {
+                        "nvmf" => {
+                            bdev.share_nvmf(Some((args.cntlid_min, args.cntlid_max)))
```
mtzaurus, line 70: Generic question: are there any restrictions applied to (args.cntlid_min, args.cntlid_min) ? Should cntlid_min be always less than cntlid_min ? If user intentionally passes a lesser cntlid_min , will it break something ? Shall the values be different or so and shall we enforce some other sanity checks against the values passed to us ?

mayastor/src/subsys/nvmf/subsystem.rs
jonathan-teh: It's not, the PR for that wasn't ready at the time, it is now openebs/spdk#12.

mayastor/src/bdev/nexus/mod.rs
jonathan-teh: Sanity checking is done in `spdk_nvmf_subsystem_set_cntlid_range`.

mayastor/src/subsys/nvmf/subsystem.rs
jonathan-teh: I've dropped the `spdk_` prefix.

gila (approved):

mayastor/src/bdev/nexus/mod.rs
```diff
@@ -29,9 +35,79 @@ pub mod nexus_module;
...
+    uri: String,
+}
+
 /// public function which simply calls register module
 pub fn register_module() {
-    nexus_module::register_module()
+    nexus_module::register_module();
+
+    jsonrpc_register(
+        "nexus_share",
```
jkryl, line 56: I'm wondering why don't we change grpc method instead of creating a whole new jsonrpc method.

jkryl (commented):
Could you explain how it is supposed to be used by control plane? I would assume that the control plane would create multiple nexus instances with non-overlapping controller id ranges. Can controllers of unrelated volumes still have the same controller id?

jonathan-teh:
> Could you explain how it is supposed to be used by control plane? I would assume that the control plane would create multiple nexus instances with non-overlapping controller id ranges.

Yes, exactly that. The control plane will be responsible for assigning the non-overlapping ranges.

> Can controllers of unrelated volumes still have the same controller id?

Yes, the controller ID only needs to be unique within the same NVMe subsystem.

mayastor/src/bdev/nexus/mod.rs
jonathan-teh: That was @gila's suggestion and that perhaps eventually gRPC methods will be replaced by JSON-RPC ones.

Hoverbear (approved):

mayastor/src/bdev/nexus/mod.rs
jkryl: until that happens I'm not sure it is reasonable to have the code duplicated. Having grpc nexus_share method and jsonrpc nexus_share method and both having different arguments is quite confusing already. Regardless if we decide to use jsonrpc or grpc in the future, if we start duplicating rpc methods now, it's going to make any change in the future more complicated. @gila any opinion?

gila: @jkryl I see your concern. But there is reason to the madness. What we want to do is figure out how we can, in a reasonable way -- implement a proper method without going out and breaking things all over the map by changing the bdev_share() and/or publish calls. So it is temporary really and subject to some churn as we go and figure out what's the best to do this.

jonathan-teh:
bors merge

PR #807, opened 2021-03-24
feat(nvmx): abstract block devices for nexus children

mtzaurus:
Nexus now treats all its replica devices as abstract block devices,
which allows clear separation, and uses abstract block device I/O
handles for serving child I/O.

In order to prevent start of controller resets in response to I/O errors
(mostly connection failures), maximum number of controller resets
introduced along with reset cool down interval.

Per-channel I/O operation accounting is introduced to have better I/O
observability/statistics.

Implements CAS-781

### DIFF
--- a/mayastor/src/bdev/dev/nvmx/channel.rs
+++ b/mayastor/src/bdev/dev/nvmx/channel.rs
@@ -202,6 +202,7 @@ pub struct NvmeIoChannelInner<'a> {
     pub qpair: Option<IoQpair>,
     io_stats_controller: IoStatsController,
     pub device: Box<dyn BlockDevice>,
+    num_pending_ios: u64,
 
     // Flag to indicate the shutdown state of the channel.
     // We need such a flag to differentiate between channel reset and shutdown.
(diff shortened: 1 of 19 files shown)


mayastor/src/bdev/dev/nvmx/channel.rs
```diff
@@ -245,6 +251,22 @@ impl NvmeIoChannelInner<'_> {
...
 
+    /// Account active I/O for channel.
+    #[inline]
+    pub fn account_io(&mut self) {
+        self.num_pending_ios += 1;
+    }
+
+    /// Discard active I/O operation for channel.
+    #[inline]
+    pub fn discard_io(&mut self) {
```
gila, line 262: nit; discard does not sound like a really good name for this IMHO.

gila (approved):

mayastor/src/bdev/dev/nvmx/controller_inner.rs
```diff
@@ -44,11 +49,26 @@ impl TryFrom<u32> for DeviceTimeoutAction {
...
 }
 
+/// Maximum number of controller reset attempts to be taken in case
+/// controller reset fails. Once this limit is reached, the next possible
+/// controller reset will be allowed only when reset cooldown interval
+/// elapses.
+/// This is done to prevent the storm of reset requests in response to
+/// frequent I/O errors in a controller (including errors while processing
+/// admin queue completions).
+const MAX_RESET_ATTEMPTS: u32 = 5;
```
Hoverbear, line 59: We will likely want to make these configurable.

PR #810, opened 2021-03-25
fix(nvmx): nexus children removed in response to events

mtzaurus:
Nexus now registers a listener for all its children and finalizes
child removal in response to device removal event.

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_child.rs
+++ b/mayastor/src/bdev/nexus/nexus_child.rs
@@ -136,7 +136,7 @@ impl Display for ChildState {
 #[derive(Serialize)]
 pub struct NexusChild {
     /// name of the parent this child belongs too
-    pub(crate) parent: String,
+    parent: String,
 
     /// current state of the child
     #[serde(skip_serializing)]
@@ -303,6 +303,11 @@ impl NexusChild {
     /// Get full name of this Nexus child.
     pub(crate) fn get_name(&self) -> &str {
         &self.name
+    }
+
+    /// Get name of the nexus this child belongs to.
+    pub fn get_nexus_name(&self) -> &str {
+        &self.parent
     }
 
     /// Online a previously offlined child.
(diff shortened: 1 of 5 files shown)


mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -489,7 +494,13 @@ impl NexusChild {
     pub(crate) async fn destroy(&self) -> Result<(), NexusBdevError> {
         if self.device.is_some() {
             self.set_state(ChildState::Destroying);
-            device_destroy(&self.name).await
+            info!("{} destroying underlaying block device", self.name);
```
cjones1024, line 497: should be "underlying" rather than underlaying

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -489,7 +494,13 @@ impl NexusChild {
...
+            let rc = device_destroy(&self.name).await;
+            info!(
+                "{} underlaying block device destroyed, rc = {:?}",
```
cjones1024, line 500: underlaying --> underlying

gila (approved):

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -305,6 +305,11 @@ impl NexusChild {
         &self.name
     }
 
+    /// Get name of the nexus this child belongs to.
+    pub fn get_nexus_name(&self) -> &str {
```
Hoverbear, line 309: This might be more appropriate as  `parent()`?

Hoverbear (approved):
Rest looks good.

mayastor/src/bdev/nexus/nexus_child.rs
mtzaurus: Indeed ! Thanks for highlighting the typo.

mayastor/src/bdev/nexus/nexus_child.rs
mtzaurus: Parent here is nexus. :-) There are no other parents for nexus children than a nexus.
We will probably change 'parent' member to 'nexus' or so.

PR #812, opened 2021-03-25
fix: change to use mayadata/fio CAS-599

blaisedias:
Update deploy/fio.yaml to use the newly minted
mayadata/fio image on docker hub

### DIFF
--- a/deploy/fio.yaml
+++ b/deploy/fio.yaml
@@ -9,8 +9,7 @@ spec:
        claimName: ms-volume-claim
   containers:
     - name: fio
-      image: nixery.dev/shell/fio/tini
-      command: [ "tini", "--" ]
+      image: mayadata/fio
       args:
         - sleep
         - "1000000"


jkryl (approved):
thanks Blaise ❤️

deploy/fio.yaml
```diff
@@ -9,8 +9,7 @@ spec:
        claimName: ms-volume-claim
   containers:
     - name: fio
-      image: nixery.dev/shell/fio/tini
-      command: [ "tini", "--" ]
+      image: mayadata/fio
       args:
         - sleep
```
arne-rusek, line 14: Shouldn't the `tini` stay somewhere? Without it it would take ages to shut down the container, wouldn't it?

Either `command` or in `args`?

```suggestion
        - tini
        - "--"
        - sleep
```

jkryl: My understanding was that the command is defined in the Dockerfile. If not we definitely need to define it here and I suggest we use tail -f /dev/null (or similar) instead of sleep with very high value.

arne-rusek: > My understanding was that the command is defined in the Dockerfile
It is not. But can be if someone puts it there :-)

arne-rusek: I've added `ENTRYPOINT ["/sbin/tini", "--"]` into the container as it makes sense to have it baked into the container and I cannot imagine a use-case where one doesn't want to have it.

When run as: `docker run mayadata/fio sleep 8000` or `docker run mayadata/fio tail -f /dev/null` it works and when I try to stop the container (which without `tini` waits for a timeout) it works and is quick.

note: I've removed suggestion from above comment as it doesn't make sense now,

jkryl: nice

arne-rusek (approved):
LGTM

jkryl:
@blaisedias : is there any particular reason why this hasn't been merged yet? In the meantime I tripped over one other place that needs to be updated and that is our gitbook (different repo). It contains verbatim copy of fio yaml file so if you can update that too, that would be great.

blaisedias:
bors merge

PR #815, opened 2021-03-27
feat(csi): be able to set io timeout on nvme initiator

jkryl:
User sets the timeout in storage class and from there it is propagated
to CSI agent that sets it on the device.

Resolves: CAS-499


csi/src/dev/nvmf.rs
```diff
@@ -90,20 +116,55 @@ impl Attach for NvmfAttach {
...
+                    DeviceError::new(&format!(
+                        "NVMe device \"{}\" does not match \"{}\"",
+                        dev_name, *DEVICE_REGEX,
+                    ))
+                })?
+                .get(1)
+                .unwrap()
+                .as_str();
+            let path = format!(
+                "/sys/class/nvme/nvme{}/nvme{}c1n1/queue",
```
Hoverbear, line 148: My devices don't have this file? 

```rust
❯ ls /sys/class/nvme/nvme
nvme0/ nvme1/ 

~ 
❯ ls /sys/class/nvme/nvme0/
address            dev                firmware_rev       numa_node          power/             rescan_controller  serial             state              subsystem/         uevent             
cntlid             device/            model              nvme0n1/           queue_count        reset_controller   sqsize             subsysnqn          transport          

~ 
❯ ls /sys/class/nvme/nvme1/
address  cntlid  dev  device  firmware_rev  model  numa_node  nvme1n1  power  queue_count  rescan_controller  reset_controller  serial  sqsize  state  subsysnqn  subsystem  transport  uevent
```

jonathan-teh: I think the `nvme{}c1n1` device requires a kernel with `CONFIG_NVME_MULTIPATH=y` and `nvme_core.multipath=1` (which is the default once the former is enabled).

jonathan-teh: I've confirmed that by reloading the `nvme_core` kernel module with `multipath=0`. This means a fallback is necessary as the `c1` bit may not be present where multipath is disabled (or not compiled in). Also it's usually `c0` for a single controller. Where multipath is actually used (not yet in csi), paths appear to be `nvme0/nvme0c0n1` and `nvme1/nvme0c1n1` for a block device accessed as `/dev/nvme0n1`.

Hoverbear (approved):
Code seems fine otherwise.

csi/src/dev/nvmf.rs
Hoverbear: Ahh okay, :) Glad I asked.

gila:
Good catch @hoverbear we probably should try both paths?

csi/src/dev/nvmf.rs
jkryl: Good catch! I actually thought about this code even when writing it (surprise :-) ) and I thought that it could break when we have different controller IDs allocated for different controllers with ANA feature. @jonathan-teh would perhaps know better. So what I could do is to use a glob pattern perhaps because I don't really care about multipathing or controller IDs. So the pattern would become `/sys/class/nvme/nvme{}/nvme{}*n1/` . The n1 at the end would stay because mayastor volumes have never more than one namespace. He does it taste to you guys? Or any other suggestion? (cc @gila )

jonathan-teh: The `c0` bit is simply a controller index maintained by the initiator. There will be a controller index as long as the kernel initiator supports multipath, regardless of whether the target advertises ANA support.
Your pattern needs a tweak. With 3 paths to the first namespace and 1 to the second, the devices available are:
```
host:/sys/class/nvme$ ls -1 /dev/nvme*
/dev/nvme0
/dev/nvme0n1
/dev/nvme1
/dev/nvme2
/dev/nvme3
/dev/nvme3n1
/dev/nvme-fabrics
host:/sys/class/nvme$ ls -d nvme?/nvme*
nvme0/nvme0c0n1  nvme1/nvme0c1n1  nvme2/nvme0c2n1  nvme3/nvme3c3n1
```
so I think it should be `/sys/class/nvme/nvme{}/nvme*n1/`, which should only match one entry.

jkryl: Perfect. Let's make it so then. Thanks @jonathan-teh !

test/grpc/test_csi.js
```diff
@@ -427,7 +444,17 @@ function csiProtocolTest (protoname, shareType, timeoutMillis) {
         client.nodeStageVolume(getDefaultArgs(), (err) => {
           if (err) return done(err);
           assert.equal(getFsType(mountTarget), 'xfs');
-          done();
+          // Other protocols than NVMF do not honor ioTimeout setting
+          if (protoName !== 'NVMF') return done();
+          let major = getFsDevice(mountTarget).match(/nvme(\d+)n1/)[1];
+          fs.readFile(
+            `/sys/class/nvme/nvme${major}/nvme${major}c1n1/queue/io_timeout`,
```
jonathan-teh, line 451: This won't work where there is e.g. `/sys/class/nvme2/nvme0c2n1` for the case of connecting to 3 paths with the same nexus/subnqn. I think it should be `/sys/class/nvme/nvme{major}/nvme*n1/...` like in the CSI plugin itself.

jonathan-teh (changes requested):

jonathan-teh (approved):

test/grpc/test_csi.js
jkryl: will fix it. thanks 👍

jkryl:
bors merge

PR #818, opened 2021-03-29
test(e2e): do not send xray reports via bors try

chriswldenyer:
Do not send XRay reports as a result of a try build.
Try builds are usually experimental so the test results
may not be meaningful.


chriswldenyer:
bors try

chriswldenyer:
Superseded by #821

Issue #772, opened 2021-03-05
failed to provision volume with StorageClass "mayastor-iscsi"


krishnakekan19:
**Describe the bug**
I am trying to install maystor openebs storage on RKE on-premise cluster, I tried to test the setup with StorageClass with PVC creation. StorageClass gets created successfully but PVC shows `STATUS as Pending`. It is not able to create a volume out of SC.
It needs to create PV dynamically with given SC and shows status as `BOUND`.

**To Reproduce**
You can follow the below setup documentation on RHEL8.2 RKE cluster:
ref: https://mayastor.gitbook.io/introduction/quickstart/prerequisites
CMDS to reproduce it:
```
Iscsi and hugepages are need to be manually set
kubectl label node node1 openebs.io/engine=mayastor   //label the worker nodes
kubectl create namespace mayastor                                   //create a namespace
kubectl create -f https://raw.githubusercontent.com/openebs/Mayastor/master/deploy/moac-rbac.yaml
kubectl apply -f https://raw.githubusercontent.com/openebs/Mayastor/master/csi/moac/crds/mayastorpool.yaml
kubectl apply -f https://raw.githubusercontent.com/openebs/Mayastor/master/deploy/nats-deployment.yaml
kubectl apply -f https://raw.githubusercontent.com/openebs/Mayastor/master/deploy/csi-daemonset.yaml
kubectl apply -f https://raw.githubusercontent.com/openebs/Mayastor/master/deploy/moac-deployment.yaml
kubectl apply -f https://raw.githubusercontent.com/openebs/Mayastor/master/deploy/mayastor-daemonset.yaml

#apply this with 3 worker nodes
cat <<EOF | kubectl create -f -
apiVersion: openebs.io/v1alpha1
kind: MayastorPool
metadata:
  name: pool-on-node-worker1
  namespace: mayastor
spec:
  disks:
  - /dev/sdb
  node: worker1
EOF

cat <<EOF | kubectl create -f -
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: mayastor-iscsi
parameters:
  # Set the number of data replicas ("replication factor")
  repl: '1'
  # Set the export transport protocol
  protocol: 'iscsi'
provisioner: io.openebs.csi-mayastor
EOF

cat <<EOF | kubectl create -f -
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: ms-volume-claim
spec:
  accessModes:
  - ReadWriteOnce
[...]

PR #832, opened 2021-04-07
fix(moac): failed to provision volume on RKE

jkryl:
Two issues, described in great depth in #772 github ticket, were
encountered by a user while trying to provision mayastor volume on
RKE cluster:

1. Mayastor config file was mounted as a directory instead as a file
2. Unhandled exception in Moac (result of partially created volume)

Both issues are fixed in this commit.

Add 1. There does not seem to be a good reason for mounting config
file instead of just the whole config directory so I have removed
that mount. It's not the first time we see this issue and although it
is probably caused by a bug in k8s/OS it can be easily avoided.

### DIFF
--- a/chart/templates/mayastor-daemonset.yaml
+++ b/chart/templates/mayastor-daemonset.yaml
@@ -65,8 +65,6 @@ spec:
           mountPath: /dev/shm
         - name: configlocation
           mountPath: /var/local/mayastor/
-        - name: config
-          mountPath: /var/local/mayastor/config.yaml
         resources:
           # NOTE: Each container must have mem/cpu limits defined in order to
           # belong to Guaranteed QoS class, hence can never get evicted in case of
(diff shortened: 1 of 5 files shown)


paulyoong (approved):

tiagolobocastro (approved):
I can't remember if there was a particular reason other than making sure we can create a file in said directory before attempting to start mayastor.
Looks like a k8s bug: https://github.com/kubernetes/kubernetes/issues/65825

jkryl:
Thanks Tiago for finding the issue! It seems the issue is rather old and it says it's fixed. The user was running 1.19 version of k8s which is not that old. Anyway for the sake of simplicity I think it's best to remove the mountpoint that we don't really need. I will proceed with merging the PR 👍

tiagolobocastro:
> Thanks Tiago for finding the issue! It seems the issue is rather old and it says it's fixed. The user was running 1.19 version of k8s which is not that old. Anyway for the sake of simplicity I think it's best to remove the mountpoint that we don't really need. I will proceed with merging the PR +1

Where does it say fixed? It's been closed due to lack of information...

jkryl:
Just my wrong conclusion, closed != fixed.

jkryl:
bors merge

jkryl:
bors retry

PR #833, opened 2021-04-07
fix(moac): cannot unpublish nexus on a node that is down

jkryl:
This introduces a new volume state "offline" that is used for
volumes that are inaccessible (the node with the nexus is down).
It is important distinction from "faulted" state that has been
used for this case previously.  "faulted" speaks for the data
(data loss), while "offline" speaks for availability of the data
(access). FSA has been fixed to handle this new state properly so
it is possible to:

1. create pod with mounted volume (on a different node than nexus),
2. bring the node with nexus down,
3. delete pod with the mounted volume,
4. recreate the pod with the same volume again.

This gives a chance to manually recover the volume by the user
action in case that the node with nexus never comes back online.

What happens in the background is that a series of CSI calls
(unpublish/publish) brings the volume to the state when it can be
used again. Making this process more smooth without requiring the
manual intervention is out of scope of this ticket.

Resolves: CAS-816

### DIFF
--- a/csi/moac/nexus.ts
+++ b/csi/moac/nexus.ts
@@ -6,6 +6,7 @@ import * as _ from 'lodash';
 const { GrpcCode, GrpcError, mayastor } = require('./grpc_client');
 const log = require('./logger').Logger('nexus');
 
+import { Node } from './node';
 import { Replica } from './replica';
 
 // Protocol used to export nexus (volume)
@@ -43,7 +44,7 @@ function compareChildren(a: Child, b: Child) {
 }
 
 export class Nexus {
-  node?: any;
+  node?: Node;
   uuid: string;
   size: number;
   deviceUri: string;
@@ -60,7 +61,7 @@ export class Nexus {
   // @param {object[]} props.children   Replicas comprising the nexus (uri and state).
   //
   constructor(props: any) {
-    this.node = null; // set by registerNexus method on node
+    this.node = undefined; // set by registerNexus method on node
     this.uuid = props.uuid;
     this.size = props.size;
     this.deviceUri = props.deviceUri;
@@ -126,7 +127,7 @@ export class Nexus {
   // When anything in nexus changes, this can be called to emit mod event
   // (a shortcut for frequently used code).
   _emitMod() {
-    this.node.emit('nexus', {
+    this.node!.emit('nexus', {
       eventType: 'mod',
       object: this
     });
@@ -139,7 +140,7 @@ export class Nexus {
   bind(node: any) {
     this.node = node;
     log.debug(`Adding "${this.uuid}" to the nexus list of node "${node}"`);
-    this.node.emit('nexus', {
+    this.node!.emit('nexus', {
       eventType: 'new',
       object: this
     });
@@ -148,12 +149,12 @@ export class Nexus {
   // Unbind the previously bound nexus from the node.
   unbind() {
     log.debug(`Removing "${this}" from the nexus list`);
-    this.node.unregisterNexus(this);
-    this.node.emit('nexus', {
+    this.node!.unregisterNexus(this);
+    this.node!.emit('nexus', {
       eventType: 'del',
       object: this
     });
-    this.node = null;
+    this.node = undefined;
   }
 
   // Set state of the nexus to offline.
@@ -163,6 +164,11 @@ export class Nexus {
     log.warn(`Nexus "${this}" got offline`);
     this.state = 'NEXUS_OFFLINE';
     this._emitMod();
+  }
+
+  // Return true if the nexus is down (unreachable).
+  isOffline() {
+    return !(this.node && this.node.isSynced());
   }
 
   // Publish the nexus to make accessible for IO.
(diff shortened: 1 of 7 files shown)


gila (commented):
@jkryl Can we more configurable within this approach? The reason I ask is, what if we have a node that has a network failure for N seconds? we might want to give *some*  variance here? what do you think?

csi/moac/nexus.ts
```diff
@@ -165,6 +166,11 @@ export class Nexus {
     this._emitMod();
   }
 
+  // Return true if the nexus is reachable (online).
+  isOffline() {
+    return !(this.node && this.node.isSynced());
+  }
```
paulyoong, line 172: The comment appears to state the opposite of the function. Doesn't the function return true if the nexus is offline?

jkryl:
The question being how soon the nexus can be replaced by a new one. I will try to answer it. A couple of factors must be taken into account as it's far from simple. What I have described above is rather eventually realised state. It does not happen immediately when the node with nexus goes down.

First thing to consider is how soon the moac is able to detect an offline node. Currently moac tries to sync a node once a minute. If the sync fails, then there are two other tunables: 1. # of retries (default 0) and 2. retry interval (default 10s). So in the current implementation it takes up to one minute to detect a node that is down.

Another thing to realise is that unless the application that uses the volume is terminated, nothing can happen to the nexus. It must stay on the node where it is even if the node is down. If the nexus is down it usually takes 10 minutes at least for IO error to be reported by the nvme initiator. Only then the pod can be terminated because all attempts to unpublish/unstage the volume will get stuck on unmounting the volume.

From that follows that replacing a bad nexus is a matter of minutes (and currently tens of minutes) rather than seconds. We don't have to worry about nexuses popping up upon every minor network outage. Though it gives a hope to users that they will be able eventually to access their data again which wasn't possible before.

I see two action items following from this discussion that are out of scope of this PR but still related:

1. provide new command line options for moac to tweak sync interval and number of retries (CAS-85),
2. extend io_timeout setting from storage class to controller timeout (ctrl loss tmo) to control when the system can be unmounted (CAS-823)

PS: there are still bugs in our csi agent that prevent us from unstaging a volume with unreachable target (CAS-817). The scope of this ticket is just the control plane.

csi/moac/nexus.ts
```diff
@@ -299,8 +311,14 @@ export class Nexus {
   // Destroy nexus on storage node.
   async destroy() {
     log.debug(`Destroying nexus "${this}" ...`);
-    await this.node.call('destroyNexus', { uuid: this.uuid });
-    log.info(`Destroyed nexus "${this}"`);
+    if (!this.node!.isSynced()) {
+      // We don't want to block the volume life-cycle in case that the node
+      // is down - it may never come back online.
+      log.warn(`Faking the destroy of "${this}" because it is unreachable`);
```
paulyoong, line 317: Am I right in thinking that if we fake the destruction of the nexus and it reappears somehow (maybe due to a transient network fault) that the FSA loop will eventually destroy it for real?

csi/moac/nexus.ts
jkryl: You are absolutely right 🤦  let me fix that 👍

csi/moac/nexus.ts
jkryl: Correct. And for the well known problem of queued IOs being written to replicas from reappearing nexus we must use ACLs or similar.

paulyoong: Thanks :+1:

paulyoong (approved):

csi/moac/replica.ts
```diff
@@ -136,10 +136,16 @@ export class Replica {
   async destroy() {
     log.debug(`Destroying replica "${this}" ...`);
     if (!this.pool) {
-      throw new Error('Cannot offline a replica that has not been bound');
+      throw new Error('Cannot destroy a replica that has not been bound');
+    }
+    if (!this.pool.node.isSynced()) {
+      // We don't want to block the volume life-cycle in case that the node
+      // is down - it may never come back online.
+      log.warn(`Faking the destroy of "${this}" because it is unreachable`);
```
tiagolobocastro, line 144: My only worry here is what happens if the old nexus is not really offline, just having some network issues, and if for some reason it tries to write to the replicas, eg: update the "golden number". 
eg, a rebuild process could be ongoing and at the end it does reconfigure and bumps up the golden number while they are are already being used by another "newer" nexus on a different node.
A way to solve this could be to have the replicas tell the nexus via the data plane, "you're out".
A temporary way forward for now, could be checking if there is no rebuild ongoing? Can we assume that the nexus would not issue any other IO to the children since we've been called to unpublish?

tiagolobocastro (approved):
Approving on the basis that we will keep track of the outstanding issue in a ticket.

jkryl:
Thank you @tiagolobocastro ! For the record the ticket number that tracks additional work that needs to be done is CAS-833. Also noted in the commit message.

jkryl:
bors merge

jkryl:
bors retry

PR #842, opened 2021-04-15
feat(moac): nexus and app must run on the same node if possible

jkryl:
This avoids some failure recovery scenarios that would be
difficult to handle otherwise. App and nexus are either both up
or down. If they are down, then k8s restarts the application on a
different node. That triggers unpublish and publish calls that
allow mayastor CSI plugin to move the nexus along with the app to
the new node. In the future, the restriction to run both on the
same node may be lifted once we provide redundancy at the nexus
level.

What users need to do to take advantage of this new code
improving stability? At this time they need to run stateful apps
strictly on storage nodes using any of the provided mechanisms
to influence k8s scheduler decisions (i.e. nodeSelector). In the
future, they won't need to do anything as the node affinity for
apps will be set by the CSI plugin. This will be added in the
followup commit.

Additional losely related change is that storage class parameters
are checked during volume creation rather than later during the
publish to detect errors rather sooner than later.

Resolves: CAS-818

### DIFF
--- a/csi/moac/test/csi_test.js
+++ b/csi/moac/test/csi_test.js
@@ -355,6 +355,72 @@ module.exports = function () {
         );
       });
 
+      it('should fail if ioTimeout is used with protocol other than nvmf', async () => {
+        createVolumeStub.resolves(returnedVolume);
+        await shouldFailWith(GrpcCode.INVALID_ARGUMENT, () =>
+          client.createVolume().sendMessage({
+            name: 'pvc-' + UUID,
+            capacityRange: {
+              requiredBytes: 10,
+              limitBytes: 20
+            },
+            volumeCapabilities: [
+              {
+                accessMode: { mode: 'SINGLE_NODE_WRITER' },
+                filesystem: {}
+              }
+            ],
+            parameters: {
+              protocol: 'iscsi',
+              ioTimeout: 30
+            }
+          })
+        );
+      });
+
+      it('should fail if ioTimeout has invalid value', async () => {
+        createVolumeStub.resolves(returnedVolume);
+        await shouldFailWith(GrpcCode.INVALID_ARGUMENT, () =>
+          client.createVolume().sendMessage({
+            name: 'pvc-' + UUID,
+            capacityRange: {
+              requiredBytes: 10,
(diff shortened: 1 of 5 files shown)


jkryl:
bors try

csi/moac/test/csi_test.js
```diff
@@ -355,6 +355,72 @@ module.exports = function () {
         );
       });
 
+      it('should fail if ioTimeout is used with other protocol than nvmf', async () => {
```
tiagolobocastro, line 358: "protocol other than"  ?

csi/moac/volume.ts
```diff
@@ -341,6 +341,19 @@ export class Volume {
         }
         return;
       }
+    } else {
+      if (this.nexus) {
+        if (this.nexus.node?.name !== this.publishedOn) {
+          // Respawn the nexus on the desired node.
+          log.info(`Recreating the nexus "${this.nexus}" on the desired node "${this.publishedOn}"`);
```
tiagolobocastro, line 348: So we destroy it now and then the fsa will recreate it on the correct node?

tiagolobocastro (approved):
Should we document the new behaviour and add recommendation on the gitbooks for example?

csi/moac/test/csi_test.js
jkryl: The storage classes provided by us reflect the accurate configuration. That is, the iscsi sc does not have it and nvmf sc does have it. Given that scsi is obsolete and will be likely removed at some point in the future (at least from the documentation) I would rather keep it simple.

csi/moac/volume.ts
jkryl: Correct

csi/moac/test/csi_test.js
jkryl: will fix. thanks 👍

jkryl:
I agree we should definitely update our gitbook. I would do that after the next PR (late binding of volume) is merged to avoid updating it twice. thanks for reviewing the fix!

paulyoong (approved):

jkryl:
bors merge

PR #850, opened 2021-04-20
feat(csi): support WaitForFirstConsumer for mayastor volumes

jkryl:
We want to support late binding of mayastor volumes because we can
optimize distribution of the volume over the nodes. In particular we
can enforce that the nexus and one of the replicas is created on
the same node as the app when creating the volume.

There is a new storage class parameter "local" that enforces the
locality if set to true (the default in our storage class yaml).
If it is not possible to schedule the app on the same node as nexus
and replica and "local" is enabled, then the pod remains in
unscheduleable state until the user rectifies the situation.

There is one downside of local volumes though. Once the node
affinity is set during the volume creation it cannot be changed
later. Assuming that all nodes holding original replicas of the
volume would be removed from the cluster, then the app becomes
unschedulable although that the replicas might have been migrated
to new nodes already.

Convert index.js to typescript to catch more errors in the file and
provide simple moac script that calls index.ts:main().

Resolves: CAS-834

### DIFF
--- a/chart/templates/moac-deployment.yaml
+++ b/chart/templates/moac-deployment.yaml
@@ -20,7 +20,7 @@ spec:
           args:
             - "--v=2"
             - "--csi-address=$(ADDRESS)"
-            - "--feature-gates=Topology=false"
+            - "--feature-gates=Topology=true"
           env:
             - name: ADDRESS
               value: /var/lib/csi/sockets/pluginproxy/csi.sock
--- a/csi/moac/.gitignore
+++ b/csi/moac/.gitignore
@@ -2,6 +2,7 @@
 /proto/
 /result
 /csi.js
+/index.js
 /nexus.js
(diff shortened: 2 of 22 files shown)


gila (approved):

jkryl:
bors try

jkryl:
bors retry

paulyoong (approved):

jkryl:
bors merge

jkryl:
bors retry

jkryl:
waiting for a bug in e2e tests to be fixed: https://github.com/mayadata-io/mayastor-e2e/pull/42

jkryl:
bors retry

jkryl:
bors merge

jkryl:
bors retry

jkryl:
bors merge

PR #854, opened 2021-04-22
fix(nexus): remove alias from bdevs that are local

gila:
If the alias we try to set, is present, we return true rather then
failing the operation.

If the bdev is destroyed, we remove the URI from the alias.

While debugging this i've also decided to remove the tracing
of yield points as they make the calls to noise and have no
real value


jkryl (approved):

jonathan-teh (approved):

jkryl:
bors merge

PR #855, opened 2021-04-22
fix(moac): fix for assertions hit while preparing demo

jkryl:


csi/moac/volume.ts
```diff
@@ -573,6 +571,8 @@ export class Volume {
     this.state = VolumeState.Unknown;
     log.info(`Volume "${this}" with ${this.replicaCount} replica(s) and size ${this.size} was created`);
     this.fsa();
+    // TODO: should we wait for the temporary nexus (created just for labeling
+    // replicas) to go away before we return back OK? IMHO we should.
```
tiagolobocastro, line 575: this would make sense... could we destroy it right after we create it?

tiagolobocastro (approved):

csi/moac/volume.ts
jkryl: yes, but it's better to leave it to fsa. fsa takes care of serializing all volume operations. We create it here and fsa will destroy it when it is no longer needed (that is when the volume is healthy and no published). So here we can just wait until the nexus is null. I'm not sure if we need any timeout but maybe we can make it simple initially and stick to the rule that create-volume returns only if the volume has been created and is healthy. Otherwise waits indefinitely.

paulyoong (approved):

jkryl:
bors merge

jkryl:
bors retry

PR #856, opened 2021-04-26
fix(mayastor): replicas are reported incorrectly for multiple pools

cjones1024:
BUG
If mayastore has multiple pools defined then, in the mayastore YAML config file,
all defined replicas are reported multiple times as belonging to each and every pool.

FIX
Each replica is now reported precisely once under the pool to which it belongs.


jkryl (approved):

tiagolobocastro (approved):

cjones1024:
bors merge

PR #857, opened 2021-04-26
feat(core): add spawn_local() for threads

gila:
This removes the local macro we have been using for a while and
replaces it with a submission function. Although this looks rather
pointless -- this enables us to run an async runtime in separate
threads and communication is done through rte_ring.

Additionally this will also allow us to run any functions along side
mayastor that does, for example, health checking etc.

IOW, lets say you have core mask 0x1 -- we will allow tokio to create
its own threads and unaffinitze them, and then use this interface to
spawn futures there who are !Send nor !Sync. The spawn_local() function
added here, returns a futures::oneshot::Receiver<T> which can be awaited
on threads that run somewhere but not on the specified mask.

Clippy started to point out that certain asserts should be written
differently so thats part of this PR too.


composer/src/lib.rs
```diff
@@ -677,10 +677,17 @@ impl Drop for ComposeTest {
     fn drop(&mut self) {
         if thread::panicking() && self.logs_on_panic {
             self.containers.keys().for_each(|name| {
-                tracing::error!("Logs from container '{}':", name);
+                tracing::error!(
```
tiagolobocastro, line 680: thanks, this will be useful to find the needle in the haystack :+1:

mayastor/src/bin/mayastor.rs
```diff
@@ -4,22 +4,42 @@ extern crate tracing;
...
 use structopt::StructOpt;
 mayastor::CPS_INIT!();
+
+fn start_tokio_runtime(args: &MayastorCliArgs) {
+    let grpc_endpoint = grpc::endpoint(args.grpc_endpoint.clone());
+    let rpc_address = args.rpc_address.clone();
+
+    Mthread::spawn_unaffinitized(move || {
+        let rt = tokio::runtime::Builder::new_multi_thread()
+            .worker_threads(4)
```
tiagolobocastro, line 22: Does this mean that we now have >1 gRPC requests running at the same time?

mayastor/src/grpc/mod.rs
```diff
@@ -1,47 +1,42 @@
...
-            Ok(res) => res,
-            Err(err) => {
-                error!("{}", crate::grpc::print_error_chain(&err));
-                return Err(err.into());
-            }
-        }
-    }};
+impl From<CoreError> for tonic::Status {
+    fn from(e: CoreError) -> Self {
+        let e = e;
```
tiagolobocastro, line 36: Is this needed?

mayastor/tests/block_device_nvmf.rs
```diff
@@ -1679,7 +1679,7 @@ async fn nvmf_reset_abort_io() {
 
     // Sleep for a few seconds to let all I/O operations be aborted.
     println!("Sleeping for 1 sec to let reset hit I/O operations");
-    tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+    tokio::time::sleep(std::time::Duration::from_secs(2)).await;
```
tiagolobocastro, line 1682: should we update the println above?

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -49,93 +120,282 @@ impl mayastor_server::Mayastor for MayastorSvc {
...
         }
 
-        sync_config(pool_grpc::create(args)).await
+        let rx = rpc_submit::<_, _, LvsError>(async move {
+            let pool = Lvs::create_or_import(args).await?;
+
+            Ok(Pool::from(pool))
+        })?;
+
+        rx.await
```
tiagolobocastro, line 129: this pattern seems to repeat a lot, could it turned into a fn's or something?

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -49,93 +120,282 @@ impl mayastor_server::Mayastor for MayastorSvc {
             return Err(Status::invalid_argument("Missing devices"));
         }
 
-        sync_config(pool_grpc::create(args)).await
+        let rx = rpc_submit::<_, _, LvsError>(async move {
```
tiagolobocastro, line 123: Do we need to update the config file on pool create/destroy operations?

mayastor/src/bin/mayastor.rs
gila: Yes, but we always had that as .await will yield.

mayastor/src/grpc/mayastor_grpc.rs
gila: Yes, we can but, I want to keep it explicit for now because I want to make it clear that we are going between the two loops.

mayastor/src/grpc/mayastor_grpc.rs
gila: Yes, in fact, we only need to go over this for "pools" but that is covered in a separate ticket. I'm just doing half the job already =)

mtzaurus (approved):

mayastor/src/grpc/mayastor_grpc.rs
tiagolobocastro: Ah ok, though I'd recommend not merging this one until the other ticket's PR comes around otherwise develop will not work with pod restarts in k8s

mayastor/src/bin/mayastor.rs
tiagolobocastro: Ah you're right, I was misremembering things

tiagolobocastro (approved):

PR #865, opened 2021-05-03
feat(csi): ctrl_loss_tmo should be set to ioTimeout from SC

jkryl:
When the node with nexus target is down or isn't accessible for any
reason, one have to wait for 10-15 minutes with default linux settings
until the IO error is propagated back to the consumer of the volume
(i.e. app or filesystem). By setting ctrl_loss_tmo to ioTimeout from
storage class, we give the control over this to user. If ioTimeout is
not set in SC then default linux settings apply.

Resolves: CAS-823

### DIFF
--- a/csi/src/dev/iscsi.rs
+++ b/csi/src/dev/iscsi.rs
@@ -116,6 +116,13 @@ impl TryFrom<&Url> for IscsiDevice {
 
 #[tonic::async_trait]
 impl Attach for IscsiAttach {
+    async fn parse_parameters(
+        &mut self,
+        _context: &HashMap<String, String>,
+    ) -> Result<(), DeviceError> {
+        Ok(())
+    }
+
     async fn attach(&self) -> Result<(), DeviceError> {
         match IscsiAdmin::find_session(&self.portal, &self.iqn) {
             Ok(found) => {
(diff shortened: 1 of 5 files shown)


csi/src/dev/iscsi.rs
```diff
@@ -116,6 +116,13 @@ impl TryFrom<&Url> for IscsiDevice {
 
 #[tonic::async_trait]
 impl Attach for IscsiAttach {
+    async fn parse_parameters(
```
tiagolobocastro, line 119: Do we need to set the same timeout for iscsi?

jkryl: ioTimeout setting in storage class is valid only for nvmf protocol. If set for another protocol then create volume request fails. So the answer is that we don't need to set it for iscsi because it's not supported. However that might change in the future if there is a motivation to improve iscsi for whatever reason.

jonathan-teh (approved):
Looks fine, the commit message should be `ctrl_loss_tmo` though. Indentation looks off from line 442, the linter should pick that up.

jkryl:
Thanks @jonathan-teh . I have fixed the commit message. As for the formatting, apparently that's how rust fmt/clippy like it. I ran them again just to verify that it passes.

csi/src/dev/iscsi.rs
tiagolobocastro: Ah ok, is it documented that the ioTimeout is only valid for nvmf?

tiagolobocastro (approved):

jkryl:
bors merge

csi/src/dev/iscsi.rs
jkryl: ioTimeout parameter as such isn't documented. It should be documented in our gitbook. good point 👍

Issue #822, opened 2021-03-30
Mounted Volume permisions not set for fsGroup


bbroniewski:
**Describe the bug**
After volume creation Kubernetes should change owner of the volume directories to group specified by fsGroup in pod "spec" field, but it is not happening. The same works fine with cStor engine.

**To Reproduce**
My cluster setup is:
-Flannel CNI - Layer 2
-Metallb - Load Balancer
-Openebs Mayastor with disk attached to the device using protocol iscsi
-One node

Configuration of default StorageClass and Mayastor tested using "Fio" pod.

During installation of Kyma, resource "ory" initiates a postgresql database with pvc which require 8GB pv.
Mayastor assigns PV and all Mayastor objects are healthy.

Docker container is trying to create directory using "mkdir" command and it is failing:
"mkdir: can't create directory '/data/pgdata': Permission denied"

**Expected behavior**
Directory should be created, because mounted directories should have group specified in "fsGroup" as owner.

**Screenshots**
![image](https://user-images.githubusercontent.com/11493147/113049624-249f0f00-91a4-11eb-9ee7-6d0ba60af57d.png)
![image](https://user-images.githubusercontent.com/11493147/113049701-3d0f2980-91a4-11eb-989f-84d260db7a0d.png)

** OS info (please complete the following information):**
Ubuntu Server 20.04
 - Kernel version 5.4.0-70-generic
 - MayaStor revision 0.8

**Additional context**
Link to the thread on slack:
https://kubernetes.slack.com/archives/CUAKPFU78/p1616749885288800

jkryl:
@bbroniewski thanks for the ticket! We need to find out why securityContext fsGroup does not work with mayastor. It has been verified to work with cstor. The investigation is in progress.

kschzt:
I have a very similar issue with TimescaleDB (PostgreSQL), the fsGroup is not honored and the pod fails trying to create the /var/lib/postgresql/data folder.

payes:
hi @kschzt @bbroniewski , can you please try adding fsType in parameters section of storage class. Something similar to:
```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: mayastor-nvmf
parameters:
  repl: '3'
  protocol: 'nvmf'
  fsType: 'xfs'
provisioner: io.openebs.csi-mayastor
```

kschzt:
@payes Happy to confirm that setting `fsType` worked! Thanks!

[...]

PR #866, opened 2021-05-04
fix(fstype): add default fstype to csi provisioner container

payes:
fixes: https://github.com/openebs/Mayastor/issues/822

Currently application deployments on mayastor volumes are not adhering to fsGroups, 
due to which the mount points are not always writable. This is because there is no 
fsType set by default.   fsType should either be mentioned in the storage class or 
should be set by default in csi provisioner container. fsType defined in Storage class 
overrides this default.

https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/csi/csi_mounter.go#L391

Signed-off-by: Payes Anand <payes.anand@mayadata.io>

### DIFF
--- a/deploy/storage-class.yaml
+++ b/deploy/storage-class.yaml
@@ -7,6 +7,8 @@ parameters:
   repl: '1'
   protocol: 'iscsi'
   local: 'yes'
+  # It is recommended to use xfs for Mayastor
+  # fsType: 'xfs'
 provisioner: io.openebs.csi-mayastor
 volumeBindingMode: WaitForFirstConsumer
 ---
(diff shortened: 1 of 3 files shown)


jkryl (approved):

deploy/storage-class.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
 parameters:
   repl: '1'
   protocol: 'iscsi'
+  # It is reconnecded to use xfs for Mayastor
```
jonathan-teh, line 8: Typo: recommended?

jonathan-teh (changes requested):

deploy/storage-class.yaml
payes: thanks @jonathan-teh

jonathan-teh (approved):

jonathan-teh:
Please squash your commits (you'll need to resolve the merge conflict anyway).

gila:
thanks @payes !

PR #868, opened 2021-05-04
fix(moac): propagate errors from FSA to callers

jkryl:
This solves the problem when nexus publish fails because create nexus
grpc call fails, but the publish operation is still waiting for
completing the publish. There is nothing that would trigger a new run
of FSA so it waits indefinitely (until k8s unpublishes the volume).
The same problem applies to create, unpublish and destroy operations
too. We introduce a new delegation mechanism for notifying blocked
callers about errors that may happen in FSA. FSA itself is responsible
for interpretting failures and notifying potential consumers.

Another improvement is that if the failure does not have any consumer
so we cannot rely for example on k8s to retry the operation, there is
a timer 30s that triggers a new run on fsa, so the operation will not
be stuck forever.

Note that if the operation fails (i.e. publish) and the error is
returned through gRPC reply, the desired state change is not undone
(i.e. publish will be retried upon next run of fsa).
That is done:
 * to keep the code simple
 * it is compatible with k8s behaviour (as it will anyway retry the
   operation anyway) and
 * avoids situations when the rollback of desired state could
   interfere with other state changes and triggering more state
   changes.

node-black-list has been removed to simplify the code. I doubt its
usefulness and it wasn't covered by the tests. The only code where it
was useful, when adding replicas to a nexus, was changed to work even
without that list. We try to add all replicas. The failures aren't
treated as fatal and continue the fsa execution afterwards.

Resolves: CAS-870

### DIFF
--- a/csi/moac/volume.ts
+++ b/csi/moac/volume.ts
@@ -12,6 +12,20 @@ import { Node } from './node';
 const log = require('./logger').Logger('volume');
 const { GrpcCode, GrpcError } = require('./grpc_client');
 
+// If state transition in FSA fails due to an error and there is no consumer
+// for the error, we set a retry timer to retry the state transition.
+const RETRY_TIMEOUT_MS = 30000;
+
+type DoneCallback = (err?: Error, res?: unknown) => void;
+
+// ID of the operation delegated to fsa() to perform.
+enum DelegatedOp {
+  Create,
+  Publish,
+  Unpublish,
+  Destroy,
+}
+
 // State of the volume
 export enum VolumeState {
   Unknown = 'unknown',
@@ -81,10 +95,10 @@ export class Volume {
   private publishedOn: string | undefined;
   // internal properties
   private emitter: events.EventEmitter;
-  private changeCallbacks: (() => void)[];
   private registry: any;
   private runFsa: number; // number of requests to run FSA
-  private nodeBlackList: Record<string, boolean>; // replicas on these nodes should be avoided
+  private waiting: Record<DelegatedOp, DoneCallback>; // ops waiting for completion
+  private retry_fsa: NodeJS.Timeout | undefined;
 
   // Construct a volume object with given uuid.
   //
@@ -116,9 +130,17 @@ export class Volume {
     this.state = state || VolumeState.Pending;
     // other properties
     this.runFsa = 0;
-    this.nodeBlackList = {};
     this.emitter = emitter;
-    this.changeCallbacks = [];
+    this.waiting = <Record<DelegatedOp, DoneCallback>> {};
+  }
+
+  // Clear the timer on the volume to prevent it from keeping nodejs loop alive.
+  deactivate() {
+    this.runFsa = 0;
+    if (this.retry_fsa) {
+      clearTimeout(this.retry_fsa);
+      this.retry_fsa = undefined;
+    }
   }
 
   // Stringify volume
@@ -141,7 +163,8 @@ export class Volume {
   getReplicas(): Replica[] {
     return Object.values(this.replicas);
   }
-  // Publish the volume. That means, make it accessible through a block device.
+
+  // Publish the volume. That means, make it accessible through a target.
   //
   // @params nodeId        ID of the node where the volume will be mounted.
   // @return uri           The URI to access the nexus.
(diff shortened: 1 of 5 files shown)


csi/moac/volume.ts
```diff
@@ -153,94 +176,83 @@ export class Volume {
...
-          log.info(`Published "${self}" at ${uri}`);
-          return resolve(uri);
-        }
-      }
-      self._onStateChange(waitForPublish.bind(self, resolve, reject));
-    }
-    return new Promise((resolve, reject) => {
-      waitForPublish(resolve, reject);
-    });
+    // Cancel any unpublish that might be in progress
```
tiagolobocastro, line 204: why do we cancel the unpublish and not a previous publish?

csi/moac/volume.ts
```diff
@@ -383,18 +474,16 @@ export class Volume {
...
-      return (!r.isOffline() &&
-        !childReplicaPairs.find((pair) => pair.r === r) &&
-        !this.nodeBlackList[r.pool!.node!.name]);
+      return (!r.isOffline() && !childReplicaPairs.find((pair) => pair.r === r));
     });
     for (let i = 0; i < newReplicas.length; i++) {
       try {
-        await this.nexus.addReplica(newReplicas[i]);
-        return;
+        childReplicaPairs.push({
```
tiagolobocastro, line 481: looks like we now add all replicas at once, does this cause multiple concurrent rebuilds?

tiagolobocastro (approved):

csi/moac/volume.ts
```diff
@@ -791,7 +971,12 @@ export class Volume {
   }
 
   // Share replicas as appropriate to allow access from the nexus.
-  async _ensureReplicaShareProtocols(nexusNode: Node, replicaSet: Replica[]) {
+  // It does not throw unless none of the replicas can be shared.
+  // Though replicas that failed to be shared are added to the blacklist.
```
paulyoong, line 975: Need to update the comment so that we don't refer to "blacklist" if we are no longer using it.

paulyoong (approved):

csi/moac/volume.ts
```diff
@@ -12,6 +12,20 @@ import { Node } from './node';
 const log = require('./logger').Logger('volume');
 const { GrpcCode, GrpcError } = require('./grpc_client');
 
+// If state transition in FSA fails due to an error and there is no consumer
+// for the error, we set a retry timer to retry the state transition.
+const RETRY_TIMEOUT = 30000;
```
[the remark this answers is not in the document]
jkryl: it is in ms. I can add unit suffix to the name if you think it would be a good idea.

csi/moac/volume.ts
```diff
@@ -12,6 +12,20 @@ import { Node } from './node';
...
+
+type DoneCallback = (err?: Error, res?: unknown) => void;
+
+// ID of the operation delegated to fsa() to perform.
+enum DelegatedOp {
+  Create,
+  Publish,
+  Unpublish,
+  Destroy,
+}
```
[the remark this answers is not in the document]
jkryl: Indeed, that's the reason. We are interested only in operations that are triggered by external csi call.

csi/moac/volume.ts
jkryl: it's up to mayastor to serialize rebuilds if it needs to do that. And indeed it does that. In the future it could do rebuilds in parallel. There has been nothing to enforce just one rebuild at a time even before. Something could have triggered fsa again resulting in a new addChild call. So I don't think the situation is worse than it was in any regard.

csi/moac/volume.ts
jkryl: there cannot be a previous publish. All CSI calls of given type for given volume are serialized. We cancel unpublish because it might be in progress, stuck or waiting for something.

csi/moac/volume.ts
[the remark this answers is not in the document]
tiagolobocastro: Yes that'd be good I think :+1:

csi/moac/volume.ts
tiagolobocastro: hmm AFAIK mayastor does not serialize rebuilds. I remember from a while back that MOAC would not add a child to the nexus if the nexus was degraded, so this must have changed before. Multiple rebuilds should work but I do remember that we did not make those tests extensive because we were told that MOAC would not do such thing. If this is now possible then we might need to add more rebuild tests to validate this.

jkryl: It was probably a misunderstanding then. MOAC would not delete a child while a rebuild is in progress. Though AFAIK there has not been such rule for adding child. I have created the ticket CAS-890 for missing tests 👍  thanks for pointing it out.

tiagolobocastro: Ah, I see, then definitely a misunderstanding!
We used to have a rebuild multiple on mayastor but it's been removed, would be good to bring that one back, I'll update that ticket to mention it.

jkryl:
bors merge

jkryl:
bors merge

jkryl:
bors merge

PR #869, opened 2021-05-06
fix: disregard msp spec modifications

tiagolobocastro:
When MOAC restarts it checks the pool resources and may then attempt to
create pools based on specs, which could cause it to recreate pools on
different nodes for instance, if the user had modified the SPEC.

This change "caches" the original spec in the status part of the
resource which then MOAC refers to, rather then the resource SPEC.
Also, if a disk URI (retrieved from mayastor) is available, always use that one.

If a SPEC change is detected, then tag a message on the MSP reason
alerting the user to the fact that it will be ignored. It is removed
if the user reverts the SPEC to the original.

Resolves CAS-855

### DIFF
--- a/csi/moac/pool_operator.ts
+++ b/csi/moac/pool_operator.ts
@@ -57,17 +57,27 @@ function poolStateFromString(val: string): PoolState {
   }
 }
 
+// Object defines spec properties of a pool resource.
+export class PoolSpec {
+  node: string;
+  disks: string[];
+
+  // Create and validate pool custom resource.
+  constructor(node: string, disks: string[]) {
+    this.node = node;
+    this.disks = disks;
+  }
+}
+
 // Object defines properties of pool resource.
 export class PoolResource extends CustomResource {
   apiVersion?: string;
   kind?: string;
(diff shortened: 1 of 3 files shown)


jkryl (approved):
Nice! 🙇‍♂️

tiagolobocastro:
bors try

csi/moac/pool_operator.ts
```diff
@@ -289,7 +322,23 @@ export class PoolOperator {
   //
   async _createPool (resource: PoolResource) {
     const name: string = resource.getName();
-    const nodeName = resource.spec.node;
+    const nodeName = resource.getSpec().node;
+
+    // Nothing prevents the user from modifying the spec part of the CRD, which could trick MOAC into recreating
+    // the pool in a different node for example.
```
paulyoong, line 328: typo: in to on

paulyoong (approved):

csi/moac/pool_operator.ts
tiagolobocastro: Done

tiagolobocastro:
bors merge

PR #870, opened 2021-05-06
feat(nexus): acquire write exclusive reservation on adding a child

jonathan-teh:
Refactor the existing reservation key code such that both code paths
that add a child call NexusChild::acquire_write_exclusive(), which
sends an NVMe Reservation Acquire command to acquire a write exclusive
reservation on the child, then a Reservation Report command to verify
that the reservation has been acquired, followed by with another
Reservation Acquire command to preempt an existing reservation, if
another host has it.

Add a test to verify that, refactoring the existing nexus_multipath
test as well as disabling it as the write exclusive reservation means
only one Nexus can write to a replica.

Fixes CAS-850

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_bdev_children.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev_children.rs
@@ -178,20 +178,21 @@ impl Nexus {
             self.name.clone(),
             Some(child_bdev),
         );
-        match child.open(self.size) {
-            Ok(name) => {
-                // we have created the bdev, and created a nexusChild struct. To
-                // make use of the device itself the
-                // data and metadata must be validated. The child
-                // will be added and marked as faulted, once the rebuild has
-                // completed the device can transition to online
-                info!("{}: child opened successfully {}", self.name, name);
+        let mut child_name = child.open(self.size);
+        if let Ok(ref name) = child_name {
+            // we have created the bdev, and created a nexusChild struct. To
+            // make use of the device itself the
+            // data and metadata must be validated. The child
+            // will be added and marked as faulted, once the rebuild has
+            // completed the device can transition to online
+            info!("{}: child opened successfully {}", self.name, name);
 
-                // FIXME: use dummy key for now
-                if let Err(e) = child.resv_register(0x12345678).await {
-                    error!("Failed to register key with child: {}", e);
-                }
-
+            if let Err(e) = child.acquire_write_exclusive().await {
+                child_name = Err(e);
+            }
+        }
+        match child_name {
+            Ok(_) => {
                 // it can never take part in the IO path
                 // of the nexus until it's rebuilt from a healthy child.
                 child.fault(Reason::OutOfSync).await;
@@ -487,14 +488,31 @@ impl Nexus {
             });
         }
 
-        // FIXME: use dummy key for now
+        // acquire a write exclusive reservation on all children,
+        // if any one fails, close all children.
+        let mut we_err: Result<(), Error> = Ok(());
         for child in self.children.iter() {
-            if let Err(error) = child.resv_register(0x12345678).await {
-                error!(
-                    "{}: failed to register key {} for child {}",
-                    self.name, child.name, error
-                );
+            if let Err(error) = child.acquire_write_exclusive().await {
+                we_err = Err(Error::ChildWriteExclusiveResvFailed {
+                    source: error,
+                    child: child.name.clone(),
+                    name: self.name.clone(),
+                });
+                break;
             }
+        }
+        if let Err(error) = we_err {
+            for child in &mut self.children {
+                if let Err(error) = child.close().await {
+                    error!(
+                        "{}: child {} failed to close with error {}",
+                        self.name,
+                        &child.name,
+                        error.verbose()
+                    );
+                }
+            }
+            return Err(error);
         }
 
         for child in self.children.iter() {
--- a/mayastor/src/bdev/nexus/nexus_child.rs
+++ b/mayastor/src/bdev/nexus/nexus_child.rs
@@ -21,16 +21,25 @@ use crate::{
         VerboseError,
     },
     core::{
+        nvme_reservation_acquire_action,
+        nvme_reservation_register_action,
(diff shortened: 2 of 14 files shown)


jonathan-teh:
bors try

jonathan-teh:
bors try

jonathan-teh:
bors try

mayastor/src/bdev/nvmx/handle.rs
```diff
@@ -1045,6 +1096,13 @@ impl BlockDeviceHandle for NvmeDeviceHandle {
             )
         };
 
+        if rc < 0 {
+            return Err(CoreError::NvmeIoPassthruDispatch {
+                source: Errno::from_i32(rc.abs()),
```
gila, line 1101: Can you make use of the `to_result()` from ffiresult.rs?

jonathan-teh: Sure, I'd forgotten that that exists.

mayastor/tests/nexus_multipath.rs
```diff
@@ -262,14 +278,142 @@ async fn nexus_multipath() {
...
         .unwrap();
-    assert!(
-        output_dis2.status.success(),
-        "failed to disconnect from remote replica, {}",
-        output_dis2.status
+
+    let mut hdls = test.grpc_handles().await.unwrap();
+
+    // create a pool on remote node 1
+    hdls[0]
```
gila, line 302: Its not certain that this is "node1" -- the grpc_handles() function may return a different order. If you need to be sure it's node1 (as in the first ms instance)  -- it's better to get a handle by name. Do you know for sure the order here does not matter?

jonathan-teh: I hadn't realised that. The node order indeed doesn't matter, as long as they are referred to consistently. I'll clarify it one way or another, either by getting the handle by name or in the comments.

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -270,21 +290,179 @@ impl NexusChild {
...
+    }
+
+    /// Get NVMe reservation report
+    /// Returns: (key, host id) of write exclusive reservation holder
+    async fn resv_report(
+        &self,
+        hdl: &dyn BlockDeviceHandle,
+    ) -> Result<Option<(u64, [u8; 16])>, ChildError> {
+        let mut buffer = hdl.dma_malloc(4096).context(HandleDmaMalloc {})?;
+        if let Err(e) = hdl.nvme_resv_report(1, &mut buffer).await {
```
mtzaurus, line 349: Would it be possible to wrap reservation data returned by 'nvme_resv_report' by a high-level data structure, to hide low-level manipulations with reservation info, like:
'buffer.as_slice().split_at(std::mem::size_of::<spdk_nvme_reservation_status_extended_data'),
ad expose only a well-defined API to access reservation info (like status, ) via this structure ?
That would let us keep NexusChild high-level enough from SPDK API and interact with the underlying block device via well-defined API

mayastor/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -476,13 +475,9 @@ impl Nexus {
...
 
-        // FIXME: use dummy key for now
         for c in &self.children {
-            if let Err(e) = c.resv_register(0x12345678).await {
-                error!(
-                    "{}: child {} failed to register key {}",
-                    self.name, c.name, e
-                );
+            if let Err(e) = c.acquire_write_exclusive().await {
+                error!("Failed to acquire write exclusive reservation on child: {}", e);
```
mtzaurus, line 480: What is our fallback scenario here if we fail to acquire exclusive access ? Do we keep using such a child ? Won't this lead to data corruption if the child is being used by other nexus ?

mayastor/src/bdev/nexus/nexus_child.rs
jonathan-teh: I agree that this function is a little low-level for `NexusChild` but it's only used internally and partly done that way for expediency. If I were to move the various manipulations to `nvme_resv_report()`, then it'd probably return something like a tuple of `Options` of `spdk_nvme_reservation_status_extended_data` and and array of `spdk_nvme_registered_ctrlr_extended_data`. Is that what you had in mind?

mayastor/src/bdev/nexus/nexus_bdev_children.rs
jonathan-teh: Currently we do blunder ahead and attempt to use it but if the write exclusive reservation is held by another host then all writes will fail and it'll be unusable. I'll change it to return an error.

mayastor/src/bdev/nexus/nexus_child.rs
mtzaurus: I was suggesting wrapping reservation info into some high-level structure(s), to let API users see only that placeholder.
If that takes longer to properly wrap it or it ends up in lots of unnecessary Options - then we can leave it as it is.

mayastor/src/bdev/nexus/nexus_bdev_children.rs
jonathan-teh: I'm going to refactor this slightly. Currently `NexusChild::acquire_write_exclusive()` is called from `Nexus::try_open_children()` and `Nexus::add_child_only()`, and I think it's better to call it from `NexusChild::open()` directly to cover both cases so if the write exclusive reservation fails, it'll fail as part of opening the child and allow the existing code to handle clean-ups. This does also mean it'll try to acquire the write exclusive reservation in `NexusChild::online()` but I think that should be fine and onlining a child is only possible via gRPC (which in turn that may be dead code).

mayastor/src/bdev/nexus/nexus_child.rs
jonathan-teh: I've opted to leave it alone for now.

jonathan-teh:
bors try

cjones1024 (approved):

gila (approved):

jonathan-teh:
bors merge

PR #873, opened 2021-05-10
feat(core): provide handle to external runtime

gila:
Mayastor has its own futures which are driven by polling of the "green" threads.
If however, we want to do something which does not need or can run on the
mayastor cores, we can dispatch it to a foreign runtime.

Additionally this also fixes a bug where we where interacting with the nvmf
target from an invalid thread.

### DIFF
--- a/mayastor/src/bin/mayastor.rs
+++ b/mayastor/src/bin/mayastor.rs
@@ -3,7 +3,7 @@ extern crate tracing;
 use futures::future::FutureExt;
 use mayastor::{
     bdev::util::uring,
-    core::{MayastorCliArgs, MayastorEnvironment, Mthread, Reactors},
+    core::{runtime, MayastorCliArgs, MayastorEnvironment, Mthread, Reactors},
     grpc,
     logger,
     subsys,
@@ -11,7 +11,7 @@ use mayastor::{
 use std::path::Path;
 use structopt::StructOpt;
 mayastor::CPS_INIT!();
-use mayastor::subsys::{message_bus_init, Registration};
+use mayastor::subsys::Registration;
 
 fn start_tokio_runtime(args: &MayastorCliArgs) {
     let grpc_address = grpc::endpoint(args.grpc_endpoint.clone());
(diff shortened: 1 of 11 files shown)


mayastor/src/bin/mayastor.rs
```diff
@@ -24,29 +24,24 @@ fn start_tokio_runtime(args: &MayastorCliArgs) {
...
-            debug!("mayastor mbus subsystem init");
-            message_bus_init();
-            mbus_api::message_bus_init_tokio(endpoint);
-            Registration::init(&node_name, &grpc_address.to_string());
-        }
+        runtime::block_on(async move {
+            let mut futures = Vec::new();
+            if let Some(endpoint) = endpoint {
+                debug!("mayastor mbus subsystem init");
+                message_bus_init();
```
tiagolobocastro, line 31: I meant to ask you about this before, do we need to call these message bus inits? IIRC the registration does the init should do the message bus stuff for us.

mayastor/src/bin/mayastor.rs
```diff
@@ -24,29 +24,24 @@ fn start_tokio_runtime(args: &MayastorCliArgs) {
...
+                Registration::init(&node_name, &grpc_address.to_string());
+                futures.push(subsys::Registration::run().boxed());
+            }
 
-        let futures = vec![
-            subsys::Registration::run().boxed_local(),
-            grpc::MayastorGrpcServer::run(grpc_address, rpc_address)
-                .boxed_local(),
-        ];
+            let futures =
```
tiagolobocastro, line 37: aren't we overwriting the previous contents of futures?

paulyoong:
> Mayastor has its own futures, which are driven by polling of the "green."
> threads. If, however, we want to do something which does not need or can
> run on the mayastor cores, we can dispatch it to a foreign runtime.
> 
> This also fixes a bug in create_replica() when the replica we wanted to create already exists. If this happened, then prior to this commit, the system would panic.

I think the comment is meant to say "...which does not need or **cannot** run on the mayastor cores..."

mayastor/src/bin/mayastor.rs
gila: I don't know, to be honest, that code has 3 different init functions and is rather confusing, to be honest. I don't understand why it cants simply be one future that does all the work and be done with it :)

gila:
bors try

gila:
bors try

gila:
bors try

gila:
Bors r+

gila:
Bors merge

PR #875, opened 2021-05-10
fix: add error message when we fail to send registration hb

tiagolobocastro:
NATS messages get queued by the nats client and so we may not get an error when we try
to send a message. So, attempt to flush the queue after each send with a default timeout
of 5 seconds. If the flush fails then we'll log the error. Subsequent sends will then be
failed by the NATS client until such a time connection is reestablished.

Also add connect/disconnect logging to the NATS server.


tiagolobocastro:
bors try

paulyoong (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #878, opened 2021-05-11
fix(moac): fix double callback in controllerPublish method

jkryl:
Callback was called twice which resulted into strange errors about
unhandled exceptions in grpc lib being printed into the log.
As part of troubleshooting I have rewritten grpc_client to
typescript to catch more errors during the compilation.


tiagolobocastro (approved):

jkryl:
bors merge

PR #880, opened 2021-05-12
fix(nexus): retire replicas in response to I/O errors

mtzaurus:
1. Added support of fail-fast I/O completion for proper I/O error propagation.
2. Concurrent Nexus retire operations are now serialised and get executed
   sequentially in case different replicas experience I/O errors at the same
   time.
3. I/O submission errors can now also trigger device retire (solves edge cases
   when I/O channels are dropped concurrently by reset operation, triggered
   by I/O-unrelated operations (admin queue timeouts, etc).
4. Fixed a few misbehaviours in generic retire logic (like retiring healthy
   replicas in response to I/O errors from the faulty replica).

Implements CAS-868, CAS-865, CAS-866

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_bdev.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev.rs
@@ -17,27 +17,28 @@ use serde::Serialize;
 use snafu::{ResultExt, Snafu};
 use tonic::{Code, Status};
 
+use crate::core::IoDevice;
+
 use rpc::mayastor::NvmeAnaState;
-use spdk_sys::{
-    spdk_bdev,
-    spdk_bdev_register,
-    spdk_bdev_unregister,
-    spdk_io_device_register,
-    spdk_io_device_unregister,
-};
+use spdk_sys::{spdk_bdev, spdk_bdev_register, spdk_bdev_unregister};
 
 use crate::{
     bdev::{
         nexus,
         nexus::{
             instances,
-            nexus_channel::{DrEvent, NexusChannel, ReconfigureCtx},
+            nexus_channel::{
+                DrEvent,
+                NexusChannel,
+                NexusChannelInner,
+                ReconfigureCtx,
+            },
             nexus_child::{ChildError, ChildState, NexusChild},
             nexus_label::LabelError,
             nexus_nbd::{NbdDisk, NbdError},
         },
     },
-    core::{Bdev, CoreError, IoType, Protocol, Reactor, Share},
+    core::{Bdev, CoreError, Cores, IoType, Protocol, Reactor, Share},
     ffihelper::errno_result_from_i32,
     nexus_uri::{bdev_destroy, NexusBdevError},
     rebuild::RebuildError,
@@ -173,6 +174,8 @@ pub enum Error {
     },
     #[snafu(display("Child {} of nexus {} not found", child, name))]
     ChildNotFound { child: String, name: String },
+    #[snafu(display("Failed to pause child {} of nexus {}", child, name))]
+    PauseChild { child: String, name: String },
     #[snafu(display("Suitable rebuild source for nexus {} not found", name))]
     NoRebuildSource { name: String },
     #[snafu(display(
@@ -302,6 +305,13 @@ pub enum NexusTarget {
     NexusIscsiTarget,
     NexusNvmfTarget,
 }
+#[derive(Debug, Eq, PartialEq)]
+enum NexusPauseState {
+    Unpaused,
+    Pausing,
+    Paused,
+    Unpausing,
+}
 
 /// The main nexus structure
 #[derive(Debug)]
@@ -327,6 +337,11 @@ pub struct Nexus {
     pub(crate) share_handle: Option<String>,
     /// enum containing the protocol-specific target used to publish the nexus
     pub nexus_target: Option<NexusTarget>,
+    /// Nexus I/O device.
+    pub io_device: Option<IoDevice>,
+    /// Nexus pause counter to allow concurrent pause/resume.
+    pause_state: NexusPauseState,
+    pause_waiters: Vec<oneshot::Sender<i32>>,
 }
 
 unsafe impl core::marker::Sync for Nexus {}
@@ -386,6 +401,45 @@ impl Drop for Nexus {
     }
 }
 
+struct UpdateFailFastCtx {
+    increment: bool,
+    sender: oneshot::Sender<bool>,
+    nexus: String,
+}
+
+fn update_failfast_cb(
+    channel: &mut NexusChannelInner,
+    ctx: &mut UpdateFailFastCtx,
+) -> i32 {
+    let old_value = channel.fail_fast;
+
+    if ctx.increment {
+        channel.fail_fast = channel
+            .fail_fast
+            .checked_add(1)
+            .expect("Fail-fast counter overflow");
+    } else {
+        channel.fail_fast = channel
+            .fail_fast
+            .checked_sub(1)
+            .expect("Fail-fast counter underflow");
+    }
+    info!(
+        "{}: fail-fast counter transition: {} -> {}",
+        ctx.nexus, old_value, channel.fail_fast
+    );
+    0
+}
+
+fn update_failfast_done(status: i32, ctx: UpdateFailFastCtx) {
+    info!(
+        "{}: Fail-fast counter update completed, increment={}, status={}",
+        ctx.nexus, ctx.increment, status
+    );
+
+    ctx.sender.send(true).expect("Receiver disappeared");
+}
+
 impl Nexus {
     /// create a new nexus instance with optionally directly attaching
     /// children to it.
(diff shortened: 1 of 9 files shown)


mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -516,54 +596,88 @@ impl NexusBio {
...
+                                    if let Err(err) = child.destroy().await {
+                                        error!(
+                                            "{}: destroying child {} failed {}",
+                                            nexus, child, err
+                                        );
+                                    }
+                                }
+                                None => {
+                                    warn!(
+                                        "{} does not belong (anymore) to nexus {}, skipping child removal",
```
jonathan-teh, line 663: Slightly clearer as "... no longer belongs to nexus ...".

jonathan-teh (commented):
Minor comments for now, will review again later.

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -699,15 +863,12 @@ impl Nexus {
...
-        unsafe {
-            spdk_io_device_register(
-                self.as_ptr(),
-                Some(NexusChannel::create),
-                Some(NexusChannel::destroy),
-                std::mem::size_of::<NexusChannel>() as u32,
-                (*self.bdev.as_ptr()).name,
-            );
-        }
+        let io_device = IoDevice::new::<NexusChannel>(
```
gila, line 866: Nice!

mayastor/src/bdev/nexus/nexus_channel.rs
```diff
@@ -31,6 +31,7 @@ pub(crate) struct NexusChannelInner {
     pub(crate) writers: Vec<Box<dyn BlockDeviceHandle>>,
     pub(crate) readers: Vec<Box<dyn BlockDeviceHandle>>,
     pub(crate) previous: usize,
+    pub(crate) fail_fast: u32,
```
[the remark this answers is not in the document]
mtzaurus: I will add 'checked_sub()' / checked_add() to make sure no underflow/overflow happens.

gila (approved):

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -633,29 +686,140 @@ impl Nexus {
...
+                if let Some(Protocol::Nvmf) = self.shared() {
+                    if let Some(subsystem) =
+                        NvmfSubsystem::nqn_lookup(&self.name)
+                    {
+                        info!(
+                            "{} pausing subsystem {}",
+                            self.name,
+                            subsystem.get_nqn()
+                        );
+                        subsystem.pause().await.unwrap();
```
tiagolobocastro, line 759: while one future enters the await, another can also get there and not the "_" case below?

mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -516,54 +596,88 @@ impl NexusBio {
...
+                                        device, nexus
+                                    );
+                                }
+                            }
+
+                            nexus.clear_failfast().await.unwrap();
+                            nexus.resume().await.unwrap();
+
+                            if nexus.status() == NexusStatus::Faulted {
+                                error!(":{} has no children left... ", nexus);
```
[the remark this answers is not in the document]
gila: I think we want it to be like this as the faulted nexus will cause the application to fail the IO.

mayastor/src/bdev/nexus/nexus_bdev.rs
mtzaurus: Nope.
We get here only if the state is Unpaused and we mark it as Unpausing, so the other futures can reach this branch when we enter await.
Also, we allow executing this function only on the first core always (see asserts in the beginning), so no concurrent executions are possible.

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -633,29 +686,140 @@ impl Nexus {
...
+                assert_eq!(self.pause_state, NexusPauseState::Paused);
             }
         }
 
         Ok(())
     }
 
+    // Abort all active I/O for target child and set I/O fail-fast flag
+    // for the child.
+    async fn update_failfast(&self, increment: bool) -> Result<(), Error> {
```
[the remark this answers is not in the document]
mtzaurus: We should panic if overflow or underflow of the fail-fast counter occurs (as per Jonathan's suggestion).
The operation itself is just increment/decrement the counter.

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -633,29 +686,140 @@ impl Nexus {
         }
     }
 
-    /// resume IO to the bdev
-    pub(crate) async fn resume(&self) -> Result<(), Error> {
+    /// Resume IO to the bdev.
+    /// Note: in order to handle cuncurrent resumes properly, this function must
```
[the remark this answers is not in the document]
mtzaurus: The comment is not 100% correct, as there might be only one resumer - the one, that has successfully paused the nexus.

mtzaurus:
bors try

mayastor/src/bdev/nexus/nexus_bdev.rs
tiagolobocastro: oh I missed the set of the pausing status above, then yes that's correct :+1:

tiagolobocastro (approved):

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -633,29 +686,140 @@ impl Nexus {
...
+            _ => {
+                info!(
+                    "{} concurrent subsystem pause detected, yielding at state: {:?}",
+                    self.name, self.pause_state,
+                );
+
+                let (s, r) = oneshot::channel::<i32>();
+                self.pause_waiters.push(s);
+
+                r.await.expect("Nexus pause sender disappeared");
```
tiagolobocastro, line 780: btw what happens on a nexus destroy, does it wait for the pause to complete? If it does not wait is it possible that while we wait here the nexus has been destroyed already which will drop all the senders?

mayastor/src/bdev/nexus/nexus_channel.rs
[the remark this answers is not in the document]
mtzaurus: Fixed.

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -633,29 +686,140 @@ impl Nexus {
...
+        let io_device = self.io_device.as_ref().expect("Nexus not opened");
+
+        io_device.traverse_io_channels(
+            update_failfast_cb,
+            update_failfast_done,
+            NexusChannel::inner_from_channel,
+            ctx,
+        );
+
+        info!("{}: Updating failfast, increment={}", self.name, increment);
```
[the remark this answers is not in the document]
mtzaurus: Fixed.

mayastor/src/bdev/nexus/nexus_io.rs
mtzaurus: Fixed.

mayastor/src/bdev/nexus/nexus_io.rs
```diff
@@ -245,12 +258,29 @@ impl NexusBio {
         if success {
             self.ctx_as_mut().num_ok += 1;
         } else {
+            error!(
```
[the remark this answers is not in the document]
mtzaurus: Yes, printing out every error is a bit too aggressive. I will remove this error!.

mayastor/src/bdev/nexus/nexus_bdev.rs
mtzaurus: Removing a nexus does not impact consistency between pauser/waiters, as they all already have a valid reference to a nexus object, obtained via 'nexus_lookup()'. I presume we will have some edge cases later when implementing configuration update for a nexus that is being removed concurrently, but our pause/unpause code is unrelated to that.
So what happens is basically that all waiters that have valid references to a Nexus will get awakened, try to update the configuration (which might fail for removed nexuses) and resume the nexus, letting other waiter to repeat the same set of actions.
We definitely need to improve the way that we expose nexus instances via 'nexus_lookup()' and have some sort of reference counted Arc, but from the retire logic perspective, that's a separate thing, as we assume that as long as nexus_lookup() gives us a valid reference to a nexus, all waiters will eventually execute.

tiagolobocastro: I guess it depends on what the code that calls pause is doing right, because pause takes a self. So if you have say:
```rust
nexus.pause().await.unwrap();
nexus.do_this_that(); <- this might panic if the nexus has already been destroyed
```
Though the one I'm most wondering about is if the nexus is destroyed I think all pause_waiters will probably be woken up and might hit the expect on the r.await.
But as you said this not related to pausing, we'll need to have some sort of RFC logic to not drop the nexus if it's still being used by the pause or any other context that is still making use of the nexus.

PR #886, opened 2021-05-17
fix(mayastor): nexus child devices must be registered before they can be closed

cjones1024:
Properly closing a nexus child requires an event listener to be registered
in order to handle the resulting device removal event.

There is a bug in Nexus::try_open_children(), in which a nexus child may be
opened and then closed again before any event listener has been registered,
causing the close() to hang.

The fix is to ensure that the event listener is registered
as soon as a child is successfully opened.


mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -883,11 +886,17 @@ pub async fn nexus_create(
 }
 
 /// Destroy list of child bdevs
-async fn destroy_child_bdevs(name: &str, list: &[String]) {
-    let futures = list.iter().map(String::as_str).map(bdev_destroy);
+async fn destroy_child_bdevs(nexus: &str, list: &[String]) {
+    let futures = list
+        .iter()
+        .map(String::as_str)
+        .filter(|name| Bdev::lookup_by_name(name).is_some())
```
mtzaurus, line 893: Does this properly find our non-bdev NVMx devices ? IIRC, Bdev::lookup_by_name() ends up in 'spdk_bdev_get_by_name()', which can't see our NVMe controllers.

cjones1024: Have replaced with device_lookup(), and also used device_destroy().
There no longer seems to be any valid use for bdev_destroy(), and that should probably be removed.

gila (changes requested):
I'd like to see a test case for this, it should have been there already ofcourse, but A test case where the nexus is destroyed properly when a child fails to connect would help.

cjones1024:
> 
> 
> I'd like to see a test case for this, it should have been there already ofcourse, but A test case where the nexus is destroyed properly when a child fails to connect would help.

This scenario is most easily triggered if a child fails to open by virtue of being too small for the specified nexus size. Have now added a new cargo test for this.

PR #887, opened 2021-05-17
feat(nexus): separate pool configuration from other mayastor configuration

cjones1024:
- Create new PoolConfig struct to represent pool configuration, 
  and remove pool information from existing Config struct
  (resolves CAS-861).
- Add new option to specify YAML file to persist pool configuration
 (resolves CAS-862)..
- Pool configuration is updated (only) when a pool is created or destroyed
  (resolves CAS-864).
- At startup, attempt to create pools specified in pool configuration YAML file.


cjones1024:
bors try

mayastor/src/subsys/config/pool.rs
```diff
@@ -0,0 +1,181 @@
...
+        fs::write(&file, config.as_bytes())
+    }
+
+    /// Export current pool configuration
+    pub async fn export() -> Result<(), std::io::Error> {
+        static MUTEX: Lazy<Mutex<u32>> = Lazy::new(|| Mutex::new(0));
+
+        if let Some(file) = get_config_location() {
+            info!("Saving pool configuration to {}", file);
+            let _guard = MUTEX.lock().await;
```
tiagolobocastro, line 75: The mutex doesn't need to be async here right?

mayastor/src/subsys/config/pool.rs
```diff
@@ -0,0 +1,181 @@
...
+/// Initialise the config file location
+fn init_config_location<P>(file: P)
+where
+    P: AsRef<Path> + Display + ToString,
+{
+    CONFIG_FILE.get_or_init(|| file.to_string());
+}
+
+#[inline]
+/// Return the config file location
```
tiagolobocastro, line 26: Btw, usually the comments go above the "#[]" macros

mayastor/src/subsys/config/pool.rs
```diff
@@ -0,0 +1,181 @@
...
+/// Pools that we create. Future work will include the ability to create RAID0
+/// or RAID5.
+pub struct Pool {
+    /// name of the pool to be created or imported
+    pub name: String,
+    /// bdevs to create outside of the nexus control
+    pub disks: Vec<String>,
+    /// list of replicas (not required, informational only)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub replicas: Option<Vec<Replica>>,
```
tiagolobocastro, line 152: I don't think we need the replicas, they're shared via the properties?

cjones1024: As the comment says, the list of replicas is informational only - it is not used on start up as the pool knows what replicas have been created. This is potentially misleading, so for now the pool config is updated everytime a replica changes in order to maintain consistency. I would like to remove all references to replicas entirely, but I really want to do that as a separate PR in order to make it easier to reinstate should we wish to at some point in the future.

mayastor/src/subsys/config/pool.rs
cjones1024: Somewhat moot as the following write is blocking, but at least the calling thread is not blocked while waiting to acquire the mutex.

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -227,6 +229,7 @@ impl mayastor_server::Mayastor for MayastorSvc {
                 }
                 Ok(lvol) => {
                     debug!("created lvol {}", lvol);
+                    sync_config().await;
```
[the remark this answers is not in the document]
cjones1024: I'm assuming we only want to do the sync when something has actually changed. So, in this case, if there's a failure and nothing gets created, we don't want to update the config file.

mayastor/src/subsys/config/pool.rs
cjones1024: I think I agree, however did a quick search through our code and we aren't consistent at all.

mayastor/src/grpc/mayastor_grpc.rs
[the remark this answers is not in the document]
tiagolobocastro: ah right, then you could test the result.and_then(sync) ? anyway, this is bikeshedding :)

mayastor/src/subsys/config/pool.rs
tiagolobocastro: we shouldn't have any waiter because the the write is blocking, so no other thread should be waiting on the lock because our reactor would be "stuck" on the write.

mayastor/src/subsys/config/pool.rs
tiagolobocastro: :+1:

tiagolobocastro (approved):

mayastor/src/core/env.rs
```diff
@@ -601,6 +608,17 @@ impl MayastorEnvironment {
         cfg.apply();
     }
 
+    /// load the pool config file.
+    fn load_pool_config(&self) -> Option<PoolConfig> {
```
gila, line 612: This should not fail mayastor to start.

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -149,9 +150,9 @@ impl mayastor_server::Mayastor for MayastorSvc {
         let args = request.into_inner();
         let rx = rpc_submit::<_, _, LvsError>(async move {
             if let Some(pool) = Lvs::lookup(&args.name) {
-                let _e = pool.destroy().await?;
+                pool.destroy().await?;
```
gila, line 153: First remove the pool from the config, please. If we fail to destroy the pool and crash halfway through, we at least are sure the pool is not loaded upon restart.

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -212,11 +213,12 @@ impl mayastor_server::Mayastor for MayastorSvc {
                     match lvol.share_nvmf(None).await {
                         Ok(s) => {
                             debug!("created and shared {} as {}", lvol, s);
+                            sync_config().await;
```
gila, line 216: This is not needed, so really no need to sync the config here. Also please consider in what context we are running this in

mayastor/src/grpc/mayastor_grpc.rs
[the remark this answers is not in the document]
gila: This should be removed completely, we store the shared state already within the lvol properties. No need to write to files for this.

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -244,14 +247,13 @@ impl mayastor_server::Mayastor for MayastorSvc {
...
-                Some(b) => {
-                    let lvol = Lvol::try_from(b)?;
-                    lvol.destroy().await.map(|_r| Null {})
-                }
-                None => Ok(Null {}),
+        let rx = rpc_submit::<_, _, LvsError>(async move {
+            if let Some(bdev) = Bdev::lookup_by_name(&args.uuid) {
+                let lvol = Lvol::try_from(bdev)?;
+                lvol.destroy().await?;
+                sync_config().await;
```
gila, line 254: Again, consider what context you are running this in. We are basically doing file IO within the context of SPDK, and its not needed to store this information here.

mayastor/src/subsys/config/mod.rs
```diff
@@ -181,7 +178,6 @@ impl Default for Config {
             nexus_opts: Default::default(),
             base_bdevs: None,
```
gila, line 179: please remove all other bdevs as well.

mayastor/src/subsys/config/mod.rs
```diff
@@ -538,19 +474,22 @@ impl Config {
...
-                warn!("Not all bdevs({}) were imported successfully", errors);
+                warn!(
+                    "Not all bdevs were imported successfully ({} errors)",
+                    errors
+                );
             }
         });
     }
 
+    #[allow(dead_code)]
```
gila, line 485: remove dead_code

mayastor/src/subsys/config/pool.rs
gila: There is really no need to inline this at all.

mayastor/src/subsys/config/pool.rs
```diff
@@ -0,0 +1,181 @@
...
+where
+    P: AsRef<Path> + Display + ToString,
+{
+    CONFIG_FILE.get_or_init(|| file.to_string());
+}
+
+#[inline]
+/// Return the config file location
+fn get_config_location() -> Option<&'static String> {
+    CONFIG_FILE.get()
```
gila, line 28: Can we call this get_path() or something?

mayastor/src/subsys/config/pool.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+#[derive(Debug, Default, PartialEq, Serialize, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct PoolConfig {
+    pub pools: Option<Vec<Pool>>,
```
gila, line 34: Why is this pub?

mayastor/src/subsys/config/pool.rs
gila: Remove what is not needed, please. Just the pool information is what we want.

mayastor/src/subsys/config/pool.rs
```diff
@@ -0,0 +1,181 @@
...
+                        error.verbose()
+                    );
+                    failures += 1;
+                }
+            }
+        }
+        failures
+    }
+
+    /// Import pools
```
gila, line 101: We do not need to run this in the context of mayastor, Read the file, and then for each file submit an RPC request

mayastor/src/subsys/config/pool.rs
```diff
@@ -0,0 +1,181 @@
...
+                std::io::ErrorKind::Other,
+                format!("pool configuration serialization error: {}", error),
+            )
+        })?;
+
+        fs::write(&file, config.as_bytes())
+    }
+
+    /// Export current pool configuration
+    pub async fn export() -> Result<(), std::io::Error> {
```
gila, line 70: Why is the export logic not run within the context of the RPC server? There really is no need to do that within the context of mayastor itself.

gila (changes requested):

cjones1024:
bors try

mayastor/src/subsys/config/pool.rs
cjones1024: no longer inligned

mayastor/src/subsys/config/pool.rs
cjones1024: No replica ingormation is generated. However if replica information is present in the yaml, rather than fail with invalid config, the replica information will be parsed but otherwise ignored.

mayastor/src/core/env.rs
cjones1024: Now issues a warning in the log rather than trigger a panic

mayastor/src/grpc/mayastor_grpc.rs
cjones1024: Done

mayastor/src/grpc/mayastor_grpc.rs
cjones1024: removed

mayastor/src/grpc/mayastor_grpc.rs
cjones1024: removed

mayastor/src/subsys/config/pool.rs
cjones1024: renamed to get_config_file

mayastor/src/subsys/config/mod.rs
gila: Can you remove the other stuff too?

gila (changes requested):

cjones1024:
bors try

cjones1024:
bors try

jkryl (approved):
I reviewed the JS part - looks good to me.

tiagolobocastro (approved):

cjones1024:
bors try

cjones1024:
bors try

gila (approved):

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors merge

cjones1024:
bors try

cjones1024:
bors merge

PR #892, opened 2021-05-19
fix(moac): moac should handle node removals better

jkryl:
If msn is removed from the cluster (can happen when mayastor on the
node restarts), then certain operations in volume and volume operator
can fail trying to reference the node structure that is being
disassociated with the object (nexus/replica). So handle these cases
gracefully.

Resolves: CAS-902


tiagolobocastro (approved):

paulyoong (approved):

jkryl:
bors merge

PR #893, opened 2021-05-19
test(e2e): fixes for nightly pipeline

chriswldenyer:
We may need to tag 'nightly' images outside of nightly runs.
Added Jenkins config option to tag a successful build as 'nightly'.
if tests pass when run manually.
Allow e2e tests to be disabled from Jenkins config.
Removed some duplicated code.
Replace deprecated 'continue' keyword with 'reinstall'.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -63,8 +63,8 @@ def isTimed() {
 }
 
 def getAliasTag() { // alternative tag for CI pushed images
-    if (isTimed() == true) {
-        return 'nightly'
+    if (isTimed() == true || params.run_as_nightly == true) {
+      return 'nightly'
     }
     return 'ci'
 }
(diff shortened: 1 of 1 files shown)


chriswldenyer:
bors try

Jenkinsfile
```diff
@@ -81,11 +81,8 @@ def getTestPlan() {
   if (params.e2e_continuous == true)  {
     return xray_continuous_testplan
   }
-  def causes = currentBuild.getBuildCauses()
-  for(cause in causes) {
-    if ("${cause}".contains("hudson.triggers.TimerTrigger\$TimerTriggerCause")) {
-      return xray_nightly_testplan
-    }
+  if (isTimed() == true) {
```
blaisedias, line 84: should this also be `if (isTimed() == true || params.run_as_nightly == true) ` ?

chriswldenyer: Only if we want to send the results to the nightly test plan. I thought about this and thought that it was inappropriate for this use case.

blaisedias (approved):

chriswldenyer:
bors try

tiagolobocastro (approved):

chriswldenyer:
bors merge

PR #895, opened 2021-05-20
feat(moac): update dependencies etc

jkryl:
The dependency update fixes a couple of sec vulnerabilities.

We switched back to official k8s client as all our fixes have been
merged to the upstream.

dirty-chai was removed because it's not possible to write TS tests
(definitions are different than expected). Instead we ignore the
eslint error which was why we started to use dirty-chai originally.

### DIFF
--- a/csi/moac/index.ts
+++ b/csi/moac/index.ts
@@ -1,13 +1,13 @@
 // Main file of our control plane for mayastor.
 // It binds all components together to create a meaningful whole.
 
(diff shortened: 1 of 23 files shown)


tiagolobocastro (approved):

paulyoong (approved):

jkryl:
bors merge

jkryl:
bors retry

PR #896, opened 2021-05-21
feat(persistence): add persistent store

paulyoong:
Add a persistent store which uses etcd as its backing store.

Each nexus saves a "NexusInfo" structure to etcd. The key for this
structure is the nexus UUID. The NexusInfo data includes the states of
the children and a "clean shutdown" flag.

The NexusInfo data structure has been defined in a new persistence.proto
file. This is necessary as the data structure is common between Mayastor
and MOAC.

There are 3 places where the persistent information is updated:
1. During nexus creation
2. When retiring a child due to an I/O error
3. During nexus destruction

An additional "-p" argument has been added to Mayastor which allows the
endpoint of the persistent store to be provided. If a port number is
omitted the default port number (2379) is used.

The "list nexus" RPC call has been modified such that it will not return
nexuses that are in the "Init" state. This prevents MOAC from publishing
a nexus before its state has been persisted (the nexus is set online
only after its NexusInfo has been saved to the persistent store).

Resolves: CAS-882, CAS-889


paulyoong:
bors try

mayastor/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -0,0 +1,146 @@
...
+                        reason: reason as i32,
+                    });
+                });
+            }
+            PersistOp::Update((uuid, state)) => {
+                // Only update the state of the child that has changed. Do not
+                // update the other children or "clean shutdown" information.
+                // This should only be called on a child state change.
+                nexus_info.children.iter_mut().for_each(|c| {
+                    let persistent_state: PersistentChildState = state.into();
```
mtzaurus, line 59: Wouldn't it be better to move 'let persistent_state' and 'let persistent_reason' under 'if c.uuid == uuid {' and instantiate them only when target replica is found ?

mayastor/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -0,0 +1,146 @@
...
+            return;
+        }
+        let mut nexus_info = self.nexus_info.lock().await;
+        match op {
+            PersistOp::Create => {
+                // Initialisation of the persistent info will overwrite any
+                // existing entries.
+                // This should only be called on nexus creation.
+                nexus_info.clean_shutdown = false;
+                self.children.iter().for_each(|c| {
```
mtzaurus, line 44: Shouldn't we also reset 'nexus_info.children' array to make sure it's clear and contains no stalled information when we repopulate it ?

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -379,6 +380,17 @@ impl NexusChild {
...
 
+    /// Extract the child UUID from its name.
+    pub(crate) fn uuid(&self) -> String {
+        let url = Url::parse(&self.name).expect("Failed to parse child url");
+        for pair in url.query_pairs() {
+            if pair.0 == "uuid" {
+                return pair.1.to_string();
+            }
+        }
+        panic!("Child name does not contain a uuid query parameter.");
```
mtzaurus, line 391: Since the presence of UUID is perceived as a 'must have' prerequisite for NexusChild object, wouldn't it be better to add such hard check to 'new()' method to assure the presence of UUID in the name ?

mtzaurus: Also, we already have 'guid' member in NexusChild. Looks like it serves the same purpose as 'uuid()' member. Can we initialise this member based on URI's UUID part and then just keep using the 'guid' member ?

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -421,6 +421,10 @@ impl mayastor_server::Mayastor for MayastorSvc {
             Ok(ListNexusReply {
                 nexus_list: instances()
                     .iter()
+                    .filter(|n| {
```
mtzaurus, line 424: Aren't we loosing observability here ? I mean, shouldn't MOAC filter out not fully initialised nexuses rather than Mayastor hide them ?

mayastor/src/persistent_store.rs
```diff
@@ -0,0 +1,250 @@
...
+            // No endpoint means no persistent store.
+            warn!("Persistent store not initialised");
+            return;
+        }
+
+        assert!(endpoint.is_some());
+
+        // An endpoint has been provided, initialise the persistent store.
+        let endpoint = Self::format_endpoint(&endpoint.unwrap());
+        let store = Self::connect_to_backing_store(&endpoint.clone()).await;
```
mtzaurus, line 56: Are there any other tuneables here like connection timeouts, etc ? Would it be reasonable to introduce something tuneable here to have better control over ETCd backed ?

mtzaurus: Ah, I see later on that we try to eternally connect to ETCd. So please disregard this comment.

mayastor/src/grpc/mayastor_grpc.rs
paulyoong: I am in agreement. However, unfortunately the Nexus Init state gets mapped to the Degraded state before being passed up to MOAC (as MOAC only knows about 3 states; online, degraded and faulted) - https://github.com/openebs/Mayastor/blob/72a273707e8f2a1f09bcdf18de999873df2db89e/mayastor/src/bdev/nexus/nexus_bdev.rs#L941

As far as MOAC is concerned the Degraded state doesn't prevent the nexus from being published and used by an application. What we needed to do here was prevent the nexus from being able to be used before its persistent information had been written to the store - which occurs whilst in the Init state. After speaking with Jeffry, the suggestion was to prevent initialising nexuses from being reported back to MOAC.

mayastor/src/bdev/nexus/nexus_persistence.rs
paulyoong: This code is only ever called when we are creating the nexus. As a result, the expectation is that the NexusInfo structure has only just been created and cannot contain anything yet. I will make this clearer by including an `assert` which checks that the vector of children is empty.

mayastor/src/grpc/mayastor_grpc.rs
gila: So we all agree here :)

gila (changes requested):

mayastor/src/bdev/nexus/nexus_persistence.rs
paulyoong: Good spot, thanks. Done.

mayastor/src/bdev/nexus/nexus_persistence.rs
paulyoong: I've added a couple of asserts here to check the values of the data structure.

mayastor/src/bdev/nexus/nexus_child.rs
paulyoong: Good idea. I have added a check in the `NexusChild::new` which checks for a UUID.

The `guid` member of NexusChild may be the same thing but I cannot be confident in that because the labelling code seems to conditionally overwrite it. By extracting it from the name I know that this will match up exactly with what MOAC sees.

mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -505,6 +517,10 @@ impl NexusChild {
         parent: String,
         device: Option<Box<dyn BlockDevice>>,
     ) -> Self {
+        if Self::uuid(&name).is_none() {
+            panic!("Child name does not contain a UUID.");
```
mtzaurus, line 521: 👍

mtzaurus (approved):

mayastor/src/persistent_store.rs
```diff
@@ -0,0 +1,250 @@
...
+                Ok(store) => {
+                    info!("Connected to etcd on endpoint {}", endpoint);
+                    return store;
+                }
+                Err(_) => {
+                    if output_err {
+                        // Only output the error on first failure to prevent
+                        // flooding the logs.
+                        error!(
+                            "Failed to connect to etcd on endpoint {}",
```
tiagolobocastro, line 90: Maybe add a "retrying.." to the message?

mayastor/src/persistent_store.rs
```diff
@@ -0,0 +1,250 @@
...
+
+    /// Retrieve a value, with the given key, from the store.
+    pub async fn get(key: &impl StoreKey) -> Result<Value, StoreError> {
+        let key_string = key.to_string();
+        let rx = Self::execute_store_op(async move {
+            info!("Getting key {} from store.", key_string);
+            match Self::backing_store().get_kv(&key_string).await {
+                Ok(value) => {
+                    info!("Successfully got key {}", key_string);
+                    Ok(Some(value))
```
tiagolobocastro, line 148: can't this be just Ok(value) and then below on 157 you don't need to unwrap?

mayastor/src/persistent_store.rs
```diff
@@ -0,0 +1,250 @@
...
+        let key_string = key.to_string();
+        let rx = Self::execute_store_op(async move {
+            info!("Deleting key {} from store.", key_string);
+            match Self::backing_store().delete_kv(&key_string).await {
+                Ok(_) => {
+                    info!(
+                        "Successfully deleted key {} from store.",
+                        key_string
+                    );
+                    Ok(None)
```
tiagolobocastro, line 171: similarly, could this be Ok(())?

tiagolobocastro (approved):

mayastor/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -0,0 +1,146 @@
...
+                Ok(_) => {
+                    // The state was saved successfully.
+                    break;
+                }
+                Err(e) => {
+                    // Output an error message on first failure. Thereafter
+                    // silently retry.
+                    if output_err {
+                        error!(
+                            "Failed to persist with error {}. Retrying...",
```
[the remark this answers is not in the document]
paulyoong: Done.

mayastor/src/persistent_store.rs
paulyoong: Done.

mayastor/src/persistent_store.rs
paulyoong: As discussed, changed the execute store op function to be generic in its return type.

mayastor/src/persistent_store.rs
paulyoong: As discussed, changed the execute store op function to be generic in its return type.

paulyoong:
bors try

paulyoong:
bors try

paulyoong:
bors try

paulyoong:
bors try

paulyoong:
bors merge

paulyoong:
bors merge

gila (approved):

paulyoong:
bors merge

PR #897, opened 2021-05-21
feat: add persistent information interface to moac

tiagolobocastro:
fix: don't add uuid to the nexus share uri
Extend the uri with the uuid query parameter only for the pool replicas.

chore: add persistence protobuf files
Update the script which copies the protobuf file and add it to the nix build derivation.
Also add a script which generates javascript and typescript definitions.

fix: make sure replicas have a real UUID in their URI
Make sure MOAC only uses replicas with a real UUID as this UUID is
required by coming the persistent nexus information.

feat: add persistent information interface to moac
Add a new interface that allows MOAC to determine which replicas can be used in the
creation of a nexus by using the persistent store (etcd).

On first creation all replicas are allowed to be used.
On recreations, only healthy replicas can be used.
If the last shutdown of the nexus was unclean, then only 1 replicas is choosen and for
the moment we choose a local replica, if possible.
The mayastor changes are being worked on separately.

On deletion of the volume we delete the entry from the persistent store.

test: add tests for the persistent store at the MOAC level
Add a persistent store with a mocked backend to the existing volume tests.

Add a new persistence test which does basic testing using a real etcd
server and uses a mock etcd client to test the various corner cases when
trying to create and recreate a nexus.

Configure mocha to use the typescript extension and replca the existing
javascript test index with a typescript file.
Add etcd to the nix-shell.

chore: add new command line argument to MOAC for the persistent store…
… endpoint

etcd-endpoint - defaults to 127.0.0.1:2379

chore: update MOAC's deployment files

Add the new command line argument to the chart's yaml file for k8s.
Update the nix build files to include the new dependencies.

chore: revert persistent proto file

Add persistent nexus information type definition directly to MOAC.
This must match what mayastor writes to ETCD so any changes will
have to be made in tandem.
Note that MOAC does not test against a real mayastor instance so any
changes on mayastor will NOT trigger a failure of the moac tests.
Also replace snake_case with camelCase because the semistandard is not
running against TS files at the moment.

Resolves: CAS-884 CAS-885


tiagolobocastro:
bors try

tiagolobocastro:
bors try

shell.nix
```diff
@@ -55,6 +55,7 @@ mkShell {
     python3
     utillinux
     xfsprogs
+    etcd
```
jkryl, line 58: would be good to keep alphabetical order

rpc/proto/persistence.proto
```diff
@@ -0,0 +1,36 @@
...
+  RebuildFailed = 3;
+  IoError = 4;
+  Rpc = 5;
+}
+
+message ChildInfo {
+  string uuid = 1;
+  ChildState state = 2;
+  Reason reason = 3;
+}
```
jkryl, line 31: It's not a problem but I'm wondering if we shouldn't keep the api minimal in a sense that we should avoid everything that is strictly not necessary. AFAIK all we need is:
* clean shutdown flag
* uuid of replica
* state of child and in particular just two states (Open and not open)

Adding new things to grpc api is easy and backward compatible but old things that are not used will become a maintenance burden.

jkryl (commented):
Looks good! Just a couple of comments.

chart/templates/moac-deployment.yaml
```diff
@@ -52,6 +52,7 @@ spec:
             - "--port=3000"
             - "--watcher-idle-timeout=600000"
             - "--message-bus=nats"{{ if .Values.moacDebug }}
+            - "--etcd-endpoint=mayastor-etcd"
```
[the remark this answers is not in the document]
tiagolobocastro: ah I was wondering what that debug at the end there did but did not think much of it, good catch thanks!

csi/moac/registry.ts
```diff
@@ -256,4 +259,8 @@ export class Registry extends events.EventEmitter {
 
     return pools;
   }
+
+  get_persistent_store() {
```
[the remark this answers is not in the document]
tiagolobocastro: thanks, I thought I got all of them, I replaced a few which semistandard caught, strange it missed this one.

shell.nix
tiagolobocastro: good point, I sorted the whole list now

rpc/proto/persistence.proto
tiagolobocastro: We tried to keep the states close to what mayastor has to gives us a bit more observability of what "went wrong" when a child is not healthy anymore. For example, in the future we might be able to do partials rebuilds so knowing that the child was not faulted but only out of sync would be useful.
Anyway, MOAC is only reading these fields from mayastor so I'll add @paulyoong who can make decisions about the data format.

csi/moac/test/event_stream_test.js
```diff
@@ -128,15 +130,15 @@ module.exports = function () {
...
         });
         registry.emit('replica', {
           eventType: 'mod',
-          object: { uuid: 'replica2' }
+          object: { uuid: 'replica2', uri: 'bdev:///replica2?uuid=2' }
         });
         registry.emit('replica', {
           eventType: 'del',
-          object: { uuid: 'replica3' }
+          object: { uuid: 'replica3', uri: 'bdev:///replica3?uuid=3' }
```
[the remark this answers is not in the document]
tiagolobocastro: Yep, added the incrementing var and at the end expect real uuid not to be undefined.

csi/moac/replica.ts
```diff
@@ -16,6 +17,9 @@ export class Replica {
   share: string;
   uri: string;
   isDown: boolean;
+  // hint that this replica was at some point faulted by the parent, so it would be wise to not reuse it
+  isFaulted: boolean;
+  realUuid?: string;
```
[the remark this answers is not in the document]
tiagolobocastro: Yep that makes things easier, done.

tiagolobocastro:
bors try

rpc/proto/persistence.proto
paulyoong: I think we should keep it. The new control plane is likely to be more opinionated on things in general and so will probably require more detailed information to be passed up from Mayastor. That future work may require the gRPC layer to be overhauled anyway, which I guess we **could** introduce as a breaking change, therefore the backwards compatibility issue wouldn't arise.

jkryl: 👍

jkryl (approved):

gila:
Why do we add the internal fields of mayastor to the gRPC protocol? This makes 0.0 sense to me as we have over 60+ lines of code that has the sole purpose of: "we don't care about that".... I understand the consumer MAY not care about this field, but I really do not like this at all.

rpc/proto/persistence.proto
```diff
@@ -0,0 +1,36 @@
+syntax = "proto3";
+
+option java_multiple_files = true;
+option java_package = "io.openebs.mayastor";
+option java_outer_classname = "PersistenceProto";
+
+package persistence;
+
+enum ChildState {
```
gila, line 9: Why is this here?

gila (changes requested):

paulyoong (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #899, opened 2021-05-24
fix(nexus): io_retire

gila:
IO errors are now handled differently such that errors on a channel, are
handled on a per-channel basis.

When we are in submission, we simply remove the device from the
channel(s) and retry the IO by resubmitting it into the nexus.

Completion IOs are -- when completed, also retried. For each failed
device, we kick off one retire request which will pause the nexus while
it does so. This prevents any new incoming IO.

### DIFF
--- a/mayastor/src/bdev/nvmx/controller.rs
+++ b/mayastor/src/bdev/nvmx/controller.rs
@@ -248,7 +248,7 @@ impl<'a> NvmeController<'a> {
 
         // Deactivate existing namespace in case it is no longer active.
         if !ns_active && !ctrlr_inner.namespaces.is_empty() {
-            info!("{}: deactivating existing namespace", self.name);
+            debug!("{}: deactivating existing namespace", self.name);
             notify_listeners = true;
         }
 
@@ -259,7 +259,7 @@ impl<'a> NvmeController<'a> {
             );
             vec![]
         } else {
-            info!("{}: namespace successfully populated", self.name);
+            debug!("{}: namespace successfully populated", self.name);
             vec![Arc::new(NvmeNamespace::from_ptr(ns))]
         };
 
@@ -318,7 +318,7 @@ impl<'a> NvmeController<'a> {
                 }
             })?;
 
-        info!(
+        debug!(
             "{} initiating controller reset, failover = {}",
             self.name, failover
         );
@@ -342,7 +342,7 @@ impl<'a> NvmeController<'a> {
             shutdown_in_progress: false,
         };
 
-        info!("{}: starting reset", self.name);
+        debug!("{}: starting reset", self.name);
         let inner = self.inner.as_mut().unwrap();
         // Iterate over all I/O channels and reset/configure them one by one.
         inner.io_device.traverse_io_channels(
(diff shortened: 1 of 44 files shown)


mayastor/src/bdev/nvmx/controller.rs
```diff
@@ -794,11 +794,11 @@ pub extern "C" fn nvme_poll_adminq(ctx: *mut c_void) -> i32 {
     let mut context = NonNull::<TimeoutConfig>::new(ctx.cast())
         .expect("ctx pointer may never be null");
     let context = unsafe { context.as_mut() };
-
     let rc =
         unsafe { spdk_nvme_ctrlr_process_admin_completions(context.ctrlr) };
     if rc < 0 {
-        context.reset_controller();
+        //  unsafe { spdk_nvme_ctrlr_fail(context.ctrlr) };
+        // context.reset_controller();
```
jonathan-teh, line 801: Is it safe to ignore admin command completion errors? Is this related to KATO being disabled (in SPDK, at least)?

mayastor/src/bdev/nvmx/controller.rs
```diff
@@ -393,11 +393,11 @@ impl<'a> NvmeController<'a> {
...
-        debug!("{} resetting NVMe controller", ctx.name);
-        let rc = unsafe { spdk_nvme_ctrlr_reset(controller.ctrlr_as_ptr()) };
-        if rc != 0 {
-            error!("{} failed to reset controller, rc = {}", ctx.name, rc);
-        }
+        //debug!("{} resetting NVMe controller", ctx.name);
+        //let rc = unsafe { spdk_nvme_ctrlr_reset(controller.ctrlr_as_ptr()) };
+        //if rc != 0 {
+        //    error!("{} failed to reset controller, rc = {}", ctx.name, rc);
+        //}
```
jonathan-teh, line 400: Why is it no longer necessary to reset controllers on shutdown?

jonathan-teh (approved):
It'd be good to have a summary of what the final large commit does.  At a glance I can see it removes the fail-fast codepaths (perhaps the now-dead code should be removed too) in favour of retrying on IO completion if there is a submission failure, faulting a bad child on IO failure, and probably some other aspects of the new IO handling I may have missed out. It also includes swapping out the standard synchronisation primitives for parking_lot.

mayastor/src/bdev/nvmx/controller.rs
gila: No, it is not. We don't want to reset because it will lock the thread up with the connection to something that is not there. Eventually, we might want to revisit this and say that we want to wait for a few seconds or retry a couple of times etc.

mayastor/src/bdev/nvmx/controller.rs
gila: I doubt it ever was.

gila:
bors try

spdk-sys/build.sh
```diff
@@ -15,9 +15,10 @@ rm libspdk.so
 	--with-iscsi-initiator \
 	--with-crypto \
 	--with-uring \
-	--enable-unit-tests
+	--enable-unit-tests \
+	--with-fio=$(which fio | sed s';bin/fio;include;')
 
-make -j $(nproc)
+bear -- make -j $(nproc)
```
jonathan-teh, line 21: `bear` is missing from the nix shell.

PR #901, opened 2021-05-26
fix(moac): timeouts when updating volume custom resource

jkryl:
The problem was that the objects (old and new one) were basically
the same except that order of entries in replicas array was different.
Problem #1 is that we do useless update of CR and #2 is that k8s
does not emit mod event in such case so volume-operator was timing
out waiting for the event.

The fix is to enforce deterministic order of entries in replicas
array so that deep-equal compares the objects correctly.


paulyoong (approved):

tiagolobocastro (approved):

jkryl:
bors merge

PR #914, opened 2021-06-02
Update demo etcd statefulset.yaml

dyasny:
- Use 3 replicas
- Add label for app:mayastor-etcd

### DIFF
--- a/deploy/etcd/statefulset.yaml
+++ b/deploy/etcd/statefulset.yaml
@@ -11,7 +11,7 @@ metadata:
     app.kubernetes.io/instance: mayastor
     app.kubernetes.io/managed-by: Helm
 spec:
-  replicas: 1
+  replicas: 3
   selector:
     matchLabels:
       app.kubernetes.io/name: etcd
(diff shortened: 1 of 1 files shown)


jkryl:
The change must be done to the charts/templates and deploy directory regenerated then. Though I'm not sure if this is what we want. Number of replicas has been set to 1 intentionally for develop test configuration. We use 3 replicas for production yamls and e2e test yamls.

tiagolobocastro:
Also, just setting the replica to 3 does not turn our develop into a cluster, instead you get 3 separate etcd instances. 
To use a cluster you can use our script to generate yamls for a "release" profile which does use an etcd cluster:
`./scripts/generate-deploy-yamls.sh release`

dyasny:
Right, I only wanted to bring these to a basic level of resilience since we reference them in the official guide. 

There's another issue actually - the domain name in https://github.com/openebs/Mayastor/blob/c27ab0e773aa5e5b214fa5ae281ebfaede226dc4/deploy/etcd/statefulset.yaml#L85 is hardcoded and so etcd fails to start. I switched it to using `MY_POD_IP` instead and that worked

PR #917, opened 2021-06-03
fix(moac): volume remains in destroyed state

jkryl:
For some reason which is out of scope of this PR, mayastor fails to
send a reply to grpc request issued by moac. That makes the whole
life-cycle of the volume stuck until the moac is restarted. It's
a general problem pertaining to all grpc operations. We need to
set a timeout on all grpc requests to prevent this from happening.

The default timeout for all grpc calls is 15s now. The destroy
replica call with timeout of 1 hour being exception because it is
known to be slow when the replica is large.

This might trigger new bugs in mayastor as some operations might
be resent now. It is a question of how mayastor will handle a new
request while the old one is still in the progress. We will fix
those as we discover them. This could have happened even before.
It's just that likelyhood has increased now.

Resolves: CAS-896

### DIFF
--- a/csi/moac/src/grpc_client.ts
+++ b/csi/moac/src/grpc_client.ts
@@ -6,11 +6,12 @@ import * as grpc from '@grpc/grpc-js';
 import { loadSync } from '@grpc/proto-loader';
 
 import { Logger } from './logger';
-import { ServiceClientConstructor } from '@grpc/grpc-js/build/src/make-client';
+import { ServiceClient, ServiceClientConstructor } from '@grpc/grpc-js/build/src/make-client';
 
 const log = Logger('grpc');
 
 const MAYASTOR_PROTO_PATH: string = path.join(__dirname, '../proto/mayastor.proto');
+const DEFAULT_TIMEOUT_MS: number = 15000;
 
 // Result of loadPackageDefinition() when run on mayastor proto file.
 class MayastorDef {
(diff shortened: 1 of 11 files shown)


paulyoong (approved):

tiagolobocastro:
Would we want to add now big timeouts for Create and Destroy calls to mayastor? Internally, mayastor will connect to etcd and if that is taking a long time (slow disk or maybe etcd is temporarily not available) then might be safer to have bigger timeouts to start with?

jkryl:
That is a question. Currently we wait 15s for all calls. That is far bigger than round-trip time on LAN that could be units of ms. I think we want to see errors in moac log if etcd is down or unreachable. I think we will need to find out what is the right value. But how about 30s? Or do you think it should be even higher?

tiagolobocastro:
Well at the moment mayastor uses a 30s timeout for etcd, after which it will timeout and attempt to reconnect again. 
Maybe 1 minute?

jkryl:
1 minute seems like a good start 👍

jkryl:
I have updated the PR and also the commit message to mention explicitly non-default timeout for create and destroy nexus calls.

tiagolobocastro (approved):
I've tested it on my cluster and the volumes are now all destroyed. It did take some time to make progress, suspect one of the gRPC calls was still blocked for a few minutes somehow.

jkryl:
bors merge

PR #918, opened 2021-06-04
fix(cli): controller cli command added

mtzaurus:
Introduced a new command for listing nvme controllers.

Implements CAS-918

### DIFF
--- a/mayastor/src/bdev/mod.rs
+++ b/mayastor/src/bdev/mod.rs
@@ -21,7 +21,12 @@ pub use nexus::{
     },
     nexus_persistence::{ChildInfo, NexusInfo},
 };
-pub use nvmx::nvme_io_ctx_pool_init;
+pub use nvmx::{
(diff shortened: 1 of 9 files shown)


tiagolobocastro (approved):
Would it be possible to show outstanding IOs to a controller?

gila:
> Would it be possible to show outstanding IOs to a controller?

That would be interesting to know indeed because I would imagine that we would use this tool if a controller is acting up. Could we traverse the IO channels and collect those stats?

mtzaurus:
We can add I/O stats for the controllers on behalf of a separate command - 'controller stats' or so.  Would that work ?

mtzaurus:
I created a separate ticket for tracking I/O stats for nvme controllers: "CAS-950 Want CLI for displaying I/O stats for NVMe controllers"

gila (approved):

PR #919, opened 2021-06-07
fix: don't readd existing children to the nexus

tiagolobocastro:
fix(moac): use filtered replicaSet

During the volume's fsa, MOAC filters out replicas which are not shared properly but it then
uses the unfiltered list to determine which replicas it should add to a nexus.
This could cause a nexus to have 2 bdev children.
Fix: use the filtered replicaSet

fix: don't add the same loopback device to the nexus

Other device types return an error when we attempt to open them twice.
The loopback's are special cases, because they're already open, eg: we
create a replica and then open the loopback device.
When a loopback device is reopened with a different UUID it ends up
overwriting the existing UUID and returning OK which sounds very bad.
Fix: Fail the open request if the UUID is not the expected one.

Related issue, because opening an existing loopback succeeds it means
that we can add nexus children to the nexus twice.
Fix: Fails NexusAddChild if the child is already part of the nexus!

### DIFF
--- a/mayastor/src/nexus_uri.rs
+++ b/mayastor/src/nexus_uri.rs
@@ -53,6 +53,12 @@ pub enum NexusBdevError {
     // Bdev create/destroy errors
     #[snafu(display("bdev {} already exists", name))]
     BdevExists { name: String },
+    #[snafu(display(
+        "bdev {} already exists with a different uuid: {}",
+        name,
+        uuid
+    ))]
+    BdevWrongUuid { name: String, uuid: String },
     #[snafu(display("bdev {} not found", name))]
     BdevNotFound { name: String },
     #[snafu(display("Invalid parameters for bdev create {}", name))]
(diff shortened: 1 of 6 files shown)


tiagolobocastro:
bors try

jkryl (approved):
Thanks a lot!

gila (approved):

tiagolobocastro:
bors merge

mayastor/src/nexus_uri.rs
```diff
@@ -53,6 +53,12 @@ pub enum NexusBdevError {
     // Bdev create/destroy errors
     #[snafu(display("bdev {} already exists", name))]
     BdevExists { name: String },
+    #[snafu(display(
+        "bdev {} already exists with a different uuid: {}",
```
paulyoong, line 57: Does the bdev name contain the uuid within it? If not, it could be useful to output both the requested uuid and the existing uuid.

paulyoong (approved):

PR #920, opened 2021-06-07
fix(nexus): removing a loopback replica that is already shared

jonathan-teh:
Removing a loopback replica that is already shared over nvmf should
result in the replica remaining shared over nvmf. NexusChild::close()
calls Descriptor::unclaim() which marks the bdev as unclaimed, causing
the replica to lose its shared state. When the replica is subsequently
destroyed, the nvmf target remains, though with no namespaces, and is
leaked.

Have Descriptor::unclaim() only unclaim a bdev that was claimed by
NEXUS_MODULE. Only test code calls Descriptor::claim().

This issue was always present but could previously only be exposed by
adding a replica over loopback, then sharing that replica over nvmf.
After changes in 347d37a, an already-shared replica could be added as
a local (loopback) child, making it more likely.

Add a cargo test to verify that the loopback replica remains shared
over nvmf after its removal from the nexus.

### DIFF
--- a/mayastor/src/core/descriptor.rs
+++ b/mayastor/src/core/descriptor.rs
@@ -68,10 +68,12 @@ impl Descriptor {
         err == 0
     }
 
-    /// unclaim a previously claimed bdev
+    /// unclaim a bdev previously claimed by NEXUS_MODULE
     pub(crate) fn unclaim(&self) {
         unsafe {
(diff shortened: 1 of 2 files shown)


jonathan-teh:
bors try

mtzaurus (approved):

jonathan-teh:
bors merge

PR #924, opened 2021-06-08
fix(moac): avoid duplicating watcher upon restart

jkryl:
It seems to be an undocumented behaviour of the watcher library that
we are using, to call error callback even if there is no error upon
the first start. The solution is to register error callback after the
watcher is started.

I also made some changes to avoid printing error messages to the log
when the watcher restart was actually intentional.

Resolves: CAS-938

### DIFF
--- a/csi/moac/src/watcher.ts
+++ b/csi/moac/src/watcher.ts
@@ -1,6 +1,7 @@
 // Implementation of a cache for arbitrary k8s custom resource in openebs.io
 // api with v1alpha1 version.
 
+import assert from 'assert';
 import * as _ from 'lodash';
 import events = require('events');
 import {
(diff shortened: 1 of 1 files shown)


csi/moac/src/watcher.ts
```diff
@@ -293,22 +300,28 @@ export class CustomResourceCache<T> extends events.EventEmitter {
 
   // Called when the connection breaks.
   _onError(err: any) {
-    log.error(`Watcher error: ${err}`);
-    this.stop();
+    // do not print the error if it was us who terminated the watcher
+    if (this.connected) {
+      log.error(`Watcher error: ${err}`);
+      this.stop();
+    }
```
paulyoong, line 307: How do we know we were the ones who terminated the watcher here? What if we were connected fine but an error occurred?

paulyoong: Thanks for the explanation.

paulyoong (approved):

tiagolobocastro (approved):

PR #921, opened 2021-06-07
feat(nexus): add kato support

gila:
Adds the ability for device removal based on KATO timeouts

PR #926, opened 2021-06-08
fix(spdk): restored iscsi initiator module in spdk binary

mtzaurus:
Iscsi initiator module was accidentally removed from spdk build process
in the latest commits, which leads to missing 'delete_iscsi_disk'
function when building custom spdk.
This fix re-adds the removed module to the resulting spdk binary.


jonathan-teh (approved):
I did point this out in #921.

PR #927, opened 2021-06-09
feat(yaml): enable mayastor trace for some images

arne-rusek:
Enable RUST_LOG=trace for develop and test images, avoid on release
images.

Does it make sense? Should help in cases when errors are not easily repeatable.

### DIFF
--- a/chart/develop/values.yaml
+++ b/chart/develop/values.yaml
@@ -15,3 +15,4 @@ etcd:
   persistence:
     enabled: false
 
+mayastorLogLevel: debug
--- a/chart/release/values.yaml
+++ b/chart/release/values.yaml
@@ -16,3 +16,4 @@ etcd:
   persistence:
     enabled: false
 
+mayastorLogLevel: info
--- a/chart/templates/mayastor-daemonset.yaml
+++ b/chart/templates/mayastor-daemonset.yaml
@@ -34,6 +34,8 @@ spec:
         image: {{ include "mayastorImagesPrefix" . }}mayadata/mayastor:{{ .Values.mayastorImagesTag }}
         imagePullPolicy: {{ .Values.mayastorImagePullPolicy }}
         env:
+        - name: RUST_LOG
+          value: mayastor={{ .Values.mayastorLogLevel }}
         - name: MY_NODE_NAME
           valueFrom:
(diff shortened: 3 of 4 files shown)


blaisedias (approved):
LGTM

jkryl (commented):
trace level means that messages can be written to a log for each IO. I don't think it is appropriate log level even for develop image. It will just flood the log. It is meant to be used when debugging specific IO problem and be always explicit. I would use debug level at most for develop.

tiagolobocastro:
I'd agree that trace is probably too much. I usually run my develop with "mayastor=debug".
What about changing the mayastorTraceLog from bool to an actual log value, set the default to "info" and for develop set it to "mayastor=debug"?

arne-rusek:
Please take another look.

chart/develop/values.yaml
```diff
@@ -15,3 +15,4 @@ etcd:
   persistence:
     enabled: false
 
+mayastorLogLevel: debug
```
tiagolobocastro, line 18: I think we want "mayastor=debug" otherwise we'll get debug from all crates/libraries we make use of?

arne-rusek: `mayastor=` is directly in the template

https://github.com/openebs/Mayastor/pull/927/files#diff-35937fcb9923c47951a5557076cfeabc39258b78547d11f878e82453dccbbcd7R38

tiagolobocastro: hmm sorry! I think what I said is slightly wrong, we'd want "**info,**mayastor=debug", otherwise we'd actually lose info level messages from the crates/libraries we make use of, including the local ones from our repo. So that basically means, info for all, and for the mayastor crate, use debug.
Example, this one:  WARN **mbus_api**::mbus_nats:mbus_nats.rs:77] Connection to the NATS server has been lost.
https://rust-lang-nursery.github.io/rust-cookbook/development_tools/debugging/config_log.html

arne-rusek: Oh I see! Thanks for clarification. Let me fix that.

jkryl (approved):

gila (approved):

gila:
bors merge

PR #928, opened 2021-06-09
test(python): add a nexus create/destroy test and update FioSpdk

jonathan-teh:
Fix docker-compose config so that container ms3 starts on a system
with only 8 cores.

Add a new test that creates 2 replicas shared over nvmf, 1 as
loopback. Create a nexus with all 3 replicas, then destroy the nexus,
for 10 repetitions.

Update the FioSpdk class to match the interface of Fio so that it
supports multiple filenames and add copies of existing Fio tests that
use FioSpdk instead, which removes the requirement for another VM.

### DIFF
--- a/test/python/common/fio_spdk.py
+++ b/test/python/common/fio_spdk.py
@@ -1,32 +1,35 @@
-import asyncio
 import os
 import shutil
-from common.command import run_cmd_async
 from urllib.parse import urlparse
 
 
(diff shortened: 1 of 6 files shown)


paulyoong (approved):

gila (approved):

gila:
bors merge

PR #929, opened 2021-06-09
chore: add memory and vcpu variables to terraform

tiagolobocastro:
These can be overridden with -var="name=$value" via the cmdline.

### DIFF
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -57,6 +57,18 @@ variable "nr_hugepages" {
   default     = "512"
 }
 
+variable "memory" {
+  type        = number
+  default     = 4096
+  description = "Amount of memory (MiB) allocated to each node - only needed for libvirt provider"
+}
+
+variable "vcpu" {
+  type        = number
+  default     = 2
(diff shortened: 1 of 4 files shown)


terraform/variables.tf
```diff
@@ -57,6 +57,18 @@ variable "nr_hugepages" {
...
+variable "memory" {
+  type        = number
+  default     = 4096
+  description = "Amount of memory (MiB) allocated to each node - only needed for libvirt provider"
+}
+
+variable "vcpu" {
+  type        = number
+  default     = 2
+  description = "Virtual CPU's allocated to each node - only needed for libvirt provider"
```
jonathan-teh, line 69: Unnecessary apostrophe.

jonathan-teh (approved):

terraform/variables.tf
tiagolobocastro: good catch thanks - fixed

cjones1024 (approved):

tiagolobocastro:
bors merge

PR #930, opened 2021-06-09
feat(moac): serialize all CSI create and publish requests

jkryl:
It is done to workaround for blocking connect in mayastor but not only
that. It will improve the overall readability of logs and
debuggability of the issues.

As part of that the CSI method implementation were rewritten to adhere
to async paradigm instead of combining callbacks with await in a
single function which was messy.

Resolves: CAS-935


csi/moac/src/csi.ts
```diff
@@ -550,42 +553,29 @@ export class CsiServer {
...
+    };
   }
 
-  async deleteVolume (args: any, cb: CsiDoneCb) {
+  async deleteVolume (args: any) {
     assert(this.volumes);
 
     log.debug(`Request to destroy volume "${args.volumeId}"`);
 
-    // If this is a duplicate request then assure it is executed just once.
```
paulyoong, line 561: Do we not need to check for duplicate requests for deletions now?

csi/moac/src/csi.ts
```diff
@@ -698,40 +666,23 @@ export class CsiServer {
...
       );
-      return cb(null, {});
-    }
-    try {
-      parseMayastorNodeId(args.nodeId);
-    } catch (err) {
-      return cb(err);
-    }
-
-    // If this is a duplicate request then assure it is executed just once.
```
paulyoong, line 709: Again, do we not need to check for duplicate unpublish requests anymore?

csi/moac/src/csi.ts
jkryl: Checking for duplicate requests makes never harm. The assumption we make is that if there is the same method with the same parameters already in progress, we wait for its completion and return the same reply as for the operation which is in progress. Because all methods should be idempotent it is not strictly necessary to detect duplicates but it saves some time and reduces log cluttering. It is also a good performance optimization because k8s can be quite aggressive with sending retries. So it is not strictly necessary for any of the methods but we do it for all.

tiagolobocastro (approved):

paulyoong (approved):

jkryl:
bors merge

PR #932, opened 2021-06-09
feat(nexus): serialize gRPC requests

gila:
gRPC requests are received on the tokio runtime and dispatched to the
reactors by sending it a message. This made us run requests
in parallel which is what we want to avoid.

Additionally, the embedded json-rpc server only checks for incoming
messages every N ms. This change also makes the adopt that model by
holding the lock while it sleeps.

The lock implemented is an RW lock so we can make it more relaxed later
if we need to.

### DIFF
--- a/mayastor/src/grpc/mayastor_grpc.rs
+++ b/mayastor/src/grpc/mayastor_grpc.rs
@@ -24,6 +24,7 @@ use crate::{
         },
         rpc_submit,
         GrpcResult,
+        Serializer,
     },
     host::{blk_device, resource},
     lvs::{Error as LvsError, Lvol, Lvs},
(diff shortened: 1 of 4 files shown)


mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -32,14 +33,46 @@ use crate::{
...
+        trace!(?self.name, "unlocked");
+        *lock = false;
+        output
+    }
+}
+
+impl MayastorSvc {
+    pub fn new(interval: Duration) -> Self {
+        Self {
+            name: String::from("CSISvc"),
```
paulyoong, line 70: Should "CSISvc" be "MayastorSvc"?

paulyoong (approved):

mayastor/src/grpc/mayastor_grpc.rs
gila: Yeah, maybe but as we have multiple services in "mayastor" calling mayastor SVC seems confusing to me?

paulyoong: Fair enough :+1:

mayastor/src/grpc/server.rs
```diff
@@ -20,7 +20,9 @@ impl MayastorGrpcServer {
     ) -> Result<(), ()> {
         info!("gRPC server configured at address {}", endpoint);
         let svc = Server::builder()
-            .add_service(MayastorRpcServer::new(MayastorSvc))
+            .add_service(MayastorRpcServer::new(MayastorSvc::new(
+                Duration::from_millis(4),
```
jonathan-teh, line 24: How was this duration determined?

jonathan-teh (approved):

mtzaurus (approved):

gila:
bors merge

PR #933, opened 2021-06-09
feat: add nvme_core module io_timeout parameter to the csi server

tiagolobocastro:
If this parameter is set, the csi server will set the parameter on startup.
The io_timeout value on the storage class overrides this value and that
should work, but we're not 100% sure that it works for all kernel
versions. This value should cover the cases when the SC value does not
work.

Added a 30s timeout value to the charts. A user may choose to comment
the value out of the yaml to preserve existing settings.

### DIFF
--- a/csi/src/server.rs
+++ b/csi/src/server.rs
@@ -137,6 +137,14 @@ async fn main() -> Result<(), String> {
                 .multiple(true)
                 .help("Sets the verbosity level"),
         )
+        .arg(
+            Arg::with_name("nvme-core-io-timeout")
+                .long("nvme-core-io-timeout")
+                .value_name("TIMEOUT")
+                .takes_value(true)
+                .required(false)
+                .help("Sets the global nvme_core module io_timeout, in seconds"),
+        )
         .get_matches();
 
     let node_name = matches.value_of("node-name").unwrap();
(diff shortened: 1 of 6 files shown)


csi/src/server.rs
```diff
@@ -137,6 +137,14 @@ async fn main() -> Result<(), String> {
                 .multiple(true)
                 .help("Sets the verbosity level"),
         )
+        .arg(
+            Arg::with_name("nvme_core io_timeout")
+                .long("nvme_io_timeout")
```
gila, line 142: nit; can we call it nvme_core_io_timeout? it will make it less ambiguous and maps to the actual kernel variable name?

tiagolobocastro: On the "with_name" or the "long"?

jkryl: Perhaps both of them and for the sake of consistency I would suggest that name of arg is `nvme_core_io_timeout` but the long parameter name to be `nvme-core-io-timeout` to avoid underscores in command line options.

csi/src/server.rs
```diff
@@ -175,6 +183,17 @@ async fn main() -> Result<(), String> {
...
     builder.init();
 
+    if let Some(nvme_io_timeout_secs) = matches.value_of("nvme_core io_timeout")
+    {
+        let io_timeout_secs: u32 = nvme_io_timeout_secs
+            .parse()
+            .expect("nvme_core io_timeout should be an integer number, representing the timeout in seconds");
+
+        if let Err(error) = dev::nvmf::set_nvmecore_iotimeout(io_timeout_secs) {
+            error!("Failed to set nvme_core io_timeout: {}", error.to_string());
```
jkryl, line 193: Shouldn't we exit with an error here? Not honouring io timeout setting can have bad consequences and message in the log file can be easily overlooked.

jkryl (approved):

csi/src/server.rs
tiagolobocastro: Yep, done!

csi/src/server.rs
tiagolobocastro: Ok, changed it to a panic.

tiagolobocastro:
bors try

tiagolobocastro:
bors cancel

jkryl:
thanks! looks good

tiagolobocastro:
bors merge

gila (approved):

PR #935, opened 2021-06-10
test(e2e): change default for e2e testing to off

chriswldenyer:
Change the default to not run e2e tests.
This is to allow the nightly tests to generate develop
images for use by the new nightly-stable pipeline.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -189,7 +189,7 @@ pipeline {
   parameters {
     booleanParam(defaultValue: false, name: 'e2e_continuous')
     booleanParam(defaultValue: false, name: 'run_as_nightly')
-    booleanParam(defaultValue: true, name: 'run_e2e_test')
+    booleanParam(defaultValue: false, name: 'run_e2e_test')
   }
   triggers {
     cron(cron_schedule)


Jenkinsfile
```diff
@@ -189,7 +189,7 @@ pipeline {
   parameters {
     booleanParam(defaultValue: false, name: 'e2e_continuous')
     booleanParam(defaultValue: false, name: 'run_as_nightly')
-    booleanParam(defaultValue: true, name: 'run_e2e_test')
+    booleanParam(defaultValue: false, name: 'run_e2e_test')
```
tiagolobocastro, line 192: can this variable not be overridden from the jenkins UI?

arne-rusek: only when run manually IMHO (i.e. "build with parameters") - bors/scheduler runs will always use the defaults; there are also global environment variables - a bit less user friendly re UI however default could be changed without touching the code

chriswldenyer: Yes it can. But we want to disable it indefinitely in the Mayastor builds. E2e tests will still be run, but in another pipeline.

arne-rusek (approved):
If doing what you really want - re @tiagolobocastro 's question about configurability - LGTM

GlennBullingham (approved):

chriswldenyer:
bors merge

chriswldenyer:
bors merge

PR #936, opened 2021-06-10
fix: add grpc soft timeout

tiagolobocastro:
During disruption tests when the nodes become inaccessible the gRPC
request are not timing out as expected - this is somewhat similar to
what we've seen with the mayastor replicas.
Fix: Add a soft timeout on top of the grpc-timeout to make sure the
requests do timeout.

### DIFF
--- a/csi/moac/src/grpc_client.ts
+++ b/csi/moac/src/grpc_client.ts
@@ -12,6 +12,7 @@ const log = Logger('grpc');
 
 const MAYASTOR_PROTO_PATH: string = path.join(__dirname, '../proto/mayastor.proto');
 const DEFAULT_TIMEOUT_MS: number = 15000;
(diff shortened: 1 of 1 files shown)


jkryl (approved):
Nice. Is there a way how we could cancel the promise that did not complete (basically cancel the grpc request)? That would seem more clean.

gila:
> Nice. Is there a way how we could cancel the promise that did not complete (basically cancel the grpc request)? That would seem more clean.

That might be nice, however, I do wonder if it is worth the effort, and perhaps simply purge all the outstanding RPCs to that node?

tiagolobocastro:
According to what I read we'd have to cancel them 1 by 1, but I think we'd also need to change how we call the methods to use the `makeUnaryRequest`.
Let's get this in first which makes things better and I'll try to see if we cancel the original request as an improvement.
bors merge

PR #937, opened 2021-06-13
fix(test): io_timeout_reset fails too early

tiagolobocastro:
Depending on timing, if the KeepAlive is triggered before we've submitted the read IO,
the controller could be failed and the IO will fail before getting submitted.
This is likely because we now don't reset and retry to connect and simply fail the controller.
In case of early error check that the error is what we expected for this case.

### DIFF
--- a/mayastor/tests/nvme_device_timeout.rs
+++ b/mayastor/tests/nvme_device_timeout.rs
@@ -41,7 +41,7 @@ struct IoOpCtx {
 async fn test_io_timeout(action_on_timeout: DeviceTimeoutAction) {
     Config::get_or_init(|| Config {
         nvme_bdev_opts: NvmeBdevOpts {
-            timeout_us: 2_000_000,
+            timeout_us: 7_000_000,
             keep_alive_timeout_ms: 5_000,
             retry_count: 2,
             ..Default::default()


mtzaurus:
You can simply increase I/O timeout to make sure it always occurs after the test I/O has been scheduled.
The timeout needs to be slightly longer than 5 secs to not interfere with replica shutdown sequence:

```
diff --git a/mayastor/tests/nvme_device_timeout.rs b/mayastor/tests/nvme_device_timeout.rs
index f9814ac..ece2f6a 100755
--- a/mayastor/tests/nvme_device_timeout.rs
+++ b/mayastor/tests/nvme_device_timeout.rs
@@ -41,7 +41,7 @@ struct IoOpCtx {
 async fn test_io_timeout(action_on_timeout: DeviceTimeoutAction) {
     Config::get_or_init(|| Config {
         nvme_bdev_opts: NvmeBdevOpts {
-            timeout_us: 2_000_000,
+            timeout_us: 7_000_000,
             keep_alive_timeout_ms: 5_000,
             retry_count: 2,
             ..Default::default()
```

tiagolobocastro:
> You can simply increase I/O timeout to make sure it always occurs after the test I/O has been scheduled.
> The timeout needs to be slightly longer than 5 secs to not interfere with replica shutdown sequence:
> 
> ```
> diff --git a/mayastor/tests/nvme_device_timeout.rs b/mayastor/tests/nvme_device_timeout.rs
> index f9814ac..ece2f6a 100755
> --- a/mayastor/tests/nvme_device_timeout.rs
> +++ b/mayastor/tests/nvme_device_timeout.rs
> @@ -41,7 +41,7 @@ struct IoOpCtx {
>  async fn test_io_timeout(action_on_timeout: DeviceTimeoutAction) {
>      Config::get_or_init(|| Config {
>          nvme_bdev_opts: NvmeBdevOpts {
> -            timeout_us: 2_000_000,
> +            timeout_us: 7_000_000,
>              keep_alive_timeout_ms: 5_000,
>              retry_count: 2,
>              ..Default::default()
> ```

I tried to avoid doing that because that would make the timing tightly coupled with the wait for pause, but sure I can do that.

mtzaurus (approved):

tiagolobocastro:
@mtzaurus I've updated the commit.

paulyoong (approved):

tiagolobocastro:
bors merge

PR #939, opened 2021-06-14
fix(msv): multiple fixes to the msv logic

tiagolobocastro:
When a volume is being destroyed, it may not have any replicas or nexuses but we still need to
delete the MSV, so make sure we don't exit early.

Add all of the msv event "del" logic to the serialization queue, otherwise it could be ran before
the "new" event has been processed.

Skip volume update for faulted volumes.

Don't import volumes which are already created. This happens right after a volume is created,
which is an unfortunate side effect, so we can just skip it.

Don't show throw invalid extending volume error when the volume size has not yet been set.

### DIFF
--- a/csi/moac/src/volumes.ts
+++ b/csi/moac/src/volumes.ts
@@ -195,9 +195,7 @@ export class Volumes extends events.EventEmitter {
   importVolume(uuid: string, spec: VolumeSpec, status?: VolumeStatus): Volume {
     let volume = this.volumes[uuid];
 
-    if (volume) {
-      volume.update(spec);
-    } else {
+    if (!volume) {
       let state = status?.state;
       let size = status?.size;
       // We don't support multiple nexuses yet so take the first one
--- a/csi/moac/src/volume.ts
+++ b/csi/moac/src/volume.ts
@@ -181,6 +181,7 @@ export class Volume {
       VolumeState.Unknown,
       VolumeState.Pending,
       VolumeState.Destroyed,
+      VolumeState.Faulted,
     ].indexOf(this.state) < 0);
   }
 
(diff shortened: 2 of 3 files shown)


csi/moac/src/volumes.ts
```diff
@@ -169,6 +169,11 @@ export class Volumes extends events.EventEmitter {
     const volume = this.volumes[uuid];
     if (!volume) return;
 
+    if (volume.state === VolumeState.Destroyed) {
+      log.warn(`Destruction of volume "${uuid}" is already in progress`);
+      return;
+    }
+
```
jkryl, line 176: I think this breaks the basic expectation that when this function completes the volume is destroyed. Calling destroy volume twice cannot probably happen with CSI because we detect duplicate requests and process just the first one. On the other hand we could fall into this little trap in the future. Is it harmful if we pass the request further?

jkryl (approved):

csi/moac/src/volumes.ts
tiagolobocastro: Hmm good point, we can indeed handle this further up.

tiagolobocastro: I've added it back to where it originally existed, in the volume_operator, but this time in the function that is "serialized" so it won't be done out of order.

tiagolobocastro: It means it's only called when the MSV was already deleted, and if the event happens to come along while the volume is still not deleted, then I think it's ok to just ignore it as a duplicate?

jkryl: looks good!

paulyoong (approved):

tiagolobocastro:
bors merge

PR #940, opened 2021-06-15
chore(mayastor): allow size of io_ctx mempools to be configurable

cjones1024:
- create functions to initialise io_ctx mempools to be called explicitly on startup
- add command line options to override the size of each mempool

resolves CAS-952

### DIFF
--- a/mayastor/src/bdev/device.rs
+++ b/mayastor/src/bdev/device.rs
@@ -10,7 +10,7 @@ use std::{
 
 use async_trait::async_trait;
 use nix::errno::Errno;
-use once_cell::sync::Lazy;
+use once_cell::sync::{Lazy, OnceCell};
 
 use spdk_sys::{
     iovec,
@@ -50,12 +50,8 @@ use crate::core::{
 static BDEV_LISTENERS: Lazy<RwLock<HashMap<String, Vec<DeviceEventListener>>>> =
     Lazy::new(|| RwLock::new(HashMap::new()));
 
-// Size of the memory pool for NVMe I/O structures.
-const IOCTX_POOL_SIZE: u64 = 64 * 1024 - 1;
-static IOCTX_POOL: Lazy<MemoryPool<IoCtx>> = Lazy::new(|| {
-    MemoryPool::<IoCtx>::create("bdev_io_ctx", IOCTX_POOL_SIZE)
-        .expect("Failed to create memory pool for bdev I/O context")
-});
+// Memory pool for bdev I/O context.
+static BDEV_IOCTX_POOL: OnceCell<MemoryPool<IoCtx>> = OnceCell::new();
 
 /// Wrapper around native SPDK block devices, which mimics target SPDK block
 /// device as an abstract BlockDevice instance.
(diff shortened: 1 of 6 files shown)


cjones1024:
bors check

cjones1024:
bors try

mayastor/src/bdev/nvmx/handle.rs
```diff
@@ -380,19 +370,32 @@ fn io_type_to_err(
     }
 }
 
+/// Initialize memory pool for allocating NVMe controller I/O contexts.
+/// This must be called before the first I/O operations take place.
+pub fn nvme_io_ctx_pool_init(size: u64) {
+    NVME_IOCTX_POOL.get_or_init(|| {
```
gila, line 376: Can we have a debug!() that shows the pool size etc? This is useful sometimes.

mayastor/src/bdev/device.rs
```diff
@@ -278,18 +274,35 @@ fn io_type_to_err(
...
+pub fn bdev_io_ctx_pool_init(size: u64) {
+    BDEV_IOCTX_POOL.get_or_init(|| {
+        MemoryPool::<IoCtx>::create("bdev_io_ctx", size).expect(
+            "Failed to create memory pool [bdev_io_ctx] for bdev I/O contexts",
+        )
+    });
+}
+
+/// Allocate a bdev I/O context from the pool.
+fn alloc_bdev_io_ctx(
```
gila, line 288: Adding tryfrom/from and impl drop would reduce potential errors here, lets work on this on a follow up commit

mayastor/src/bdev/device.rs
```diff
@@ -50,12 +50,8 @@ use crate::core::{
...
     Lazy::new(|| RwLock::new(HashMap::new()));
 
-// Size of the memory pool for NVMe I/O structures.
-const IOCTX_POOL_SIZE: u64 = 64 * 1024 - 1;
-static IOCTX_POOL: Lazy<MemoryPool<IoCtx>> = Lazy::new(|| {
-    MemoryPool::<IoCtx>::create("bdev_io_ctx", IOCTX_POOL_SIZE)
-        .expect("Failed to create memory pool for bdev I/O context")
-});
+// Memory pool for bdev I/O context.
+static BDEV_IOCTX_POOL: OnceCell<MemoryPool<IoCtx>> = OnceCell::new();
```
gila, line 54: The drop for the mempool is never called as it's static. IMHO, the accounting it does is very useful so let us try to call it explicitly on shutdown such that we know, at least during the test that all IO's are accounted for. (we can do this in a follow up change)

gila (approved):

tiagolobocastro (approved):

mayastor/src/bdev/nvmx/handle.rs
cjones1024: We already get the following message from the MemPool::create()
``[2021-06-15T12:02:45.825229905+01:00  INFO mayastor::core::mempool:mempool.rs:51] Memory pool 'bdev_io_ctx' with 65535 elements (24 bytes size) successfully created``
I wasn't sure if there is anything else that can usefully be added without sounding repetitive.

PR #942, opened 2021-06-15
fix: fetch git history and tags on Jenkins

tiagolobocastro:


gila (approved):

Jenkinsfile
```diff
@@ -200,7 +200,14 @@ pipeline {
       agent { label 'nixos-mayastor' }
       steps {
         cleanWs()
-        checkout scm
+        checkout([
+          $class: 'GitSCM',
+          branches: scm.branches,
+          extensions: scm.extensions.findAll{!(it instanceof jenkins.plugins.git.GitSCMSourceDefaults)} + [
+            [$class: 'CloneOption', noTags: false, reference: '', shallow: false, honorRefspec: false],
```
arne-rusek, line 207: Do you need deep checkout? I.e. will you be switching? If so the whole stash/unstash will probably stop working as sources are stashed right after checkout. If not it would be faster to do shallow checkout.

tiagolobocastro:
bors try

Jenkinsfile
tiagolobocastro: From a couple of comments online I thought a deep clone was needed, but turns out it's not, so I've set shallow to true :+1:

tiagolobocastro:
bors try

Jenkinsfile
tiagolobocastro: I forgot to push... so we do need the deep clone indeed...

arne-rusek (approved):
As agreed, please drop the refspec, otherwise LGTM

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

PR #945, opened 2021-06-15
fix(moac): volume gets stuck in offline state when doing failover

jkryl:
The problem was that the mayastor unregistered itself and node object
was removed from moac's internal state before all events related to
offlining the node could be processed. One of the previous complaints
about msn records was that they disappear whenever the storage node
is rebooted which is confusing for users. This fix is addressing both
issues at once. Instead of removing the node upon deregistration, it
disconnects the grpc client and switches the node to offline state.
So it stays around along with all other objects that might be on the
node (pools, replicas, nexuses) and events can be processed as
normally. The obvious downside is that users have to remove msn
records manually in case they know that the storage node will never
join the cluster again. But that's just a small inconvenience
compared to what this brings.

A note for insiders. The way how you can tell the difference between
offline node that is unreachable and offline node that unregistered
itself is by looking at grpcEndpoint in msn spec. It will be empty
string for the later.

Resolves: CAS-954

### DIFF
--- a/csi/moac/src/nats.ts
+++ b/csi/moac/src/nats.ts
@@ -134,7 +134,7 @@ export class MessageBus {
       return;
     }
     log.trace(`"${id}" requested deregistration`);
-    this.registry.removeNode(id);
+    this.registry.disconnectNode(id);
   }
 
   _subscribe () {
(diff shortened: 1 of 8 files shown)


tiagolobocastro (approved):
nice one :+1:

paulyoong (approved):

jkryl:
bors merge

PR #948, opened 2021-06-17
feat(moac): decouple moac from mayastor

jkryl:
It is a preparation step for moving moac to a separate repository.
Major update to the documentation.

This commit solves two major problems:
1. Sharing proto files between mayastor and moac
2. Removing dependency on nix and using docker to build moac image

add 1. We use curl to download proto files from the mayastor repo.
It does not make sense now but will in the future. Everytime a change
to proto files in mayastor repo is made, the commit hash in a script
needs to be changed to pull these new changes. It is a conscious
decision resulting into a deterministic build of moac.

add 2. We still provide nix-shell for setting up dev environment for
moac, however that's all what we use the nix for. The script
csi/moac/scripts/release.sh is a simplified version of its
mayastor counterpart.

Resolves: CAS-958


jkryl:
bors try

gila (approved):

nix/pkgs/images/default.nix
```diff
@@ -83,6 +82,9 @@ in
     contents = [ busybox mayastor-adhoc ];
   });
 
+  # The algorithm for placing packages into the layers is not optimal.
+  # There are a couple of layers with negligable size and then there is one
```
tiagolobocastro, line 86: s/negligable/negligible

csi/moac/scripts/download_protos.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env sh
+
+set -e
+
+DIR="$(realpath "$(dirname "$0")")"
+COMMIT=7f5d5a10c22d20d4502775d0639e94903773fdcd
```
tiagolobocastro, line 6: Could this commit be added to somewhere more obvious, maybe in the root of moac? Could we add it, for example, to the package.json?

tiagolobocastro (approved):

csi/moac/doc/volume-operator.md
```diff
@@ -0,0 +1,62 @@
...
+
+The volume operator relies on following information when making decisions:
+
+* info from CSI requests (i.e. create volume parameters from CSI RPC method)
+* info obtained from storage nodes (i.e. list of created replicas and their parameters)
+* info obtained from msv k8s custom resources (i.e. was the volume mounted)
+
+When moac starts up it reads all available information from storage nodes and
+then tries to create/modify objects according to what it believes is the desired
+state. The source of the information about the desired are msv resources. For
```
paulyoong, line 19: Need to add "state" -> "...about the desired **state** are msv resources."

csi/moac/scripts/download_protos.sh
jkryl: I could probably pass it as an argument from package.json. Would that be ok?

paulyoong (approved):

csi/moac/scripts/download_protos.sh
tiagolobocastro: yeah that sounds good!

nix/pkgs/images/default.nix
jkryl: fixed. thanks

csi/moac/doc/volume-operator.md
jkryl: fixed. thanks

jkryl:
bors merge

PR #954, opened 2021-06-21
test(py): add precommit hooks and pytest-bdd 

gila:


paulyoong (commented):
There appears to be a number of empty files. Was this intended?

gila:
> There appears to be a number of empty files. Was this intended?

They are not empty, GH just reports nonsense here.

tiagolobocastro (approved):

tiagolobocastro:
> > There appears to be a number of empty files. Was this intended?
> 
> They are not empty, GH just reports nonsense here.

It does actually report the permissions change. Rather than "Empty file. " it should just display nothing, that would be clearer I think.

paulyoong (approved):

PR #956, opened 2021-06-23
fix: remove broken symlink to pool CRD

jkryl:
It should have been removed when moac was removed from the repo.  We
can live without the link because pool crd is created by moac when it
starts. People who wish to create pools before the moac is fully up
and running can employ different methods (i.e. waiting in loop until
the definition of crd appears or wait until moac liveness probe
confirms that moac has been fully started).


blaisedias (approved):
LGTM

paulyoong (approved):

jkryl:
bors merge

PR #958, opened 2021-06-28
test(py): add pytest-bdd feature tests for pools and replicas

cjones1024:
roughly translated from test/grpc/test_replica.js

resolves CAS-961

### DIFF
--- a/test/python/features/pool.feature
+++ b/test/python/features/pool.feature
@@ -1,0 +1,35 @@
+Feature: Mayastor pool management
+
+  Background:
+    Given a mayastor instance "ms0"
+
+  Scenario: creating a pool using disk with invalid block size
+    When the user attempts to create a pool specifying a disk with an invalid block size
+    Then the pool creation should fail
+
+  Scenario: creating a pool with multiple disks
+    When the user attempts to create a pool specifying multiple disks
+    Then the pool creation should fail
+
+  Scenario: creating a pool with an AIO disk
+    When the user creates a pool specifying a URI representing an aio disk
+    Then the pool should be created
+
+  Scenario: creating a pool with a name that already exists
+    Given a pool "p0"
+    When the user creates a pool with the name of an existing pool
+    Then the pool creation should succeed
+
+  Scenario: listing pools
+    Given a pool "p0"
+    When the user lists the current pools
+    Then the pool should appear in the output list
+
+  Scenario: destroying a pool
+    Given a pool "p0"
+    When the user destroys the pool
+    Then the pool should be destroyed
+
+  Scenario: destroying a pool that does not exist
+    When the user destroys a pool that does not exist
+    Then the pool destroy command should succeed
--- a/test/python/features/replica.feature
+++ b/test/python/features/replica.feature
@@ -1,0 +1,88 @@
+Feature: Mayastor replica management
+
+  Background:
+    Given a mayastor instance "ms0"
+    And a pool "p0"
+
+  Scenario: creating a replica
+    When the user creates an unshared replica
+    Then the replica is created
+    And the share state is unshared
+
+  Scenario: creating a replica shared over "iscsi"
+    When the user attempts to create a replica shared over "iscsi"
+    Then the create replica command should fail
+
+  Scenario: creating a replica with a name that already exists
+    Given a replica
+    When the user creates a replica that already exists
+    Then the create replica command should succeed
+
+  Scenario: listing replicas
+    Given a replica
+    When the user lists the current replicas
+    Then the replica should appear in the output list
+
+  Scenario: sharing a replica over "nvmf"
+    Given a replica that is unshared
+    When the user shares the replica over "nvmf"
+    Then the share state should change to "nvmf"
+
+  Scenario: sharing a replica over "iscsi"
+    Given a replica that is unshared
+    When the user attempts to share the replica over "iscsi"
+    Then the share replica command should fail
+
+  Scenario: sharing a replica that is already shared with the same protocol
+    Given a replica shared over "nvmf"
+    When the user shares the replica with the same protocol
+    Then the share replica command should succeed
+
+  Scenario: sharing a replica that is already shared with a different protocol
+    Given a replica shared over "nvmf"
+    When the user attempts to share the replica with a different protocol
+    Then the share replica command should fail
+
+  Scenario: unsharing a replica
+    Given a replica shared over "nvmf"
+    When the user unshares the replica
+    Then the share state should change to unshared
+
+  Scenario: destroying a replica
+    Given a replica
+    When the user destroys the replica
+    Then the replica should be destroyed
+
+  Scenario: destroying a replica that does not exist
+    When the user destroys a replica that does not exist
+    Then the replica destroy command should succeed
+
+  Scenario: listing replica stats
+    Given a replica
+    When the user gets replica stats
+    Then the stats for the replica should be listed
+
+  Scenario: creating a replica shared over "nvmf"
+    When the user creates a replica shared over "nvmf"
+    Then the replica is created
+    And the share state is "nvmf"
+
+  Scenario: listing replicas
+    Given a replica
+    When the user lists the current replicas
+    Then the replica should appear in the output list
+
+  Scenario: writing to a shared replica
+    Given a replica shared over "nvmf"
+    When the user writes to the replica
+    Then the write operation should succeed
+
+  Scenario: reading from a shared replica
+    Given a replica shared over "nvmf"
+    When the user reads from the replica
+    Then the read operation should succeed
+
+  Scenario: recreating a replica
+    Given a replica
+    When the user attempts to recreate the existing replica
+    Then the old data should have been reset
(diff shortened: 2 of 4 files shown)


test/python/test_pool.py
```diff
@@ -0,0 +1,164 @@
+
```
gila, line 1: IMHO it probably better to put this in a subdir called pool, with its own dockerfile perhaps?

test/python/test_pool.py
```diff
@@ -0,0 +1,164 @@
...
+@pytest.fixture
+def create_image_file():
+    filename = '/tmp/ms0-aio.img'
+    run_cmd(f"rm -f '{filename}'", True)
+    run_cmd(f"truncate -s 64M '{filename}'", True)
+    yield filename
+    run_cmd(f"rm -f '{filename}'", True)
+
+
+@pytest.fixture
```
gila, line 67: I see what you are trying to do here, but why not yield in create pool and then have it destroyed? Fixtures are nice but my (short) experience is that they can make things complicated and surprising when not needed.

test/python/test_pool.py
```diff
@@ -0,0 +1,164 @@
...
+def create_pool(mayastor_instance, mayastor_state):
+    mayastor_state['pool'] = mayastor_instance.ms.CreatePool(pb.CreatePoolRequest(name='p0', disks=['malloc:///disk0?size_mb=100']))
+
+
+@when('the user attempts to create a pool specifying a URI representing an aio disk')
+def create_pool_from_aio_disk(mayastor_instance, create_image_file, mayastor_state):
+    try:
+        mayastor_state['pool'] = mayastor_instance.ms.CreatePool(pb.CreatePoolRequest(name='p0', disks=[f"aio://{create_image_file}"]))
+    except Exception as error:
+        mayastor_state['error'] = error
```
gila, line 89: I see several of these, why not raise the exception? (i.e test failed).

test/python/test_pool.py
```diff
@@ -0,0 +1,164 @@
...
+@when('the user lists the current pools')
+def list_pools(mayastor_instance, mayastor_state):
+    try:
+        mayastor_state['pool_list'] = mayastor_instance.ms.ListPools(pb.Null(), wait_for_ready=True).pools
+    except Exception as error:
+        mayastor_state['error'] = error
+
+
+@then('the pool creation should fail')
+def pool_creation_should_fail(find_pool, mayastor_state):
```
gila, line 142: why not call this when the test should fail rather than storing an error which might be clobbered?

gila (changes requested):

test/python/test_pool.py
cjones1024: Yes I think this should happen eventually, but I also think there might be scope for further reorganisation as well.

test/python/test_replica.py
```diff
@@ -0,0 +1,317 @@
+
+import pytest
```
gila, line 2: The files do not seem to be properly formatted. I've used `black` on the other files and its also run by the precommit hooks

gila (changes requested):

test/python/test_pool.py
cjones1024: Some (although by no means all) tests require the gRPC command to fail so, so raising (or rather not catching) the exception is not always appropriate. The real reason however is an attempt to adhere to the BDD flow: where a 'given' clause sets up the environment, the 'when' clause performs the action under test, and it is the 'then' clause that is responsible for validating the behaviour.

test/python/test_pool.py
cjones1024: create_pool is a (given) step function, and is definitely not a fixture. Placing a yield here causes an error (even if nothing happens subsequently. I'm not sure if this is a bug as such or by design.
In any case create_pool is not supposed to destroy the pool as its job is to create the pool in order to test the DestroyPool command.

test/python/test_pool.py
cjones1024: As I understand it, this is not really a normal function to be invoked directly, it is much like a fixture and is invoked appropriately via the pytest infrastructure - which ensures that the argument values (again fixtures) are only evaluated once and the results cached appropriately.

test/python/test_pool.py
```diff
@@ -0,0 +1,164 @@
...
+        mayastor_state['pool'] = mayastor_instance.ms.CreatePool(pb.CreatePoolRequest(name='p0', disks=[f"aio://{create_image_file}"]))
+    except Exception as error:
+        mayastor_state['error'] = error
+
+
+@when('the user attempts to create a pool specifying a disk with an invalid block size')
+def create_pool_from_disk_with_an_invalid_block_size(mayastor_instance, mayastor_state):
+    try:
+        mayastor_state['pool'] = mayastor_instance.ms.CreatePool(pb.CreatePoolRequest(name='p0', disks=['malloc:///disk0?size_mb=100&blk_size=1024']))
+    except Exception as error:
```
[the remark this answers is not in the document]
cjones1024: There is no straightforward mechanism for passing results from the 'when' step to the 'then' step. Creating a context representing the state of the test allows this communication. It is not Ideal but I have not yet mangaged to come up with a better approach

test/python/test_replica.py
cjones1024: reformatted via black as per precommit hook

test/python/test_pool.py
```diff
@@ -0,0 +1,171 @@
...
+    "Destroying a pool that does not exist."
+
+
+@scenario("features/pool.feature", "listing pools")
+def test_listing_pools():
+    "Listing pools."
+
+
+@pytest.fixture(scope="module")
+def mayastor_instance(mayastor_mod):
```
gila, line 45: Perhaps naming this ms0 is more clear?

test/python/test_pool.py
```diff
@@ -0,0 +1,171 @@
...
+
+@pytest.fixture(scope="module")
+def find_pool(mayastor_instance):
+    def lookup(name):
+        for pool in mayastor_instance.ms.ListPools(pb.Null()).pools:
+            if pool.name == name:
+                return pool
+        return None
+
+    yield lookup
```
gila, line 57: There is no reason to yield here, only yield what you want something run after the function exits.

test/python/test_pool.py
```diff
@@ -0,0 +1,171 @@
...
+    )
+
+
+@when("the user attempts to create a pool specifying a URI representing an aio disk")
+def create_pool_from_aio_disk(mayastor_instance, create_image_file, mayastor_state):
+    mayastor_state["pool"] = mayastor_instance.ms.CreatePool(
+        pb.CreatePoolRequest(name="p0", disks=[f"aio://{create_image_file}"])
+    )
+
+
```
gila, line 95: The two functions above both create a pool and one may clobber the other. Why not write a simple function like "get_pool()" and then call the gRPC method?

test/python/test_pool.py
```diff
@@ -0,0 +1,171 @@
...
+def create_pool_that_already_exists(mayastor_instance, mayastor_state):
+    mayastor_instance.ms.CreatePool(
+        pb.CreatePoolRequest(name="p0", disks=["malloc:///disk0?size_mb=100"])
+    )
+
+
+@when("the user attempts to destroy a pool that does not exist")
+def destroy_pool_that_does_not_exist(mayastor_instance, mayastor_state):
+    mayastor_instance.ms.DestroyPool(pb.DestroyPoolRequest(name="p0"))
+
```
gila, line 133: Why not simply pass pool name?

test/python/test_pool.py
```diff
@@ -0,0 +1,171 @@
...
+
+@when("the user attempts to destroy the pool")
+def destroy_pool(mayastor_instance, mayastor_state):
+    mayastor_instance.ms.DestroyPool(
+        pb.DestroyPoolRequest(name=mayastor_state["pool"].name)
+    )
+    del mayastor_state["pool"]
+
```
gila, line 141: Same as above 👆

test/python/test_pool.py
```diff
@@ -0,0 +1,171 @@
...
+@then("the pool destroy command should succeed")
+@then("the pool should be destroyed")
+def pool_destruction_should_succeed(find_pool):
+    assert find_pool("p0") == None
+
+
+@then("the pool should appear in the output list")
+def pool_should_appear_in_output(mayastor_state):
+    assert mayastor_state["pool"].name in [
+        pool.name for pool in mayastor_state["pool_list"]
```
gila, line 170: Again why not call list pool and over that?

gila (commented):
@cjones1024 I don't fully get the reasoning behind the cache still. In fact pytest itself already does this where applicable

test/python/test_pool.py
cjones1024: done

cjones1024:
After refactoring, some of the above comments are no longer applicable. Please treat effectively as a new PR

jonathan-teh (approved):

gila (approved):

cjones1024:
bors merge

PR #959, opened 2021-06-29
fix(grpc): assert in gRPC serialiser

mtzaurus:
- Mayastor gRPC serializer now uses enriched context to capture detailed
information about the method being invoked and its arguments.
- Assert in serializer's guard removed.
- Configurable gRPC client timeout for Mayastor handles in Python tests.

Implements CAS-956


test/python/common/hdl.py
```diff
@@ -12,16 +14,30 @@ class MayastorHandle(object):
     def __init__(self, ip_v4):
         """Init."""
         self.ip_v4 = ip_v4
+        self.timeout = config["grpc"]["client_timeout"]
```
gila, line 17: nice! thanks

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -32,36 +33,75 @@ use crate::{
...
 use tonic::{Request, Response, Status};
 #[derive(Debug)]
 struct UnixStream(tokio::net::UnixStream);
 
+use ::function_name::named;
+use std::panic::AssertUnwindSafe;
+
+impl GrpcClientContext {
+    #[track_caller]
+    pub fn new<T>(req: &Request<T>, fid: &str) -> Self
```
gila, line 49: Im fine with this. However, IMHO, if we have T + Debug, then we don't need the fid. Through the request, we *know* what method it should have been, right? Does it matter? Not really,  but fewer crates to carry around is always better.

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -32,36 +33,75 @@ use crate::{
...
+        // request. Even though AssertUnwindSafe() allows us to
+        // intercept asserts in underlying method strategies, such a
+        // situation can still happen when the high-level future that
+        // represents gRPC call at the highest level (i.e. the one created
+        // by gRPC server) gets cancelled (due to timeout or somehow else).
+        // This can't be properly intercepted by 'locked' function itself in the
+        // first place, so the state needs to be cleaned up properly
+        // upon subsequent gRPC calls.
+        if let Some(c) = guard.replace(ctx) {
+            warn!("{}: gRPC method timed out, args: {}", c.id, c.args);
```
gila, line 86: We did not timeout; the callee lost its patience, small nit but can be confusing for people when they read this? how about, "grpc client disappeared" it something like that?

test/python/test_rpc.py
```diff
@@ -0,0 +1,54 @@
+from common.hdl import MayastorHandle
```
gila, line 1: Nice, thank you!

mayastor/src/grpc/mayastor_grpc.rs
mtzaurus: I am not sure if it's possible to get method's name from tonic::Request, as it primarily represents the method arguments themselves.
In fact, printing the request via {:?} doesn't show any presence of the method name:
```
Request { metadata: MetadataMap { headers: {"te": "trailers", "content-type": "application/grpc", "user-agent": "tonic/0.4.3"} }, message: DestroyPoolRequest { name: "xyz" }, extensions: Extensions }
```

I presume the decision about method name has already been made by a higher-level logic of Tonic before dispatching the request to the method, so the method name is not needed in the request.

mayastor/src/grpc/mayastor_grpc.rs
mtzaurus: Sure thing, fixed.

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -32,36 +33,75 @@ use crate::{
...
+        }
+
+        let fut = AssertUnwindSafe(f).catch_unwind();
+        let r = fut.await;
+
+        // Request completed, remove the marker.
+        let ctx = guard.take().expect("gRPC context disappeared");
```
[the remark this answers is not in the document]
mtzaurus: Yes, this should never ever happened, as the lock is still held. So we could use 'unwrap()' here, but 'expect()' looks safer to me. "Trust no one" (c)

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -32,36 +33,75 @@ use crate::{
...
+        // represents gRPC call at the highest level (i.e. the one created
+        // by gRPC server) gets cancelled (due to timeout or somehow else).
+        // This can't be properly intercepted by 'locked' function itself in the
+        // first place, so the state needs to be cleaned up properly
+        // upon subsequent gRPC calls.
+        if let Some(c) = guard.replace(ctx) {
+            warn!("{}: gRPC method timed out, args: {}", c.id, c.args);
+        }
+
+        let fut = AssertUnwindSafe(f).catch_unwind();
```
tiagolobocastro, line 89: hmm so we allow the gRPC to be retried because the tonic future but the actual mayastor future sent with "rpc_submit" is still running, so we loose the "serialization" that this code was adding to begin with.
Is there anything that can be done about that?

mayastor/src/grpc/mayastor_grpc.rs
gila: Yes but, look:
```
DestroyPoolRequest {}...
```

That gives us all the info already, no?

gila: Yes but notice that in the Request we see it's a DestroyPoolRequest... so we have all we need?

mtzaurus: If we don't want to display the name of the method which failed (like: "pool_destroy(): method failed, args: {}", but display only args, then it's fine.

mayastor/src/grpc/mayastor_grpc.rs
mtzaurus: This situation is two-fold. On one hand we have a running rpc_submit-related future, which might result in a de-serialised invocation as you are pointing out. So it would be nice to make sure this future has completed before proceeding. And this is doable, IMHO (we could introduce one-shot channel and make the receiver part be part of GRPC context, so we can complete the call the next time we see the existing context). But then another question arises: for how long should we wait to not block forever, as we want to avoid a situation when one blocking call blocks all other methods.
We definitely can improve our gRPC serialisation (use fine-grained locks instead of one global lock, etc), but I would do it on behalf of another changeset, where we can also handle this 'serialisation escape' question.

mayastor/src/grpc/mayastor_grpc.rs
gila: As discussed lets leave it as-is :+1:

gila (approved):

mayastor/src/grpc/mayastor_grpc.rs
tiagolobocastro: Yeah, I'm not sure how we'd want to handle the case a call is taking forever, though we could argue it can still happen anyway if the gRPC client does not set a timeout.
But yeah, not sure what we'd want to do... @gila?

tiagolobocastro (approved):

PR #966, opened 2021-07-02
ci(fix): remove the need to specify cargosha256

gila:
By not pulling in dependencies from git directly we can get rid of the
need to calculate the sha and use nix builtins.

The git dependencies we used where:

 - h2 (to fix a k8s issue)
 - partition_id
 - blkid

 The changes to h2 are perhaps not needed anymore as the newer k8s
 clients should have the issue of authority IDs fixed.

 To get around the partition and blkid dependencies, I've implemented
 the required pieces myself basing it those crates which are not
 actively maintained.

### DIFF
--- a/devinfo/src/partition/mod.rs
+++ b/devinfo/src/partition/mod.rs
@@ -1,0 +1,330 @@
+use self::PartitionSource::{Path as SourcePath, *};
+use std::{
+    borrow::Cow,
+    fmt::{self, Display, Formatter},
+    fs,
+    io::{Error, ErrorKind},
+    path::{Path, PathBuf},
+    str::FromStr,
+};
+
+/// Describes a partition identity.
+///
+/// A device path may be recovered from this.
+///
+/// # Notes
+///
+/// This is a struct instead of an enum to make access to the `id` string
+/// easier for situations where the variant does not need to be checked.
+#[derive(Clone, Debug, Hash, Eq, PartialEq)]
+pub struct PartitionID {
+    pub variant: PartitionSource,
+    pub id: String,
+}
+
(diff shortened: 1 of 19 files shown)


gila:
bors try

gila:
bors try

tiagolobocastro (approved):

gila:
bors try

gila:
bors try

gila:
bors try

gila:
bors try

gila:
bors try

paulyoong (approved):

devinfo/src/partition/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+            ::std::thread::sleep(::std::time::Duration::from_millis(wait));
+            tried += 1;
+        } else {
+            return result;
+        }
+    }
+}
+
+fn canonicalize(path: &Path) -> Cow<'_, Path> {
+    // NOTE: It seems that the kernel may intermittently error.
```
jonathan-teh, line 276: Is there a link to an upstream bug report? Same question for the skipping directories issues below.

jonathan-teh (approved):

devinfo/src/partition/mod.rs
gila: I don't know. This code is mostly imported from the unmaintained partition identity crate. I've only made changes that were needed in order to get it to work for our use case and made sure we only imported stuff we we actually need/use. I'm not even sure if this is stuff we need in the long run as udev can also get all of this info for us.

gila:
bors merge

PR #969, opened 2021-07-07
feat(nexus): add create_v2 gRPC with NVMe parameters

jonathan-teh:
Create a separate NexusNvmeParams structure with NVMe controller ID
range for sharing over NVMf and the NVMe reservation key for children,
add it to the Nexus structure, and add it to nexus_create_v2() as an
additional argument.

Fixes CAS-971


jonathan-teh:
Sending for review while I work on Python tests.

jonathan-teh:
bors try

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -1080,12 +1100,40 @@ pub async fn nexus_create(
         return Ok(());
     }
 
+    let (min_cntlid, max_cntlid, resv_key) =
+        nvme_params.unwrap_or((NVME_MIN_CNTLID, NVME_MAX_CNTLID, 0x1234_5678));
+    if min_cntlid < NVME_MIN_CNTLID
```
cjones1024, line 1105: There is some redundancy here: we only really need: min_cntlid < NVME_MIN_CNTLID || max_cntlid < min_cntlid || NVME_MAX_CNTLID < max_cntlid

jonathan-teh: Good spot. The upstream SPDK code is actually similarly redundant.

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -248,6 +251,10 @@ pub enum Error {
     InvalidShareProtocol { sp_value: i32 },
     #[snafu(display("Invalid NvmeAnaState value {}", ana_value))]
     InvalidNvmeAnaState { ana_value: i32 },
+    #[snafu(display("Invalid NVMe controller ID range for nexus {}", name))]
+    InvalidNvmeControllerIdRange { name: String },
+    #[snafu(display("Invalid NVMe reservation key for nexus {}", name))]
+    InvalidNvmeReservationKey { name: String },
```
gila, line 257: Do we really need to keep adding new errors for invalid arguments? Why can't we have that FAR more generic? like InvalidArgument { my details } ?

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -355,6 +362,12 @@ pub struct Nexus {
     pub(crate) child_count: u32,
     /// vector of children
     pub children: Vec<NexusChild>,
+    /// minimum NVMe controller ID for sharing over NVMf
```
gila, line 365: How about we put this in a struct?

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -1065,6 +1084,7 @@ pub async fn nexus_create(
     size: u64,
     uuid: Option<&str>,
     children: &[String],
+    nvme_params: Option<(u16, u16, u64)>,
```
gila, line 1087: This is rather cumbersome no? If we put this in a struct it is much easier later one when we add/remove arguments?

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -1080,12 +1100,40 @@ pub async fn nexus_create(
         return Ok(());
     }
 
+    let (min_cntlid, max_cntlid, resv_key) =
```
gila, line 1103: This validation could be done then as an impl on the struct rather than adding this

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -1080,12 +1100,40 @@ pub async fn nexus_create(
...
+        nvme_params.unwrap_or((NVME_MIN_CNTLID, NVME_MAX_CNTLID, 0x1234_5678));
+    if min_cntlid < NVME_MIN_CNTLID
+        || min_cntlid > NVME_MAX_CNTLID
+        || max_cntlid < NVME_MIN_CNTLID
+        || max_cntlid > NVME_MAX_CNTLID
+        || min_cntlid > max_cntlid
+    {
+        error!(
```
gila, line 1111: Likewise, the the error could put in the function that does the checking.

mayastor/src/bdev/nexus/nexus_config.rs
```diff
@@ -1,110 +0,0 @@
-use std::{ffi::CStr, ptr::null_mut, str::FromStr};
```
gila, line 1: Thanks for getting rid of that :+1:

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -1065,6 +1084,7 @@ pub async fn nexus_create(
     size: u64,
     uuid: Option<&str>,
     children: &[String],
```
gila, line 1086: I thought we wanted *new* method nexus_create_v2(). This will also then allow us to fix more buried issues we have with the nexus in terms of adding locks?

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -1080,12 +1100,40 @@ pub async fn nexus_create(
...
+    }
+
     // Create a new Nexus object, and immediately add it to the global list.
     // This is necessary to ensure proper cleanup, as the code responsible for
     // closing a child assumes that the nexus to which it belongs will appear
     // in the global list of nexus instances. We must also ensure that the
     // nexus instance gets removed from the global list if an error occurs.
-    nexus_list.push(Nexus::new(name, size, uuid, None));
+    nexus_list.push(Nexus::new(
+        name, size, uuid, None, min_cntlid, max_cntlid, resv_key,
```
gila, line 1135: If we put this in a struct it will be much nicer...

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -1080,12 +1100,40 @@ pub async fn nexus_create(
         return Ok(());
     }
 
+    let (min_cntlid, max_cntlid, resv_key) =
+        nvme_params.unwrap_or((NVME_MIN_CNTLID, NVME_MAX_CNTLID, 0x1234_5678));
```
gila, line 1104: in a struct we could return defaults.

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -1080,12 +1100,40 @@ pub async fn nexus_create(
...
+    if min_cntlid < NVME_MIN_CNTLID
+        || min_cntlid > NVME_MAX_CNTLID
+        || max_cntlid < NVME_MIN_CNTLID
+        || max_cntlid > NVME_MAX_CNTLID
+        || min_cntlid > max_cntlid
```
gila, line 1109: We perhaps should check these numbers before pass the RPC request to mayastor?

mayastor/src/bdev/nexus/nexus_bdev.rs
jonathan-teh: Switched to `InvalidArguments` error type common to both.

mayastor/src/bdev/nexus/nexus_bdev.rs
jonathan-teh: Done, `NexusNvmeParams`.

mayastor/src/bdev/nexus/nexus_bdev.rs
jonathan-teh: Done, for now it only has an extra `NexusNvmeParams` parameter.

mayastor/src/bdev/nexus/nexus_bdev.rs
jonathan-teh: There is now a `impl Default for NexusNvmeParams`.

mayastor/src/bdev/nexus/nexus_bdev.rs
jonathan-teh: I prefer to validate at the lowest level to avoid any codepath in the future that manages to bypass that checking.

jonathan-teh:
Now includes Python tests.

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -1065,6 +1107,59 @@ pub async fn nexus_create(
...
+        NexusNvmeParams::default(),
+        children,
+    )
+    .await
+}
+
+/// As create_nexus with additional parameters:
+/// min_cntlid, max_cntldi: NVMe controller ID range when sharing over NVMf
+/// resv_key: NVMe reservation key for children
+pub async fn nexus_create_v2(
```
tiagolobocastro, line 1124: Will this include the changes we spoke about in the call the other day about each nexus having a unique uuid and a separate uuid for the namespace? Or will that be part of a separate PR?
Can/Should the control plane make use of this new method for non ANA volumes?

jonathan-teh: No, the UUID changes are not part of this ticket, it will have to be part of a separate ticket (has one been raised).
The new method can be used as long as `NexusNvmeParams` contains default values as that's what the existing method does.

gila (approved):

cjones1024 (approved):

mtzaurus (approved):

jonathan-teh:
bors merge

PR #973, opened 2021-07-13
feat(nexus): change create_v2 gRPC to have separate name and uuid

jonathan-teh:
Add a separate name field to the CreateNexusV2 gRPC, which is a
freeform string used directly in the Nexus structure in Rust instead
of being derived from the uuid field by prefixing "nexus-" to it.

Fixes CAS-1007

### DIFF
--- a/mayastor/src/grpc/nexus_grpc.rs
+++ b/mayastor/src/grpc/nexus_grpc.rs
@@ -73,6 +73,22 @@ impl Nexus {
             rebuilds: RebuildJob::count() as u32,
         }
     }
+
+    pub fn to_grpc_v2(&self) -> rpc::NexusV2 {
+        rpc::NexusV2 {
+            name: name_to_uuid(&self.name).to_string(),
+            uuid: self.bdev.uuid().to_string(),
+            size: self.size,
+            state: rpc::NexusState::from(self.status()) as i32,
+            device_uri: self.get_share_uri().unwrap_or_default(),
+            children: self
+                .children
+                .iter()
+                .map(|ch| ch.to_grpc())
+                .collect::<Vec<_>>(),
+            rebuilds: RebuildJob::count() as u32,
+        }
+    }
 }
 
 /// Convert nexus name to uuid.
@@ -99,10 +115,14 @@ pub fn uuid_to_name(uuid: &str) -> Result<String, Error> {
     }
 }
 
-/// Lookup a nexus by its uuid prepending "nexus-" prefix. Return error if
-/// uuid is invalid or nexus not found.
+/// Look up a nexus by its uuid prepending "nexus-" prefix, or by name if not
+/// a valid uuid (if created by nexus_create_v2).
+/// Return error if nexus not found.
 pub fn nexus_lookup(uuid: &str) -> Result<&mut Nexus, Error> {
-    let name = uuid_to_name(uuid)?;
+    let name = match uuid_to_name(uuid) {
+        Ok(name) => name,
+        Err(_) => uuid.to_string(),
+    };
 
     if let Some(nexus) = instances().iter_mut().find(|n| n.name == name) {
         Ok(nexus)
--- a/rpc/proto/mayastor.proto
+++ b/rpc/proto/mayastor.proto
@@ -43,6 +43,7 @@ service Mayastor {
   rpc CreateNexusV2 (CreateNexusV2Request) returns (Nexus) {}
   rpc DestroyNexus (DestroyNexusRequest) returns (Null) {}
   rpc ListNexus (Null) returns (ListNexusReply) {}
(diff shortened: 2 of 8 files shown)


jonathan-teh:
bors try

mayastor/src/grpc/nexus_grpc.rs
```diff
@@ -73,6 +73,22 @@ impl Nexus {
             rebuilds: RebuildJob::count() as u32,
         }
     }
+
+    pub fn to_grpc_v2(&self) -> rpc::NexusV2 {
```
tiagolobocastro, line 77: Should this be an impl From<&Nexus> to rpc::NexusV2 ?

rpc/proto/mayastor.proto
```diff
@@ -258,6 +260,23 @@ message ListNexusReply {
   repeated Nexus nexus_list = 1;
 }
 
+// represents a nexus device
+message NexusV2 {
+  string name = 1;             // name of the nexus
+  string uuid = 2;             // UUID of the bdev
+  uint64 size = 3;             // size of the volume in bytes
+  NexusState state = 4;        // current state of the nexus
```
tiagolobocastro, line 268: Now that we have a NexusV2 (so MOAC won't be affected), could we return the actual nexus and children states and not the "simplified" states that MOAC wanted?

tiagolobocastro (approved):

mayastor/src/grpc/nexus_grpc.rs
jonathan-teh: There's actually a code comment above `to_grpc(&self)`:
```
    /// We cannot use From trait because it is not value to value conversion.
    /// All we have is a reference to nexus.
```

rpc/proto/mayastor.proto
jonathan-teh: From what I can tell, it's only the `ChildState` that's mapped into fewer `rpc::ChildState`s, do you mean you want `ChildState`s return directly (I assume the reason is dropped from the `Faulted` state)?

mayastor/src/grpc/nexus_grpc.rs
tiagolobocastro: hmm that comment doesn't seem quite right, you can have traits for references?

rpc/proto/mayastor.proto
tiagolobocastro: The NexusState is also mapped IIRC, there's no Init or Closed in the protobuf.
Yes I think at some point we discussed getting the states returned directly @gila @paulyoong  ?
Yeah the reason is dropped from the Faulted state, I'm actually not sure how we'd do that in protobuf :/

mayastor/src/grpc/nexus_grpc.rs
jonathan-teh: Seems to the Jan's comment. Anyway, if I do have an `impl From<&Nexus> for rpc::NexusV2` and try to use an `into`:
```
error[E0277]: the trait bound `NexusV2: From<&Box<nexus_bdev::Nexus>>` is not satisfied
   --> mayastor/src/grpc/mayastor_grpc.rs:652:32
    |
652 |                     .map(|n| n.into())
    |                                ^^^^ the trait `From<&Box<nexus_bdev::Nexus>>` is not implemented for `NexusV2`
    |
    = help: the following implementations were found:
              <NexusV2 as From<&nexus_bdev::Nexus>>
    = note: required because of the requirements on the impl of `Into<NexusV2>` for `&Box<nexus_bdev::Nexus>`
```
You can try double-dereferencing the `&Box` to get a `nexus_bdev::Nexus`, then you either change to `impl From<&Nexus>` and get a missing Copy trait, or try `&(**n).into()` and get `expected struct ``NexusV2``, found reference`. I'm just going to leave it alone.

rpc/proto/mayastor.proto
jonathan-teh: In the case of `NexusState` the mapping is done in `nexus_bdev` itself, not just in the RPC layer. This is looking to be more than a quick bug fix, please raise a separate ticket with some requirements.

mayastor/src/grpc/nexus_grpc.rs
tiagolobocastro: Box impl's deref, so just need to ".map(|n| n.deref().into()".
But yeah leave it at that for now

rpc/proto/mayastor.proto
tiagolobocastro: We have this old one, I guess it fits: https://mayadata.atlassian.net/browse/CAS-608

paulyoong (approved):

jonathan-teh:
bors merge

PR #974, opened 2021-07-13
chore(build): rust-overlay and container image

gila:
This removes the test binaries from the container image as they are not needed.
Also, add /bin to PATH and wrap mayastor-client to mctl.

Lastly, the mozilla-overlay isn't very flexible to use. Switch to a
more flexible overlay which, for example, does not take forever to copy
HTML files when building images.

The second commit bumps cargo deps.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -46,9 +46,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.41"
+version = "1.0.42"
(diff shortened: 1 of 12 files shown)


gila:
bors try

paulyoong (approved):

gila:
bors try

tiagolobocastro (approved):

PR #975, opened 2021-07-15
test(py): add pytest-bdd feature tests for nexus

cjones1024:
- roughly translated from test/grpc/test_nexus.js

- also added non-bdd tests which tests create/publish/unpublish/destroy
  sequence of operations for each of the main protocols

- resolves CAS-968

### DIFF
--- a/test/python/features/nexus.feature
+++ b/test/python/features/nexus.feature
@@ -1,0 +1,97 @@
+Feature: Mayastor nexus management
+
+  Background:
+    Given a local mayastor instance
+    And a remote mayastor instance
+
+  Scenario: creating a nexus
+    Given a list of child devices
+    When creating a nexus
+    Then the nexus should be created
+
+  Scenario: creating the same nexus
+    Given a nexus
+    When creating an identical nexus
+    Then the operation should succeed
+
+  Scenario: creating a different nexus with the same children
+    Given a nexus
+    When attempting to create a new nexus
+    Then the operation should fail with existing URIs in use
+
+  Scenario: creating a nexus without children
+    When attempting to create a nexus with no children
+    Then the operation should fail
+
+  Scenario: creating a nexus with missing children
+    When attempting to create a nexus with a child URI that does not exist
+    Then the operation should fail
+
+  Scenario: creating a nexus from children with mixed block sizes
+    When attempting to create a nexus from child devices with mixed block sizes
+    Then the nexus should not be created
+
+  Scenario: creating a nexus that is larger than its children
+    Given a list of child devices
+    When attempting to create a nexus with a size larger than that of any of its children
+    Then the nexus should not be created
+
+  Scenario: destroying a nexus
+    Given a nexus
+    When destroying the nexus
+    Then the nexus should be destroyed
+
+  Scenario: destroying a nexus without first unpublishing
+    Given a nexus published via "nvmf"
+    When destroying the nexus without unpublishing it
+    Then the nexus should be destroyed
+
+  Scenario: destroying a nexus that does not exist
+    When destroying a nexus that does not exist
+    Then the operation should succeed
+
+  Scenario: listing nexuses
+    Given a nexus
+    When listing all nexuses
+    Then the nexus should appear in the output list
+
+  Scenario: removing a child from a nexus
+    Given a nexus
+    When removing a child
+    Then the child should be successfully removed
+
+  Scenario: adding a child to a nexus
+    Given a nexus
+    When removing a child
+    And adding the removed child
+    Then the child should be successfully added
+
+  Scenario: publishing a nexus
+    Given a nexus
+    When publishing a nexus via "nvmf"
+    Then the nexus should be successfully published
+
+  Scenario: unpublishing a nexus
+    Given a nexus published via "nvmf"
+    When unpublishing the nexus
+    Then the nexus should be successfully unpublished
+
+  Scenario: unpublishing a nexus that is not published
+    Given an unpublished nexus
+    When unpublishing the nexus
+    Then the request should succeed
+
+  Scenario: republishing a nexus with the same protocol
+    Given a nexus published via "nvmf"
+    When publishing the nexus with the same protocol
+    Then the nexus should be successfully published
+
+  Scenario: republishing a nexus with a different protocol
+    Given a nexus published via "nvmf"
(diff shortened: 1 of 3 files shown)


test/python/test_bdd_nexus.py
```diff
@@ -0,0 +1,494 @@
...
+        for nexus in mayastor_mod[nexus_instance].ms.ListNexus(pb.Null()).nexus_list:
+            if nexus.uuid == uuid:
+                return nexus
+        return None
+
+    yield find
+
+
+@pytest.fixture
+def local_bdev_with_different_blocksize(mayastor_mod, nexus_instance):
```
jonathan-teh, line 215: Perhaps name this "with_512b_block_size" as it's not immediately clear what different means.

jonathan-teh (approved):

test/python/test_bdd_nexus.py
cjones1024: done

test/python/test_bdd_nexus.py
```diff
@@ -0,0 +1,494 @@
...
+
+
+@then("the nexus should not be created")
+def nexus_should_not_be_created(find_nexus, nexus_uuid):
+    assert find_nexus(nexus_uuid) == None
+
+
+@then("the nexus should be destroyed")
+def nexus_should_be_destroyed(find_nexus, nexus_uuid):
+    assert find_nexus(nexus_uuid) == None
```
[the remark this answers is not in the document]
cjones1024: While I agree, I'm inclined to leave it as is for now.

test/python/test_bdd_nexus.py
```diff
@@ -0,0 +1,494 @@
+import pytest
+from pytest_bdd import given, scenario, then, when, parsers
+
+from collections import namedtuple
+import os
```
gila, line 5: Does not seem to be used? Python has very rich tooling in terms of linters, checkers etc... maybe install a few?

test/python/test_bdd_nexus.py
```diff
@@ -0,0 +1,494 @@
...
+
+@pytest.fixture(scope="module")
+def find_nexus(mayastor_mod, nexus_instance):
+    def find(uuid):
+        for nexus in mayastor_mod[nexus_instance].ms.ListNexus(pb.Null()).nexus_list:
+            if nexus.uuid == uuid:
+                return nexus
+        return None
+
+    yield find
```
gila, line 211: what does this yield do?

test/python/test_nexus_publish.py
```diff
@@ -0,0 +1,210 @@
...
+):
+    for n in range(5):
+        uuid = get_uuid(n)
+        create_nexus(uuid, megabytes(64), nexus_children)
+        publish_nexus(uuid, share_protocol)
+        unpublish_nexus(uuid)
+        destroy_nexus(uuid)
+    assert nexus_count() == 0
+
+
```
gila, line 196: @pytest.mark.parametrize('repeat', range(5))

test/python/test_bdd_nexus.py
cjones1024: This is fixture that returns a function as its value. It is used to check if a nexus "currently" exists. We cannot create a simple fixture to return the nexus, as it would only be evaluated once per test (the result is cached and reused everywhere the fixture is referenced), and this is usually not what we want. We could use an ordinary function, but then it could only be called within a fixture and each calling fixture would have to request all of the fixtures whose values the function requires. Doing it this way means the dependent fixtures are only declared once.

test/python/test_bdd_nexus.py
cjones1024: Will look into adding extra linter(s) to the commit hook, but have removed the unused import for now.

test/python/test_nexus_publish.py
cjones1024: done

test/python/features/nexus.feature
```diff
@@ -0,0 +1,98 @@
...
+    Then the nexus should be successfully published
+
+  Scenario: republishing a nexus with a different protocol
+    Given a nexus published via "nvmf"
+    When attempting to publish the nexus with a different protocol
+    Then the request should fail
+
+  Scenario: publishing a nexus with a crypto-key
+    Given a nexus published via "nvmf"
+    Given an unpublished nexus
```
paulyoong, line 96: Did you intend to have two "Given" statements here? If so, I assume the "When" statement refers to the unpublished nexus?

test/python/test_bdd_nexus.py
```diff
@@ -0,0 +1,493 @@
...
+
+
+@when("removing a child")
+def remove_child(mayastor_mod, nexus_instance, nexus_uuid, nexus_children):
+    mayastor_mod[nexus_instance].ms.RemoveChildNexus(
+        pb.RemoveChildNexusRequest(uuid=nexus_uuid, uri=nexus_children[0])
+    )
+
+
+@when("adding the removed child")
```
paulyoong, line 424: It says adding a removed child but it seems like we are only adding a child.

paulyoong (approved):

test/python/features/nexus.feature
cjones1024: Yes this is a mistake. The first "Given" clause should not be there.  Have  now corrected the "scenario" accordingly

test/python/test_bdd_nexus.py
cjones1024: This is a bit confusing when seen out of context. Hopefully it makes sense when you read the entire scenario to which it belongs

test/python/features/nexus.feature
```diff
@@ -0,0 +1,98 @@
...
+
+  Scenario: destroying a nexus
+    Given a nexus
+    When destroying the nexus
+    Then the nexus should be destroyed
+
+  Scenario: destroying a nexus without first unpublishing
+    Given a nexus published via "nvmf"
+    When destroying the nexus without unpublishing it
+    Then the nexus should be destroyed
```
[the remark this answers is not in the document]
cjones1024: I could easily argue this caes yes. However the fact is that, at present, if you destroy a nexus that is still published, mayastor will implicitly unpublish first and then (successfully) destroy the nexus. The test therefore accurately reflects current behaviour.

cjones1024:
bors merge

PR #977, opened 2021-07-15
fix(build): ship mayastor-csi in container

gila:
All the containers in fact use the same images. So the CSI agent should
actually be part of that.


chriswldenyer (approved):

blaisedias (approved):
LGTM

PR #980, opened 2021-07-22
feat(nexus): allow 2nd nexus to connect to replica

jonathan-teh:
Change the NVMe reservation type to write exclusive, all registrants,
which allows all registrants (any host that registers a key, any key),
write access to the replica. This also treats all registrants as
reservation holders.

Update the existing nexus_resv_acquire cargo test to verify that both
nexuses are registered to the replica's NVMe target. Add a python test
to verify that using the kernel NVMe initiator.

Fixes CAS-973

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_child.rs
+++ b/mayastor/src/bdev/nexus/nexus_child.rs
@@ -435,7 +435,7 @@ impl NexusChild {
                 key,
                 0,
                 nvme_reservation_acquire_action::ACQUIRE,
-                nvme_reservation_type::WRITE_EXCLUSIVE,
+                nvme_reservation_type::WRITE_EXCLUSIVE_ALL_REGS,
             )
             .await
         {
@@ -457,7 +457,7 @@ impl NexusChild {
                     key,
                     pkey,
                     nvme_reservation_acquire_action::PREEMPT,
-                    nvme_reservation_type::WRITE_EXCLUSIVE,
+                    nvme_reservation_type::WRITE_EXCLUSIVE_ALL_REGS,
                 )
                 .await?;
                 if let Some((_, hostid)) = self.resv_report(&*hdl).await? {
--- a/mayastor/tests/nexus_multipath.rs
+++ b/mayastor/tests/nexus_multipath.rs
@@ -25,6 +25,7 @@ static NXNAME: &str = "nexus0";
 static UUID: &str = "cdc2a7db-3ac3-403a-af80-7fadc1581c47";
 static HOSTNQN: &str = "nqn.2019-05.io.openebs";
 static HOSTID0: &str = "53b35ce9-8e71-49a9-ab9b-cba7c5670fad";
+static HOSTID1: &str = "c1affd2d-ef79-4ba4-b5cf-8eb48f9c07d0";
 
 fn nvme_connect(target_addr: &str, nqn: &str) {
     let status = Command::new("nvme")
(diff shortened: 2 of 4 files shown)


mayastor/src/bdev/nexus/nexus_child.rs
```diff
@@ -435,7 +435,7 @@ impl NexusChild {
                 key,
                 0,
                 nvme_reservation_acquire_action::ACQUIRE,
-                nvme_reservation_type::WRITE_EXCLUSIVE,
+                nvme_reservation_type::WRITE_EXCLUSIVE_ALL_REGS,
```
tiagolobocastro, line 438: Do we always want WRITE_EXCLUSIVE_ALL_REGS? Would there be a case for WRITE_EXCLUSIVE, eg to ensure only 1 nexus can access the replicas?

jonathan-teh: The existing code does only allow 1 nexus to access the replicas, and adding the replicas to a new nexus pre-empts the old nexus and prevents it from writing the replicas. My understanding is we want to move away from that behaviour and allow more than 1 nexus to connect to a replica. If we need to maintain 2 different behaviours, then a flag needs to be added to the Nexus structure to signal that and a new field in NexusCreateV2, or perhaps using V2 to signal a "HA" nexus and V1 for the old single-writer Nexus.

tiagolobocastro: Ok, it seems to be a bit "risky" to allow all nexus to access the replicas concurrently unless we have a use case and solution for doing RR, but I'm not in the loop of what we're aiming for here so maybe I need to revisit the ANA docs.

mayastor/tests/nexus_multipath.rs
```diff
@@ -433,40 +440,43 @@ async fn nexus_resv_acquire() {
...
-    // Verify that it has grabbed the write exclusive reservation
+    // Verify that the second nexus has registered
     let v2 = get_nvme_resv_report(&rep_dev);
-    assert_eq!(v2["rtype"], 1, "should have write exclusive reservation");
-    assert_eq!(v2["regctl"], 1, "should have 1 registered controller");
+    assert_eq!(
+        v2["rtype"], 5,
+        "should have write exclusive, all registrants reservation"
+    );
+    assert_eq!(v2["regctl"], 2, "should have 2 registered controllers");
```
tiagolobocastro, line 449: If possible it'd be nice to define an actual type with human friendly fields, or maybe some friendly getter methods, though it may be more work than it's worth since this is only a test.

test/python/test_nexus_multipath.py
```diff
@@ -0,0 +1,230 @@
...
+    yield replicas
+    try:
+        hdls["ms1"].replica_destroy(UUID)
+        hdls["ms2"].replica_destroy(UUID)
+    except Exception as e:
+        logging.debug(e)
+
+
+@pytest.mark.timeout(60)
+async def test_nexus_multipath(
```
tiagolobocastro, line 188: Do we also need a test for the case where you have a nexus but the second/third nexuses use a different reservation key?

tiagolobocastro (approved):

mayastor/src/bdev/nexus/nexus_child.rs
jonathan-teh: At the moment any nexus, or indeed any NVMe client, can access any replica anyway, unless the `NEXUS_NVMF_RESV_ENABLE` environment variable is set.

test/python/test_nexus_multipath.py
jonathan-teh: There is a separate ticket for 3rd Nexus, I'll add it then.

mayastor/tests/nexus_multipath.rs
jonathan-teh: I'll leave it as-is, it's only a test and it's still pretty readable.

test/python/test_nexus_multipath.py
tiagolobocastro: :+1:

cjones1024 (approved):

jonathan-teh:
bors merge

PR #982, opened 2021-07-25
test(py): run pytest based tests as part of CI

cjones1024:
- added file ./test/python/setup.sh to setup python environment
 - added file ./scripts/pytest-tests.sh to run "pytest" tests
 - added "pytest tests" stage to Jenkinsfile

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -139,6 +139,7 @@ if (params.e2e_continuous == true) {
   run_linter = false
   rust_test = false
   grpc_test = false
+  pytest_test = false
   e2e_test_profile = "continuous"
   // use images from dockerhub tagged with e2e_continuous_image_tag instead of building from current source
   e2e_build_images = false
@@ -150,6 +151,7 @@ if (params.e2e_continuous == true) {
   run_linter = true
   rust_test = true
   grpc_test = true
+  pytest_test = true
   // Some long e2e tests are not suitable to be run for each PR
   e2e_test_profile = (env.BRANCH_NAME != 'staging' && env.BRANCH_NAME != 'trying') ? "nightly" : "ondemand"
   e2e_build_images = true
@@ -270,6 +272,42 @@ pipeline {
             always {
               junit '*-xunit-report.xml'
               sh './scripts/check-coredumps.sh --since "${START_DATE}"'
+            }
+          }
+        }
+        stage('pytest tests') {
+          when {
+            beforeAgent true
+            expression { pytest_test == true }
+          }
+          agent { label 'nixos-mayastor-pytest' }
+          stages {
+            stage('checkout') {
+              steps {
+                checkout([
+                  $class: 'GitSCM',
+                  branches: scm.branches,
+                  extensions: scm.extensions.findAll{!(it instanceof jenkins.plugins.git.GitSCMSourceDefaults)} + [[$class: 'CloneOption', noTags: false, reference: '', shallow: false]],
+                  userRemoteConfigs: scm.userRemoteConfigs
+                ])
+              }
+            }
+            stage('build') {
+              steps {
+                sh 'printenv'
+                sh 'nix-shell --run "cargo build --bins"'
+              }
+            }
+            stage('python setup') {
+              steps {
+                sh 'nix-shell --run "./test/python/setup.sh"'
+              }
+            }
+            stage('run tests') {
+              steps {
+                sh 'printenv'
+                sh 'nix-shell --run "./scripts/pytest-tests.sh"'
+              }
             }
           }
         }
--- a/scripts/pytest-tests.sh
+++ b/scripts/pytest-tests.sh
@@ -1,0 +1,18 @@
+#! /usr/bin/env bash
+
+set -euxo pipefail
+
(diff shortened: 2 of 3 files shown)


cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try-

cjones1024:
bors try

cjones1024:
bors try

Jenkinsfile
```diff
@@ -273,6 +275,40 @@ pipeline {
...
+                  $class: 'GitSCM',
+                  branches: scm.branches,
+                  extensions: scm.extensions.findAll{!(it instanceof jenkins.plugins.git.GitSCMSourceDefaults)} + [[$class: 'CloneOption', noTags: false, reference: '', shallow: false]],
+                  userRemoteConfigs: scm.userRemoteConfigs
+                ])
+              }
+            }
+            stage('build') {
+              steps {
+                sh 'nix-shell --run "cargo build --bins"'
```
tiagolobocastro, line 297: are all the binaries in the project needed for the pytests?

tiagolobocastro (approved):

paulyoong (approved):

Jenkinsfile
cjones1024: Certainly not all artifacts are required. However the docker-compose stuff currently requires access to a "working tree". It should be reasonably straight forward to determine which artifacts are required and then copy them to the new agent. I believe the brief is to get these tests working in CI now so that we have as much test coverage as we can as soon as we can and worry about efficiency later. For now the simplest approach (ie. which doesn't require significant changes to the existing code) is to perform another checkout + build on the "pytest" agent.

cjones1024:
bors merge

cjones1024:
bors merge-

cjones1024:
bors merge

Jenkinsfile
```diff
@@ -273,6 +275,40 @@ pipeline {
...
+        stage('pytest tests') {
+          when {
+            beforeAgent true
+            expression { pytest_test == true }
+          }
+          agent { label 'nixos-mayastor-pytest' }
+          stages {
+            stage('checkout') {
+              steps {
+                checkout([
```
tiagolobocastro, line 287: Rather than checking out here could we just unstash?
```
cleanWs()
unstash 'source'
```
The CI system seems to be "messed up", with stashes going missing, I wonder if this is somehow causing that?

tiagolobocastro (changes requested):

Jenkinsfile
```diff
@@ -273,6 +275,40 @@ pipeline {
             }
           }
         }
+        stage('pytest tests') {
```
tiagolobocastro, line 278: Can we add a "sh 'printenv'" like on the other stages?
Also, I see this on the logs:
"[2021-07-26T12:24:36Z ERROR mayastor_csi::nodeplugin_grpc] gRPC server failed with error: transport error: error creating server listener: Address already in use (os error 98)"
Do the pytests clean up after themselves?

Jenkinsfile
cjones1024: These tests are run on a different agent - and my understanding is this implicitly creates a new workspace anyway. We could try stashing rather than a clean checkout. I can't see how this could mess up the other stages, but that doesn't men its not possible.

Jenkinsfile
cjones1024: They should cleanup after themselves yes. All the containers should be destroyed. Unfortuantely, I have seen a case before wher the mayastor_csi causes a kernel hang and cannot be killed. This is one of the reasons why we ultimately want to run the tests completely within a VM.

cjones1024: Will add a sh printenv though

Jenkinsfile
tiagolobocastro: Oh right, it's a different machine... If we don't ever have other tests on these agent then I guess it's ok with checkout, otherwise we could also checkout on a python init stage. But yeah, we don't need it now then :+1:

tiagolobocastro (approved):
you're aware of it now, please run bors try a few times to make sure the hang is not from this PR

cjones1024:
bors check

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
testing again after kernel downgrade on pytest01

bors try

Jenkinsfile
cjones1024: Thinking about the varios ways in which a Jenkins pipeline may fail or be aborted. There is no guarantee that pytest will perform all the necessary teardown in the case of an abnormal termination. Need to investigate what control pytest gives us in such situations. In any case we probably should include a "post" stage in the Jenkins pipeline as a catch all to try and ensure that all necessary cleanup operations have been performed regardless of the test result. However, I think this is better done under a separate Jira.

cjones1024:
bors try

cjones1024:
bors merge

PR #987, opened 2021-08-03
fix(lvol): ensure superblock info is wiped

gila:
When the base bdev does not support UNMAP, writing zero's is costly
and very time consuming.


gila:
bors try

gila:
bors retry

gila:
bors try

gila:
bors try

gila:
bors try

mayastor/src/lvs/lvol.rs
```diff
@@ -223,6 +221,48 @@ impl Lvol {
...
+                    source: Errno::ENOMEM,
+                    name: self.name(),
+                }
+            })?;
+            // write zero to the first 8MB which whipes the metadata and the
+            // first 4MB of the data partition
+            let range =
+                std::cmp::min(self.as_bdev().size_in_bytes(), (1 << 20) * 8);
+            debug!(?self, ?range, "zeroing range");
+            for offset in 0 .. (range >> 12) {
```
mtzaurus, line 253: We perform 2048 independent I/O operations here ? ( ((1 << 20) * 8) / (1 << 12) = 2048). Can we increase the size of DMA buffer to something like 0.5MB - 1MB to reduce the number of I/Os or there are some other restrictions which I have overlooked ?

mayastor/src/lvs/lvol.rs
```diff
@@ -239,18 +279,18 @@ impl Lvol {
...
             let sender =
                 unsafe { Box::from_raw(sender as *mut oneshot::Sender<i32>) };
             sender.send(errno).unwrap();
         }
 
-        let name = self.name();
-
         // we must always unshare before destroying bdev
         let _ = self.unshare().await;
+        self.wipe_super().await?;
```
mtzaurus, line 291: In case we can't wipe out the superblock (due to I/O errors on superblock or so), should that block us from removing the underlaying LVOL ?

mayastor/src/lvs/lvol.rs
gila: There is a max on the buffer size we can submit to the bdev layer (64KB iirc) but it would amount to the same thing really. The most important thing here is that buffer is 4K aligned as that is needed for aio and uring. When submitting 64K IOs we would need to ensure that the size of volume is indeed, 64K aligned as well which is not .. i guess all that obvious?

mayastor/src/lvs/lvol.rs
gila: Yeah I thought about that too, we could swallow the error but IMHO, it might hide some other issues if we do that. We can always think about it later IMHO.

paulyoong (approved):

mayastor/src/lvs/lvol.rs
mtzaurus: I mean, if we're OK with issuing 2048 independent I/O operations, then it's fine.
I heard that modern disks support writing more than 64K of data at once. :-)

mayastor/src/lvs/lvol.rs
mtzaurus: No problem, at least we log all the errors. LGTM.

mtzaurus (approved):

gila:
bors merge

tiagolobocastro (approved):

mayastor/src/lvs/lvol.rs
tiagolobocastro: The max is 64k (SPDK_BDEV_LARGE_BUF_MAX_SIZE).

PR #991, opened 2021-08-09
feat(lvm): add support for lvm pool

shubham14bajpai:
This PR adds support for pooltype `lvm` which are backed by lvm `volume groups`.
While creating volume groups the tag `mayastor` is added to identify the volume
groups owned by mayastor. 
This PR also adds tests for lvm & lvs pool creation/deletion and listing them 
together.

Ref: CAS-1015

Signed-off-by: shubham <shubham.bajpai@mayadata.io>


zeenix:
@shubham14bajpai Great work! Could you please split the patch into [smaller logical ones](https://dev.to/paulinevos/atomic-commits-will-help-you-git-legit-35i7)? Pro-tip: when your PR/commit description has sentence that with "also adds/changes", it's a good indication that part can be split into its own commit at least. You could also split `lvm.rs` addition into a separate commit with its use in the following commit. Addition of some of the types (e.g `PoolType`) can also go into their own commits. The smaller the commits can be, the better, as long as each of them build and don't break anything.

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -228,17 +246,50 @@ impl mayastor_server::Mayastor for MayastorSvc {
...
+                            .map(Response::new)
+                    },
+                    PoolBackend::Lvm => {
+                        // check if a lvs pool already exists with the same name
+                        if let Some(_pool) = Lvs::lookup(args.name.as_str()) {
+                            return Err(Status::invalid_argument("lvs pool with the same name already exists"))
+                        };
+                        // check if the disks are used by existing lvs pool
+                        if Lvs::iter()
+                            .map(|l| l.base_bdev().name()).any(|d: String| args.disks.contains(&d)){
```
zeenix, line 279: is the `String` type annotation needed here? Compiler can't infer? If it can, convention is to not specify.

shubham14bajpai: Yes, it can. I was learning iters at this point and forgot to remove it.

zeenix: > Yes, it can. 

Cool.

> I was learning iters at this point and forgot to remove it.

NP, we all make these mistakes all the time. This is why reviews are important. :)

zeenix:
That was super fast work! LGTM

zeenix (approved):

zeenix:
@shubham14bajpai But let's wait for someone more familiar with the code, have a look before merging.

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -228,17 +246,50 @@ impl mayastor_server::Mayastor for MayastorSvc {
...
-                rx.await
-                    .map_err(|_| Status::cancelled("cancelled"))?
-                    .map_err(Status::from)
-                    .map(Response::new)
+                let resp = match PoolBackend::try_from(args.pooltype)? {
+                    PoolBackend::Lvs => {
+                        if  MayastorFeatures::get_features().lvm {
+                            // check if a lvm pool already exists with the same name
+                            if let Some(_pool) = lookup_lvm_vg(args.name.as_str()).await {
+                                return Err(Status::invalid_argument("lvm pool with the same name already exists"))
```
mtzaurus, line 254: Can you please add pool name in the error message(s), that would simplify analysing error logs, as it clearly shows what name is conflicting.

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -228,17 +246,50 @@ impl mayastor_server::Mayastor for MayastorSvc {
...
+                            PoolConfig::capture().export().await;
+                            Ok(Pool::from(pool))
+                        })?;
+                        rx.await
+                            .map_err(|_| Status::cancelled("cancelled"))?
+                            .map_err(Status::from)
+                            .map(Response::new)
+                    },
+                    PoolBackend::Lvm => {
+                        // check if a lvs pool already exists with the same name
```
mtzaurus, line 273: Shall we reject creation of an LVM pool if 'MayastorFeatures::get_features().lvm' is not set ?

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -228,17 +246,50 @@ impl mayastor_server::Mayastor for MayastorSvc {
...
+                        if let Some(_pool) = Lvs::lookup(args.name.as_str()) {
+                            return Err(Status::invalid_argument("lvs pool with the same name already exists"))
+                        };
+                        // check if the disks are used by existing lvs pool
+                        if Lvs::iter()
+                            .map(|l| l.base_bdev().name()).any(|d| args.disks.contains(&d)){
+                                return Err(Status::invalid_argument("a lvs pool already uses the disk"))
+                            };
+                        let pool = match import_or_create_lvm_vg(args).await {
```
mtzaurus, line 282: You can shorten this construction by doing transformations directly on 'Result' object returned here, using method chaining.

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -254,23 +305,38 @@ impl mayastor_server::Mayastor for MayastorSvc {
...
-                let rx = rpc_submit::<_, _, LvsError>(async move {
-                    if let Some(pool) = Lvs::lookup(&args.name) {
-                        // Remove pool from current config and export to file.
-                        // Do this BEFORE we actually destroy the pool.
-                        let mut config = PoolConfig::capture();
-                        config.delete(&args.name);
-                        config.export().await;
-
-                        pool.destroy().await?;
+                let resp = match PoolBackend::try_from(args.pooltype)? {
```
mtzaurus, line 308: I am in a doubt whether we should use pool type when removing the pool, as we can't have 2 pools of different types with the same name. IMHO, we should use only pool name as input and perform pool type traversal inside gRPC call, hiding it from user. That will bring us benefits when calling our methods via REST API, as user normally uses pool name as its ID when calling 'DEL /pools/:id', which then gets simply passed to our gRPC without performing one extra call for resolving pool type.

test/python/docker-compose.yml
```diff
@@ -6,10 +6,11 @@ version: '3'
 services:
   ms0:
     container_name: "ms0"
-    image: rust:latest
+    image: shubham1494/lvm-test:latest
```
gila, line 9: don't think this is what we want, do we?

test/python/lvm-test.dockerfile
```diff
@@ -0,0 +1,4 @@
+# rust image with lvm utils installed
```
gila, line 1: for what purpose is this docker file?

mayastor/src/grpc/mayastor_grpc.rs
shubham14bajpai: Makes sense. The workflow would be something like 
```
if lvs_lookup(name){
  lvs_del_pool(name)
  return
}

if lvm_enabled && lvm_lookup(name){
  lvm_del_pool(name)
  return
}
```
?

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+use rpc::mayastor::{CreatePoolRequest, Pool, PoolType, PoolState};
+
+const PVCREATE_COMMAND: &str = "pvcreate";
+const PVS_COMMAND: &str = "pvs";
+const PVREMOVE_COMMAND: &str = "pvremove";
+const VGCREATE_COMMAND: &str = "vgcreate";
+const VGS_COMMAND: &str = "vgs";
+const VGREMOVE_COMMAND: &str = "vgremove";
+
+pub async fn lookup_lvm_vg(name: &str) -> Option<Pool> {
```
gila, line 13: Functions should have /// style comments to explain what they do. It gets added to our docs that way. It is clear what the function does in this case, so a simple short brief is more than sufficient.

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+    let pools = list_lvm_vg().await.ok()?;
+    match pools.iter().find(|p| p.name == name) {
+        Some(p) => {
+            return Some(Pool{
```
gila, line 17: this is fine *but* -- you could write this as `pools.iter().filter(|p| p.name == name) if you where to implement the TryFrom trait and/or Into. Those are just minor things that do not change anything ofcours.

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+                capacity: p.capacity,
+                used: p.used,
+                pooltype: PoolType::Lvm.into(),
+                state: PoolState::PoolOnline.into(),
+            });
+        }
+        None => None
+    }
+}
+
```
gila, line 29: If I call recall directly, we were not planning to "create" pools rather only import them. I really do not want to get into the work related to created pools at this point. It's simple to create pools but even simpler to create them incorrectly when considering JBOD redundancy for example. We should document that this is purely for the simplest cases.

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+struct PV {
+    pv: Vec<VgPvMap>
+}
+#[derive(Debug, Serialize, Deserialize)]
+struct VgPvMap {
+    vg_name: String,
+    pv_name: String
+}
+
+pub async fn get_lvm_vg(name: String) -> Result<Pool, Error> {
```
gila, line 177: IIRC we use verbs like list rather than get. Get is used to get properties of a volume instance but not to get all the volumes, dont we?

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+        used: capacity - free,
+        pooltype: PoolType::Lvm.into(),
+        state: PoolState::PoolOnline.into(),
+    };
+    Ok(pool)
+}
+
+pub async fn remove_lvm_vg(name: String) -> Result<(), Error> {
+    let pool = match get_lvm_vg(name.clone()).await{
+        Ok(p) => p,
```
gila, line 246: Why do we take a string here (and the fn above?)

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+            Ok(s) => s,
+            Err(_) => "failed to execute pv_remove".to_string(),
+        };
+        return Err(Error::FailedExec{err: msg.to_string()})
+    }
+    info!("pool '{}' has been destroyed successfully", name);
+    Ok(())
+}
+
+pub async fn lookup_lvm_disk(name: &str) -> Option<String> {
```
gila, line 279: it seems this function is only used to determine if the disk in use, as such -- would a pool not suffice? if not, and we need to return more context, why not return the while struct Pool?

gila (changes requested):

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -284,18 +350,35 @@ impl mayastor_server::Mayastor for MayastorSvc {
...
+                let rec  =  rx.await
+                        .map_err(|_| Status::cancelled("cancelled"))?
+                        .map_err(Status::from);
+
+                let mut lvs_pools = match rec {
+                    Ok(pools) => pools,
+                    Err(e) => return Err(e),
+                };
+
+                if  MayastorFeatures::get_features().lvm {
```
mtzaurus, line 371: Can we retain all found LVS pools and return the partial result to the user in case we see error upon LVM pools enumeration ? I would suggest just log the error.

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+const PVCREATE_COMMAND: &str = "pvcreate";
+const PVS_COMMAND: &str = "pvs";
+const PVREMOVE_COMMAND: &str = "pvremove";
+const VGCREATE_COMMAND: &str = "vgcreate";
+const VGS_COMMAND: &str = "vgs";
+const VGREMOVE_COMMAND: &str = "vgremove";
+
+pub async fn lookup_lvm_vg(name: &str) -> Option<Pool> {
+    let pools = list_lvm_vg().await.ok()?;
+    match pools.iter().find(|p| p.name == name) {
```
mtzaurus, line 15: You can use '.map()' method to transform target Option without the need of extra 'match'.

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+            });
+        }
+        None => None
+    }
+}
+
+pub async fn import_or_create_lvm_vg(req: CreatePoolRequest) -> Result<Pool, Error> {
+    let name = req.name.as_str();
+    let disks = req.disks.as_slice();
+    let pool = match import_lvm_vg(CreatePoolRequest{
```
mtzaurus, line 33: We could use 'Result::or_else()' combinator to handle the absence of the pool and eliminate intensive use of the 'match' operator.

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+            _ => return Err(e)
+        }
+    };
+    info!("The pool '{}' has been created on {:?}", name, disks);
+    Ok(pool)
+}
+
+pub async fn import_lvm_vg(req: CreatePoolRequest) -> Result<Pool, Error> {
+    let pools = list_lvm_vg().await?;
+    match pools.iter().find(|p| p.name == req.name) {
```
mtzaurus, line 54: I would suggest using '.map()' after 'find()' to remove extra 'match'.

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+        None => Err(Error::NotFound {err: format!("lvm volume {} not found", req.name)}),
+    }
+}
+
+pub async fn create_lvm_vg(req: CreatePoolRequest) -> Result<Pool, Error> {
+    let output = Command::new(PVCREATE_COMMAND)
+        .args(&req.disks)
+        .output()
+        .await
+        .expect("failed to execute pv_create");
```
mtzaurus, line 77: Could we handle PVCREATE_COMMAND invocation failures in such a way that we don't crash via 'expect()' ?
We could propagate error to the caller as the reason why we can't create the pool.
.expect() sometimes is too optimistic. :-)

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    let r: VGListReport = match serde_json::from_slice(output.stdout.as_slice()){
+        Ok(r) => r,
+        Err(e) => return Err(Error::FailedParsing{err: e.to_string()}),
+    };
+
+    let mut res: Vec<Pool> = vec![];
+
+    for v in r.report[0].vg.as_slice() {
```
mtzaurus, line 138: What if r.report is empty, i.e. there are no existing LVMs ? In such a case 'r.report[0].vg' could cause panic, something like:
```
panicked at 'index out of bounds: the len is 0 but the index is 0'
```
Can we check against the length of r.report before performing enumeration ?

mayastor/src/lvm/lvm.rs
zeenix: Also:

*  it's best to avoid using abbreviations as much possible IMO (unless they're well known even to people unfamiliar with the tech in question). Similarly for all for the identifiers, e.g `VolGroupList` (even though they're internal, it add to code readability).

* since this function is already in the `lvm` module, you don't need to have `_lvm_` in the names, right?

```suggestion
pub async fn lookup_vol_group(name: &str) -> Option<Pool> {
```

mayastor/src/bin/mayastor-client/pool_cli.rs
```diff
@@ -25,6 +25,14 @@ pub fn subcommands<'a, 'b>() -> App<'a, 'b> {
...
                 .index(2)
                 .help("Disk device files"),
+        )
+        .arg(
+            Arg::with_name("pooltype")
+                .short("t")
+                .help("the type of the pool")
+                .required(false)
+                .possible_values(&["lvs", "lvm"])
+                .default_value("lvs"),
```
zeenix, line 35: Oh and pool_cli changes look like a separate logical change to me. If so, you know what to do. :)

mayastor/src/lvm/error.rs
```diff
@@ -0,0 +1,16 @@
...
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility = "pub(crate)")]
+pub enum Error {
+    #[snafu(display("Failed to parse {}", err))]
+    FailedParsing { err: String },
+    #[snafu(display("Failed to execute command {}", err))]
+    FailedExec { err: String },
+    #[snafu(display("{} Not found", err))]
+    NotFound { err: String},
```
zeenix, line 11: Already covered by [`std::io::ErrorKind::NotFound`](https://doc.rust-lang.org/nightly/std/io/enum.ErrorKind.html#variant.NotFound). We recently discussed how we should stop adding more error types and variants so we can start here. :) 

```suggestion
    #[snafu(display("I/O error: {}", err))]
    Io { err: std::io::Error },
```

This way, in the future you could emit any of the errors already covered by `std::io::Error` w/o adding new variants. It'll also make it easier to unify/merge error types (where needed) in the future.

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+                return Ok(Pool{
+                    name: p.name.as_str().to_owned(),
+                    disks: p.disks.as_slice().to_owned(),
+                    capacity: p.capacity,
+                    used: p.used,
+                    pooltype: PoolType::Lvm.into(),
+                    state: PoolState::PoolOnline.into(),
+                });
+            }
+            Err(Error::ImportErr {err: format!("lvm vg {} disks not match", req.name)})
```
zeenix, line 66: Best for readability if you special case the error case:

```suggestion
            if p.disks != req.disks {
                return Err(Error::ImportErr {err: format!("lvm vg {} disks not match", req.name)});
            }

            Ok(Pool{
                    name: p.name.as_str().to_owned(),
                    disks: p.disks.as_slice().to_owned(),
                    capacity: p.capacity,
                    used: p.used,
                    pooltype: PoolType::Lvm.into(),
                    state: PoolState::PoolOnline.into(),
                });
```

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -208,6 +225,7 @@ impl From<MayastorFeatures> for rpc::mayastor::MayastorFeatures {
     fn from(f: MayastorFeatures) -> Self {
         Self {
             asymmetric_namespace_access: f.asymmetric_namespace_access,
+            lvm: f.lvm,
```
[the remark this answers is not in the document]
akhilerm: >  This flag can be manipulated by changing the environment. I could, for example, set it and unset it later, right?

Environments are set while starting the process right? So setting, unsetting should not affect.

Also, Is using environment variable the best way to define feature gates?

mayastor/src/lvm/lvm.rs
shubham14bajpai: Few doubts about pools:
- Will we only import existing `volumegroups` as mayastor pools? 
- Or will we allow the creation of `volumegroups` from mayastor as well by passing unused disks?
- Right now this code only supports creation of vgs and while creating we add the tag `mayastor` to identify it.
- For importing we can add the `mayastor`  tag to the existing volumegroups so that it will be identified as being in use by mayastor as pools. 
- Also for importing a volumegroup the disks argument is not required, how should we handle that case?

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    let r: PVsReport = match serde_json::from_slice(output.stdout.as_slice()){
+        Ok(r) => r,
+        Err(e) => return Err(Error::FailedParsing{err: e.to_string()}),
+    };
+
+    let mut disks: Vec<String> = vec![];
+
+    for p in r.report[0].pv.as_slice() {
```
mtzaurus, line 227: We might need checks against r.report not being empty to avoid panics.

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+        .expect("failed to execute vgs");
+    if !output.status.success() {
+        let msg = match String::from_utf8(output.stderr){
+            Ok(s) => s,
+            Err(_) => "failed to execute vgs".to_string(),
+        };
+        return Err(Error::FailedExec{err: msg.to_string()})
+    }
+
+    let r: VGListReport = match serde_json::from_slice(output.stdout.as_slice()){
```
zeenix, line 131: Descriptive names for variables please.

```suggestion
    let report: VGListReport = match serde_json::from_slice(output.stdout.as_slice()){
```

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+            Err(_) => "failed to execute vg_create".to_string(),
+        };
+        return Err(Error::FailedExec{err: msg.to_string()})
+    }
+
+    get_lvm_vg(req.name).await
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct VGListReport {
```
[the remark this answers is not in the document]
zeenix: Actually, this type isn't really needed. `VGList` is enough.

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+    report: Vec<VGList>
+}
+#[derive(Debug, Serialize, Deserialize)]
+struct VGList {
+    vg: Vec<VGName>
+}
+#[derive(Debug, Serialize, Deserialize)]
+struct VGName {
+    vg_name: String
+}
```
zeenix, line 115: Since these types don't really have an impl (i-e they don't abstract the underlying type) and are private even, they're good candidates for type alias instead:

```suggestion
type VolGroupNameList = Vec<VolGroupName>;
type VolGroupName = String;
```

(note: i'm assuming descriptive names as per my previous suggestions here).

mayastor/src/lvm/lvm.rs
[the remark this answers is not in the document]
zeenix: oops, sorry I misunderstood the code. It should actually be a type alias, like I suggested below for `VGList` and `VGName`.

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+        Err(e) => return Err(Error::FailedParsing{err: e.to_string()}),
+    };
+
+    let mut res: Vec<Pool> = vec![];
+
+    for v in r.report[0].vg.as_slice() {
+        let pool = match get_lvm_vg(v.vg_name.clone()).await {
+            Ok(p) => p,
+            Err(e) => return Err(e),
+        };
```
zeenix, line 142: Why not make use of `?` here?

```suggestion
        let pool = get_lvm_vg(v.vg_name.clone()).await?;
```

mayastor/src/lvm/lvm.rs
shubham14bajpai: I am not sure but we need the same name as the son fields for `serde.rs` to work. The json returned by this command would look like 

```
$ sudo vgs --options=vg_name --reportformat=json @mayastor
  {
      "report": [
          {
              "vg": [
                  {"vg_name":"pool"}
              ]
          }
      ]
  }
``` 
The struct names can be improved as suggested but I guess the fields should remain the same.

mayastor/src/lvm/lvm.rs
zeenix: yeah. probably a good place to make use of `Option::ok_or`

mayastor/src/lvm/lvm.rs
zeenix: Ah, sorry I missed on that detail. :)

test/python/docker-compose.yml
shubham14bajpai: Once the dockerfile is approved we can have this image in openebs docker hub

test/python/lvm-test.dockerfile
shubham14bajpai: As we are using lvm binaries from the code we need those installed in the test setup. This docker file is used to create that image.

shubham14bajpai:
Addressed most of the review comments. Still trying to figure out how to manage errors properly, need some guidance on that.

For the pool import part the current code can import an existing vol group (which is nothing but adding mayastor tag to the vol group) or create one:
```
$ sudo pvcreate /dev/sdc
  Physical volume "/dev/sdc" successfully created.

$ sudo vgcreate volgrp /dev/sdc
  Volume group "volgrp" successfully created

$ sudo vgs
  VG     #PV #LV #SN Attr   VSize   VFree  
  volgrp   1   0   0 wz--n- <14.32g <14.32g

$ ./mayastor-client pool create -t lvm volgrp /dev/sdc
volgrp

$ ./mayastor-client pool create -t lvm pool /dev/sdb
pool

$ ./mayastor-client pool list
NAME   TYPE STATE     CAPACITY USED DISKS   
pool   lvm  online 15372124160    0 /dev/sdb
volgrp lvm  online 15372124160    0 /dev/sdc

$ sudo vgs
  VG     #PV #LV #SN Attr   VSize   VFree  
  pool     1   0   0 wz--n- <14.32g <14.32g
  volgrp   1   0   0 wz--n- <14.32g <14.32g
```

For the pool destroy should we just remove the mayastor tag so that it is no longer available to mayastor or actually call the vgremove and pvremove commands to clean the disks. Currently it calls the commands to do so.

test/python/lvm-test.dockerfile
gila: What I mean is - we don't have docker files anywhere why do we need one here?

mayastor/src/lvm/lvm.rs
gila: Perhaps later we will add the ability to "create" VGs, but for now, we do not. If the disk argument is not needed simple mark the argument name as disks: _disk and it will indicate an unused variable.

test/python/docker-compose.yml
gila: We don't need to have docker files at all for the tests.

mayastor/src/grpc/mayastor_grpc.rs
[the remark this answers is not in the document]
gila: Well, it depends a bit. When running in containers and no config file is given you kinda end up with either long arguments or ENV variables.

mayastor/src/bin/mayastor-client/pool_cli.rs
```diff
@@ -205,3 +218,22 @@ fn pool_state_to_str(idx: i32) -> &'static str {
         rpc::PoolState::PoolFaulted => "faulted",
     }
 }
+
+fn pooltype_to_str(idx: i32) -> &'static str {
```
gila, line 222: calling pool_type_to_str(3) will crash the program the whole idea here is that you can only pass valid values to PoolType::from_i32().  So it's probably better to write something as:

`PoolType::from_i32(idx)?` and handle the error

zeenix:
> Addressed most of the review comments.

:+1: Please do resolve the conversations that you've addressed as suggested by the reviewer in question.

> Still trying to figure out how to manage errors properly, need some guidance on that.

We can try to help if we know what problems you're facing. :)

mayastor/src/bin/mayastor-client/pool_cli.rs
shubham14bajpai: This function is only for printing pooltype in the mayastor-client `pool list`. So I guess the wrong index should not come.

mayastor/src/core/env.rs
```diff
@@ -142,8 +142,14 @@ impl MayastorFeatures {
             Err(_) => false,
         };
 
+        let lvm = match std::env::var("LVM_ENABLE") {
```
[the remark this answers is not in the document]
shubham14bajpai: If I am correct the above code means true if the env is present, no matter the value. But we want true only if the env value is 1?

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -158,6 +166,14 @@ impl From<LvsError> for Status {
     }
 }
 
+impl From<LvmError> for Status {
+    fn from(e: LvmError) -> Self {
+        match e {
+            _ => Status::internal(e.to_string()),
```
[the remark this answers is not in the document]
shubham14bajpai: Fixed by adding more arms :+1:

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+    vg: Vec<VGName>
+}
+#[derive(Debug, Serialize, Deserialize)]
+struct VGName {
+    vg_name: String
+}
+pub async fn list_lvm_vg() -> Result<Vec<Pool>, Error> {
+    let output = Command::new(VGS_COMMAND)
+        .args(&["--options=vg_name", "--reportformat=json"])
+        .arg("@mayastor")
```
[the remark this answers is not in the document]
shubham14bajpai: This tag is to identify all the vol group owned by mayastor

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,285 @@
...
+       .expect("failed to execute vgs");
+    if !output.status.success() {
+        let msg = match String::from_utf8(output.stderr){
+            Ok(s) => s,
+            Err(_) => "failed to execute pvs".to_string(),
+        };
+        return Err(Error::FailedExec{err: msg.to_string()})
+    }
+
+    let r: PVsReport = match serde_json::from_slice(output.stdout.as_slice()){
```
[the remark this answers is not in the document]
shubham14bajpai: I couldn't find any function in serde_json which could do this in option

test/python/lvm-test.dockerfile
akhilerm: All the tests are currently using `rust:latest` image for the docker containers. For testing the lvm related changes we need the lvm binaries to be present inside the container, which is not available in `rust:latest` image. So we created another image with this docker file and used it in the tests. 

If we dont want to have the dockerfile, we should choose an image with lvm preinstalled.

mayastor/src/lvm/lvm.rs
shubham14bajpai: This get function is used to call the vgs and pvs commands to find out the capacity, used & disks for a vol group and convert that into a rpc::Pool struct, so I guess get makes more sense here

mayastor/src/lvm/lvm.rs
shubham14bajpai: The list_vol_group gets the pools which have `mayastor` tag on them. This function list all the vol groups in the system and check if the disk provided in `lvs` pool creation is not in use by a vol group

mayastor/src/bin/mayastor-client/pool_cli.rs
shubham14bajpai: These tie in with the grpc changes

shubham14bajpai:
>  Please do resolve the conversations that you've addressed as suggested by the reviewer in question.

Done :+1: 

> We can try to help if we know what problems you're facing. :)

It was mostly how to use combinators, a little practice helped me :slightly_smiling_face: . Also added the `std::io:Error` changes, please take a look

test/python/lvm-test.dockerfile
shubham14bajpai: We don't need the dockerfile, just an image with lvm installed to run the lvm commands from mayastor. Once we push this image to some openebs common repo we can remove the docker file.

mayastor/src/core/env.rs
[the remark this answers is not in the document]
tiagolobocastro: You can just compare it I guess
```rust
if std::env::var("LVM_ENABLE") == Ok("1".to_string()) {
      println!("OK");
  }
```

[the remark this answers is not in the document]
zeenix: @tiagolobocastro I don't think you need `.to_string()` here though. `String` implements `PartialEq<str>`.

zeenix:
> > Please do resolve the conversations that you've addressed as suggested by the reviewer in question.
> 
> Done +1

Great. Could you please squash the fixups to you patches to the original ones so it's possible to review per-commit?

mayastor/src/core/env.rs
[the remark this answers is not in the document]
shubham14bajpai: Combining all the above suggestions :slightly_smiling_face: 
```rust
let lvm = matches!(std::env::var("LVM_ENABLE"), Ok(x) if x == "1")
```

zeenix:
bors try

zeenix:
@shubham14bajpai as the test run points out the format is wrong in some places. Did you run `cargo fmt` on it? I guess like me (and perhaps other?) you remove the default pre-commit hook installed by nix-shell for you? :)

shubham14bajpai:
> @shubham14bajpai as the test run points out the format is wrong in some places. Did you run `cargo fmt` on it? I guess like me (and perhaps other?) you remove the default pre-commit hook installed by nix-shell for you? :)

Yes when I enable it a lot of files are throwing errors. Is there a way to just run `cargo fmt` on my changes?

zeenix:
> Yes when I enable it a lot of files are throwing errors.

Oh for you as well? I wonder how that's been working since CI uses the same rust and fmt version as well so the last rust tools update in nix PR should have caught all those unrelated errors? @gila this is what I was talking about yesterday btw.

> Is there a way to just run `cargo fmt` on my changes?

I think so. I think you pass it the paths? Anyway, best if you fix the `develop` branch (using `cargo fmt`) and submit a spearate PR for that and then base this PR/branch on top of that.

test/python/lvm-test.dockerfile
shubham14bajpai: Tried mounting the `vgs` command related binaries into the docker compose file
```
$ ldd /sbin/vgs
	linux-vdso.so.1 (0x00007ffc561a1000)
	libblkid.so.1 => /lib/x86_64-linux-gnu/libblkid.so.1 (0x00007fc229f7d000)
	libaio.so.1 => /usr/lib/x86_64-linux-gnu/libaio.so.1 (0x00007fc229f78000)
	libselinux.so.1 => /lib/x86_64-linux-gnu/libselinux.so.1 (0x00007fc229f4d000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fc229dfe000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fc229c0c000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fc22a4ef000)
	libpcre2-8.so.0 => /usr/lib/x86_64-linux-gnu/libpcre2-8.so.0 (0x00007fc229b7c000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fc229b74000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fc229b51000)
```
Still lvm commands are not working from the mayastor binary. Should we go with the `lvm-test` image approach as it straightforward? This image won't be built again and again, only needs to be pushed to openebs repo

mayastor/src/lvm/lvm.rs
gila: I understand, but it is not what I tried asking... Why do we return String? Why not Option<Pool>? Also should we not use udev to obtain this information? It would be a lot nice to report it that way maybe. We can improve the udev usage later.

mayastor/src/lvm/lvm.rs
```diff
@@ -0,0 +1,321 @@
...
+        self.vg_free
+    }
+
+    /// returns the used capacity
+    pub fn used(&self) -> u64 {
+        self.capacity() - self.available()
+    }
+
+    /// delete a given vol group and its corresponding physical vols
+    pub async fn remove(self) -> Result<(), Error> {
```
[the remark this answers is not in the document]
shubham14bajpai: One doubt regarding the destroy function. Should we call the vgremove & pvremove command and wipe all the lvm related headers from the disks? Or we just remove the mayastor tag as we only import the vg by adding tag not actually create it?

Right now this calls the vgremove & pvremove.

shubham14bajpai:
bors try

gila:
Bors try

tiagolobocastro:
I've taken this PR and re-raised on #1587

PR #994, opened 2021-08-11
chore: drop smol dep

zeenix:
* In mbus-api: it can be dropped completely. This drops two other indirect deps as well.
* In mayastor: we only need async-channel.


mbus-api/src/lib.rs
```diff
@@ -22,9 +22,8 @@ pub use mbus_nats::{
 pub use receive::*;
 pub use send::*;
 use serde::{de::StdError, Deserialize, Serialize};
-use smol::io;
 use snafu::{ResultExt, Snafu};
-use std::{fmt::Debug, marker::PhantomData, str::FromStr, time::Duration};
+use std::{fmt::Debug, io, marker::PhantomData, str::FromStr, time::Duration};
```
mtzaurus, line 26: Just out of my curiosity: why do we need std::io here now but didn't need it before ?
Sorry if I overlooked that somehow in your PR synopsis.

mtzaurus: Ah, I see: smol::io has been removed and replaced with std::io.

mtzaurus (approved):

tiagolobocastro (approved):

zeenix:
bors merge

PR #1004, opened 2021-08-24
chore: Split developer nix shell from CI

zeenix:
Developer and CI are a bit of a different targets, with the former preferring at least some of the tools to come from their host system and the latter preferring all the tools to be provided in one image that's created by `nix-build`. So this PR splits the CI nix shell configuration into a separate file and drops a bunch of tools from the default nix shell configuration.


zeenix:
bors try

zeenix:
bors try

shell.nix
```diff
@@ -24,13 +24,9 @@ mkShell {
...
     clang_11
     cowsay
-    docker
-    docker-compose
     e2fsprogs
     etcd
     fio
-    gdb
-    git
     kubernetes-helm
```
gila, line 30: This one can be removed as well?

shell.nix
```diff
@@ -51,8 +47,7 @@ mkShell {
     python3
     utillinux
     xfsprogs
```
gila, line 49: and these two?

zeenix: right, I also dropped `e2fsprogs`.

zeenix: As said on the slack, I re-added `utillinux`. I also updated the commit log with info on that.

shell.nix
```diff
@@ -24,13 +24,9 @@ mkShell {
   buildInputs = [
     clang_11
     cowsay
-    docker
-    docker-compose
     e2fsprogs
     etcd
```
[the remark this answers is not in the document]
zeenix: As discussed on slack 2 weeks ago, it seems the persistence tests:

```
    persist_clean_shutdown
    persist_io_failure
    persist_unexpected_restart
    persistent_store_connection
```

fail when `etcd` is not part of the nix-shell:

```
[FATAL tini (7)] exec /usr/bin/etcd failed: No such file or directory
```

So gave up on this. Let's revisit when `etcd` from nix gives anyone any problems. :)

PR #1002, opened 2021-08-23
feat: build against SPDK 21.07

jonathan-teh:
Build mayastor against SPDK 21.07. This fixes heap corruption when a 
nexus is unpublished or destroyed whilst an NVMe client is still 
connected to it.

The upgrade required fixes to leaked bdev aliases and corrects the use
of spdk_nvme_probe_poll_async().

Fixed CAS-1042, CAS-1056.

PR #1008, opened 2021-09-01
chore(test): validate remaining pytest tests

cjones1024:
Include all current pytests in scripts/pytest-tests.sh.

Identify, and temporarily disable, those tests that
are currently failing and require additional work.

### DIFF
--- a/scripts/pytest-tests.sh
+++ b/scripts/pytest-tests.sh
@@ -1,6 +1,24 @@
 #! /usr/bin/env bash
 
-set -euxo pipefail
+set -eu -o pipefail
+
+function run_tests()
+{
+  while read name extra
+  do
+    case "$name" in
+      test_*)
+        if [ -f "$name.py" ]
+        then
+        (
+          set -x
+          pytest --tc-file=test_config.ini "$name.py"
+        )
+        fi
+      ;;
+    esac
+  done
+}
 
 if [ "${SRCDIR:-unset}" = unset ]
 then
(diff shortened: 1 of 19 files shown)


cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
The following tests are excluded, as they are currently failing:
- test_cli_controller (2nd test fails due to error during teardown)
- test_rpc
- test_multi_nexus
- test_nexus
- test_null_nexus

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

gila (changes requested):
There seems to be an empty file? Also, is there a reason we reduce the number of nexus volumes from 100 to 3?

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
> 
> 
> There seems to be an empty file? Also, is there a reason we reduce the number of nexus volumes from 100 to 3?

There were a couple of files that had changes to their permissions only (the test_*.py files should not be executable).
The nunber of GUIDs/volumes was actually 10 rather than 3. This was initially an oversight in that my development VM could not cope with 100. However, after adding extra memory/CPUs/hugepages, I still cannot get this test to run with the full 100.  It also appears to be failing for the same reason when run in the Jenkins pipeline. The main issue seems to be that pytest is capturing the log output in memory, and this test is generating a *lot* of messages. I am hoping that pytest can be coerced into writing the output generated directly to files rather than caching in memory, but I need to investigate this further.  Have disabled this test for now, and will raise a separate Jira/PR to address the resource issues.

cjones1024:
bors try

jonathan-teh (approved):
Nice refactoring for commonality and readability.
There were some changes to `test_nexus_resv_key` to bring the test up-to-date, which I'd overlooked in #980 and #1002.

zeenix:
> The following tests are excluded, as they are currently failing:
> 
>     * test_cli_controller
> 
>     * test_rpc
> 
>     * test_nexus_multipath

Ah, so I wasn't alone there after all. :) How come the CI is not failing? :thinking:

cjones1024:
> 
> 
> > The following tests are excluded, as they are currently failing:
> > ```
> > * test_cli_controller
> > 
> > * test_rpc
> > 
> > * test_nexus_multipath
> > ```
> 
> Ah, so I wasn't alone there after all. :) How come the CI is not failing? 🤔

There are only two tests remaining that are failing. They are test_rpc and test_null_nexus. These and a number of other tests were not included in the "CI" script that invokes the tests as they were relying of an additional VM in which to connect to the published taqrgets and run fio. 
The purpose of this particular PR is to remove that requirement and ensure that all existing tests will be included in the CI pipeline.
This is now the case with the exception of the two tests mentioned above.

cjones1024:
> 
> 
> There seems to be an empty file? Also, is there a reason we reduce the number of nexus volumes from 100 to 3?

The number of nexuses has now been set to 10 in order to ensure that the test passes while I am currently investigating why the test fails with much larger number of nexues.

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

gila (approved):

cjones1024:
bors merge

cjones1024:
bors try

PR #1012, opened 2021-09-13
feat(grpc): add mayastor v1 service for pools

shubham14bajpai:
- adds the implementation of mayastor v1 service for Pool.
- adds separate rpcs to import and export pools
- operations on non-existing pool will return an error instead of null

Signed-off-by: shubham <shubham14bajpai@gmail.com>

### DIFF
--- a/mayastor/src/grpc/mayastor/v1.rs
+++ b/mayastor/src/grpc/mayastor/v1.rs
@@ -1,0 +1,196 @@
+use crate::{
+    core::Share,
+    grpc::{rpc_submit, GrpcClientContext, GrpcResult, Serializer},
+    lvs::{Error as LvsError, Lvs, PoolArgs},
+    subsys::PoolConfig,
+};
+use futures::FutureExt;
+use nix::errno::Errno;
+use std::{convert::TryFrom, fmt::Debug, time::Duration};
+use tonic::{Request, Response, Status};
+
+use rpc::mayastor::v1::*;
+
+#[derive(Debug)]
+struct UnixStream(tokio::net::UnixStream);
+
+use ::function_name::named;
+use std::panic::AssertUnwindSafe;
+
+#[derive(Debug)]
+pub struct MayastorSvc {
+    name: String,
+    interval: Duration,
+    client_context: tokio::sync::Mutex<Option<GrpcClientContext>>,
+}
+
+#[async_trait::async_trait]
+impl<F, T> Serializer<F, T> for MayastorSvc
+where
+    T: Send + 'static,
+    F: core::future::Future<Output = Result<T, Status>> + Send + 'static,
+{
+    async fn locked(&self, ctx: GrpcClientContext, f: F) -> Result<T, Status> {
+        let mut context_guard = self.client_context.lock().await;
+
+        // Store context as a marker of to detect abnormal termination of the
+        // request. Even though AssertUnwindSafe() allows us to
+        // intercept asserts in underlying method strategies, such a
+        // situation can still happen when the high-level future that
+        // represents gRPC call at the highest level (i.e. the one created
+        // by gRPC server) gets cancelled (due to timeout or somehow else).
+        // This can't be properly intercepted by 'locked' function itself in the
+        // first place, so the state needs to be cleaned up properly
+        // upon subsequent gRPC calls.
+        if let Some(c) = context_guard.replace(ctx) {
+            warn!("{}: gRPC method timed out, args: {}", c.id, c.args);
+        }
+
+        let fut = AssertUnwindSafe(f).catch_unwind();
+        let r = fut.await;
+
+        // Request completed, remove the marker.
+        let ctx = context_guard.take().expect("gRPC context disappeared");
+
+        match r {
+            Ok(r) => r,
+            Err(_e) => {
+                warn!("{}: gRPC method panicked, args: {}", ctx.id, ctx.args);
+                Err(Status::cancelled(format!(
+                    "{}: gRPC method panicked",
+                    ctx.id
+                )))
+            }
+        }
+    }
+}
+
+impl TryFrom<CreatePoolRequest> for PoolArgs {
+    type Error = LvsError;
+    fn try_from(pool: CreatePoolRequest) -> Result<Self, Self::Error> {
+        if pool.disks.is_empty() {
+            Err(LvsError::Invalid {
+                source: Errno::EINVAL,
+                msg: "Missing devices".to_string(),
+            })
+        } else {
+            Ok(Self {
+                name: pool.name,
+                disks: pool.disks,
+            })
+        }
+    }
+}
+
+impl MayastorSvc {
+    pub fn new(interval: Duration) -> Self {
+        Self {
+            name: String::from("CSISvc"),
+            interval,
+            client_context: tokio::sync::Mutex::new(None),
+        }
+    }
+}
+
+impl From<Lvs> for Pool {
+    fn from(l: Lvs) -> Self {
+        Self {
+            name: l.name().into(),
+            disks: vec![l.base_bdev().bdev_uri().unwrap_or_else(|| "".into())],
+            state: PoolState::PoolOnline.into(),
+            capacity: l.capacity(),
+            used: l.used(),
+        }
+    }
+}
+
+#[tonic::async_trait]
+impl mayastor_server::Mayastor for MayastorSvc {
+    #[named]
+    async fn create_pool(
+        &self,
+        request: Request<CreatePoolRequest>,
+    ) -> GrpcResult<Pool> {
+        self.locked(
+            GrpcClientContext::new(&request, function_name!()),
+            async move {
+                let args = request.into_inner();
+
+                if args.disks.is_empty() {
+                    return Err(Status::invalid_argument("Missing devices"));
+                }
+
+                let rx = rpc_submit::<_, _, LvsError>(async move {
+                    let pool = Lvs::create_or_import(PoolArgs::try_from(args)?)
+                        .await?;
+                    // Capture current pool config and export to file.
+                    PoolConfig::capture().export().await;
+                    Ok(Pool::from(pool))
+                })?;
+
+                rx.await
+                    .map_err(|_| Status::cancelled("cancelled"))?
+                    .map_err(Status::from)
+                    .map(Response::new)
+            },
+        )
+        .await
+    }
+
+    #[named]
+    async fn destroy_pool(
+        &self,
+        request: Request<DestroyPoolRequest>,
+    ) -> GrpcResult<Null> {
+        self.locked(
+            GrpcClientContext::new(&request, function_name!()),
+            async move {
+                let args = request.into_inner();
+                info!("{:?}", args);
+                let rx = rpc_submit::<_, _, LvsError>(async move {
+                    if let Some(pool) = Lvs::lookup(&args.name) {
+                        // Remove pool from current config and export to file.
+                        // Do this BEFORE we actually destroy the pool.
+                        let mut config = PoolConfig::capture();
+                        config.delete(&args.name);
+                        config.export().await;
+
+                        pool.destroy().await?;
+                    }
+                    Ok(Null {})
+                })?;
+
+                rx.await
+                    .map_err(|_| Status::cancelled("cancelled"))?
+                    .map_err(Status::from)
+                    .map(Response::new)
+            },
+        )
+        .await
+    }
+
+    #[named]
+    async fn list_pools(
+        &self,
+        request: Request<Null>,
+    ) -> GrpcResult<ListPoolsReply> {
+        self.locked(
+            GrpcClientContext::new(&request, function_name!()),
+            async move {
+                let rx = rpc_submit::<_, _, LvsError>(async move {
+                    Ok(ListPoolsReply {
+                        pools: Lvs::iter()
+                            .map(|l| l.into())
+                            .collect::<Vec<Pool>>(),
+                    })
+                })?;
+
+                rx.await
+                    .map_err(|_| Status::cancelled("cancelled"))?
+                    .map_err(Status::from)
+                    .map(Response::new)
+            },
+        )
+        .await
+    }
+}
--- a/mayastor/src/grpc/mayastor_grpc.rs
+++ b/mayastor/src/grpc/mayastor_grpc.rs
@@ -32,7 +32,7 @@ use crate::{
         Serializer,
     },
     host::{blk_device, resource},
-    lvs::{Error as LvsError, Lvol, Lvs},
+    lvs::{Error as LvsError, Lvol, Lvs, PoolArgs},
     nexus_uri::NexusBdevError,
     subsys::PoolConfig,
 };
@@ -119,6 +119,23 @@ impl MayastorSvc {
             name: String::from("CSISvc"),
             interval,
             rw_lock: tokio::sync::RwLock::new(None),
+        }
+    }
+}
+
+impl TryFrom<CreatePoolRequest> for PoolArgs {
+    type Error = LvsError;
+    fn try_from(pool: CreatePoolRequest) -> Result<Self, Self::Error> {
+        if pool.disks.is_empty() {
+            Err(LvsError::Invalid {
+                source: Errno::EINVAL,
+                msg: "Missing devices".to_string(),
+            })
+        } else {
+            Ok(Self {
+                name: pool.name,
+                disks: pool.disks,
+            })
         }
     }
 }
@@ -241,7 +258,8 @@ impl mayastor_server::Mayastor for MayastorSvc {
                 }
 
                 let rx = rpc_submit::<_, _, LvsError>(async move {
-                    let pool = Lvs::create_or_import(args).await?;
+                    let pool = Lvs::create_or_import(PoolArgs::try_from(args)?)
+                        .await?;
                     // Capture current pool config and export to file.
                     PoolConfig::capture().export().await;
                     Ok(Pool::from(pool))
--- a/mayastor/src/grpc/mayastor_grpc.rs
+++ b/mayastor/src/grpc/mayastor_grpc.rs
@@ -135,6 +135,7 @@ impl TryFrom<CreatePoolRequest> for PoolArgs {
             _ => Ok(Self {
                 name: args.name,
                 disks: args.disks,
+                uuid: None,
             }),
         }
     }
--- a/mayastor/src/grpc/mayastor_grpc.rs
+++ b/mayastor/src/grpc/mayastor_grpc.rs
@@ -32,8 +32,9 @@ use crate::{
         Serializer,
     },
     host::{blk_device, resource},
-    lvs::{Error as LvsError, Lvol, Lvs, PoolArgs},
+    lvs::{Error as LvsError, Lvol, Lvs},
     nexus_uri::NexusBdevError,
+    pool::PoolArgs,
     subsys::PoolConfig,
 };
 
@@ -125,17 +126,16 @@ impl MayastorSvc {
 
 impl TryFrom<CreatePoolRequest> for PoolArgs {
     type Error = LvsError;
-    fn try_from(pool: CreatePoolRequest) -> Result<Self, Self::Error> {
-        if pool.disks.is_empty() {
-            Err(LvsError::Invalid {
+    fn try_from(args: CreatePoolRequest) -> Result<Self, Self::Error> {
+        match args.disks.len() {
+            0 => Err(LvsError::Invalid {
                 source: Errno::EINVAL,
-                msg: "Missing devices".to_string(),
-            })
-        } else {
-            Ok(Self {
-                name: pool.name,
-                disks: pool.disks,
-            })
+                msg: "invalid argument, missing devices".to_string(),
+            }),
+            _ => Ok(Self {
+                name: args.name,
+                disks: args.disks,
+            }),
         }
     }
 }
--- a/mayastor/src/grpc/mod.rs
+++ b/mayastor/src/grpc/mod.rs
@@ -40,6 +40,9 @@ mod json_grpc;
 mod mayastor_grpc;
 mod nexus_grpc;
 mod server;
+pub mod mayastor {
+    pub mod v1;
+}
 
 mod v1 {
     pub mod bdev;
--- a/mayastor/src/grpc/mod.rs
+++ b/mayastor/src/grpc/mod.rs
@@ -40,8 +40,8 @@ mod json_grpc;
 mod mayastor_grpc;
 mod nexus_grpc;
 mod server;
-pub mod mayastor {
-    pub mod v1;
+pub mod v1 {
+    pub mod pool;
 }
 
 mod v1 {
--- a/mayastor/src/grpc/mod.rs
+++ b/mayastor/src/grpc/mod.rs
@@ -41,12 +41,9 @@ mod mayastor_grpc;
 mod nexus_grpc;
 mod server;
 pub mod v1 {
-    pub mod pool;
-}
-
-mod v1 {
     pub mod bdev;
     pub mod json;
+    pub mod pool;
 }
 
 #[derive(Debug)]
--- a/mayastor/src/grpc/v1/pool.rs
+++ b/mayastor/src/grpc/v1/pool.rs
@@ -160,6 +160,7 @@ impl PoolRpc for PoolSvc {
             GrpcClientContext::new(&request, function_name!()),
             async move {
                 let args = request.into_inner();
+                info!("{:?}", args);
                 match PoolBackend::try_from(args.pooltype)? {
                     PoolBackend::Lvs => {
                         let rx = rpc_submit::<_, _, LvsError>(async move {
@@ -250,7 +251,7 @@ impl PoolRpc for PoolSvc {
                     } else {
                         return Err(LvsError::Invalid {
                             source: Errno::EINVAL,
-                            msg: format!("pool {} not found", args.name,),
+                            msg: format!("pool {} not found", args.name),
                         });
                     }
                     Ok(())
@@ -274,7 +275,7 @@ impl PoolRpc for PoolSvc {
             GrpcClientContext::new(&request, function_name!()),
             async move {
                 let args = request.into_inner();
-
+                info!("{:?}", args);
                 let rx = rpc_submit::<_, _, LvsError>(async move {
                     let pool = Lvs::import_from_args(PoolArgs::try_from(args)?)
                         .await?;
--- a/mayastor/src/grpc/v1/pool.rs
+++ b/mayastor/src/grpc/v1/pool.rs
@@ -1,7 +1,8 @@
 use crate::{
     core::Share,
     grpc::{rpc_submit, GrpcClientContext, GrpcResult, Serializer},
-    lvs::{Error as LvsError, Lvs, PoolArgs},
+    lvs::{Error as LvsError, Lvs},
+    pool::{PoolArgs, PoolBackend},
     subsys::PoolConfig,
 };
 use futures::FutureExt;
@@ -9,7 +10,7 @@ use nix::errno::Errno;
 use std::{convert::TryFrom, fmt::Debug, time::Duration};
 use tonic::{Request, Response, Status};
 
-use rpc::mayastor::v1::*;
+use rpc::mayastor::v1::pool::*;
 
 #[derive(Debug)]
 struct UnixStream(tokio::net::UnixStream);
@@ -18,14 +19,14 @@ use ::function_name::named;
 use std::panic::AssertUnwindSafe;
 
 #[derive(Debug)]
-pub struct MayastorSvc {
+pub struct PoolSvc {
     name: String,
     interval: Duration,
     client_context: tokio::sync::Mutex<Option<GrpcClientContext>>,
 }
 
 #[async_trait::async_trait]
-impl<F, T> Serializer<F, T> for MayastorSvc
+impl<F, T> Serializer<F, T> for PoolSvc
 where
     T: Send + 'static,
     F: core::future::Future<Output = Result<T, Status>> + Send + 'static,
@@ -67,22 +68,37 @@ where
 
 impl TryFrom<CreatePoolRequest> for PoolArgs {
     type Error = LvsError;
-    fn try_from(pool: CreatePoolRequest) -> Result<Self, Self::Error> {
-        if pool.disks.is_empty() {
-            Err(LvsError::Invalid {
+    fn try_from(args: CreatePoolRequest) -> Result<Self, Self::Error> {
+        match args.disks.len() {
+            0 => Err(LvsError::Invalid {
                 source: Errno::EINVAL,
-                msg: "Missing devices".to_string(),
-            })
-        } else {
-            Ok(Self {
-                name: pool.name,
-                disks: pool.disks,
-            })
+                msg: "invalid argument, missing devices".to_string(),
+            }),
+            _ => Ok(Self {
+                name: args.name,
+                disks: args.disks,
+            }),
         }
     }
 }
 
-impl MayastorSvc {
+impl TryFrom<ImportPoolRequest> for PoolArgs {
+    type Error = LvsError;
+    fn try_from(args: ImportPoolRequest) -> Result<Self, Self::Error> {
+        match args.disks.len() {
+            0 => Err(LvsError::Invalid {
+                source: Errno::EINVAL,
+                msg: "invalid argument, missing devices".to_string(),
+            }),
+            _ => Ok(Self {
+                name: args.name,
+                disks: args.disks,
+            }),
+        }
+    }
+}
+
+impl PoolSvc {
     pub fn new(interval: Duration) -> Self {
         Self {
             name: String::from("CSISvc"),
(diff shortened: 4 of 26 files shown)


rpc/src/lib.rs
```diff
@@ -38,3 +38,24 @@ pub mod mayastor {
 
     include!(concat!(env!("OUT_DIR"), "/mayastor.rs"));
 }
+
+#[allow(dead_code)]
+#[allow(clippy::type_complexity)]
+#[allow(clippy::unit_arg)]
+#[allow(clippy::redundant_closure)]
+#[allow(clippy::upper_case_acronyms)]
+pub mod mayastorv1 {
```
gila, line 47: what if we tried something like this:
```
pub mod mayastor {
   mod v1 { ... }
   mod v2 { ..}
}
```

Then we can use it as mayastor::v1::Pool and mayastor::v2::Pool ?

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -125,6 +125,23 @@ impl MayastorSvc {
     }
 }
 
+impl TryFrom<CreatePoolRequest> for PoolArgs {
+    type Error = LvsError;
```
gila, line 129: I don't think this should be a LvsError rather a gRPC Error?

zeenix:
@shubham14bajpai Could you please add some context/rationale/description for this change in the commit log? Also, is the `refactor` the appropriate tag here? Isn't it more like a `feat` or `chore`?

rpc/src/lib.rs
```diff
@@ -38,3 +38,24 @@ pub mod mayastor {
...
+
+    #[derive(Debug)]
+    pub enum Error {
+        ParseError,
+    }
+
+    impl From<()> for Null {
```
zeenix, line 54: what's this all about?

zeenix: it seems very generic so why in this module? :thinking:

rpc/src/lib.rs
zeenix, line 47: nit: why not just name it `v1`? I think @gila pointed that out for another similar named module.

rpc/src/lib.rs
gila: It has to be, as the you cant impl traits otherwise....

rpc/src/lib.rs
```diff
@@ -38,3 +38,24 @@ pub mod mayastor {
...
+
+#[allow(dead_code)]
+#[allow(clippy::type_complexity)]
+#[allow(clippy::unit_arg)]
+#[allow(clippy::redundant_closure)]
+#[allow(clippy::upper_case_acronyms)]
+pub mod mayastorv1 {
+
+    #[derive(Debug)]
+    pub enum Error {
```
zeenix, line 50: what's this error with only 1 variant needed for? Why in this module? This also seems very generic.

mayastor/src/grpc/mod.rs
```diff
@@ -40,6 +40,9 @@ mod json_grpc;
 mod mayastor_grpc;
 mod nexus_grpc;
 mod server;
+pub mod mayastor {
```
zeenix, line 43: do we really need the outer `mayastor` mod? We only have 1 API here, even if there could be multiple versions of it.

mayastor/src/grpc/mayastor/v1.rs
```diff
@@ -0,0 +1,196 @@
...
+struct UnixStream(tokio::net::UnixStream);
+
+use ::function_name::named;
+use std::panic::AssertUnwindSafe;
+
+#[derive(Debug)]
+pub struct MayastorSvc {
+    name: String,
+    interval: Duration,
+    rw_lock: tokio::sync::RwLock<Option<GrpcClientContext>>,
```
zeenix, line 24: The content of the lock are more important than the lock type itself so I'd name identifiers after the data, so `client_context` would be better IMO.

mayastor/src/grpc/mayastor/v1.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+#[async_trait::async_trait]
+impl<F, T> Serializer<F, T> for MayastorSvc
+where
+    T: Send + 'static,
+    F: core::future::Future<Output = Result<T, Status>> + Send + 'static,
+{
+    async fn locked(&self, ctx: GrpcClientContext, f: F) -> Result<T, Status> {
+        let mut guard = self.rw_lock.write().await;
```
zeenix, line 34: similarly here: I'd suggest re-using `client_context` (or simply `context`), or `context_guard` if you like.

mayastor/src/grpc/mayastor/v1.rs
zeenix: Also, why a `RwLock` if you never take a read lock on it? :thinking:

mayastor/src/grpc/mayastor/v1.rs
```diff
@@ -0,0 +1,196 @@
...
+    name: String,
+    interval: Duration,
+    rw_lock: tokio::sync::RwLock<Option<GrpcClientContext>>,
+}
+
+#[async_trait::async_trait]
+impl<F, T> Serializer<F, T> for MayastorSvc
+where
+    T: Send + 'static,
+    F: core::future::Future<Output = Result<T, Status>> + Send + 'static,
```
zeenix, line 31: This would look more readable with a separate where clause for the `Output`:

```suggestion
    F: core::future::Future + Send + 'static,
    F::Output: Result<T, Status>,
```

rpc/src/lib.rs
zeenix: > It has to be,

What has to be? :) Which traits?

mayastor/src/grpc/mayastor/v1.rs
```diff
@@ -0,0 +1,196 @@
...
+            capacity: l.capacity(),
+            used: l.used(),
+        }
+    }
+}
+
+#[tonic::async_trait]
+impl mayastor_server::Mayastor for MayastorSvc {
+    #[named]
+    async fn create_pool(
```
akhilerm, line 110: Dont we need to update the pool_cli also to point to the new functions?

mayastor/src/grpc/mayastor_grpc.rs
shubham14bajpai: This call is made inside a spdk thread so the returned error should be LvsError 
https://github.com/openebs/mayastor/blob/067c7f4270835dd4c475be6de631d1dbc37fa5af/mayastor/src/grpc/mayastor_grpc.rs#L245
If we take grpcError we need to convert it again.

Also the PoolArgs struct is declared inside the lvs lib so taking LvsError

rpc/src/lib.rs
shubham14bajpai: I guess this is not needed. Removing it.

mayastor/src/grpc/mayastor/v1.rs
```diff
@@ -0,0 +1,196 @@
...
+    #[named]
+    async fn destroy_pool(
+        &self,
+        request: Request<DestroyPoolRequest>,
+    ) -> GrpcResult<Null> {
+        self.locked(
+            GrpcClientContext::new(&request, function_name!()),
+            async move {
+                let args = request.into_inner();
+                info!("{:?}", args);
```
mtzaurus, line 149: Shall we  make it debug() ?

mayastor/src/grpc/mayastor/v1.rs
shubham14bajpai: I guess Result is not a trait so the where clause is throwing `not a trait` error. Is there another way to use the where clause?

mayastor/src/grpc/mayastor/v1.rs
shubham14bajpai: IMO info makes more sense as user should get an history of the volumes which got deleted. The details of deletion process can be in debug.

mayastor/src/grpc/mod.rs
shubham14bajpai: The mayastor mod is the folder that is introduced inside the grpc folder which will have the v1, v2 etc. Also we put all the mods into one we may end up conflicting structs and methods as all will be having the same names

mayastor/src/grpc/mayastor/v1.rs
zeenix: ah, didn't think of that. :) nm.

mayastor/src/grpc/mayastor/v1.rs
```diff
@@ -0,0 +1,196 @@
...
+            client_context: tokio::sync::Mutex::new(None),
+        }
+    }
+}
+
+impl From<Lvs> for Pool {
+    fn from(l: Lvs) -> Self {
+        Self {
+            name: l.name().into(),
+            disks: vec![l.base_bdev().bdev_uri().unwrap_or_else(|| "".into())],
```
cjones1024, line 99: Not sure this is the right thing to do here. What does it mean to have a disk with no name/uri ?

gila (approved):

mayastor/src/grpc/mayastor/v1.rs
shubham14bajpai: We should wait for all the code to migrate otherwise we may have two different clients in the cli

mayastor/src/grpc/mayastor/v1.rs
shubham14bajpai: This function is used to convert the `lvs` pool into `rpc` pool for ListPools. I tried changing it to try_from instead but in that case we might need to ignore the entire pool on error which is not ideal I guess.

mayastor/src/grpc/mod.rs
akhilerm: Does the mayastor mod need to be public? Only the v1 is needed to be pub right?

akhilerm:
As the proto changes for bdev/json rpc versioning are merged in mayastor-api, #1022  needs to be merged first, before merging this PR. Else build will fail with `not imlpemented` error.

mayastor/src/grpc/v1/pool.rs
```diff
@@ -0,0 +1,301 @@
...
+                uuid: None,
+            }),
+        }
+    }
+}
+
+impl TryFrom<ImportPoolRequest> for PoolArgs {
+    type Error = LvsError;
+    fn try_from(args: ImportPoolRequest) -> Result<Self, Self::Error> {
+        match args.disks.len() {
```
gila, line 89: I think this implementation is somewhat hard to read and it seems its mostly a if statement mode here. Rust has support for that so perhaps we can make use of that to make it easier on the eyes. 

```rust

if args.disk.empty() {
   return Err(....)
}
```

```

File: mayastor/src/grpc/v1/pool.rs [hunk 13]
```diff
@@ -0,0 +1,301 @@
...
+            0 => Err(LvsError::Invalid {
+                source: Errno::EINVAL,
+                msg: "invalid argument, missing devices".to_string(),
+            }),
+            _ => {
+                let uuid = if let Some(UuidValue::Uuid(uuid)) = args.uuid_value
```
gila, line 95: This can be expressed simpler; by hoisting it out of the match. We then, in effect, have two if statements that are easy to understand for everyone. And then return the proper value. Also, let us say we, in the future, want to be more precise with the values the UUID field can have (i.e., v4 checking). It would be clear where that part is handled.

mayastor/src/grpc/v1/pool.rs
```diff
@@ -0,0 +1,301 @@
...
+                let args = request.into_inner();
+                match PoolBackend::try_from(args.pooltype)? {
+                    PoolBackend::Lvs => {
+                        let rx = rpc_submit::<_, _, LvsError>(async move {
+                            let pool = Lvs::create_or_import(
+                                PoolArgs::try_from(args)?,
+                            )
+                            .await?;
+                            // Capture current pool config and export to file.
+                            PoolConfig::capture().export().await;
```
gila, line 154: We dont need to capture this anymore, it can be removed.

mayastor/src/grpc/v1/pool.rs
```diff
@@ -0,0 +1,301 @@
...
+            GrpcClientContext::new(&request, function_name!()),
+            async move {
+                let args = request.into_inner();
+                info!("{:?}", args);
+                let rx = rpc_submit::<_, _, LvsError>(async move {
+                    if let Some(pool) = Lvs::lookup_by_uuid(&args.uuid) {
+                        // Remove pool from current config and export to file.
+                        // Do this BEFORE we actually destroy the pool.
+                        let mut config = PoolConfig::capture();
+                        config.delete(&args.uuid);
```
gila, line 184: all the config.* stuff can be removed.

mayastor/src/lvs/lvs_pool.rs
```diff
@@ -244,6 +257,85 @@ impl Lvs {
         }
     }
 
+    /// imports a pool based on its name, uuid and base bdev name
+    pub async fn import_from_args(args: PoolArgs) -> Result<Lvs, Error> {
```
gila, line 261: Is there are reason this fn is not called "import(args: poolArgs)"  ?

mayastor/src/lvs/lvs_pool.rs
```diff
@@ -244,6 +257,85 @@ impl Lvs {
...
+        if args.disks.len() != 1 {
```
gila, line 262: Within the try_from() parts, this is already checked, so this can not happen, can it?

mayastor/src/lvs/lvs_pool.rs
```diff
@@ -244,6 +257,85 @@ impl Lvs {
...
+                msg: format!(
+                    "invalid number {} of devices {:?}",
+                    args.disks.len(),
+                    args.disks
+                ),
+            });
+        }
+
+        // default to aio if no specific driver scheme is specified
+        let disks = args
```
gila, line 274: I think we want to move this into the try_from() part as well.

mayastor/src/lvs/lvs_pool.rs
```diff
@@ -244,6 +257,85 @@ impl Lvs {
...
+            .map(|d| {
+                if Url::parse(d).is_err() {
+                    format!("aio://{}", d)
+                } else {
+                    d.clone()
+                }
+            })
+            .collect::<Vec<_>>();
+
+        let parsed = uri::parse(&disks[0]).map_err(|e| Error::InvalidBdev {
```
gila, line 286: same here.

mayastor/src/lvs/lvs_pool.rs
shubham14bajpai: No, thinking a little ahead for lvm, the try_from only checks for 0 devices. Here we are checking devices should be exactly 1 in case of lvs.

mayastor/src/lvs/lvs_pool.rs
shubham14bajpai: Similar I want to keep the PoolArgs generic for both lvm and lvs that way we don't repeat the common checks

mayastor/src/lvs/lvs_pool.rs
shubham14bajpai: There is another function called "import(name: &str, bdev: &str)". If we change that now then a lot of rust test may break.

test/python/tests/replica/features/pool.feature
```diff
@@ -18,7 +18,7 @@ Feature: Mayastor pool management
   Scenario: creating a pool with a name that already exists
     Given a pool "p0"
     When the user creates a pool with the name of an existing pool
-    Then the pool creation should succeed
+    Then the pool create command should fail
```
akhilerm, line 21: Should we have this change in old API also, because consumers of old API does not expect pool creation to fail if pool already exists?

akhilerm (commented):
Given a few comments @shubham14bajpai

mayastor/src/lvs/lvs_pool.rs
```diff
@@ -244,6 +257,85 @@ impl Lvs {
...
+            });
+        }
+
+        // default to aio if no specific driver scheme is specified
+        let disks = args
+            .disks
+            .iter()
+            .map(|d| {
+                if Url::parse(d).is_err() {
+                    format!("aio://{}", d)
```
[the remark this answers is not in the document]
shubham14bajpai: The current default is aio so taking it from there

test/python/tests/replica/features/pool.feature
shubham14bajpai: We are taking these in develop and we are eventually moving towards changing the control plane according. Taking this will help us find any control plane changes needed early.

tiagolobocastro: Fail is a bit generic, it should specify the error code/type, otherwise it will be difficult to determine if it's a failure creating the pool (let's say device error) or if it's because it already exists.
Will there be a documentation for the semantic changes?
Seems that now Pool's have a uuid as well? should it also be checked if it already exists?

shubham14bajpai: For create if it exists we return a `EEXIST` and rest of the checks will return `EINVAL`. And yes there should be documentation for it. I don't have one but will come up with one as soon as possible.

For the uuid, the create request does not have the uuid field. The import rpc is now separate and has a optional uuid field that will check for the uuid if provided.

tiagolobocastro: Would it be possible to add the uuid for the create?

shubham14bajpai: IMO for the create request uuid should not be needed as we won't have a uuid with us before the pool is created. For import the uuid can be helpful as we know the uuid of an exported pool we need to import.

tiagolobocastro: hmm, so what happens if the create times out, but ends up succeeding. Now the control plane doesn't know the uuid of the pool, so how does it reconcile things?

shubham14bajpai: As we added a filter for the name in ListPools, we can check for the pool before creating it.
The reconcile loop would be like
```
  pool = ListPools(ListPoolOptions(name="tpool"))
  if pool == nil {
      CreatePool(name="tpool", disks=["/dev/sdc"])
  }

```

tiagolobocastro: right but then we're reverting to just checking by name and not uuid. So in that case the control plane could instead just set the name to a uuid (assuming the name can be that long) and keep its own name<->uuid mapping.

shubham14bajpai: Can there be multiple pools with the same name? If no how can uuid help with reconciling? Why do we need to keep a different name mapping in the control plane?

tiagolobocastro: I don't know if it can or not. If it can then the uuid would be the tie breaker if you will. Let's say you plug a pool in the mayastor node which happens to have the same name as another pool, how does the control plane know which one is which without a uuid? Ofcourse the uuid can also be the same, but in that case nothing will help :D

shubham14bajpai: Ohhh now it make sense, it is better to have it and not use it than to not have it :sweat_smile: . On the same lines can you take a look at the proto file for the messages https://github.com/openebs/mayastor-api/pull/8/files

shubham14bajpai: So for the Create request the uuid will be optional and it will be checked only if provided. And we still need to fail if the pool exists for the given name or uuid. right?

shubham14bajpai: Yes both lvm and lvs have a uuid, not in the same format but unique in the way spdk and lvm generates one.

shubham14bajpai: Sure we can add a check to lookup_by_uuid and reject the request if one already exists.

akhilerm: FYI: For LVM pools and volumes the uuid will always be auto generated, we can never set it manually.

mayastor/src/grpc/v1/pool.rs
```diff
@@ -0,0 +1,311 @@
...
+        request: Request<CreatePoolRequest>,
+    ) -> GrpcResult<Pool> {
+        self.locked(
+            GrpcClientContext::new(&request, function_name!()),
+            async move {
+                let args = request.into_inner();
+                match PoolBackend::try_from(args.pooltype)? {
+                    PoolBackend::Lvs => {
+                        let rx = rpc_submit::<_, _, LvsError>(async move {
+                            let pool = Lvs::create_or_import(
```
tiagolobocastro, line 145: since we now have import, do we still need create_or_import, or could we just have a create?

rpc/src/lib.rs
```diff
@@ -37,4 +25,31 @@ pub mod mayastor {
...
+    pub mod v1 {
+
+        // dont export the raw pb generated code
+        mod pb {
+            include!(concat!(env!("OUT_DIR"), "/mayastor.v1.rs"));
+        }
+
+        pub mod pool {
+            pub use super::pb::{
+                destroy_pool_request::UuidValue as DestroyUuid,
```
tiagolobocastro, line 38: Is there anyway to make these the same and have a single UuidValue type?

test/python/tests/replica/features/pool.feature
tiagolobocastro: Yep @shubham14bajpai explained this to me offline. We may have a solution for LVM adding the uuid as a tag, and this way hopefully be able to keep the same uuid format for different pool types.

akhilerm:
bors try

shubham14bajpai:
bors try

mtzaurus:
bors try

mayastor/src/grpc/v1/pool.rs
```diff
@@ -0,0 +1,326 @@
...
+    ) -> GrpcResult<()> {
+        self.locked(
+            GrpcClientContext::new(&request, function_name!()),
+            async move {
+                let args = request.into_inner();
+                info!("{:?}", args);
+                let rx = rpc_submit::<_, _, LvsError>(async move {
+                    if let Some(pool) = Lvs::lookup(&args.name) {
+                        if let Some(uuid) = args.uuid {
+                            if uuid != pool.uuid() {
```
tiagolobocastro, line 191: nit: would a single if work?
```rust
if args.uuid.is_some() && args.uuid != Some(pool.uuid()
```
not sure if it's better or not though

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -123,6 +124,23 @@ impl MayastorSvc {
...
+    fn try_from(args: CreatePoolRequest) -> Result<Self, Self::Error> {
+        match args.disks.len() {
+            0 => Err(LvsError::Invalid {
+                source: Errno::EINVAL,
+                msg: "invalid argument, missing devices".to_string(),
+            }),
+            _ => Ok(Self {
+                name: args.name,
+                disks: args.disks,
+                uuid: None,
```
tiagolobocastro, line 138: Shouldn't the uuid be filled from `args` ?

tiagolobocastro:
bors try

mayastor/src/grpc/v1/pool.rs
shubham14bajpai: We will wait until we make appropriate changes in the control plane. Once we start making use of import there I will remove the import part from create. That was my thought.

mayastor/src/grpc/v1/pool.rs
```diff
@@ -0,0 +1,326 @@
...
+        Ok(Self {
+            name: args.name,
+            disks: args.disks,
+            uuid: args.uuid,
+        })
+    }
+}
+
+impl PoolSvc {
+    pub fn new(interval: Duration) -> Self {
```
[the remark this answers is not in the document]
shubham14bajpai: Actually, it is no longer needed. :+1:

mayastor/src/grpc/mayastor_grpc.rs
shubham14bajpai: This is from the old service so did not want to change the existing behavior. This will only be taken for mayastor.v1 services.

mayastor/src/grpc/v1/pool.rs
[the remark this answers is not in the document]
shubham14bajpai: It was earlier used here https://github.com/openebs/mayastor/commit/487df468b069b53c166ec67162914416fe810935 but that code was removed I guess

tiagolobocastro:
bors delegate @shubham14bajpai

mayastor/src/grpc/mayastor_grpc.rs
tiagolobocastro: Ah right, makes sense!

mayastor/src/grpc/v1/pool.rs
tiagolobocastro: hmm yeah could do that, but we could just start with the create ("only") as well since the control-plane will not make use of the v1 yet, so when we start making use of it we can also start to use the separate create/import logic.

mayastor/src/grpc/v1/pool.rs
```diff
@@ -0,0 +1,324 @@
...
+
+    #[named]
+    async fn import_pool(
+        &self,
+        request: Request<ImportPoolRequest>,
+    ) -> GrpcResult<Pool> {
+        self.locked(
+            GrpcClientContext::new(&request, function_name!()),
+            async move {
+                let args = request.into_inner();
```
tiagolobocastro, line 276: are we missing an info! here?

shubham14bajpai:
bors try

mayastor/src/grpc/v1/pool.rs
```diff
@@ -0,0 +1,324 @@
...
+    #[named]
+    async fn export_pool(
+        &self,
+        request: Request<ExportPoolRequest>,
+    ) -> GrpcResult<()> {
+        self.locked(
+            GrpcClientContext::new(&request, function_name!()),
+            async move {
+                let args = request.into_inner();
+                info!("{:?}", args);
```
tiagolobocastro, line 235: so we have infos on enter, is there somewhere an info or other log with the result of the call?

tiagolobocastro:
bors delegate

tiagolobocastro:
bors d+

mayastor/src/grpc/v1/pool.rs
shubham14bajpai: Yes there is a log in the destroy function called in this. I will try to make the logs consistent for all different RPCs. Preferably moving the logs from grpc file to lvs lib with proper sentences instead of just dumping the logs.

shubham14bajpai:
bors try

mayastor/src/grpc/v1/pool.rs
tiagolobocastro: hmm potential problem with moving it to the lvs, is that if a pool could not be found for example, we would not get the log, unless we have a specific log for the not found case...
but where will the log of the result go?

tiagolobocastro: I think it may be worth making use of tracing here because I can give us things like automatic logs on error etc.

test/python/tests/replica/test_bdd_pool.py
```diff
@@ -147,7 +152,11 @@ def pool_creation_should_fail(find_pool):
     assert find_pool("p0") == None
 
 
-@then("the pool creation should succeed")
+@then("the pool create command should fail")
+def the_pool_destroy_command_should_fail(create_pool_that_already_exists):
```
akhilerm, line 156:
```suggestion
def the_pool_create_command_should_fail(create_pool_that_already_exists):
```

shubham14bajpai:
bors try

mayastor/src/grpc/v1/pool.rs
shubham14bajpai: Added `info!("{:?}", args);` to all the rpc functions as it looked much cleaner

mayastor/src/grpc/v1/pool.rs
shubham14bajpai: Added tracing to all the lvs functions called in the rpc function

tiagolobocastro: any reason to adding it to the lvs instead of the grpc? In the grpc we'd miss some errors, eg the ret error on 241.

shubham14bajpai:
bors try

mayastor/src/grpc/v1/pool.rs
shubham14bajpai: It shows error like this:
```
[2021-12-03T12:33:52.484652812+05:30 ERROR mayastor::grpc::v1::pool:pool.rs:232] error=status: InvalidArgument, message: "errno EINVAL: Invalid argument: pool tpool not found", details: [], metadata: MetadataMap { headers: {} }
```

As this will be returned as the grpc status anyways and no actual mayastor gets triggered I thought we can ignore such errors.

tiagolobocastro: hmm yeah good point, don't know... one pro to keeping it is that we have the info! message on start, so might look confusing if there is no corresponding end message.

mayastor/src/grpc/v1/pool.rs
tiagolobocastro: Do we also need an info! for the success case at the end?

mayastor/src/grpc/v1/pool.rs
shubham14bajpai: The lvs function create_or_import is common between both current and v1 rpcs. I did not want to break the existing one and same time didn't want to write a entire new function. This will also help us test run v1 with the control plane when we finally start replacing the rpcs. Will refactor this slowly after that.

mayastor/src/grpc/v1/pool.rs
shubham14bajpai: I checked all the called functions have info! message at the end in them and they look correct. I think adding an exit message will make it repetitive. Another issue I faced was once we pass the args then at the end we get the `already moved` error which again will require hacks like cloning the fields etc.

mayastor/src/grpc/v1/pool.rs
shubham14bajpai: There is a success message in all the called functions.

shubham14bajpai:
bors try

mayastor/src/lvs/lvs_pool.rs
```diff
@@ -243,25 +257,130 @@ impl Lvs {
...
+                } else {
+                    d.clone()
+                }
+            })
+            .collect::<Vec<_>>();
```
tiagolobocastro, line 285: I think we may be able to simplify this code a bit, since we already know that we expect 1 and only 1 disk. So rather than doing the check with `if args.disks.len() != 1 {` we might be able to:
```rust
  let disk = match args.disks.first() {
      Some(disk) if args.disks.len() == 1 => {
          disk
      },
      _ => return err "LEN NOT 1",
  };
```

mayastor/src/lvs/lvs_pool.rs
```diff
@@ -243,25 +257,130 @@ impl Lvs {
...
+
+        let parsed = uri::parse(&disks[0]).map_err(|e| Error::InvalidBdev {
+            source: e,
+            name: args.name.clone(),
+        })?;
```
tiagolobocastro, line 290: What about having a common parse function that to use aio::// if the "raw disk" fails to parse and use it everywhere we want to try aio to avoid code duplication?

tiagolobocastro: Maybe even the initial check for disk len 1 can be made common since that is also used elsewhere.

mayastor/src/pool.rs
```diff
@@ -131,3 +136,32 @@ impl From<Pool> for rpc::Pool {
...
 }
+
+/// PoolArgs is used to translate the input for the grpc
+/// Create/Import requests which comtains name, uuid & disks.
+/// This help us avoid importing grpc structs in the actual lvs mod
+#[derive(Clone, Debug)]
+pub struct PoolArgs {
+    pub name: String,
+    pub disks: Vec<String>,
+    pub uuid: Option<String>,
```
tiagolobocastro, line 147: Rather than String, should this be PoolUuid or a uuid::Uuid?

mayastor/src/pool.rs
```diff
@@ -131,3 +136,32 @@ impl From<Pool> for rpc::Pool {
         }
     }
 }
+
+/// PoolArgs is used to translate the input for the grpc
+/// Create/Import requests which comtains name, uuid & disks.
+/// This help us avoid importing grpc structs in the actual lvs mod
+#[derive(Clone, Debug)]
+pub struct PoolArgs {
+    pub name: String,
```
tiagolobocastro, line 145: Should we define a PoolName type? Eg: are there constraints to how short/long the name can be?

mayastor/src/pool.rs
```diff
@@ -131,3 +136,32 @@ impl From<Pool> for rpc::Pool {
...
+    pub disks: Vec<String>,
```
tiagolobocastro, line 146: can we add getters rather than having these as pub?

tiagolobocastro (approved):

mayastor/src/lvs/lvs_pool.rs
shubham14bajpai: `uri::parse(&disk)` returns a object of type `Box<dyn BdevCreateDestroy<Error = NexusBdevError>>` which I am not sure we should be included in lvs_pool
Keeping this part separate for now.

mayastor/src/pool.rs
shubham14bajpai: For now spdk does not have limit I guess but we can do this future

mayastor/src/pool.rs
shubham14bajpai: This will require some cleanup in `subsys/config` code will take that in separate PR.

shubham14bajpai:
bors try

akhilerm (approved):
LGTM

PR #1007, opened 2021-08-31
test(e2e): support for coverage builds

chriswldenyer:
Allow optional building of a coverage-enabled
version of mayastor image. Option --coverage
added to release.sh.
Use mayastor image as host for mayastor-csi
for CI testing.

PR #1013, opened 2021-09-15
test(e2e): use mayastor image for mayastor-csi daemonset

chriswldenyer:
Use mayastor image as host for mayastor-csi for CI testing.
Use the "command" field in the yamls for mayastor and mayastor-csi to specify unambiguously which binary to run.
Part of PR #1007 moved to separate PR as requested in review.
This relates to mq-1989.

### DIFF
--- a/chart/templates/mayastor-daemonset.yaml
+++ b/chart/templates/mayastor-daemonset.yaml
@@ -56,6 +56,8 @@ spec:
         - "-P/var/local/mayastor/pools.yaml"
         - "-l{{ include "mayastorCpuSpec" . }}"
         - "-pmayastor-etcd"
+        command:
+        - mayastor
         securityContext:
           privileged: true
         volumeMounts:
--- a/chart/templates/csi-daemonset.yaml
+++ b/chart/templates/csi-daemonset.yaml
@@ -28,7 +28,7 @@ spec:
       # the same.
       containers:
       - name: mayastor-csi
-        image: {{ include "mayastorImagesPrefix" . }}mayadata/mayastor-csi:{{ .Values.mayastorImagesTag }}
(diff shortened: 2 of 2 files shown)


blaisedias (approved):
LGTM

chart/templates/mayastor-daemonset.yaml
```diff
@@ -56,6 +56,8 @@ spec:
         - "-P/var/local/mayastor/pools.yaml"
         - "-l{{ include "mayastorCpuSpec" . }}"
         - "-pmayastor-etcd"
+        command:
+        - mayastor
```
zeenix, line 60: * nit: this specific change is a separate logical change so best to put it in a separate patch/commit with its own description/rationale. 

* Talking of rationales, the commit log is missing that. Please also provide it on the commits, not just the PR so it can be view quickly and/or offline.

zeenix (changes requested):
Otherwise LGTM.

chart/templates/mayastor-daemonset.yaml
blaisedias: I don't think it is a separate logical change, we will be using the same image for mayastor and maystor-csi.
So for consistency both yaml files should have the command argument.

chriswldenyer: I've just updated the git commit message to match the PR description.
The change in the mayastor-daemonset.yaml is related to the change in mayastor-csi-daemonset.yaml in that they both use the same image and so should use the "command" field to specify the binary to run. I have updated the comments to reflect this.

zeenix: > I don't think it is a separate logical change, we will be using the same image for mayastor and maystor-csi.

Separate logical change in the sense of commit, not PR.

GlennBullingham (approved):
LGTM

PR #1015, opened 2021-09-15
chore(ci): refactor python tests to allow custom docker-compose.yml

cjones1024:
Reorganise related test_*.py files in their own separate directory.
This also allows the addition of their own (custom)
docker-compose.yml file.
Note that, consequently, the way the tests are invoked has now changed slightly,
See test/python/README.md and scripts/pytest-tests.sh for details


cjones1024:
bors try

test/python/tests/ana_client/docker-compose.yml
```diff
@@ -60,15 +62,19 @@ services:
...
         ipv4_address: 10.0.0.4
     cap_add:
+      # NUMA related
       - SYS_ADMIN
       - SYS_NICE
+      # uring needs mmap
       - IPC_LOCK
     security_opt:
+      # we can set this to a JSON file to allow per syscall access
       - seccomp:unconfined
```
jonathan-teh, line 78: As we're changing this file anyway, could you change this to `=` as this is in the system logs:
```
dockerd[1018]: time="2021-09-15T12:09:43.820452714+01:00" level=warning msg="Security options with `:` as a separator are deprecated and will be completely unsupported in 17.04, use `=` instead."
```

jonathan-teh (approved):

gila (approved):

scripts/pytest-tests.sh
```diff
@@ -30,24 +34,18 @@ cd "$SRCDIR/test/python" && source ./venv/bin/activate
...
-test_nexus_multipath
-
-test_multi_nexus
-test_nexus
-test_null_nexus
+tests/replica
+tests/publish
+tests/rebuild
+
+# tests/csi
```
akhilerm, line 41: are the CSI tests disabled now?

scripts/pytest-tests.sh
```diff
@@ -30,24 +34,18 @@ cd "$SRCDIR/test/python" && source ./venv/bin/activate
...
+
+tests/ana_client
+tests/cli_controller
+tests/replica_uuid
```
akhilerm, line 45: Should the replica_uuid tests also come under `tests/replica`?

test/python/tests/rpc/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
+#
+# {SRCDIR} should point to your working tree which should be your current pwd
+#
+
+version: '3'
+services:
+  ms1:
```
akhilerm, line 7: Is there a specific format that we are following with the naming of test containers? I saw that some docker-compose files uses ms0 as the first contatiner name. 

Or is this done so that tests can be run parallely?

test/python/tests/cli_controller/docker-compose.yml
```diff
@@ -0,0 +1,92 @@
+#
+# {SRCDIR} should point to your working tree which should be your current pwd
+#
+
+version: '3'
+services:
+  ms1:
```
akhilerm, line 7: Does the naming of containers follow a certain format? here it starts with ms1, but in some files it starts with ms0

akhilerm (commented):
Some general questions:
1. Are these tests in separate directories run in parallel?
2. The replica_uuid feature mentions about v1 api [here](https://github.com/openebs/mayastor/blob/51367353c05ec4743c01a0d520a920a6c5c68cac/test/python/tests/replica_uuid/features/replica-guid.feature#L17
), Shouldn't it be renamed v2?

scripts/pytest-tests.sh
cjones1024: Yes, but this is a temporary measure while I resolve issues with running the mayastor-csi plugin from within a container.

test/python/tests/rpc/docker-compose.yml
cjones1024: Prior to this PR there has been a single docker-compose.yml file, and the tests have followed a tacit convention that containers are "numbered", starting from ms0. They also have fixed IP addresses where ms0 has address 10.0.0.2, ms1 has address 10.0.0.3, etc.

Since each (new) test now has its own custom docker-compose.yml, there is no reason why this convention *must* be followed. However, I would recommend that we continue do do so unless there is a good reason for doing otherwise. Having said that, there was one test that did not use "ms0" as its first container and, as I did not write it, decided to leave it as is - at least for now.

There is no intention at this time to run the pytests in parallel.

cjones1024:
> 
> 
> Some general questions:
> 
>     1. Are these tests in separate directories run in parallel?
> 

There is no intention at this time to run the pytests in parallel.
The reason they are in separate directories is so that each directory may contain a docker-compose.yml file.
Thus all tests that require the same docker-compose.yml file *can* reside in the same directory.
Although there is no requirement that they *must* do so as the directory structure can be used to group logically related tests.
The only rule that needs to be observed is that tests with different container configurations must reside in reside directories

>     2. The replica_uuid feature mentions about v1 api [here](https://github.com/openebs/mayastor/blob/51367353c05ec4743c01a0d520a920a6c5c68cac/test/python/tests/replica_uuid/features/replica-guid.feature#L17), Shouldn't it be renamed v2?

Possibly. That question should perhaps be directed to the test's author @mtzaurus. However, I would prefer that is addressed in a separate Jira/PR

scripts/pytest-tests.sh
cjones1024: This is indeed possible from a technical perspective. I have kept them as "logically" distinct for now as they were authored by different people. As mentioned below, there are other minor changes to this test that perhaps need addressing. If so I would prefer that "moving" the test be done at that time instead of here and now

test/python/tests/ana_client/docker-compose.yml
cjones1024: Done

test/python/tests/cli_controller/docker-compose.yml
cjones1024: See comments above. Yes this test probably should use "ms0" as its first container. However, I did not write this test, and decided to leave it as is - at least for now.

akhilerm (approved):
LGTM

cjones1024:
bors merge

cjones1024:
bors merge

cjones1024:
bors merge

PR #1016, opened 2021-09-15
fix(nvmx): handle failures from processing completions for admin comands

jonathan-teh:
Handle it by scheduling a future once, and only once, to destroy the
device if spdk_nvme_ctrlr_process_admin_completions() returns an
error, usually -ENXIO if the remote child is unreachable when
handling admin commands such as keep-alive.

This means an unreachable remote child on an idle nexus is now noticed
and the child is marked as degraded. In the case where the nexus is
under IO load and 2 remote replicas are unreachable at almost the same
time, both replicas are now consistently degraded.

Fixes CAS-941

### DIFF
--- a/mayastor/src/bdev/nvmx/controller.rs
+++ b/mayastor/src/bdev/nvmx/controller.rs
@@ -56,6 +56,7 @@ use crate::{
         IoDevice,
         OpCompletionCallback,
         OpCompletionCallbackArg,
+        Reactors,
     },
     ffihelper::{cb_arg, done_cb},
     nexus_uri::NexusBdevError,
@@ -503,6 +504,10 @@ impl<'a> NvmeController<'a> {
                 source: Errno::EBUSY,
             }
         })?;
+        // Prevent racing device destroy
+        unsafe {
+            self.timeout_config.as_mut().start_device_destroy();
+        }
 
         debug!("{} shutting down the controller", self.name);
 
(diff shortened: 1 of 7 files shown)


jonathan-teh:
bors try

gila (approved):

mayastor/src/bdev/nvmx/controller.rs
```diff
@@ -821,22 +822,36 @@ extern "C" fn aer_cb(ctx: *mut c_void, cpl: *const spdk_nvme_cpl) {
...
+    //
+    // ENXIO: the qpair is not connected or when the controller is
+    // marked as failed.
+    //
+    // EAGAIN: returned whenever the controller is being reset.
     let result = context.process_adminq();
 
     if result < 0 {
-        //error!("{}: {}", context.name, Errno::from_i32(result.abs()));
+        error!(
```
mtzaurus, line 842: Don't we see lots of errors flooding the console ? I do recall some situations where logs were full of error messages related to failures upon processing adminq completions, that's why it was commented out.

mayastor/src/bdev/nvmx/controller.rs
```diff
@@ -821,22 +822,36 @@ extern "C" fn aer_cb(ctx: *mut c_void, cpl: *const spdk_nvme_cpl) {
...
+            "process adminq: {}: {}",
+            context.name,
+            Errno::from_i32(result.abs())
+        );
+        if context.start_device_destroy() {
```
mtzaurus, line 847: Can we add a new message to denote the scheduling of device removal, so that it's clear that the operation has been started ? Straight after the 'if ...' ?
We could also display here the error that caused the failure and it will be printed only once without flooding the console (not sure if we care about all the subsequent errors since the device was identified as faulted and removal has been scheduled anyways).

mayastor/src/bdev/nvmx/controller.rs
jonathan-teh: Yes, without scheduling the device destroy future the log will flood. With it, this is seen only once on an idle nexus, maybe 3-5 times on a nexus with an IO load.

mayastor/src/bdev/nvmx/controller.rs
jonathan-teh: Good idea, will do.

mayastor/src/bdev/nvmx/controller.rs
mtzaurus: Good then, thanks.

mtzaurus (approved):

jonathan-teh:
This seems to have revealed another issue, the `persist_io_failure` cargo test is failing this assert:
```
    // Expect child2 to be faulted due to an I/O error.
    let child = child_info(&nexus_info, &uuid(&child2));
    assert!(!child.healthy);
```
because the persisted entry for the child in etcd must still be in state `ChildOpen` whereas the state in Mayastor is `ChildDegraded`.

jonathan-teh:
bors try

jonathan-teh:
bors try

jonathan-teh:
bors try

cjones1024 (approved):

jonathan-teh:
bors try

jonathan-teh:
bors try

jonathan-teh:
bors merge

PR #1017, opened 2021-09-16
chore: speed up replica creation/destruction

paulyoong:
Improve the performance of replica creation/destruction by:
1. Increasing the size of the buffer used to wipe the superblock from 4K to 64K.
2. Wiping the superblock on creation only.

### DIFF
--- a/mayastor/src/lvs/lvol.rs
+++ b/mayastor/src/lvs/lvol.rs
@@ -249,7 +249,7 @@ impl Lvol {
                     name: self.name(),
                 }
             })?;
-            // write zero to the first 8MB which whipes the metadata and the
+            // write zero to the first 8MB which wipes the metadata and the
             // first 4MB of the data partition
             let range =
                 std::cmp::min(self.as_bdev().size_in_bytes(), (1 << 20) * 8);
@@ -292,7 +292,6 @@ impl Lvol {
 
         // we must always unshare before destroying bdev
         let _ = self.unshare().await;
-        self.wipe_super().await?;
 
         let name = self.name();
 
--- a/mayastor/src/lvs/lvol.rs
+++ b/mayastor/src/lvs/lvol.rs
@@ -225,7 +225,7 @@ impl Lvol {
         }
     }
 
-    // wipe the first MB if unmap is not supported on failure the operation
+    // wipe the first 8MB if unmap is not supported on failure the operation
     // needs to be repeated
     pub async fn wipe_super(&self) -> Result<(), Error> {
         if !unsafe { self.0.as_ref().clear_method == LVS_CLEAR_WITH_UNMAP } {
@@ -238,7 +238,7 @@ impl Lvol {
                         name: self.name(),
                     }
                 })?;
-            let buf = hdl.dma_malloc(1 << 12).map_err(|e| {
+            let buf = hdl.dma_malloc(1 << 16).map_err(|e| {
                 error!(
                     ?self,
                     ?e,
@@ -254,7 +254,7 @@ impl Lvol {
             let range =
                 std::cmp::min(self.as_bdev().size_in_bytes(), (1 << 20) * 8);
             debug!(?self, ?range, "zeroing range");
-            for offset in 0 .. (range >> 12) {
+            for offset in 0 .. (range >> 16) {
                 hdl.write_at(offset * buf.len(), &buf).await.map_err(|e| {
                     error!(?self, ?e);
                     Error::RepDestroy {
--- a/mayastor/src/lvs/lvol.rs
+++ b/mayastor/src/lvs/lvol.rs
@@ -22,6 +22,7 @@ use spdk_sys::{
     vbdev_lvol_destroy,
     vbdev_lvol_get_from_bdev,
     LVS_CLEAR_WITH_UNMAP,
+    SPDK_BDEV_LARGE_BUF_MAX_SIZE,
 };
 
 use crate::{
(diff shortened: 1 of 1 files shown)


paulyoong:
bors try

mayastor/src/lvs/lvol.rs
```diff
@@ -292,7 +292,6 @@ impl Lvol {
 
         // we must always unshare before destroying bdev
         let _ = self.unshare().await;
-        self.wipe_super().await?;
```
zeenix, line 295: More a n00b question rather than a comment on your changes: how do we ensure no user data gets leaked? I see that `wipe_super` zeroes the first 4MB of the data. Why only 4MB? :thinking:

paulyoong: `wipe_super` is also zeroing out the superblock such that it would make any filesystem already on the disk unusable.

I'd have to defer to @gila as to why we chose zero out 4MB of the data partition, but in general we didn't want to be zeroing out the entire disk as this would be time consuming. Specifically from the commit message (https://github.com/openebs/Mayastor/commit/40545fff42055db242ad1bc10463b3ebf6951b59):
"When the base bdev does not support UNMAP, writing zero's is costly and very time consuming"

Also, the creation/deletion can be called from a gRPC request, so if the operations are taking a significant amount of time the gRPC calls can timeout. This change is actually trying to address this as we were seeing timeouts in the control plane due to this.

tiagolobocastro: If we don't do this on destroy, should we reduce the DestroyVolume specific timeout?

paulyoong: Yes I think we should do that. Then, if at a later stage we start seeing the DestroyVolume timing out, we'll have a relatively quick feedback loop.

zeenix: Thanks @paulyoong for explaining in detail. I've a good understanding of your changes now and why we avoid zeroing out the whole volume (which I kinda guessed already) but my main question remains: how do we ensure no user data gets leaked? Or is that not a concern?

zeenix: by "leaked", I mean remaining on the drive, in case no future volumes overwrite it, which could be sensitive. Perhaps data is always encrypted?

paulyoong: You are correct - at the moment the data remains on the disk. I remember there being a discussion about this previously and I think it was decided that this should be addressed in the future.

zeenix (approved):
LGTM.

mayastor/src/lvs/lvol.rs
```diff
@@ -238,7 +238,7 @@ impl Lvol {
                         name: self.name(),
                     }
                 })?;
-            let buf = hdl.dma_malloc(1 << 12).map_err(|e| {
+            let buf = hdl.dma_malloc(1 << 16).map_err(|e| {
```
akhilerm, line 241: question: Even if this block size is changed, Will the actual IO to the disk be still done in 4k blocks?

paulyoong: It's not the block size that is being changed, rather the size of the buffer that is being passed to SPDK. SPDK can handle buffers up to a size of `SPDK_BDEV_LARGE_BUF_MAX_SIZE` which is 64KB (https://github.com/spdk/spdk/blob/49911c47fb9bdae96cecd6a3f299cf87c4fba200/include/spdk/bdev.h#L56).

How the I/Os get submitted to the underlying disk is up to SPDK, but by increasing the buffer size we are able to make fewer calls into SPDK and let them perform work against this larger buffer.

Note: Rebuild also uses this buffer size to improve performance.

mayastor/src/lvs/lvol.rs
```diff
@@ -249,12 +249,12 @@ impl Lvol {
...
                 }
             })?;
-            // write zero to the first 8MB which whipes the metadata and the
+            // write zero to the first 8MB which wipes the metadata and the
             // first 4MB of the data partition
             let range =
                 std::cmp::min(self.as_bdev().size_in_bytes(), (1 << 20) * 8);
             debug!(?self, ?range, "zeroing range");
-            for offset in 0 .. (range >> 12) {
+            for offset in 0 .. (range >> 16) {
```
tiagolobocastro, line 257: Nit: Can we derive this from 8MiB together with the buffer size we're using? To make it easier to change if we ever need to?

mayastor/src/lvs/lvol.rs
tiagolobocastro: Should we actually use `SPDK_BDEV_LARGE_BUF_MAX_SIZE` ?

tiagolobocastro (approved):

mtzaurus:
I have a general question: is it ok to leave the data after deleting the LVOL ? Is wiping out all data upon volume removal not what we are supposed to do in the first place ?
Shall we have a tuneable option to allow data removal once an LVOL is removed ?

mayastor/src/lvs/lvol.rs
paulyoong: Good idea - done.

mayastor/src/lvs/lvol.rs
paulyoong: Sure - done.

paulyoong:
> I have a general question: is it ok to leave the data after deleting the LVOL ? Is wiping out all data upon volume removal not what we are supposed to do in the first place ?
> Shall we have a tuneable option to allow data removal once an LVOL is removed ?
From the discussion that were had, it seemed ok to remove the wiping of the data from the destruction. Wiping out all the data on volume removal would be good but it takes a long time and things tend to time out. We had code previously that attempted to do this (implementation of unmap) but there were problems with this. 

I definitely think this is something that needs revisting and that we can improve on.

paulyoong:
bors merge

PR #1018, opened 2021-09-17
fix(pytest): create explicit readiness_check in MayastorHandle

cjones1024:
Attempt to work around gRPC bug by retrying once on failure.

### DIFF
--- a/test/python/common/hdl.py
+++ b/test/python/common/hdl.py
@@ -18,8 +18,7 @@ class MayastorHandle(object):
         self.channel = grpc.insecure_channel(("%s:10124") % self.ip_v4)
         self.bdev = rpc.BdevRpcStub(self.channel)
         self.ms = rpc.MayastorStub(self.channel)
-        self.bdev_list()
-        self.pool_list()
+        self._readiness_check()
 
     def set_timeout(self, timeout):
         self.timeout = timeout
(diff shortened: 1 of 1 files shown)


cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

cjones1024:
bors try

jonathan-teh (approved):

test/python/common/hdl.py
```diff
@@ -34,12 +33,21 @@ def install_stub(self, name):
                 setattr(stub, f, partial(h, timeout=self.timeout))
         return stub
 
+    def _readiness_check(self):
+        try:
```
tiagolobocastro, line 37: shouldn't we wait for a set amount of time until mayastor is ready? otherwise the tests will still fail if we're not immediately ready?

cjones1024: Although it's not immediately clear but both bdev_list() and pool_list() actually invoke the appropriate gRPC call with wait_for_ready=True, which should queue the requests until the channel is ready. The problem is that there appears to be a bug in the python gRPC code where in some situations this doesn't always work as it is supposed to. It has been claimed that an immediate retry does seem to resolve the issue. That's what this change does, but it is difficult to verify as I have been unable to reproduce the conditions under which the bug is triggered. At the very least, the change should not make matters worse.

tiagolobocastro (approved):

test/python/common/hdl.py
cjones1024: Will update the commit message to be more informative

cjones1024:
bors merge

PR #1020, opened 2021-09-21
chore(ci): modify CSI pytests to run mayastor-csi plugin inside container

cjones1024:
resolves CAS-1109

### DIFF
--- a/test/python/tests/csi/test_bdd_csi.py
+++ b/test/python/tests/csi/test_bdd_csi.py
@@ -91,8 +91,9 @@ def mayastor_instance(container_mod):
 
 @pytest.fixture(scope="module")
 def csi_container(container_mod):
-    container = container_mod.get("ms1")
-    yield container.get("NetworkSettings.Networks.mayastor_net.IPAddress")
+    # wait until container has address
+    container_mod.get("ms1").get("NetworkSettings.Networks.mayastor_net.IPAddress")
+    yield
 
 
 @pytest.fixture(scope="module")
--- a/test/python/tests/csi/test_bdd_csi.py
+++ b/test/python/tests/csi/test_bdd_csi.py
@@ -1,21 +1,17 @@
 import pytest
 from pytest_bdd import given, scenario, then, when, parsers
 
-import logging
 import os
-import signal
 import subprocess
-import threading
 import time
 
+from common.mayastor import container_mod
 from common.hdl import MayastorHandle
 from common.csi_hdl import CsiHandle
 
 import grpc
 import csi_pb2 as pb
 import csi_pb2_grpc as rpc
-
-pytest_plugins = ["docker_compose"]
 
 
 class Nexus:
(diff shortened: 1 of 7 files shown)


cjones1024:
bors try

test/python/tests/csi/test_bdd_csi.py
```diff
@@ -86,63 +82,27 @@ def get_volume_capability(volume, read_only):
...
     )
-    yield handles
 
 
 @pytest.fixture(scope="module")
-def mayastor_instance(handles_mod):
-    yield handles_mod["ms0"]
+def csi_container(container_mod):
+    container = container_mod.get("ms1")
+    yield container.get("NetworkSettings.Networks.mayastor_net.IPAddress")
```
mtzaurus, line 95: Does it yield container itself or the container's IP address ? In case of IP address, I think we can rename the fixture to 'csi_container_address' or so, to make it more consistent, IMHO.

test/python/tests/csi/test_csi.py
```diff
@@ -1,83 +1,43 @@
...
+    yield MayastorHandle(
+        container.get("NetworkSettings.Networks.mayastor_net.IPAddress")
     )
-    yield handles
 
 
 @pytest.fixture(scope="module")
-def mayastor_instance(handles_mod):
-    yield handles_mod["ms0"]
+def csi_container(container_mod):
```
mtzaurus, line 28: The function was changed to not return the container and return 'NetworkSettings.Networks.mayastor_net.IPAddress' instead. In case I overlooked it: who/what is the primary consumer of this new information in test modules ?
The only thing I saw was 'fix_socket_permissions', but it doesn't rely on fixture's output in any fashion.

test/python/tests/csi/test_bdd_csi.py
cjones1024: It is actually used to enforce dependencies and so the return value is not really useful. Will change so that it does not return a value.

test/python/tests/csi/test_csi.py
cjones1024: See above comment, It is used to enforce a dependency for timing purposes. We need to know that the csi_container is running before we start messing with the socket. As noted above shall change so that it does not produce a value.

mtzaurus: Thanks.

cjones1024:
bors try

mtzaurus:
Are we going to test CSI controller plugin as part of Mayastor ? To be properly tested, it requires all the Control Plane components to be deployed.
CSI controller BDD tests are added on behalf of CAS-1123, they be run as part of Control Plane BDD tests, with CSI controller being launched via the deployer app.

cjones1024:
> 
> 
> Are we going to test CSI controller plugin as part of Mayastor ? To be properly tested, it requires all the Control Plane components to be deployed. CSI controller BDD tests are added on behalf of CAS-1123, they be run as part of Control Plane BDD tests, with CSI controller being launched via the deployer app.

These tests are for the node plugin only. They should not be considered part of tests for the controller. Even if there ends up being some overlap, we could remove (or disable) the "duplicate" tests. The duration of the tests is sufficiently short however that it is probably worth retaining them anyway.

cjones1024:
bors try

PR #1026, opened 2021-09-30
Two random fixes

zeenix:


mayastor/src/bdev/dev.rs
```diff
@@ -17,30 +17,39 @@
...
+    use std::convert::TryFrom;
+
+    use snafu::ResultExt;
+
+    use crate::{
+        bdev::BdevCreateDestroy,
+        nexus_uri::{self, NexusBdevError},
+    };
+
+    use super::super::{aio, loopback, malloc, null, nvme, nvmx, uring};
```
tiagolobocastro, line 44: might tempted to suggest `crate::bdev::` instead of `super::super::`, not sure what's better

.pre-commit-config.yaml
```diff
@@ -12,7 +12,7 @@ repos:
     -   id: rust-style
         name: Rust style
         description: Run cargo fmt on files included in the commit. rustfmt should be installed before-hand.
-        entry: cargo fmt --all -- --check
+        entry: cargo +nightly fmt --all -- --check
```
tiagolobocastro, line 15: might want to change the Jenkins file to use +nightly as well?

tiagolobocastro (approved):

.pre-commit-config.yaml
zeenix: the problem is that `cargo` from nix (which is what CI uses) doesn't have `+nightly` option for some reason. Probably cause it's nightly by default. :shrug: `rustfmt` is backwords-compatible though so I'm not sure if we need to (somehow) setup different pre-commit config for the `default.nix` and `ci.nix`. Let's see what the CI says:

bors try

tiagolobocastro: Oh that's interesting, wonder why it works like that - but yeah good point, it's already nightly :+1:

jonathan-teh (approved):

zeenix:
bors merge

PR #1027, opened 2021-10-01
fix(nvmx): ensure child is faulted after handling admin command failure

jonathan-teh:
Ensure that a child that is inaccessible ends up in the state faulted due to IO error, regardless of whether the error is first noticed in the IO or admin completion paths.

Fixes CAS-1128


jonathan-teh:
bors try

jonathan-teh:
This has survived 3 hours of `test_nexus_2_remote_mirror_kill_all_fio`.

mayastor/src/bdev/nvmx/controller.rs
```diff
@@ -843,19 +850,17 @@ pub extern "C" fn nvme_poll_adminq(ctx: *mut c_void) -> i32 {
...
+            error!(
+                "process adminq: {}: {}",
+                context.name,
+                Errno::from_i32(result.abs())
+            );
+            info!("dispatching nexus fault and retire: {}", context.name);
             let dev_name = context.name.to_string();
             Reactors::master().send_future(async move {
-                let _ = destroy_device(dev_name).await;
+                nexus::nexus_fault_child(dev_name).await;
```
mtzaurus, line 862: I am a bit concerned that controller has knowledge about the holding nexus directly.
Can't we rather add indirection here by utilising existing event mechanism in NVMx controller to notify all the listeners about device failure ?
Nexus already registers itself as a listener to monitor events on its children devices, so we could probably extend the event handler to handle also device I/O failures.

mayastor/src/bdev/nexus/mod.rs
```diff
@@ -125,6 +132,38 @@ pub fn instances() -> &'static mut Vec<Box<Nexus>> {
...
+                    == c.state.compare_exchange(
+                        ChildState::Open,
+                        ChildState::Faulted(Reason::IoError),
+                    )
+            })
+        {
+            info!("{} retiring {}", nexus.name, child_name);
+            // Error indicates it is already paused and another thread is
+            // processing the fault
+            let _ = nexus.child_retire(child_name).await;
```
paulyoong, line 160: Should we output a message in the event of an error? It might help with future debugging.

mayastor/src/bdev/nvmx/controller.rs
jonathan-teh: Actually it doesn't, that function is iterating through all nexuses, and their children, looking for a child whose device name matches this controller name.

mayastor/src/bdev/nexus/mod.rs
jonathan-teh: Sure, I'll add that.

mayastor/src/bdev/nvmx/controller.rs
mtzaurus: The idea behind the controller event mechanism was to decouple controller events from the logic that processes it.
This situation with faulting nexus child looks like a very good example of that. We might want to have more than one consumer of such an event potentially.

jonathan-teh: I had a look at the `DeviceEventType::DeviceRemoved` issued at the end of `destroy_device()` (i.e. the previous fix for this issue) and this event is sent after the device is removed in order for the child to respond to it, in this case by removing the child from the nexus.
I could introduce a new event type of `FaultChild` then move `nexus::nexus_fault_child()` into `child_event_listener()` so the faulting is handled as an event instead of sending a `Future`.

mtzaurus: Sounds great ! Thanks !

jonathan-teh:
bors try

mtzaurus (approved):
Thanks !

mayastor/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -422,6 +423,56 @@ impl Nexus {
                     }
                 }
             }
+            DeviceEventType::AdminCommandCompletionFailed => {
+                let cn = &device;
+                for nexus in nexus::instances() {
+                    if nexus
+                        .children
+                        .iter()
+                        .filter(|c| c.state() == ChildState::Open)
```
tiagolobocastro, line 432: what if the child was rebuilding? So it's state would be Faulted(OutOfSync) IIRC?

jonathan-teh: Good point. This code that marks the child as `Faulted` is basically from `NexusChannelInner::fault_child()` but we can't easily call that from here because we don't have a channel as we're not performing IO. Presumably it'll just ignore the child and as the child is rebuilding anyway the failure will be noticed in the IO path and acted upon there.

jonathan-teh:
bors try

mayastor/src/bdev/nexus/nexus_bdev_children.rs
jonathan-teh: I've factored out the iterating over the children of a Nexus to mark it as faulted but I've left its functionality alone. A child being rebuilt, which then becomes inaccessible, is a separate case and I'm inclined to only leave this as fixing the this issue whilst ensuring the original issue of handling admin command completion failure remains fixed, without having to fix all possible issues in the area.

mayastor/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -422,6 +424,38 @@ impl Nexus {
...
+                            "{}: retiring child {} in response to admin command completion failure event",
+                            nexus.name,
+                            device,
+                        );
+
+                        let child_dev = device.to_string();
+                        Reactors::master().send_future(async move {
+                            // Error indicates it is already paused and another
+                            // thread is processing
+                            // the fault
```
tiagolobocastro, line 441: was this new line intentional?

.pre-commit-config.yaml
```diff
@@ -12,7 +12,7 @@ repos:
     -   id: rust-style
         name: Rust style
         description: Run cargo fmt on files included in the commit. rustfmt should be installed before-hand.
-        entry: cargo +nightly fmt --all -- --check
+        entry: cargo fmt --all -- --check
```
tiagolobocastro, line 15: any reason to remove the +nightly?

tiagolobocastro (approved):

mayastor/src/bdev/nexus/nexus_bdev_children.rs
jonathan-teh: Doh, not sure how that crept in, will remove.

.pre-commit-config.yaml
jonathan-teh: It doesn't work, I get
```
error: no such subcommand: `+nightly`
```
when trying to commit any Rust code.

PR #1028, opened 2021-10-05
fix(csi): use base hostname in csi plugin instead of full domain name

mtzaurus:
CSI node plugin now reports only its base name instead of fully
qualified domain name, to eliminates mismatches in node name comparison
when Kubernets tries to locate a Mayastor node to perform volume
binding.

Resolves: CAS-1095


mtzaurus:
bors merge

PR #1032, opened 2021-10-11
fix(csi): node_unpublish_volume() can fail (while reporting success)

cjones1024:
node_unpublish_volume() can fail (while reporting success)
if the nvmf target is unreachable, causing an endless loop
with node_unpublish_volume() being called repeatedly.

Cause:
A stat() call on the mountpoint can fail if the associated
nvmf target device is unreachable.
This failure is incorrectly interpreted to mean that the mountpoint
does not exist, and that the device has already been unmounted
when this is not the case. This causes node_unpublish_volume()
to report success without actually doing anything, and therefore
leaving the device mounted.

Fix:
The fix is to explicitly check for an error when calling stat(),
so as to to distinguish when there is an error other than 'NotFound',
and to still check for a mount in that case.

### DIFF
--- a/csi/src/node.rs
+++ b/csi/src/node.rs
@@ -1,6 +1,7 @@
 use std::{
     boxed::Box,
     collections::HashMap,
+    io::ErrorKind,
     path::Path,
     time::Duration,
     vec::Vec,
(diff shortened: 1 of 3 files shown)


cjones1024:
bors try

PR #1036, opened 2021-10-15
fix(nvmx): support for Unmap and WriteZeroes IoTypes

jonathan-teh:
Make support for these IoTypes dependent on the controller's ONCS (Optional NVM Command Support) field instead of hardcoding.

Fixes CAS-1164


jonathan-teh:
bors merge

PR #1037, opened 2021-10-18
fix(csi): node_unpublish_volume() fails if nvmf target is unreachable

cjones1024:
When performing I/O on a volume has failed due to the underlying nvmf
target being unreachable, subsequent attempts to unpublish the volume are
ineffective, resulting in an endless loop where node_unpublish_volume()
is called repeatedly with no effect.

Cause:
When the underlying nvmf target has timed out, and is thus disconnected,
we are left with a mount to a device that no longer exists. In the
case of a filesystem mount, any attempt to access the corresponding
mount_point fails with an EIO error, although an umount() would succeed
at this point. At present, node_unpublish_volume() implicitly performs a
stat() on the mount_point to determine what type of bind mount is present
(either a block special file or directory). It then proceeds to perform
an umount(), followed by deletion of the file or directly that is the
mountpoint. If the underyling nvmf device has been disconnected, then the
initial stat() can fail with an EIO. Unfortunately, this is incorrectly
interpreted to mean hat the mountpoint does not exist, and that the
device has already been unmounted. Thus node_unpublish_volume() assumes
that there is nothing to do and returns (reporting success). The result
is repeated calls to node_unpublish_volume() that have no material effect.

Fix:
The fix is to change the order in which the stat() and umount() are
performed. The required unmount operation is identical in this case,
regardless as to whether the mount_point is a file or directory.
The umount() should succeed even if the nvme device no longer
exists, and a subsequent stat() on the mount_point no longer fails,
thus allowing the deletion of the directory or file as appropriate.

resolves CAS-817


csi/src/node.rs
```diff
@@ -298,31 +296,74 @@ impl node_server::Node for Node {
...
+
+        let mount_point = Path::new(&msg.target_path);
+        match fs::metadata(mount_point) {
+            Ok(metadata) => {
+                if metadata.is_dir() {
+                    // Mount Volume
+                    debug!("Removing directory {}", msg.target_path);
+                    if let Err(error) = fs::remove_dir(mount_point) {
+                        return Err(failure!(
+                            Code::Internal,
```
tiagolobocastro, line 320: I guess there's not a more appropriate error code than `Internal` for this?

csi/src/node.rs
```diff
@@ -298,31 +296,74 @@ impl node_server::Node for Node {
...
+                    msg.volume_id,
+                    msg.target_path,
+                    error
+                ));
+            }
+        }
+
+        let mount_point = Path::new(&msg.target_path);
+        match fs::metadata(mount_point) {
+            Ok(metadata) => {
```
tiagolobocastro, line 314: nit: might be able to reduce if/else nesting by using:
```rust
Ok(metadata) if metadata.is_dir() => ...
Ok(metadata) if metadata.is_file() => ...
```

tiagolobocastro (approved):

csi/src/node.rs
cjones1024: Hadn't really given it too much thought as that was essentially what was used already, but no there is not really a better alternative.

csi/src/node.rs
cjones1024: I can change this if you feel strongly, but I think it doesn't really make much difference one way or the other.

tiagolobocastro: no not really, it's just personal preference nit :)

cjones1024:
bors try

test/python/k8s/bug/cas-817/node_unpublish_bug.py
```diff
@@ -0,0 +1,315 @@
+#! /usr/bin/python3
+
+import libvirt
```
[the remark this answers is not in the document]
cjones1024: removed for now

PR #1041, opened 2021-11-01
test(e2e): add support to build images which generate coverage data

blaisedias:


chriswldenyer (approved):
LGTM

gila (changes requested):
These coverage aspects do not belong in the YAML files and should be part of the testing bits. 

Also, why do we need a new image type and don't we set the RUSTFLAGS as an argument?

scripts/release.sh
```diff
@@ -108,20 +110,32 @@ while [ "$#" -gt 0 ]; do
...
+      shift
+      ;;
     *)
       echo "Unknown option: $1"
       exit 1
       ;;
   esac
 done
 
+if [ -n "$COVERAGE" ] && [ -n "$DEBUG" ]; then
```
gila, line 124: Why not ?

gila (changes requested):

scripts/release.sh
blaisedias: see https://github.com/openebs/mayastor/pull/1007 which you approved

nix/pkgs/mayastor/cargo-package.nix
```diff
@@ -83,15 +87,25 @@ let
...
-    buildType = "release";
-    buildInputs = buildProps.buildInputs ++ [ libspdk ];
-    SPDK_PATH = "${libspdk}";
-  });
-  debug = rustPlatform.buildRustPackage (buildProps // {
-    buildType = "debug";
-    buildInputs = buildProps.buildInputs ++ [ libspdk-dev ];
-    SPDK_PATH = "${libspdk-dev}";
-  });
+  release = rustPlatform.buildRustPackage
```
gila, line 90: Why is the indentation changed here?

scripts/release.sh
```diff
@@ -30,6 +30,7 @@ Options:
   -d, --dry-run              Output actions that would be taken, but don't run them.
   -h, --help                 Display this text.
   --registry <host[:port]>   Push the built images to the provided registry.
+  --coverage                 Build coverage-enabled version of images where possible.
```
gila, line 33: Shall we just add the coverage option to debug builds?

csi/src/server.rs
```diff
@@ -237,12 +237,29 @@ async fn main() -> Result<(), String> {
...
+    let mut signal_term = tokio::signal::unix::signal(
+        tokio::signal::unix::SignalKind::terminate(),
+    )
+    .expect("Failed to egister handler for SIGTERM");
+    let mut signal_int = tokio::signal::unix::signal(
+        tokio::signal::unix::SignalKind::interrupt(),
+    )
+    .expect("Failed to register handler for SIGINT");
+
+    tokio::select! {
```
gila, line 249: In line with comments elsewhere, I do not understand or see, why it helps to override the signal handler with nothing.

gila (changes requested):

scripts/release.sh
blaisedias: The first requirement is for coverage builds matching release code.
That is what this change is for.

nix/pkgs/mayastor/cargo-package.nix
blaisedias: fixed https://github.com/openebs/mayastor/pull/1041/commits/778fc7c2c74269b75e51c5e6f809f8a4466576e6

PR #1042, opened 2021-11-01
fix(nats): nats tcp stream stuck

tiagolobocastro:
The tcp stream to the nats-server can get stuck if the server is not
shutdown gracefully. As such Even the connection lost callback does not get
triggered because the stream is somehow stuck.

Adding the timeout if we do not receive a message with a timeout helps
address the issue: when the timeout expires the client now sends a ping
to the server and expects to receive a pong within the same timeout.
If a pong does not arrive in time then the connection is deemed lost.

As a side effect this may cause additional strain on the nats server when
too many clients are idle as the timeouts will be fired.

Don't cache connect_url's returned by the NATS INFO message as they may
get outdated and point to bogus IP addresses.

Also add connection handlers to highlight whenever the services connection
to the NATS server changes.


gila (approved):

mtzaurus (approved):

tiagolobocastro:
bors merge

PR #1045, opened 2021-11-02
fix(nexus_v2): add uuid field to the nexus

tiagolobocastro:
fix(nexus_v2): add uuid field to the nexus

The CreateV2 was changed to use uuid's but all the lookups are made using
the name, even though the arg is called uuid...
This should be properly addressed as part of 1107 but for now use the
workaround of looking up by uuid if by name fails.

When a nexus is published the NQN is set to the name but the namespace uuid's are
set to the nexus UUID. This breaks the CSI plugin logic, as it expects them to be
the same...
As a WA for the moment, when using the V2 api set the bdev uuid to match
the name, if the name is a UUID. The uuid arg is then used as a means to identify
the nexus only.
This means the nqn and the namespace uuid is now using the nexus name, which in
case of the control plane on k8s, is the volume uuid.

Also tidiy up some corner cases with UUID's which may be reused with
different names - spdk does not seem to treat UUID as unique...

With V2 there is no "nexus-" prefix to add env variable to the CSI plugin
so that we may keep the same behaviour for MOAC testing.

--------------
chore: update the deploy yamls

### DIFF
--- a/chart/templates/csi-daemonset.yaml
+++ b/chart/templates/csi-daemonset.yaml
@@ -43,7 +43,9 @@ spec:
             fieldRef:
               fieldPath: status.podIP
         - name: RUST_BACKTRACE
-          value: "1"
+          value: "1"{{ if .Values.moac }}
+        - name: MOAC
+          value: "true"{{ end }}
         args:
         - "--csi-socket=/csi/csi.sock"
         - "--node-name=$(MY_NODE_NAME)"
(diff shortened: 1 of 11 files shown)


tiagolobocastro:
bors try

tiagolobocastro:
bors try

gila:
Bors merge

gila (approved):

gila:
Bors merge

paulyoong (approved):

paulyoong:
bors merge

PR #1047, opened 2021-11-09
feat(libnvme-rs): new crate, implement connect and disconnect 

jonathan-teh:
New crate that uses the libnvme development C library. Reuse NvmeTarget, and associated code like NvmeDevice and NvmeError, from nvmeadm and implement connect and disconnect to an NVMe target using libnvme.

Fixes CAS-1218

### DIFF
--- a/libnvme-rs/src/error.rs
+++ b/libnvme-rs/src/error.rs
@@ -1,0 +1,65 @@
+use snafu::Snafu;
+
+#[derive(Debug, Snafu)]
+#[allow(missing_docs)]
+#[snafu(visibility = "pub(crate)")]
+pub enum NvmeError {
+    #[snafu(display("IO error:"))]
+    IoError { source: std::io::Error },
+    #[snafu(display("Failed to parse {}: {}, {}", path, contents, error))]
+    ValueParseError {
+        path: String,
+        contents: String,
+        error: String,
+    },
+    #[snafu(display("Failed to parse value"))]
+    ParseError {},
+    #[snafu(display("File IO error: {}, {}", filename, source))]
+    FileIoError {
+        filename: String,
+        source: std::io::Error,
+    },
+    #[snafu(display("nqn: {} not found", text))]
+    NqnNotFound { text: String },
+    #[snafu(display("No nvmf subsystems found"))]
+    NoSubsystems,
+    #[snafu(display("Connect in progress"))]
+    ConnectInProgress,
+    #[snafu(display("NVMe connect failed: {}, {}", filename, source))]
+    ConnectError {
+        source: std::io::Error,
+        filename: String,
+    },
+    #[snafu(display("Lookup host failed: {}", rc))]
+    LookupHostError { rc: i32 },
+    #[snafu(display("Add controller failed: {}", rc))]
+    AddCtrlrError { rc: i32 },
+    //#[snafu(display("IO error during NVMe discovery"))]
+    //DiscoveryError { source: nix::Error },
+    #[snafu(display("Controller with nqn: {} not found", text))]
+    CtlNotFound { text: String },
+    #[snafu(display("Invalid path {}: {}", path, source))]
+    InvalidPath {
+        source: std::path::StripPrefixError,
+        path: String,
+    },
+    #[snafu(display("NVMe subsystems error: {}, {}", path_prefix, source))]
+    SubSysError {
+        source: glob::PatternError,
+        path_prefix: String,
+    },
+    #[snafu(display("NVMe URI invalid: {}", source))]
+    UrlError { source: url::ParseError },
+    #[snafu(display("Transport type {} not supported", trtype))]
+    TransportError { trtype: String },
+    #[snafu(display("Invalid parameter: {}", text))]
+    InvalidParam { text: String },
+}
+
+impl From<std::io::Error> for NvmeError {
+    fn from(source: std::io::Error) -> NvmeError {
+        NvmeError::IoError {
+            source,
+        }
+    }
+}
--- a/libnvme-rs/src/lib.rs
+++ b/libnvme-rs/src/lib.rs
@@ -1,0 +1,40 @@
+#![allow(non_upper_case_globals)]
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
+
+#[allow(clippy::all)]
+mod bindings {
+    include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
+}
+pub use bindings::*;
+
+pub mod error;
+pub mod nvme_namespaces;
+mod nvme_tree;
+mod nvme_uri;
+
+use error::{IoError, NvmeError};
+use snafu::ResultExt;
+use std::{fs, path::Path, str::FromStr};
+
+pub use nvme_uri::NvmeTarget;
+
+/// Read and parse value from a sysfs file
+pub fn parse_value<T>(dir: &Path, file: &str) -> Result<T, NvmeError>
+where
+    T: FromStr,
+    T::Err: ToString,
+{
+    let path = dir.join(file);
+    let s = fs::read_to_string(&path).context(IoError {})?;
+    let s = s.trim();
+
+    match s.parse() {
+        Ok(v) => Ok(v),
+        Err(e) => Err(NvmeError::ValueParseError {
+            path: path.as_path().to_str().unwrap().to_string(),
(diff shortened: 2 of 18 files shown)


jonathan-teh:
bors try

jonathan-teh:
Looks like it doesn't like environments that have missing files in `/etc/nvme`.

jonathan-teh:
It appears that the current version of `nvme-cli` fails to connect if the `/etc/nvme/hostnqn` or `hostid` files are missing. A number of cargo, grpc and python tests do call `nvme connect` directly, which suggests that they will start failing when `nvme-cli` is upgraded to a newer version. The same applies to the `discover` command.

jonathan-teh:
This doesn't affect the latest 1.14 release of `nvme-cli`, which is what CI is running, and appears to have been introduced when the use of `libnvme` as submodule was introduced. Given that upstream has introduced a requirement on those files, CI needs to be fixed. I'll push a fix here to fix that segfault.

libnvme-rs/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+mod nvme_uri;
+
+use error::{IoError, NvmeError};
+use snafu::ResultExt;
+use std::{fs, path::Path, str::FromStr};
+
+pub use nvme_uri::NvmeTarget;
+
+/// Read and parse value from a sysfs file
+pub fn parse_value<T>(dir: &Path, file: &str) -> Result<T, NvmeError>
```
akhilerm, line 23: We already have a similar function in [sysfs](https://github.com/openebs/mayastor/blob/322aac84f19c0fcb12b422dd35bd5ff4ec8f44ac/sysfs/src/lib.rs#L13) crate. Can it be reused?

libnvme-rs/src/nvme_namespaces.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+impl NvmeDevice {
+    /// Construct a new NVMe device from a given path. The [struct.NvmeDevice]
+    /// will fill in all the details defined within the structure or return an
+    /// error if the value for the structure could not be found.
+    fn new(p: &Path) -> Result<Self, NvmeError> {
+        let name = p.file_name().unwrap().to_str().unwrap();
+        let devpath = format!("/sys/block/{}", name);
+        let subsyspath = format!("/sys/block/{}/device", name);
```
akhilerm, line 39: This actually gives the path of the controller right?

libnvme-rs/src/nvme_namespaces.rs
```diff
@@ -0,0 +1,92 @@
...
+            fw_rev: parse_value(subsys, "firmware_rev")?,
+            subsysnqn: parse_value(subsys, "subsysnqn")?,
+            model: parse_value(subsys, "model")?,
+            serial: parse_value(subsys, "serial")?,
+            size: parse_value(source, "size")?,
+            // /* NOTE: during my testing, it seems that NON fabric devices
+            //  * do not have a UUID, this means that local PCIe devices will
+            //  * be filtered out automatically. We should not depend on this
+            //  * feature or, bug until we gather more data
+            uuid: parse_value(source, "uuid")
```
akhilerm, line 54: In my system the local PCIe device had a uuid, not sure if we should depend on this for filtering.
```bash
akhil@MayaData:nvme0n1 $ ls
alignment_offset  capability  device             eui     events_async       ext_range  holders   integrity  mq     nsid       nvme0n1p2  nvme0n1p4  queue  removable  size    stat       trace   uuid
bdi               dev         discard_alignment  events  events_poll_msecs  hidden     inflight  lightnvm   nguid  nvme0n1p1  nvme0n1p3  power      range  ro         slaves  subsystem  uevent  wwid
akhil@MayaData:nvme0n1 $ cat uuid
01000000-0000-0000-8ce3-8e0400a2c71b
```

libnvme-rs/src/nvme_namespaces.rs
```diff
@@ -0,0 +1,92 @@
...
+        let name = p.file_name().unwrap().to_str().unwrap();
+        let devpath = format!("/sys/block/{}", name);
+        let subsyspath = format!("/sys/block/{}/device", name);
+        let source = Path::new(devpath.as_str());
+        let subsys = Path::new(subsyspath.as_str());
+
+        Ok(NvmeDevice {
+            path: p.display().to_string(),
+            fw_rev: parse_value(subsys, "firmware_rev")?,
+            subsysnqn: parse_value(subsys, "subsysnqn")?,
```
akhilerm, line 46: is this available only for fabric devices?

libnvme-rs/src/nvme_namespaces.rs
```diff
@@ -0,0 +1,92 @@
...
+            wwid: parse_value(source, "wwid")?,
+            nsid: parse_value(source, "nsid")?,
+        })
+    }
+}
+/// The DeviceList of all NVMe devices found that provide all properties as
+/// defined in the [struct.NvmeDevice]
+#[derive(Debug, Default)]
+pub struct NvmeDeviceList {
+    devices: Vec<String>,
```
akhilerm, line 65: Cant this be a `Vec<NvmeDevice>` ?

libnvme-rs/src/nvme_tree.rs
```diff
@@ -0,0 +1,130 @@
+pub struct NvmeRoot {
```
akhilerm, line 1: comments/references on this section will be good

libnvme-rs/src/nvme_uri.rs
```diff
@@ -0,0 +1,231 @@
+use std::{convert::TryFrom, time::Duration};
```
akhilerm, line 1: In this file also, please add comments / references.

libnvme-rs/src/error.rs
```diff
@@ -0,0 +1,65 @@
+use snafu::Snafu;
```
akhilerm, line 1: Is this same as the [errors](https://github.com/openebs/mayastor/blob/develop/nvmeadm/src/error.rs) in nvmeadm

akhilerm (commented):
Have provided a few questions and comments.

jonathan-teh:
bors try

jonathan-teh:
bors try

libnvme-rs/src/nvme_namespaces.rs
jonathan-teh: This field is mandatory for any device that supports NVMe 1.2.1 or later.

libnvme-rs/src/error.rs
jonathan-teh: Yes, with the addition of `LookupHostError` and `AddCtrlrError`.

libnvme-rs/src/nvme_namespaces.rs
jonathan-teh: Yes.

libnvme-rs/src/nvme_namespaces.rs
jonathan-teh: On the CI worker, the `uuid` field is indeed missing for the PCIe device. As the comment indicates, it's not a reliable indicator but it's not something that needs to be addressed here.

libnvme-rs/src/lib.rs
jonathan-teh: This function comes from `nvmeadm` and is in fact newer, with revised error results.

libnvme-rs/src/nvme_namespaces.rs
jonathan-teh: > Have provided a few questions and comments.

Thanks for your comments. Most of them pertain to code that is actually from `nvmeadm`. Perhaps I should refactor this and simply use `nvme_namespace` and `error` from `nvmeadm` instead of copying the code.

jonathan-teh:
> Have provided a few questions and comments.

Thanks for your comments. Most of them pertain to code that is actually from `nvmeadm`. Perhaps I should refactor this and simply use `nvme_namespace` and `error` from `nvmeadm` instead of copying the code.

libnvme-rs/src/nvme_namespaces.rs
jonathan-teh: This was left as-is to avoid having to modify the callers.

jonathan-teh:
Closing as @gila points out that the old interface should not be re-used.

PR #1049, opened 2021-11-12
feat(libnvme-rs): new crate, implement connect and disconnect

jonathan-teh:
New crate that uses the libnvme development C library. Reuse NvmeTarget, and some NvmeErrors, from nvmeadm and implement connect and disconnect to an NVMe-oF target using libnvme.

Fixes CAS-1218

### DIFF
--- a/libnvme-rs/src/nvme_tree.rs
+++ b/libnvme-rs/src/nvme_tree.rs
@@ -1,0 +1,149 @@
+/// RAII Wrapper for nvme_root_t
+pub(crate) struct NvmeRoot {
+    root: *mut crate::bindings::nvme_root,
+}
+
+impl NvmeRoot {
+    pub(crate) fn new(root: *mut crate::bindings::nvme_root) -> Self {
+        NvmeRoot {
+            root,
+        }
+    }
+    pub(crate) fn as_mut_ptr(&self) -> *mut crate::bindings::nvme_root {
+        self.root
+    }
+}
+
+impl Drop for NvmeRoot {
+    fn drop(&mut self) {
+        unsafe { crate::nvme_free_tree(self.root) }
+    }
+}
+
+/// Iterator for nvme_host_t
+pub(crate) struct NvmeHostIterator<'a> {
+    root: &'a NvmeRoot,
+    host: *mut crate::bindings::nvme_host,
+}
+
+impl<'a> NvmeHostIterator<'a> {
+    pub(crate) fn new(root: &'a NvmeRoot) -> Self {
+        Self {
+            root,
+            host: std::ptr::null_mut(),
+        }
+    }
+}
+
+impl<'a> Iterator for NvmeHostIterator<'a> {
+    type Item = *mut crate::bindings::nvme_host;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.host = if self.host.is_null() {
+            unsafe { crate::nvme_first_host(self.root.as_mut_ptr()) }
+        } else {
+            unsafe { crate::nvme_next_host(self.root.as_mut_ptr(), self.host) }
+        };
+        if self.host.is_null() {
+            None
+        } else {
+            Some(self.host)
+        }
+    }
+}
+
+/// Iterator for nvme_subsystem_t
+pub(crate) struct NvmeSubsystemIterator {
+    host: *mut crate::bindings::nvme_host,
+    subsys: *mut crate::bindings::nvme_subsystem,
+}
+
+impl NvmeSubsystemIterator {
+    pub(crate) fn new(host: *mut crate::bindings::nvme_host) -> Self {
+        Self {
+            host,
+            subsys: std::ptr::null_mut(),
+        }
+    }
+}
+
+impl Iterator for NvmeSubsystemIterator {
+    type Item = *mut crate::bindings::nvme_subsystem;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.subsys = if self.subsys.is_null() {
+            unsafe { crate::nvme_first_subsystem(self.host) }
+        } else {
+            unsafe { crate::nvme_next_subsystem(self.host, self.subsys) }
+        };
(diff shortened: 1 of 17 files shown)


jonathan-teh:
bors try

jonathan-teh:
Compared to #1047 this uses `libnvme` to find the NVMe namespace that it connected to and uses that to return the path of the block device. There is now an `NvmeTransportType` enum so there is no need to check for a valid `trtype` in `NvmeTarget::connect()`.

jonathan-teh:
bors try

jonathan-teh:
bors try

jonathan-teh:
bors try

paulyoong (approved):

libnvme-rs/src/nvme_uri.rs
```diff
@@ -0,0 +1,277 @@
...
+        let ret = unsafe {
+            crate::nvmf_add_ctrl(h, nvme_ctrl, &cfg, cfg.disable_sqflow)
+        };
+        if ret != 0 {
+            return Err(NvmeError::AddCtrlrError {
+                rc: ret,
+            });
+        }
+
+        let mut retries = 10;
```
gila, line 185: What is the reason for the loop? We have the controller connect here, and now we try to find the namespaces? A controller can have zero namespaces.

libnvme-rs/src/nvme_uri.rs
```diff
@@ -0,0 +1,277 @@
...
+            nr_poll_queues: 0,
+            tos: -1,
+            duplicate_connect: false,
+            disable_sqflow: false,
+            hdr_digest: false,
+            data_digest: false,
+        };
+        let ret = unsafe {
+            crate::nvmf_add_ctrl(h, nvme_ctrl, &cfg, cfg.disable_sqflow)
+        };
```
gila, line 178: What is the contract here? ret == 0 means we have, connected to the controller -- and its namespaces? or just the controller?

libnvme-rs/src/nvme_uri.rs
```diff
@@ -0,0 +1,277 @@
...
+                    std::ffi::CStr::from_ptr(crate::nvme_subsystem_get_nqn(
+                        subsys,
+                    ))
+                };
+                if cstr.to_str().unwrap() != self.subsysnqn {
+                    continue;
+                }
+                let ctrlriter = NvmeCtrlrIterator::new(subsys);
+                for c in ctrlriter {
+                    if unsafe { crate::nvme_disconnect_ctrl(c) == 0 } {
```
gila, line 248: What does != 0 mean here? cant be found? or, can be found but can't disconnect? or.... ? if the disconnect call would yield EAGAIN or EBUSY, then yeah I can imagine we MAY want to retry but perhaps it's better to handle that further up, and let the consumer decide -- not the library.

libnvme-rs/src/nvme_tree.rs
```diff
@@ -0,0 +1,149 @@
...
+    root: *mut crate::bindings::nvme_root,
+}
+
+impl NvmeRoot {
+    pub fn new(root: *mut crate::bindings::nvme_root) -> Self {
+        NvmeRoot {
+            root,
+        }
+    }
+    pub fn as_mut_ptr(&self) -> *mut crate::bindings::nvme_root {
```
gila, line 12: The goal of the lib should be to NOT have to deal with these raw dangling pointers, why is this method here and public?

libnvme-rs/src/nvme_uri.rs
jonathan-teh: That was carried over from `nvmeadm`, I guess it's to allow for slow connections to the target. The assumption here is that the target will have at least one namespace.

libnvme-rs/src/nvme_uri.rs
jonathan-teh: This means it has successfully written to `/dev/nvme-fabrics` and read a response back, which AIUI just means it has connected to the controller.

libnvme-rs/src/nvme_uri.rs
jonathan-teh: This basically writes `1` to `/sys/class/nvme/{}/delete_controller` so `!=0` would an IO error of some description. I'll try to pass the error back up.

libnvme-rs/src/nvme_tree.rs
jonathan-teh: That's an oversight, it's meant to be `pub(crate)`, will fix.

libnvme-rs/src/nvme_uri.rs
gila: I just looked into the function and it does a whole lot more than that. Perhaps we are looking at different versions but master seems to add the controller (or returns <0) and then goes on and configure it and adds the controller to the subsystem. From there,  it seems to me we can use nvme_ctrL-first_ns() would be a lot better than this manual looping.

jonathan-teh: `NvmeNamespaceIterator::next()` essentially calls `nvme_subsystem_first_ns()` or `nvme_subsystem_next_ns()`.

gila: Then I dont understand why the loop within intervals of 1 second, creates a new root and iterates over it. Once the controller is connected it is part of the root tree, and so we can just list the namespaces once, right?

jonathan-teh: AIUI the kernel connects to the namespace asynchronously after connecting to the controller. We have a similar race condition in one of the unit tests which uses nvme-cli where a `nvme list` immediately after `nvme connect` returns no NVMe devices.

libnvme-rs/src/nvme_uri.rs
jonathan-teh: I've passed the error back up as just the raw return code as there's no easy way to tell where it failed.

PR #1051, opened 2021-11-25
fix(nvmx): delay controller destruction until channels have been destroyed

jonathan-teh:
Implement Tiago's fix by essentially holding a reference to the nvmx controller in channel, giving it up on channel shutdown, and waiting in controller::device_destroy() until it holds the only reference before dropping the controller.

Fixes CAS-1177

### DIFF
--- a/mayastor/src/bdev/nvmx/channel.rs
+++ b/mayastor/src/bdev/nvmx/channel.rs
@@ -264,6 +264,10 @@ pub struct NvmeIoChannelInner<'a> {
     poller: poller::Poller<'a>,
     io_stats_controller: IoStatsController,
     pub device: Box<dyn BlockDevice>,
+    /// to prevent the controller from being destroyed before the channel
+    ctrl: Option<
+        std::sync::Arc<parking_lot::Mutex<crate::bdev::NvmeController<'a>>>,
+    >,
     num_pending_ios: u64,
 
     // Flag to indicate the shutdown state of the channel.
@@ -309,6 +313,7 @@ impl NvmeIoChannelInner<'_> {
         let rc = self.reset();
         if rc == 0 {
             self.is_shutdown = true;
+            self.ctrl.take();
         }
         rc
     }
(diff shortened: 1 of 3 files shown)


jonathan-teh:
bors try

mayastor/src/bdev/nvmx/channel.rs
```diff
@@ -264,6 +264,10 @@ pub struct NvmeIoChannelInner<'a> {
     poller: poller::Poller<'a>,
     io_stats_controller: IoStatsController,
     pub device: Box<dyn BlockDevice>,
+    /// to prevent the controller from being destroyed before the channel
+    pub ctrl: Option<
```
tiagolobocastro, line 268: guess now it doesn't need the pub

mayastor/src/bdev/nvmx/controller.rs
```diff
@@ -927,13 +929,34 @@ pub(crate) async fn destroy_device(name: String) -> Result<(), NexusBdevError> {
...
+    let mut carc = carc;
+    loop {
+        match Arc::try_unwrap(carc) {
+            Ok(i) => {
+                drop(i);
+                break;
+            }
+            Err(ret) => {
+                warn!(?name, "delaying controller destroy");
+                let rx = mayastor_sleep(std::time::Duration::from_secs(1));
```
tiagolobocastro, line 952: From the logs the channels are usually ready very quickly when we hit this, maybe start with a smaller number? 100ms or something like that?

tiagolobocastro (approved):

mayastor/src/bdev/nvmx/controller.rs
jonathan-teh: On a 100 iteration test, 100ms does seem to be more than sufficient as there were 11 instances of controller destruction being delayed and all on separate iterations. However, this is on an otherwise idle system so I've settled on 250ms as hopefully a long enough delay, without being too long.

jonathan-teh:
bors try

paulyoong (commented):
FYI: I did test these changes with my test that was previously failing and it seems to have fixed the issue for me.

paulyoong (approved):

jonathan-teh:
bors merge

PR #1055, opened 2021-12-06
chore(kato timeout): increase from 1 to 15 seconds

paulyoong:
Increase the keepalive timeout to make it more resilient to temporary
failures such as network failures and will prevent the nexus child from
being prematurely faulted.


paulyoong:
bors try

tiagolobocastro:
bors try

paulyoong:
Closing because the KATO increase doesn't seem to be having the desired effect.

PR #1052, opened 2021-11-26
chore(spdk): using spdk-rs crate

dsavitskiy:

PR #1057, opened 2021-12-07
chore: merge release/1.0 into develop

jonathan-teh:


jonathan-teh:
bors try

tiagolobocastro (approved):

.pre-commit-config.yaml
```diff
@@ -12,14 +12,14 @@ repos:
     -   id: rust-style
         name: Rust style
         description: Run cargo fmt on files included in the commit. rustfmt should be installed before-hand.
-        entry: cargo +nightly-2021-11-30 fmt --all -- --check
+        entry: cargo fmt --all -- --check
```
akhilerm, line 15: This change was introduced in the [PR](https://github.com/openebs/mayastor/pull/1052) to make use of `spdk-rs` crate. Wont switching it back cause issues in those set of commits in develop?

chart/templates/csi-daemonset.yaml
```diff
@@ -43,7 +43,9 @@ spec:
             fieldRef:
               fieldPath: status.podIP
         - name: RUST_BACKTRACE
-          value: "1"
+          value: "1"{{ if .Values.moac }}
+        - name: MOAC
+          value: "true"{{ end }}
```
akhilerm, line 48:
```suggestion
          value: "1"
{{ if .Values.moac }}
        - name: MOAC
          value: "true"
{{ end }}
```

`if` in helm can be written in the following format to make it more readable.

deploy/csi-daemonset.yaml
```diff
@@ -30,7 +30,7 @@ spec:
       # the same.
       containers:
       - name: mayastor-csi
-        image: mayadata/mayastor-csi:develop
+        image: mayadata/mayastor:develop
```
akhilerm, line 33: Do we run mayastor-csi also from the mayastor-binary?

akhilerm (commented):
A few questions on the charts and usage of rust nightly

.pre-commit-config.yaml
jonathan-teh: No, this was previously reverted in cfe04ce1, it mainly affects those (like me) who don't use the `rustup` toolchain. The `+xyz` bit actually applies to `rustup` and so doesn't work if you don't have it installed. Without that, it simply runs the current default toolchain.

chart/templates/csi-daemonset.yaml
jonathan-teh: This is a merge request, I only went as far as making it build and pass pre-commit and CI tests. Feel free to raise another PR to fix this.

deploy/csi-daemonset.yaml
jonathan-teh: It would appear so. This and the previous comment refer to code originally part of #1045.

mtzaurus (approved):

jonathan-teh:
bors merge

PR #1058, opened 2021-12-10
fix(mayastor): prevent auto-unpin of nexus

dsavitskiy:
This PR is a follow-up for #1052. It fixes a couple of things that I forgot to do in the original commit.

Mark BdevContainer (spdk-rs) and Nexus structures as "non-unpinnable". This leads to several fixes in Nexus use (replacing &mut self with Pin<&mut Self>). If a struct is not explicitly marked with marker::PhantomPinned, Rust assumes it is unpinnable and one can easily get a &mut to it. We don't want that for Bdev contexts, Nexus struct instances in particular.
Some trivial clean-ups of things like excessive "pub" use, etc.
Please also review the spdk-rs submodule part of this commit: mayadata-io/spdk-rs#2.

### DIFF
--- a/mayastor/src/bdev/nexus/mod.rs
+++ b/mayastor/src/bdev/nexus/mod.rs
@@ -53,7 +53,7 @@ pub use nexus_child::{
     NexusChild,
     Reason,
 };
-pub(crate) use nexus_io::{nexus_submit_io, NexusBio, NioCtx};
+pub(crate) use nexus_io::{nexus_submit_request, NioCtx};
 pub use nexus_iter::{
     nexus_iter,
     nexus_iter_mut,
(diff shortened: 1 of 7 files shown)


dsavitskiy:
bors try

dsavitskiy:
bors try

dsavitskiy:
This PR is a follow-up for https://github.com/openebs/mayastor/pull/1052. It fixes a couple of things that I forgot to do in the original commit.
- Mark BdevContainer (spdk-rs) and Nexus structures as "non-unpinnable". This leads to several fixes in Nexus use (replacing &mut self with Pin<&mut Self>). If a struct is not explicitly marked with marker::PhantomPinned, Rust assumes it is unpinnable and one can easily get a &mut to it. We don't want that for Bdev contexts, Nexus struct instances in particular.
- Some trivial clean-ups of things like excessive "pub" use, etc. 

Please also review the spdk-rs submodule part of this commit: https://github.com/mayadata-io/spdk-rs/pull/2.

jonathan-teh (approved):

mtzaurus (approved):

PR #1066, opened 2022-01-14
feat(grpc): adding v1 host service

akhilerm:
Signed-off-by: Akhil Mohan <akhil.mohan@datacore.com>

### DIFF
--- a/mayastor/src/grpc/controller_grpc.rs
+++ b/mayastor/src/grpc/controller_grpc.rs
@@ -1,132 +1,72 @@
 use crate::{
-    bdev::{nexus, NvmeController, NvmeControllerState, NVME_CONTROLLERS},
+    bdev::{NvmeController, NvmeControllerState, NVME_CONTROLLERS},
     core::{BlockDeviceIoStats, CoreError},
     ffihelper::{cb_arg, done_cb},
-    grpc::{rpc_submit, GrpcResult},
 };
-
-use ::rpc::mayastor as rpc;
 use futures::channel::oneshot;
-use std::convert::From;
-use tonic::{Response, Status};
+
+#[derive(Debug)]
+pub struct NvmeControllerInfo {
+    pub name: String,
+    pub state: NvmeControllerState,
+    pub size: u64,
+    pub blk_size: u32,
+}
 
 impl<'a> NvmeController<'a> {
-    fn to_grpc(&self) -> rpc::NvmeController {
+    fn to_info(&self) -> NvmeControllerInfo {
         let (size, blk_size) = self
             .namespace()
             .map_or((0, 0), |ns| (ns.size_in_bytes(), ns.block_len() as u32));
 
-        rpc::NvmeController {
+        NvmeControllerInfo {
             name: self.name.to_string(),
-            state: rpc::NvmeControllerState::from(self.get_state()) as i32,
+            state: self.get_state(),
             size,
             blk_size,
         }
     }
 }
 
-impl From<NvmeControllerState> for rpc::NvmeControllerState {
-    fn from(state: NvmeControllerState) -> Self {
-        match state {
-            NvmeControllerState::New => rpc::NvmeControllerState::New,
-            NvmeControllerState::Initializing => {
-                rpc::NvmeControllerState::Initializing
-            }
-            NvmeControllerState::Running => rpc::NvmeControllerState::Running,
-            NvmeControllerState::Faulted(_) => {
-                rpc::NvmeControllerState::Faulted
-            }
-            NvmeControllerState::Unconfiguring => {
-                rpc::NvmeControllerState::Unconfiguring
-            }
-            NvmeControllerState::Unconfigured => {
-                rpc::NvmeControllerState::Unconfigured
+/// Returns the IO stats of the given NVMe Controller
+pub async fn controller_stats(
+    controller_name: &str,
+) -> Result<BlockDeviceIoStats, CoreError> {
+    if let Some(ctrlr) = NVME_CONTROLLERS.lookup_by_name(controller_name) {
+        let (s, r) =
+            oneshot::channel::<Result<BlockDeviceIoStats, CoreError>>();
+        {
+            let ctrlr = ctrlr.lock();
+            if let Err(e) = ctrlr.get_io_stats(
+                |stats, ch| {
+                    done_cb(ch, stats);
+                },
+                cb_arg(s),
+            ) {
+                error!(
+                    "{}: failed to get stats for NVMe controller: {:?}",
+                    controller_name, e
+                );
+                return Err(e);
             }
         }
+        r.await.expect("Failed awaiting at io_stats")
+    } else {
+        Err(CoreError::BdevNotFound {
+            name: controller_name.to_string(),
+        })
     }
 }
 
-impl From<BlockDeviceIoStats> for rpc::NvmeControllerIoStats {
-    fn from(b: BlockDeviceIoStats) -> Self {
-        Self {
-            num_read_ops: b.num_read_ops,
-            num_write_ops: b.num_write_ops,
-            bytes_read: b.bytes_read,
-            bytes_written: b.bytes_written,
-            num_unmap_ops: b.num_unmap_ops,
-            bytes_unmapped: b.bytes_unmapped,
-        }
-    }
-}
-
-pub async fn controller_stats() -> GrpcResult<rpc::StatNvmeControllersReply> {
-    let rx = rpc_submit::<_, _, nexus::Error>(async move {
-        let mut res: Vec<rpc::NvmeControllerStats> = Vec::new();
-        let controllers = NVME_CONTROLLERS.controllers();
-        for name in controllers.iter() {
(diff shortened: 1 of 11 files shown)


shubham14bajpai (approved):
LGTM

akhilerm:
@mtzaurus @tiagolobocastro Can you review this PR please?

mayastor/src/grpc/controller_grpc.rs
```diff
@@ -1,132 +1,70 @@
...
-            NvmeControllerState::Running => rpc::NvmeControllerState::Running,
-            NvmeControllerState::Faulted(_) => {
-                rpc::NvmeControllerState::Faulted
-            }
-            NvmeControllerState::Unconfiguring => {
-                rpc::NvmeControllerState::Unconfiguring
-            }
-            NvmeControllerState::Unconfigured => {
-                rpc::NvmeControllerState::Unconfigured
+pub async fn controller_stats(
```
tiagolobocastro, line 31: please doc comment pub fn's

mayastor/src/grpc/controller_grpc.rs
```diff
@@ -1,132 +1,70 @@
...
-            .filter_map(|n| {
-                NVME_CONTROLLERS
-                    .lookup_by_name(n)
-                    .map(|c| c.lock().to_grpc())
-            })
-            .collect::<Vec<_>>();
-
-        Ok(rpc::ListNvmeControllersReply {
-            controllers,
+pub async fn list_controllers() -> Vec<NvmeControllerInfo> {
```
tiagolobocastro, line 60: please doc comment pub fn's

mayastor/src/grpc/controller_grpc.rs
```diff
@@ -1,132 +1,70 @@
...
+                .lookup_by_name(n)
+                .map(|c| c.lock().to_info())
         })
-    })?;
-
-    rx.await
-        .map_err(|_| Status::cancelled("cancelled"))?
-        .map_err(Status::from)
-        .map(Response::new)
+        .collect::<Vec<_>>()
```
tiagolobocastro, line 69: This is very confusing to read.. maybe `NVME_CONTROLLERS` should implement Iter or something along those lines... but the code was already like that so it's fine.

mayastor/src/grpc/mayastor_grpc.rs
```diff
@@ -242,6 +246,110 @@ impl From<MayastorFeatures> for rpc::mayastor::MayastorFeatures {
...
+    }
+}
+
+impl From<blk_device::FileSystem> for block_device::Filesystem {
+    fn from(fs: blk_device::FileSystem) -> Self {
+        let mountpoint = if fs.mountpoints.is_empty() {
+            String::from("")
+        } else {
+            fs.mountpoints.get(0).unwrap().to_string()
+        };
```
tiagolobocastro, line 273: might be able to tidy it up a bit but having something like this instead:
```suggestion
        let mountpoint = fs.mountpoints.get(0).unwrap_or_default();
```

mayastor/src/grpc/v1/mayastor.rs
```diff
@@ -0,0 +1,306 @@
...
+};
+use futures::FutureExt;
+use git_version::git_version;
+use rpc::mayastor::v1::mayastor as rpc;
+use std::panic::AssertUnwindSafe;
+use tonic::{Request, Response, Status};
+
+use ::function_name::named;
+
+/// RPC service for mayastor pool operations
```
tiagolobocastro, line 25: copy/paste error?

mayastor/src/grpc/v1/mayastor.rs
```diff
@@ -0,0 +1,306 @@
...
+#[derive(Debug)]
+#[allow(dead_code)]
```
tiagolobocastro, line 27: do we need this?

mayastor/src/grpc/v1/mayastor.rs
akhilerm: my bad. fill will it.

mayastor/src/grpc/v1/mayastor.rs
akhilerm: will remove the dead code attribute.

mayastor/src/grpc/v1/mayastor.rs
akhilerm: my bad, will fix it.

mayastor/src/grpc/v1/mayastor.rs
akhilerm: @tiagolobocastro We have the name field of the service declared, which is currently not used anywhere. Thats why dead code attribute is enabled.

akhilerm: Done.

mayastor/src/grpc/controller_grpc.rs
```diff
@@ -1,132 +1,72 @@
...
-            }
-            NvmeControllerState::Unconfiguring => {
-                rpc::NvmeControllerState::Unconfiguring
-            }
-            NvmeControllerState::Unconfigured => {
-                rpc::NvmeControllerState::Unconfigured
+/// Returns the IO stats of the given NVMe Controller
+pub async fn controller_stats(
+    controller_name: &str,
+) -> Result<BlockDeviceIoStats, CoreError> {
```
mtzaurus, line 34: I might have overlooked it, but could you please remind why we return here 'BlockDeviceIoStats' instead of returning 'NvmeControllerIoStats' directly ? All callers of 'controller_stats()' aim to end up with 'NvmeControllerIoStats', nobody uses raw 'BlockDeviceIoStats'.

```
let stats = controller_stats(&ctrl.name).await;
                        if stats.is_ok() {
                            res.push(NvmeControllerStats {
                                name: ctrl.name,
                                stats: stats
                                    .ok()
                                    .map(NvmeControllerIoStats::from),
                            });
                        }
```

Could we make this function return 'NvmeControllerIoStats' directly, so that we can eliminate unnecessary '.map()' (unless I overlooked some other use-cases of this function that require 'BlockDeviceIoStats' ?
Probably this will also not needed then:

```
 'impl From<BlockDeviceIoStats> for rpc::NvmeControllerIoStats {'
```

mayastor/src/host/blk_device.rs
```diff
@@ -28,17 +28,43 @@ use std::{
...
-    BlockDevice,
-};
 use udev::{Device, Enumerator};
 
 // Struct representing a property value in a udev::Device struct (and possibly
 // elsewhere). It is used to provide conversions via various "From" trait
 // implementations below.
 struct Property<'a>(Option<&'a OsStr>);
 
+pub struct Partition {
```
mtzaurus, line 38: Could you please clarify why we have local copies of the structures defined via '.proto' files ?
It's a bit of unusual to have a copy of auto-generated GRPC structure.

akhilerm: The local copies are maintained because once the proto files are versioned, we cannot use the proto definitions in here.  So we use the local copies and covert to the proto definition struct when we want. Also we wanted to have the proto structs to be confined only to the code sections / modules which use the grpc.

eg: if we use rpc::Partition here, then for the v1 version, we will have a different struct which we have to convert to.  So this struct acts like kind of a middleware which you can convert to structs in different proto versions (v0::Partition, v1::Partition etc).

mayastor/src/grpc/controller_grpc.rs
akhilerm: The change was done so that we dont have to use the rpc::NvmeControllerIoStats directly. (since we are creating a v1 version.)  

What we can do is create an `NvmeControllerIoStats` similar to the `NvmeControllerInfo` defined in this same file and then convert it to the rpc::NvmeControllerIoStats. This will avoid passing the `BlockDeviceIoStats` as is. This struct will have the exact same fields of  the `BlockDeviceIoStats`, thats why used it directly.

mayastor/src/host/blk_device.rs
mtzaurus: Thanks for the explanation.
Could you please add comments to these local structures about that ?

mayastor/src/grpc/controller_grpc.rs
mtzaurus: Thanks for the explanation. I think we're good now.

mtzaurus (approved):

tiagolobocastro (approved):

mayastor/src/host/blk_device.rs
akhilerm: Done.

akhilerm:
Can I get a `bors d+` on this PR?
 cc: @tiagolobocastro

tiagolobocastro:
bors d+

akhilerm:
bors try

akhilerm:
bors try

akhilerm:
bors try

akhilerm:
bors try

akhilerm:
bors merge

PR #1084, opened 2022-01-31
fix(nexus): do not persist faulted state of the last replica

mtzaurus:
In case the last replica fails, it is not marked as not healthy
in the ETCD, which allows Caontrol Plane to recover the nexus later
without ambiguity (as the replica with the most recent user data
remains known).

Resolves: CAS-1284

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_bdev.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev.rs
@@ -967,8 +967,43 @@ impl Nexus {
             // schedule the deletion of the child eventhough etcd has not been
             // updated yet we do not need to wait for that to
             // complete anyway.
-            MWQ.enqueue(Command::RemoveDevice(self.name.clone(), name));
-            self.persist(PersistOp::Update((uri.clone(), child.state())))
+            MWQ.enqueue(Command::RemoveDevice(self.name.clone(), name.clone()));
+
+            // Do not persist child state in case it's the last healthy child of
+            // the nexus: let Control Plane reconstruct the nexus
+            // using this device as the replica with the most recent
+            // user data.
+            self.persist(PersistOp::UpdateCond(
+                (uri.clone(), child.state(), &|nexus_info| {
+                    // Determine the amount of healthy replicas in the persistent state and
+                    // check against the last healthy replica remaining.
+                    let num_healthy = nexus_info.children.iter().fold(0, |n, c| {
+                        if c.healthy {
+                            n + 1
+                        } else {
+                            n
+                        }
+                    });
+
+                    match num_healthy {
+                        0 => {
+                            warn!(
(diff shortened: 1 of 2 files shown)


paulyoong (commented):
Could we add test cases to validate the behaviour?

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -967,8 +967,43 @@ impl Nexus {
...
+
+            // Do not persist child state in case it's the last healthy child of
+            // the nexus: let Control Plane reconstruct the nexus
+            // using this device as the replica with the most recent
+            // user data.
+            self.persist(PersistOp::UpdateCond(
+                (uri.clone(), child.state(), &|nexus_info| {
+                    // Determine the amount of healthy replicas in the persistent state and
+                    // check against the last healthy replica remaining.
+                    let num_healthy = nexus_info.children.iter().fold(0, |n, c| {
```
tiagolobocastro, line 980: nit: might be able to simplify with:
`let num_healthy = nexus_info.children.iter().filter(|c| c.healthy).count();`

mayastor/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -87,6 +92,22 @@ impl Nexus {
...
+                }
+
+                let uuid =
+                    NexusChild::uuid(&uri).expect("Failed to get child UUID.");
+
+                nexus_info.children.iter_mut().for_each(|c| {
+                    if c.uuid == uuid {
+                        c.healthy = Self::child_healthy(&state);
+                    }
+                });
```
tiagolobocastro, line 109: rather than duplicating, would it be possible to make use of the same code above (used in the Update)? Though we can't simply call self.persist again because we've taken the lock.

tiagolobocastro (approved):
Would be good to add a testcase, should be fairly straightforward with the composer.

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -967,8 +967,43 @@ impl Nexus {
...
+                            n + 1
+                        } else {
+                            n
+                        }
+                    });
+
+                    match num_healthy {
+                        0 => {
+                            warn!(
+                                "nexus {}: no healthy replicas persent in persistent store when retiring replica {}:
```
jonathan-teh, line 991: typo: present

jonathan-teh (approved):
Looks fine, needs a test case though.

mayastor/src/bdev/nexus/nexus_persistence.rs
mtzaurus: Yes, we can't use reentrance here as we're holding the lock.

mtzaurus:
bore merge

tiagolobocastro:
bors merge

PR #1105, opened 2022-02-18
chore: don't add default max qpairs per controller

tiagolobocastro:
Limiting the qpairs to a low value means we cannot then allocate more for the rebuilds.


paulyoong (approved):

blaisedias (approved):

tiagolobocastro:
bors merge

mtzaurus (approved):

PR #1062, opened 2021-12-30
Update nats images

xin3liang:
This PR is related to Arm64 support https://github.com/openebs/mayastor/issues/1063

PR #1106, opened 2022-02-19
Fix Helm chart.

reitermarkus:
Previously, the chart could not be installed since the default `values.yaml` was missing the necessary fields which were only present in `release/values.yaml`. This moves all default values into `values.yaml` so the chart can be installed without any changes.

This also updates NATS to the latest current version and simplifies YAML configuration for it. This also adds ARM support for NATS, since the old version did not have an ARM image. 

Closes https://github.com/openebs/mayastor/pull/1062, which also updates NATS.


chart/release/values.yaml
```diff
@@ -1,17 +1,3 @@
...
-etcd:
-  auth:
-    rbac:
-      enabled: false
-  replicaCount: 3
-
-mayastorLogLevel: info
+mayastorPools:
+  - node: "NODE_NAME"
+    device: "DEVICE"
```
tiagolobocastro, line 3: Do you think it's useful to have this @reitermarkus? I guess it's good to see an example, but if it will error since these are dummies.

chart/develop/values.yaml
```diff
@@ -1,16 +1,12 @@
 nats:
-  cluster:
-    enabled: false
+  replicas: 1
```
tiagolobocastro, line 2: Thanks, this has been a source of confusion for some people, setting replicas>1 but with cluster false...

tiagolobocastro:
@reitermarkus thanks this looks, we'll have to hold on merging for a bit as we're doing a bit of internal maintenance of the ci, so we're "freezing" things for a bit - hopefully not for too long. I'll keep you posted if things go awry.
PS - Could you please change your commits slightly, you can have a look [here](https://github.com/openebs/mayastor/runs/5351827155?check_suite_focus=true).

wibed:
@tiagolobocastro [tiagolobocastro](https://github.com/tiagolobocastro)

any update on this?

tiagolobocastro:
@wibed you can find the new chart for develop here: https://github.com/openebs/mayastor-extensions, it has not been released yet though.

PR #1110, opened 2022-02-23
feat(nexus info key): use supplied key if present

paulyoong:
Provide the ability to receive a key under which the NexusInfo structure
should be persisted. If provided by the control plane, this key will be
used otherwise it will default to the nexus UUID to ensure backwards
compatibility.

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_persistence.rs
+++ b/mayastor/src/bdev/nexus/nexus_persistence.rs
@@ -5,6 +5,30 @@ use std::time::Duration;
 
 type ChildUri = String;
 
+/// Information associated with the persisted NexusInfo structure.
+pub struct PersistentNexusInfo {
+    // Structure that is written to the persistent store.
+    inner: NexusInfo,
+    // Key to use to persist the NexusInfo structure.
+    // If `Some` the key has been supplied by the control plane.
+    key: Option<String>,
+}
+
+impl PersistentNexusInfo {
+    /// Create a new instance of PersistentNexusInfo.
+    pub(crate) fn new(key: Option<String>) -> Self {
+        Self {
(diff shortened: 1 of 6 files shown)


paulyoong:
bors try

mayastor/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -112,7 +139,7 @@ impl<'n> Nexus<'n> {
                 nexus_info.clean_shutdown = true;
             }
         }
-        self.save(&nexus_info).await;
+        self.save(nexus_info, &key).await;
```
tiagolobocastro, line 142: could we pass in persistent_nexus_info instead, which has both nexus_info and key?

mayastor/src/grpc/v1/nexus.rs
```diff
@@ -251,6 +251,7 @@ impl NexusRpc for NexusService {
                             },
                         },
                         &args.children,
+                        None,
```
tiagolobocastro, line 254: Shouldn't this be same as the "v0"?

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

mayastor/src/bdev/nexus/nexus_persistence.rs
paulyoong: Good idea. Done.

mayastor/src/grpc/v1/nexus.rs
paulyoong: Good spot, thanks. Done.

paulyoong:
bors merge

mtzaurus (approved):

PR #1113, opened 2022-02-25
Script to deploy mayastor to an AKS cluster

mukhoakash:


scripts/deploy/actions/hugepage-enabler-daemonset.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: apps/v1
```
akhilerm, line 1: Shouldn't this be part of azure node setup, instead of being in mayastor deployment scripts?

scripts/deploy/actions/kured-config.yaml
```diff
@@ -0,0 +1,141 @@
...
+            # Pass in the name of the node on which this pod is scheduled
+            # for use with drain/uncordon operations and lock acquisition
+            - name: KURED_NODE_ID
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+          command:
+            - /usr/bin/kured
+            - --period=0m30s
+#            - --force-reboot=false
```
akhilerm, line 111: Please remove the commented out lines if we really are not using them. 

Also another question, is this used for the sole purpose of node reboot after updating huge pages. If thats the case, cant we have it as part of node setup instead of mayastor specific scripts?

scripts/deploy/cluster-up.sh
```diff
@@ -0,0 +1,300 @@
...
+#
+echo "Starting Mayastor specific operations..."
+kubectl create namespace mayastor
+kubectl apply -f https://raw.githubusercontent.com/openebs/mayastor-control-plane/master/deploy/operator-rbac.yaml
+kubectl apply -f https://raw.githubusercontent.com/openebs/mayastor-control-plane/master/deploy/mayastorpoolcrd.yaml
+
+#
+# Deploy NATS
+#
+kubectl apply -f https://raw.githubusercontent.com/openebs/mayastor/master/deploy/nats-deployment.yaml
```
akhilerm, line 254: NATS is being removed now, so I think this step should be changed.

akhilerm (commented):
Have a few questions and suggestions on the changes. 

Also, IMHO, the azure specific scripts for bringing up the cluster, multi-az cluster etc should be part of documentation and we should refer to it, rather than adding those scripts in mayastor repo.

PR #1116, opened 2022-02-28
fix(Jenkins): build correct images

paulyoong:
Ensure the correct source code is used to build the images. This should
ensure that the latest images are pushed to the registry.


Abhinandan-Purkait (approved):

paulyoong:
bors merge

PR #1119, opened 2022-03-02
fix(python tests): cleanup containers

paulyoong:
Before running the python tests ensure that all existing containers are
destroyed. This will ensure that the python tests are run in a clean
environment.

It has been observed on CI that if containers are left around (due to
previous test failures) subsequent python tests will fail. The error
message will look like:
"compose.project.ProjectError: Encountered errors while bringing up the
project."

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -332,6 +332,9 @@ pipeline {
             stage('run tests') {
               steps {
                 sh 'printenv'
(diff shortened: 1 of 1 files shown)


Abhinandan-Purkait (approved):

tiagolobocastro (approved):

paulyoong:
bors merge

Issue #1063, opened 2022-01-04
 Add arm64 platform support 


xin3liang:
Hi, we are evaluating/verifying Mayastor on arm64 platforms.  But we found that there is no arm64 container images on the docker hub [mayadata repo](https://hub.docker.com/r/mayadata/).
Please provide arm64 container images as well. That will be appreciated！

xin3liang:
FYI, for dependent arm64 Bitnami etcd images. See this [upstream issue](https://github.com/bitnami/charts/issues/7305#issuecomment-1003344728). Bitnami is working on it and there is a workaround to build arm64 images.

neoaggelos:
Here's the code that fails https://github.com/openebs/mayastor/blob/85af174309b45be72cfb36d0630be82bc460154e/mayastor/src/core/env.rs#L682-L697, apparently the spdk subsystem fails to initialize.

Not sure if this is a missing module/configuration, or an issue with arm builds in general.

neoaggelos:
And here's the error in the C code of spdk. https://github.com/openebs/spdk/blob/ec2e6e2b91771abe5136f6f8b19a21130977772c/module/bdev/crypto/vbdev_crypto.c#L384-L391

I really am out of my depth as to what might me missing, I would appreciate any pointers.

xin3liang:
> I've been trying to get arm64 mayastor to work, but I'm getting the following error:
> 
> ```
> # microk8s.kubectl logs -n mayastor mayastor-7vgjh
> Defaulted container "mayastor" out of: mayastor, registration-probe (init), etcd-probe (init), initialize-pool (init)
> [2022-05-12T08:42:53.222263249+00:00  INFO mayastor:mayastor.rs:94] free_pages: 1024 nr_pages: 1024
> [2022-05-12T08:42:53.222373227+00:00  INFO mayastor:mayastor.rs:133] Starting Mayastor version: v1.0.0-117-g77f837aa7f7d
> [2022-05-12T08:42:53.222445241+00:00  INFO mayastor:mayastor.rs:134] kernel io_uring support: yes
> [2022-05-12T08:42:53.222457220+00:00  INFO mayastor:mayastor.rs:138] kernel nvme initiator multipath support: yes
> [2022-05-12T08:42:53.222498991+00:00  INFO mayastor::core::env:env.rs:600] loading mayastor config YAML file /var/local/mayastor/config.yaml
[...]

PR #1121, opened 2022-03-10
fix(build): fix arm64 build

xin3liang:
Fix build on arm64 platform. Intel-ipsec-mb lib is only specific to
x64_64 platform.
And Mayastor can run on arm64 now, not just x86_64.

Signed-off-by: Xinliang Liu <xinliang.liu@linaro.org>

This PR is related to Mayastor arm64 support: https://github.com/openebs/mayastor/issues/1063

### DIFF
--- a/nix/pkgs/libspdk/default.nix
+++ b/nix/pkgs/libspdk/default.nix
@@ -133,6 +133,7 @@ let
       install -v build/lib/pkgconfig/*.pc        $out/lib/pkgconfig/
       install -v dpdk/build/lib/*.a              $out/lib/
       install -v dpdk/build/lib/pkgconfig/*.pc   $out/lib/pkgconfig/
+    '' + lib.optionalString targetPlatform.isx86_64 ''
       install -v intel-ipsec-mb/lib/*.a          $out/lib/
 
       # fix paths in pkg config files
--- a/chart/templates/mayastor-daemonset.yaml
+++ b/chart/templates/mayastor-daemonset.yaml
@@ -24,7 +24,6 @@ spec:
       dnsPolicy: ClusterFirstWithHostNet
       nodeSelector:
         openebs.io/engine: mayastor
-        kubernetes.io/arch: amd64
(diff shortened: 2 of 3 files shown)


Abhinandan-Purkait (approved):

nix/pkgs/libspdk/default.nix
```diff
@@ -133,6 +133,7 @@ let
       install -v build/lib/pkgconfig/*.pc        $out/lib/pkgconfig/
       install -v dpdk/build/lib/*.a              $out/lib/
       install -v dpdk/build/lib/pkgconfig/*.pc   $out/lib/pkgconfig/
+    '' + lib.optionalString targetPlatform.isx86_64 ''
       install -v intel-ipsec-mb/lib/*.a          $out/lib/
```
akhilerm, line 137: q: Is there an alternative to ipsec library in arm64? Just thinking if its used in only x64,  what is the corresponding in arm64.

xin3liang: AFAIK, intel-ipsec-mb only supports intel x64 CPUs, because it is implemented in[ intel individual acceleration instructions](https://github.com/intel/intel-ipsec-mb/tree/main/lib). And for arm64 there is [one armv8 crypto lib ](https://doc.dpdk.org/guides/cryptodevs/armv8.html )in DPDK which haven been integrated in Mayastor yet.

xin3liang: And how Mayastor use intel-ipsec-mb? Because I [remove the crypto lib ](https://github.com/mayadata-io/spdk-rs/pull/9/files#diff-d0d98998092552a1d3259338c2c71e118a5b8343dd4703c0c7f552ada7f9cb42R138)it still can be compiled and run on arm64.

tiagolobocastro (approved):

xin3liang:
Ping

tiagolobocastro:
bors merge

Issue #1155, opened 2022-07-07
Build container image mayastor-io-engine-dev failed


xin3liang:
**Describe the bug**
Build container image mayastor-io-engine-dev failed at io-engine-bench build

**To Reproduce**
```
mayastor$ nix-build -A images.mayastor-io-engine-dev
...
   Compiling concurrent-queue v1.2.2
   Compiling libudev-sys v0.1.4
   Compiling byte-unit v4.0.14
   Compiling yaml-rust v0.4.5
error: failed to run custom build command for `io-engine-bench v0.1.0 (/build/mayastor/io-engine-bench)`

Caused by:
  process didn't exit successfully: `/build/mayastor/target/debug/build/io-engine-bench-ff413fc1eedf48ff/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: NotPresent', io-engine-bench/build.rs:4:57
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
error: build failed
error: builder for '/nix/store/0cblhgrwvsnqisbhr41xgdbvf2mx1ihz-io-engine.drv' failed with exit code 101
error: 1 dependencies of derivation '/nix/store/jp2bl7wppcl4x7ai6x1cm2ck5lclj5j5-docker-layer-mayastor-io-engine-dev.drv' failed to build
error: 1 dependencies of derivation '/nix/store/3yalh08i9ps2b5fwrm5zxs6vsm6zv2lj-docker-image-mayastor-io-engine-dev.tar.gz.drv' failed to build
```

**Expected behavior**
success

** OS info (please complete the following information):**
 - Distro: [e.g. NixOS]
```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.2 LTS
Release:        20.04
Codename:       focal
```
 - Kernel version: 
```
$ uname -a
Linux liuxl-focal 5.4.0-109-generic #123-Ubuntu SMP Fri Apr 8 09:12:14 UTC 2022 aarch64 GNU/Linux
```
 - MayaStor revision or container image: branch develop 

**Additional context**
cat io-engine-bench/build.rs
```
  1 fn main() {
  2     let profile = std::env::var("PROFILE").unwrap();
  3     let spdk_rpath =
[...]

PR #1156, opened 2022-07-07
fix: build only the binaries for debug builds

tiagolobocastro:
This avoids building the benchmarks, which we don't need for container images anyway.
The benchmark is excluded explicitly anyway just to make it clear.

Resolves: #1155


tiagolobocastro:
bors merge

xin3liang:
Thanks for the fix.

PR #1161, opened 2022-07-25
Nexus code cleanup

dsavitskiy:
Lots of changes in Nexus and related code aimed to clean up code structure, to simplify adding of new functionality of partial rebuild and thing provisioning.

- Nexus child list access made more strict.
- Rebuild job is now explicitly owned by Nexus child (before, rebuild job was decoupled from Nexus code, but it was still implicitly owned by nexus child).
- SPDK channel and thread usage improved.
- Error definitions cleaned up.


dsavitskiy:
bors try

io-engine/src/rebuild/rebuild_error.rs
```diff
@@ -0,0 +1,61 @@
+use crate::{bdev_api::BdevError, core::CoreError};
+use snafu::Snafu;
+use spdk_rs::{BdevDescError, DmaError};
+
+#[derive(Debug, Snafu, Clone)]
+#[snafu(visibility(pub(crate)), context(suffix(false)))]
```
tiagolobocastro, line 6: what does `context(suffix(false))` do?

io-engine/tests/lock_lba_range.rs
```diff
@@ -173,66 +110,42 @@ fn multiple_locks() {
...
     reactor_poll!(r);
 
     // Poll reactor to allow the second lock to be obtained
-    reactor_poll!(100);
+    let lock2 = recv_from(lock_receiver);
 
-    // First lock released, second lock should succeed.
-    assert!(lock_receiver.try_recv().is_ok());
+    // // First lock released, second lock should succeed.
+    // assert!(lock_receiver.try_recv().is_ok());
```
tiagolobocastro, line 143: does this need to be commented?

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -202,140 +225,175 @@ impl<'n> Nexus<'n> {
...
     ) -> Result<(), Error> {
-        let recovering_child =
-            self.as_mut().get_child_by_name(&job.destination)?;
+        if !self.rebuild_job(child_uri)?.state().done() {
+            // Leave all states as they are
+            return Ok(());
+        }
+
+        let dst_child = self.as_mut().child_mut(child_uri)?;
+        let job = dst_child.remove_rebuild_job().unwrap();
```
tiagolobocastro, line 336: should this return an error? or maybe not since we're in the callback we know the job is there?

io-engine/src/core/thread.rs
```diff
@@ -1,281 +1 @@
...
-        unsafe {
-            spdk_set_thread(self.0.as_ptr());
-            // set that we *want* to exit, but we have not exited yet
-            spdk_thread_exit(self.0.as_ptr());
-
-            // now wait until the thread is actually exited the internal
-            // state is updated by spdk_thread_poll()
-            while !spdk_thread_is_exited(self.0.as_ptr()) {
-                spdk_thread_poll(self.0.as_ptr(), 0, 0);
-            }
```
tiagolobocastro, line 131: Ah so that changes on the other repo came from here...
My understanding is very limited, but is this safe to poll in a tight loop without letting the reactor get a chance to run too?

io-engine/Cargo.toml
```diff
@@ -75,7 +75,7 @@ serde_json = "1.0.66"
 serde_yaml = "0.8.18"
 sha2 = "0.10.2"
 signal-hook = "0.3.9"
-snafu = "0.6.10"
+snafu = "0.7.1"
```
tiagolobocastro, line 78: Should we update snafu on the other crates in the workspace to avoid building two different versions?

tiagolobocastro (approved):
I might have missed something since this is huge so someone else should review as well

io-engine/src/rebuild/rebuild_error.rs
dsavitskiy: New version of Snafu adds a suffix "Error" for context structs it generates. E.g., for Error::NexusNotFound enum, Snafu generates a struct named "NexusNotFoundError" to be used with .context() method.
This breaks compatibility with existing error enums. The easiest way is to revert to old behaviour by context(suffix(false)).
But, there's a caveat: although Snafu won't add "Error" suffix with this option, it still _removes_ Error if it is already there. E.g. for MyError::NameExistsError enum Snafu generates "NameExists" struct.
Fortunately, we only had a few enums ending in "Error" in data plane, so it was not hard to rename them.

The reason why I updated to newer Snafu is that it allows to put generated struct in a namespace, preventing module's namespace pollution, and making exporting error definitions easier.

tiagolobocastro: Thanks for explaining :+1: we might also upgrade snafu on control-plane at some point

io-engine/src/core/thread.rs
dsavitskiy: Sorry Tiago, I cannot answer this question :-) I just moved code around to make structure more clear. Right now, I cannot reason how good any particular part of our reactor/thread code it. I remember Jeffry telling me that some parts are actually not quite good, and should probably be reworked. He wanted Zeeshan to do that, but unfortunately he didn't finish his work.

Perhaps, we should add a ticket to our Tech Debt epic to make investigation in that area. As for EB2 release, I'd rather keep as is.

io-engine/tests/lock_lba_range.rs
dsavitskiy: I forgot to remove these lines. 
Now, lock lba range methods work a bit differently: lock_lba_range() returns a "lock" object which must be passed to unlock_lba_range(). This makes code safer because caller doesn't have to manually keep track on passing exactly the same context pointer and range to both lock() and unlock(). The "lock" object does it automatically.
Because of that, some tests were removed (the one testing for error handling when different args were passed to lock/unlock). Other tests were worked to use the new API.

io-engine/Cargo.toml
dsavitskiy: As I mentioned above, this version of Snafu breaks compatibility, so it might require fixes in these crates. I'll try, if it is not too much to change, I will fix it.

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
dsavitskiy: The child is checked to have a job running in a line above.
But it won't hurt to add a check and a warning here instead of unwrap.

io-engine/Cargo.toml
dsavitskiy: It took a bit more effort that I expected but I updated anyway:
PR for crates in util/dependencies: https://github.com/openebs/mayastor-dependencies/pull/5
PR for spdk-rs: https://github.com/mayadata-io/spdk-rs/pull/12

dsavitskiy:
bors try

mtzaurus (approved):

dsavitskiy:
bors merge

PR #1162, opened 2022-07-25
feat(logger): adding compact log output format

dsavitskiy:
Default logger output format can be hard to follow because each line
has a verbose prefix of date, time and code location.
For development purposes, a alternative compact format is introduced.

### DIFF
--- a/io-engine/src/core/env.rs
+++ b/io-engine/src/core/env.rs
@@ -90,6 +90,9 @@ pub struct MayastorCliArgs {
     #[structopt(short = "L")]
     /// Enable logging for sub components.
     pub log_components: Vec<String>,
+    #[structopt(short = "F")]
+    /// Log format.
+    pub log_format: Option<logger::LogFormat>,
     #[structopt(short = "m", default_value = "0x1")]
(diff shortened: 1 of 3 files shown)


paulyoong (approved):

io-engine/src/core/env.rs
```diff
@@ -561,7 +565,7 @@ impl MayastorEnvironment {
     }
 
     /// initialize the logging subsystem
-    fn init_logger(&mut self) -> Result<()> {
+    fn init_logger(&mut self) {
```
tiagolobocastro, line 568: Any reason for not making failure an error now?

tiagolobocastro (approved):
Please sign your commits before merging.

io-engine/src/core/env.rs
dsavitskiy: This result was never checked in a meaningful way: it was just unwraped, causing panic if spdk_log_set_flag() failed.

dsavitskiy:
bors try

dsavitskiy:
bors merge

PR #1166, opened 2022-08-02
feat(nexus): nexus-level fault injections

dsavitskiy:
Basic support for injecting faults at nexus level has been added.

This feature is indented for developers who need to test code dealing with nexus child failures, retirement logic, rebuild logic, etc.

gRPC methods and CLI commands have been added to "inject" failures of device(s). Devices are "failed" on nexus I/O level.

This is a very basic support for fault injection, intended to be extended in future.

This functionality is implemented via Cargo features, and are opted out by default. To enable, build with "cargo build --features fault_injection".


io-engine/src/core/device_monitor.rs
```diff
@@ -26,26 +45,32 @@ pub async fn device_monitor_loop() {
     loop {
         interval.tick().await;
         if let Some(w) = device_cmd_queue().take() {
-            info!(?w, "executing command");
+            info!("Device monitor exexuting command: {}", w);
```
tiagolobocastro, line 48: s/exexuting/executing
Again, any reason to remove w from the field names?

tiagolobocastro (commented):
What kind of tests will make use of this? Eg: if we want to test this using the docker compose framework, we'd need to change the test scripts to build with the feature enabled? 
Would it be easier to control fault_injection enablement using a cmdline argument instead?

io-engine/src/bdev/nexus/nexus_io_subsystem.rs
```diff
@@ -60,9 +86,13 @@ impl<'n> NexusIoSubsystem<'n> {
...
     pub(super) async fn suspend(&mut self) -> Result<(), Error> {
-        assert_eq!(Cores::current(), Cores::first());
+        assert_eq!(
+            Cores::current(),
+            Cores::first(),
+            "NexusIoSubsystem::suspend() must called on the first core"
+        );
 
-        trace!(?self.name, "pausing nexus I/O");
+        trace!("{:?}: pausing I/O...", self);
```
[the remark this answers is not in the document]
dsavitskiy: We have had inconsistent, often poorly formatted log messages: some print nexus name, some don't, some use notation like "?self.name", some use Display trait, some use Debug, some format in other ways.
I'm trying to make log messages consistent and clear. In order to accomplish that, I implement Debug trait for our structures a consistent way, and change logging to use it.
As for nexus, Debug trait for it now prints name and status.

io-engine/src/core/device_monitor.rs
dsavitskiy: Reason is poorly formatted, hard-to-follow output.

dsavitskiy:
> What kind of tests will make use of this? Eg: if we want to test this using the docker compose framework, we'd need to change the test scripts to build with the feature enabled?
> Would it be easier to control fault_injection enablement using a cmdline argument instead?

In production, injection code must be excluded completely, because checks for faults stay in I/O code path. This cannot be done effectively by cmdline args. For our component-level testing, we can enable this feature by default in test (and debug) builds.

Actually, my primary goals for this functionality are development process, and simple component level automated testing. E.g., when debugging partial rebuild code, I want to have an easy way to run I/O on nexus, and have a child faulted in the middle.

However, for proper testing of production build, we have to use SanBlaze H/W to have "more real" I/O failures, rather than rely on this kind of fault injections.

io-engine/src/core/device_monitor.rs
dsavitskiy: I will fix typo.

mtzaurus:
As an initial implementation it's fine, IMHO.
Thought, in the future we might want to extend the functionality.

mtzaurus (approved):

io-engine/src/core/device_monitor.rs
tiagolobocastro: The possible poor format is only because the way we format the prints. Field names are output using debug/display traits, depending on ? or %.
I'm fine with these changes but we really start loosing the advantages of tracing and every info/debug becomes a simple text/log message

dsavitskiy: I see your point. I'll take a look if it possible to disable field names output in logger formatter.

tiagolobocastro (approved):

dsavitskiy:
bors merge

PR #1169, opened 2022-08-04
fix(nexus): fixing crash during nexus child retire

dsavitskiy:
When a nexus child is an SPDK block device, retiring a child can cause
use-after-free of a wrapper object which refers to SPDK Bdev.
SPDK Bdev itself is not destroyed at that point, only the wrapper is
invalidated.

Signed-off-by: Dmitry Savitskiy <dmitry.savitsky@datacore.com>


tiagolobocastro (approved):

dsavitskiy:
bors try

mtzaurus (approved):

PR #1167, opened 2022-08-02
feat(thin): passing thin flag down to spdk

dsavitskiy:

PR #1170, opened 2022-08-07
chore(spdk): updating to spdk 22.05

dsavitskiy:
SPDK 22.05 stable has been released a couple of months ago.
It has a feature we need for thin provisioning: passing errno from blob store internals to bdev I/O layer.
However, this feature is half-baked: the errno is not propagated further, is not stored, and therefore cannot be accessed via Bdev API. To support this, I have added a patch for our SPDK branch, which I will submit via a separate PR. We need to integrate SPDK 22.05 first.

This PR contains commits from https://github.com/openebs/mayastor/pull/1167, which I'm going to merge first, and after that I'll update with PR to have only single commit SPDK 22.05.

### DIFF
--- a/io-engine/src/bdev/nvme.rs
+++ b/io-engine/src/bdev/nvme.rs
@@ -94,7 +94,7 @@ impl CreateDestroy for NVMe {
                 context.count,
                 Some(nvme_create_cb),
                 cb_arg(sender),
-                std::ptr::null_mut(), // context.prchk_flags,
+                std::ptr::null_mut(),
                 std::ptr::null_mut(),
                 false,
             )
@@ -156,7 +156,6 @@ const MAX_NAMESPACES: usize = 1;
 struct NvmeCreateContext {
     trid: spdk_nvme_transport_id,
     names: [*const c_char; MAX_NAMESPACES],
-    // prchk_flags: u32,
     count: u32,
 }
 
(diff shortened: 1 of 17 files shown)


dsavitskiy:
bors try

mtzaurus (approved):

dsavitskiy:
bors try

dsavitskiy:
bors try

io-engine/src/bdev/nvme.rs
```diff
@@ -159,7 +156,7 @@ const MAX_NAMESPACES: usize = 1;
 struct NvmeCreateContext {
     trid: spdk_nvme_transport_id,
     names: [*const c_char; MAX_NAMESPACES],
-    prchk_flags: u32,
+    // prchk_flags: u32,
```
tiagolobocastro: why is this bit commented out now?

tiagolobocastro (approved):

io-engine/src/bdev/nvme.rs
dsavitskiy: "prchk_flags" were here to be passed to bdev_nvme_create(), but we never assigned anything meaningful.
In SPDK 22.05, "prchk_flags" argument is no longer passed directly to bdev_nvme_create(). It is part of "nvme_ctrlr_opts" arg instead, which we do not use.
I'll remove the commented code. If we need this parameter in future, we'll add it again somewhere.

dsavitskiy: fixed

dsavitskiy:
bors merge

PR #1176, opened 2022-08-14
feat(tests): making a separate crate of common tests modules

dsavitskiy:
Io-engine tests and benchmarks use the same common module.
Previously, benchmark used it via a symlink, which cause
problems with dependencies.
This commit also added helpers for v1 API.


dsavitskiy:
bors try

io-engine-tests/Cargo.toml
```diff
@@ -0,0 +1,91 @@
+[package]
+name = "io-engine-tests"
+description = "I/O test support library"
+version = "1.0.0"
+edition = "2018"
+authors = [
```
tiagolobocastro: nit: it's no longer recommended to add authors to crates, and it might become deprecated and removed in the future.

tiagolobocastro (approved):

mtzaurus (approved):

Cargo.toml
```diff
@@ -7,6 +7,7 @@ members = [
 	"libnvme-rs",
 	"io-engine",
     "io-engine-bench",
+    "io-engine-tests",
```
Abhinandan-Purkait: nit: is the indentation incorrect here?

Abhinandan-Purkait (approved):

io-engine-tests/Cargo.toml
dsavitskiy: I'll remove it.

Cargo.toml
dsavitskiy: It's okay on my machine. Probably, a tab is here. I'll re-format the file.

dsavitskiy:
bors merge

PR #1180, opened 2022-08-18
chore: add child state reasons to the v0 api

tiagolobocastro:
This is done because the control-plane does not support the v1 api yet.
This way we can proceed with supporting enospc.

### DIFF
--- a/io-engine/src/bin/io-engine.rs
+++ b/io-engine/src/bin/io-engine.rs
@@ -45,17 +45,6 @@ fn start_tokio_runtime(args: &MayastorCliArgs) {
     Mthread::spawn_unaffinitized(move || {
         runtime::block_on(async move {
             let mut futures = Vec::new();
-            if let Some(registration_addr) = registration_addr {
-                debug!("mayastor grpc registration init");
-                Registration::init(
-                    &node_name,
-                    &grpc_address.to_string(),
-                    registration_addr,
-                    api_versions.clone(),
-                );
-                futures.push(Registration::run().boxed());
-            }
-
             PersistentStore::init(persistent_store_endpoint).await;
             runtime::spawn(device_monitor_loop());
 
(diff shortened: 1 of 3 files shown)


dsavitskiy (approved):

io-engine/src/bin/io-engine.rs
```diff
@@ -45,29 +45,28 @@ fn start_tokio_runtime(args: &MayastorCliArgs) {
...
                 grpc::MayastorGrpcServer::run(
                     grpc_address,
                     rpc_address,
-                    api_versions,
+                    api_versions.clone(),
                 )
                 .boxed(),
             );
 
+            if let Some(registration_addr) = registration_addr {
```
Abhinandan-Purkait, line 60: Question: Why would this matter? The grpc server for this client is in the core-agent right, that will start before this anyhow because of the init-containers.

Abhinandan-Purkait (approved):

io-engine/src/bin/io-engine.rs
tiagolobocastro: Not sure I don't understand what you mean? The core will be ready indeed, but the grpc server in mayastor might not, which is the problem...

Abhinandan-Purkait: Okay. So the flow we want is register only if mayastor grpc server is ready. I was talking in terms of dependency between registration client and mayastor grpc server. Registration client is not dependent on mayastor grpc server.

tiagolobocastro: Right, it's not dependent - though maybe we should make it dependent, or at least, only register once the server side is ready to receive incoming. Though this is really only seen on CI so maybe this simple reorder is enough..

Abhinandan-Purkait: Yep.

tiagolobocastro:
bors merge

PR #1183, opened 2022-08-22
fix(nexus): fixing assert failure when removing non-existing child

dsavitskiy:

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev_children.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev_children.rs
@@ -277,7 +277,10 @@ impl<'n> Nexus<'n> {
         let paused = self.as_mut().pause_rebuild_jobs(uri).await;
 
         let idx = match self.children_iter().position(|c| c.uri() == uri) {
-            None => return Ok(()),
+            None => {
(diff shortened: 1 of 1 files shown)


io-engine/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -277,7 +277,10 @@ impl<'n> Nexus<'n> {
         let paused = self.as_mut().pause_rebuild_jobs(uri).await;
 
         let idx = match self.children_iter().position(|c| c.uri() == uri) {
-            None => return Ok(()),
+            None => {
```
tiagolobocastro, line 280: would it be easier to move 277 below getting idx? Then we can just return Ok() here.

tiagolobocastro (approved):

io-engine/src/bdev/nexus/nexus_bdev_children.rs
dsavitskiy: Probably yes, but I wouldn't touch that now.

mtzaurus (approved):

GlennBullingham:
bors merge

PR #1186, opened 2022-08-30
feat(thin): rebuild of thin provisioned replicas

dsavitskiy:


io-engine/src/bdev/nvmx/handle.rs
```diff
@@ -503,23 +507,38 @@ impl BlockDeviceHandle for NvmeDeviceHandle {
...
+            _ => Err(CoreError::ReadFailed {
+                offset,
+                len: buffer.len(),
+            }),
         };
         inner.discard_io();
         ret
     }
 
+    /// TODO
```
tiagolobocastro, line 534: can we document this?

io-engine/src/bdev/nvmx/handle.rs
```diff
@@ -503,23 +507,38 @@ impl BlockDeviceHandle for NvmeDeviceHandle {
...
+    fn set_read_mode(&mut self, mode: ReadMode) {
+        self.prchk_flags = match mode {
+            ReadMode::Normal => 0,
```
tiagolobocastro, line 537: is there an actual `SPDK_NVME_IO_FLAGS_**` to use instead of 0?

io-engine/src/rebuild/rebuild_job.rs
```diff
@@ -395,12 +398,20 @@ impl<'n> RebuildJob<'n> {
...
-            .read_at(blk * self.block_size, copy_buffer)
-            .await
-            .context(ReadIoFailed {
-                bdev: &self.src_uri,
-            })?;
+        source_hdl.set_read_mode(ReadMode::UnwrittenFail);
+
+        let res = source_hdl.read_at(blk * self.block_size, copy_buffer).await;
+
+        if let Err(CoreError::ReadingUnallocatedBlock {
```
tiagolobocastro, line 405: what happens when only part of the read range is unallocated?

io-engine/src/bdev/nvmx/handle.rs
dsavitskiy: No, it is just 0.

io-engine/src/rebuild/rebuild_job.rs
dsavitskiy: It will still fail. Probably, it would be a good idea to limit read range to be less or equal to cluster size when SPDK_NVME_IO_FLAGS_UNWRITTEN_READ_FAIL is specified.

io-engine/src/bdev/nvmx/handle.rs
dsavitskiy: Ok.

dsavitskiy:
bors try

io-engine/src/rebuild/rebuild_job.rs
tiagolobocastro: But even with the same size it'd be tricky because if it's not aligned we're back to error?

dsavitskiy: Sorry, I meant both aligned to cluster boundar,y and equal or less than cluster size.

dsavitskiy:
bors try

tiagolobocastro (approved):
Once you add the check for rebuild IO smaller and multiple of ClusterSize, maybe add a comment on the commit as well for future reference.

mtzaurus (approved):

dsavitskiy:
bors try

dsavitskiy:
bors merge

PR #1193, opened 2022-09-06
feat(reactor): freeze detection support for reactors

mtzaurus:
This fix introduces reactor heartbeat loop to detect frozen reactors, which is opted by the “diagnostics” feature (not active by default). New CLI command --diagnose-stack is added to allow dumping stack for all threads in running I/O agent, which allows stack trace collection when a frozen reactor is detected.

### DIFF
--- a/io-engine/Cargo.toml
+++ b/io-engine/Cargo.toml
@@ -10,6 +10,8 @@ build = "build.rs"
 [features]
 # Enables fault injection code.
 fault_injection = []
+# Enables diagnostics operations.
+diagnostics = ["async-process", "rstack"]
 
 [[bin]]
 name = "io-engine"
@@ -93,6 +95,9 @@ tracing-log = "0.1.2"
 tracing-subscriber = "0.2.20"
 udev = "0.6.2"
 url = "2.2.2"
+gettid = "0.1.2"
+async-process = { version = "1.5.0", optional = true }
+rstack =  { version = "0.3.2", optional = true }
 
 jsonrpc = { path = "../jsonrpc"}
 mayastor-api = { path = "../rpc/mayastor-api" }
--- a/io-engine/src/core/diagnostics.rs
+++ b/io-engine/src/core/diagnostics.rs
@@ -1,0 +1,102 @@
+use crate::core::{MayastorCliArgs, Reactor};
+use async_process::Command;
+use rstack::TraceOptions;
+use std::env;
+
+/// Get command path from process CLI arguments.
+fn get_io_agent_path() -> String {
+    env::args().next().as_ref().map(String::from).unwrap()
+}
+
+/// Dump stack for current I/O engine instance and log it.
+async fn dump_self_stack() {
+    let pid = std::process::id();
+
+    info!(pid, "Collecting stack for I/O agent process");
+
+    let out = Command::new(get_io_agent_path())
+        .arg("--diagnose-stack")
+        .arg(pid.to_string())
+        .output()
+        .await;
+
+    match out {
+        Err(error) => {
+            error!(
+                %error,
+                "Failed to collect process stack"
+            );
+        }
+        Ok(output) => {
+            let l = String::from_utf8(output.stdout).unwrap();
+            l.split('\n').for_each(|s| {
+                if !s.is_empty() {
+                    if s.starts_with("thread ") {
+                        info!("\n{}", s);
+                    } else {
+                        info!("{}", s);
+                    }
+                }
+            });
+            info!("Process stack collected successfully");
+        }
+    }
+}
+
+/// Dump detailed diagnostic information for the frozen reactor.
+/// As of now print only basic information about the reactor,
+/// in the future might print stack traces too.
+pub fn diagnose_reactor(reactor: &Reactor) {
+    info!(
+        core=reactor.core(),
+        tid=reactor.tid(),
+        state=%reactor.get_state(),
+        "Reactor is frozen"
+    );
+
+    // Spawn a task to perform stack collection.
+    tokio::spawn(async move {
+        dump_self_stack().await;
+    });
+}
+
+/// Collect sracktraces for all stack frames for all threads in target process
+/// and dump it to stdout.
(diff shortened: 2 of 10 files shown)


io-engine/Cargo.toml
```diff
@@ -93,6 +95,9 @@ tracing-log = "0.1.2"
 tracing-subscriber = "0.2.20"
 udev = "0.6.2"
 url = "2.2.2"
+async-process = "1.5.0"
+gettid = "0.1.2"
+rstack =  "0.3.2"
```
tiagolobocastro:
```suggestion
async-process = version = { "1.5.0", optional = true }
gettid = version = { "0.1.2", optional = true }
rstack =  version = { "0.3.2", optional = true }
```

io-engine/Cargo.toml
```diff
@@ -10,6 +10,8 @@ build = "build.rs"
 [features]
 # Enables fault injection code.
 fault_injection = []
+# Enables diagnostics operations.
+diagnostics = []
```
tiagolobocastro:
```suggestion
diagnostics = [ "async-process", "gettid", "rstack" ]
```

io-engine/src/core/diagnostics.rs
```diff
@@ -0,0 +1,102 @@
...
+use async_process::Command;
+use rstack::TraceOptions;
+use std::env;
+
+/// Get command path from process CLI arguments.
+fn get_io_agent_path() -> String {
+    env::args().next().as_ref().map(String::from).unwrap()
+}
+
+/// Dump stack for current I/O agent instance and log it.
```
tiagolobocastro: nit: I/O agent or I/O engine?

io-engine/src/core/reactor.rs
```diff
@@ -624,3 +635,101 @@ impl Future for &'static Reactor {
...
+
+    /// Metadata for every reactor being monitored by the reactor monitor.
+    struct ReactorRecord {
+        frozen: bool,
+        reactor: &'static Reactor,
+        reactor_tick: &'static AtomicU64,
+        core: u32,
+    }
+
+    let mut interval = tokio::time::interval(Duration::from_secs(1));
```
tiagolobocastro, line 659: Should this and `MAX_MISSED_HEARTBEATS` be configurable via cmdline?

io-engine/src/core/reactor.rs
```diff
@@ -288,15 +288,18 @@ impl Reactor {
             incoming: crossbeam::queue::SegQueue::new(),
             lcore: core,
             flags: Cell::new(ReactorState::Init),
+            tid: Cell::new(0),
```
tiagolobocastro, line 291: does it need to be a Cell?

io-engine/src/core/reactor.rs
```diff
@@ -624,3 +635,101 @@ impl Future for &'static Reactor {
...
+        frozen: bool,
+        reactor: &'static Reactor,
+        reactor_tick: &'static AtomicU64,
+        core: u32,
+    }
+
+    let mut interval = tokio::time::interval(Duration::from_secs(1));
+    let mut tick: u64 = 0;
+    let mut reactor_state: Vec<ReactorRecord> =
+        Vec::with_capacity(Cores::count().id() as usize);
```
tiagolobocastro: shouldn't this be `Cores::count() as usize` ?

io-engine/src/core/reactor.rs
```diff
@@ -624,3 +635,101 @@ impl Future for &'static Reactor {
         }
     }
 }
+
+/// Maximum number of heartbeats a reactor is allowed to miss
+/// before it is classified as frozen.
+#[cfg(feature = "diagnostics")]
+const MAX_MISSED_HEARTBEATS: u64 = 3;
```
tiagolobocastro: What's the rationale behind using a max missed hearbeats rather than simply bumping the interval to 3s and waiting for that to elapse once for example?

tiagolobocastro (commented):
Once we have the event bus in place we can these and other health things we might check to the event bus :)

io-engine/src/core/reactor.rs
mtzaurus: We need to set the value here in static RO context, as reactor instances are created before we know what thread handles it.

io-engine/Cargo.toml
mtzaurus: Good point, will change.

io-engine/src/core/reactor.rs
```diff
@@ -624,3 +635,101 @@ impl Future for &'static Reactor {
...
+            }
+        }
+
+        // Wait till heartbeat check interval elapses and check ticks
+        // reported by every reactor.
+        interval.tick().await;
+        tick += 1;
+
+        for r in &mut reactor_state {
+            if r.frozen {
```
tiagolobocastro, line 711: Should we "send" this knowledge to the registration subsystem? Would that help? Eg: if we know a reactor is stuck should the control-plane stop trying to send grpc to this node? Atm the registration keep-alive keep being sent so the control-plane thinks all is good.

dsavitskiy (approved):

io-engine/src/core/reactor.rs
mtzaurus: Probably it's simpler, yes.

io-engine/src/core/reactor.rs
mtzaurus: ::count() has a bit different semantics in our SPDK implementation:

```    pub fn count() -> Self {
        Cores(Self::get_core(Core::Count))    
}
```

The closest possible solution for us is:
```
Cores::get_core(Core::Count) as usize
```

io-engine/src/core/reactor.rs
mtzaurus: Yes, I will add such a command line options for 'diagnostics' feature.

io-engine/src/core/reactor.rs
tiagolobocastro: Thanks for explaining :+1:

tiagolobocastro (approved):

mtzaurus:
bors merge

PR #1197, opened 2022-09-15
feat(nexus): nexus shutdown operation

mtzaurus:
Implemented shutdown operation for nexuses, which lets keeping nexus published but with all replicas disconnected and all rebuild jobs cancelled. Also a new nexus state ‘shutdown’ is added to denote nexuses that have been shutdown.

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev.rs
@@ -72,6 +72,15 @@ pub enum NexusTarget {
     NexusNvmfTarget,
 }
 
+/// Sensitive nexus operations that might require extra checks against
+/// current nexus state in order to be performed.
+#[derive(Debug)]
+pub enum NexusOperation {
+    ReplicaAdd,
+    ReplicaRemove,
+    ReplicaOnline,
+}
+
 /// TODO
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub enum NvmeAnaState {
@@ -197,6 +206,10 @@ pub enum NexusStatus {
     Degraded,
     /// Online
     Online,
+    /// Shutdown in progress
+    ShuttingDown,
+    /// Shutdown
+    Shutdown,
 }
 
 impl Display for NexusStatus {
@@ -208,6 +221,8 @@ impl Display for NexusStatus {
                 NexusStatus::Degraded => "degraded",
                 NexusStatus::Online => "online",
                 NexusStatus::Faulted => "faulted",
+                NexusStatus::ShuttingDown => "shutting_down",
+                NexusStatus::Shutdown => "shutdown",
             }
         )
     }
@@ -224,6 +239,10 @@ pub enum NexusState {
     Open,
     /// reconfiguring internal IO channels
     Reconfiguring,
+    /// Shutdown in progress
+    ShuttingDown,
+    /// nexus has been shutdown
+    Shutdown,
 }
 
 impl Display for NexusState {
@@ -236,6 +255,8 @@ impl Display for NexusState {
                 NexusState::Closed => "closed",
                 NexusState::Open => "open",
                 NexusState::Reconfiguring => "reconfiguring",
+                NexusState::ShuttingDown => "shutting_down",
+                NexusState::Shutdown => "shutdown",
             }
         )
     }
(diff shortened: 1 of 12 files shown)


io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -452,6 +467,21 @@ impl<'n> Nexus<'n> {
         self.children.len()
     }
 
+    /// Check whether nexus can perform target operation.
+    pub(crate) fn check_nexus_operation(
```
tiagolobocastro, line 477: Not related to this pr, but I think we may need something like this to solve the issue of having a single lock for all nexuses at the grpc level.
Another issue with the current locks is that IIRC on timeout we drop them (or at least we used to IIRC), so we may end up with 2 concurrent operations on a resource.

io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -602,6 +632,68 @@ impl<'n> Nexus<'n> {
...
     }
 
+    pub async fn shutdown(mut self: Pin<&mut Self>) -> Result<(), Error> {
+        // Check if nexus is not already shut down and mark nexus
+        // as being shutdown, preserving current nexus state.
+        let prev_state = {
+            let mut s = self.state.lock();
+
+            if *s == NexusState::Shutdown {
+                return Err(Error::OperationNotAllowed {
```
tiagolobocastro: what does this get mapped back up to via grpc so control plane can determine the call failed because we were already shutdown?

test/python/tests/nexus/features/shutdown.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Mayastor nexus shutdown feature
+
+Background:
+    Given a published and connected nexus
+
+Scenario: shutting down a nexus with active fio client
+    Given fio client is running against target nexus
```
tiagolobocastro, line 7: should this be tested for both block and filesystem?

io-engine/src/bdev/nexus/nexus_bdev.rs
tiagolobocastro: Should this be a different error case than on `check_nexus_operation` ?

test/python/tests/nexus/features/shutdown.feature
mtzaurus: That's not needed, as in the end they both will interact with NVMe target via the same initiator code path in kernel.

io-engine/src/bdev/nexus/nexus_bdev.rs
mtzaurus: Good point, we can put it here too.

mtzaurus: > what does this get mapped back up to via grpc so control plane can determine the call failed because we were already shutdown?

IIRC, as of now it will be mapped to 'Status::internal' error code.

mtzaurus:
bors try

gila:
I fail to see the use of such a state. How is it different from a nexus which is not shared?

io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -197,6 +206,8 @@ pub enum NexusStatus {
     Degraded,
     /// Online
     Online,
+    /// Shutdown
+    Shutdown,
```
gila, line 212: Shutdown is very misleading IMHO. What is the difference between unshared and shutdown for example. What is the use case for such a state?

mtzaurus: Such a state is used in nexus switchover: we keep nexus published but pause its I/O subsystem (which keeps the connected initiator alive while we're recreating the nexus on another node).

gila: So why the new state then? IIRC there already is a paused state for the nexus?

mtzaurus: Shutdown operation performs a bit more than just pauses nexus (like closing all its replicas and cancelling all rebuild jobs). This new state is to denote that nexus underwent the former.

dsavitskiy (approved):

mtzaurus:
bors try

io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -602,6 +640,66 @@ impl<'n> Nexus<'n> {
...
     }
 
+    pub async fn shutdown(mut self: Pin<&mut Self>) -> Result<(), Error> {
+        let prev_state = {
+            let mut s = self.state.lock();
+
+            match *s {
+                // If nexus is already shutdown, operation is idempotent.
+                NexusState::Shutdown => return Ok(()),
+                t => {
```
tiagolobocastro, line 664: `NexusState::ShuttingDown => Err(Error::OperationNotAllowed {})` ?

io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -452,6 +473,23 @@ impl<'n> Nexus<'n> {
...
+    pub(crate) fn check_nexus_operation(
+        &self,
+        _op: NexusOperation,
+    ) -> Result<(), Error> {
+        match *self.state.lock() {
+            // When nexus under shutdown or is shutdown, no further nexus
+            // operations allowed.
+            NexusState::ShuttingDown | NexusState::Shutdown => {
+                Err(Error::OperationNotAllowed {
+                    reason: "Nexus is shutdown".to_string(),
```
tiagolobocastro, line 486: nit:
let state = ....
match ....
reason: format!("Nexus is {:?}", state) ?

tiagolobocastro (approved):

mtzaurus:
bors try

mtzaurus:
bors merge

PR #1198, opened 2022-09-16
fix(grpc-socket): use socketAddr type instead of string

tiagolobocastro:
In case the grpc port is not specified we were sending the endpoint with only an ip address. Switch to a socketAddr to make we always use a port.


abhilashshetty04 (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #1210, opened 2022-09-28
feat: add more reservation control to create nexus

tiagolobocastro:
This allows the create to specify whether we want to preempt the existing holder when the existing holder is not known.
EG: If a nexus create times out, we don't know if the reservations have changed or not, so if we attempt to create a nexus, which reservation key do we preempt? With `NexusNvmePreemption::Holder` we can preempt the existing holder, whomever that might be.

This logic is already kind used by the dataplane, but only if the existing holder is type 1, which we cannot create at the moment as we hardcode to 5, which is a "shared" scheme. By passing the `resvType` as an arg, it allows us to control that, and also might be useful for the future.

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev.rs
@@ -5,6 +5,7 @@
 //! application needs synchronous mirroring may be required.
 
 use std::{
+    convert::TryFrom,
     fmt::{Debug, Display, Formatter},
     marker::PhantomPinned,
     os::raw::c_void,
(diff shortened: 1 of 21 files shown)


io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -99,17 +100,60 @@ impl NvmeAnaState {
...
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
+        Ok(match value {
+            0 => Self::Reserved,
+            1 => Self::WriteExclusive,
+            2 => Self::ExclusiveAccess,
+            3 => Self::WriteExclusiveRegsOnly,
+            4 => Self::ExclusiveAccessRegsOnly,
+            5 => Self::WriteExclusiveAllRegs,
+            6 => Self::ExclusiveAccessAllRegs,
+            _ => return Err(()),
```
mtzaurus: Shall we return a string error here that provides details ?

tiagolobocastro: Yep, we should indeed.

tiagolobocastro: Done

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

tiagolobocastro:
bors try

mtzaurus (approved):

tiagolobocastro:
Seems the rebuild grpc failure is still haunting us
bors try

tiagolobocastro:
bors try

datacore-peter (approved):

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

PR #1212, opened 2022-09-29
fix: upgrade dockerhub url v2 to check for tags

tiagolobocastro:
This allows us to avoid re-publishing release tags "proper". Move the check till after the build is done, so we can still push the branch/alias tags.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -124,9 +124,9 @@ cd $SCRIPTDIR/..
 
 if [ -z "$IMAGES" ]; then
   if [ -z "$DEBUG" ]; then
-    IMAGES="mayastor-io-engine mayastor-io-engine-client mayastor-fio-spdk"
+    IMAGES="mayastor-io-engine mayastor-fio-spdk"
   else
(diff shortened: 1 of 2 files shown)


Ab-hishek (approved):

scripts/release.sh
```diff
@@ -11,16 +11,17 @@ set -euo pipefail
 
 # Test if the image already exists in dockerhub
 dockerhub_tag_exists() {
-  curl --silent -f -lSL https://index.docker.io/v1/repositories/$1/tags/$2 1>/dev/null 2>&1
+  curl --silent -f -lSL https://hub.docker.com/v2/repositories/$1/tags/$2 1>/dev/null 2>&1
 }
 
 # Derives tag name from the git repo. That is git tag value or short commit
```
chriswldenyer: This comment needs updating, I think

chriswldenyer (approved):
Looks reasonable

scripts/release.sh
tiagolobocastro: Good catch, done on this and other repos

tiagolobocastro:
bors merge

PR #1230, opened 2022-10-10
fix(nvmf): cherry-pick - don't enable ANA reporting for lvol replicas

tiagolobocastro:
We only need ANA reporting on the nexus and having the replicas ANA seems to cause some issues when the nexus connects to them.
Adding a new `ShareProps` which can be used for various share properties - also added a allowed host option which for the moment defaults to any that will likely be helpful to gate access to the targets.


Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

PR #1234, opened 2022-10-17
feat(grpc): per object nexus locking

mtzaurus:
feat(grpc): per nexus instance locks in grpc api
    
    Nexus gRPC API is now protected on per-nexus basis instead of
    global API serialisation, which provides better scalability.
    
    Generic lock manager implemented to provide common API for
    protecting sensitive resources.
    
### DIFF
--- a/io-engine/src/core/lock.rs
+++ b/io-engine/src/core/lock.rs
@@ -1,0 +1,189 @@
+use std::{
+    collections::hash_map::DefaultHasher,
+    hash::{Hash, Hasher},
+    time::Duration,
+};
+
+use futures::lock::{Mutex, MutexGuard};
+use once_cell::sync::OnceCell;
+
+/// Common IO engine resource subsystems.
+pub struct ProtectedSubsystems;
+impl ProtectedSubsystems {
+    pub const NEXUS: &'static str = "nexus";
+}
+
+/// Configuration parameters for initialization of the Lock manager.
+#[derive(Debug, Default)]
+pub struct ResourceLockManagerConfig {
+    /// Configs for subsystems: denote id and maximum amount of lockable
+    /// resources.
+    subsystems: Vec<(String, usize)>,
+}
+
+impl ResourceLockManagerConfig {
+    /// Add resource subsystem to the config.
+    /// Panics if another subsystem with the same id already exists.
+    pub fn with_subsystem<T: AsRef<str>>(
+        mut self,
+        id: T,
+        num_objects: usize,
+    ) -> Self {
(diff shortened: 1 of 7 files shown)


mtzaurus:
bors try

mtzaurus:
bors try

mtzaurus:
bors try

mtzaurus:
bors try

mtzaurus:
bors try

io-engine/src/core/lock.rs
```diff
@@ -0,0 +1,170 @@
...
+impl ProtectedSubsystems {
+    pub const NEXUS: &'static str = "nexus";
+}
+
+/// Configuration parameters for initialization of the Lock manager.
+#[derive(Debug, Default)]
+pub struct ResourceLockManagerConfig {
+    /// Configs for subsystems: denote id and maximum amount of lockable
+    /// resources.
+    subsystems: Vec<(String, usize)>,
```
tiagolobocastro, line 21: nit: If `ProtectedSubsystems` is an enum instead, then here it can be the key by type.

io-engine/src/grpc/v0/mayastor_grpc.rs
```diff
@@ -132,6 +133,46 @@ impl MayastorSvc {
...
+        let _resource_guard = lock_manager
+            .get_subsystem(ProtectedSubsystems::NEXUS)
+            .lock_resource(nexus_uuid)
+            .await;
+        let r = fut.await;
+
+        match r {
+            Ok(r) => r,
+            Err(_e) => {
+                warn!("{}: gRPC method panicked, args: {}", ctx.id, ctx.args);
```
tiagolobocastro: hmm I wonder if we run fut as a non-cancellable future, or perhaps disable grpc timeout. Otherwise if we have a timeout we'll drop the lock again and open ourselves up to problem we're trying to solve because the mayastor "reactor" futures are not cancellable.

tiagolobocastro (commented):
Do we need to also add locks to internal nexus operations? Eg: if we get a grpc call and at the same time we're retiring a child and pausing the subsystem, etc?

mtzaurus:
bors try

mtzaurus:
bors try

mtzaurus:
bors try

mtzaurus:
bors try

mtzaurus:
bors try

tiagolobocastro (approved):

dsavitskiy (approved):

mtzaurus:
bors merge

PR #1239, opened 2022-11-02
fix(csi-node): don't unstage if bind mount exists

tiagolobocastro:
If an application creates a bind mount for our volume unstaging can be very troublesome as it
leaves the mount in a really bad state.
Even if we reconnect the device, it's very unlikely things won't recover.

This commit adds a check on node_unstage for unknown mounts (as in, not a csi publish mount).
Should there be any we output this as a trace and simply fail the unstage.
It's up to the user to clean up non-csi mounts, otherwise we can never unstage.

### DIFF
--- a/ci.nix
+++ b/ci.nix
@@ -3,7 +3,7 @@ let
   sources = import ./nix/sources.nix;
   pkgs = import sources.nixpkgs {
     overlays =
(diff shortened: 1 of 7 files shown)


mtzaurus (approved):

blaisedias (approved):
Looks plausible

tiagolobocastro:
bors merge

PR #1242, opened 2022-11-08
fix: remove host path from csi daemonset

tiagolobocastro:
This interferes with the "unknown" mount finder and we no longer need the host path anyway. todo: wa this even when such a thing happens


Abhinandan-Purkait (approved):

blaisedias (approved):
IIRC this was required to support iSCSI - and that is not supported anymore 
Associated with supporting iSCSI was the fact that the containers were privileged - should we consider/have we removed that as well

tiagolobocastro:
bors merge

PR #1243, opened 2022-11-09
fix(cli): mark uuid as an optional argument during

simrankureel:
Making uuid as an optional argument during nexus create.

Currently it is mandatory to pass a uuid during nexus create, so first developer has to create a uuid and then pass it during nexus create which is a two-step process.

With the help of this change the uuid will get autogenerated if it is not present.

Performed Tests (manually) :

Ran the standalone io-engine (data plane) binary, used io-engine client to interact with the binary.

- Created a nexus without specifying uuid.
- Created a nexus by specifying uuid and then crosschecked using bdev list which shows the same uuid used to create a nexus.

Signed-off-by: simrankureel <simrankurel@microsoft.com>

### DIFF
--- a/io-engine/src/bin/io-engine-client/nexus_cli.rs
+++ b/io-engine/src/bin/io-engine-client/nexus_cli.rs
@@ -11,6 +11,7 @@ use colored_json::ToColoredJson;
 use mayastor_api::{v0, v1};
 use snafu::ResultExt;
 use tonic::{Code, Status};
+use uuid::Uuid;
 
 pub fn subcommands<'a, 'b>() -> App<'a, 'b> {
     let create = SubCommand::with_name("create")
@@ -19,7 +20,7 @@ pub fn subcommands<'a, 'b>() -> App<'a, 'b> {
             Arg::with_name("uuid")
                 .required(true)
                 .index(1)
-                .help("uuid for the nexus"),
+                .help("uuid for the nexus, if uuid is not known please provide \"\" to autogenerate"),
         )
         .arg(
             Arg::with_name("size")
@@ -46,7 +47,7 @@ pub fn subcommands<'a, 'b>() -> App<'a, 'b> {
         .arg(
             Arg::with_name("uuid")
                 .required(true)
-                .help("uuid for the nexus"),
+                .help("uuid for the nexus, if uuid is not known please provide \"\" to autogenerate"),
         )
         .arg(
             Arg::with_name("size")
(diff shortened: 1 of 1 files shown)


tiagolobocastro:
Hi @simrankureel, IIRC the reason uuid is required atm is to be able to achieve idempotency.
Unfortunately you changes are they stand, seem to be a breaking change because you're changing positional arguments is that right?

simrankureel:
> Hi @simrankureel, IIRC the reason uuid is required atm is to be able to achieve idempotency. Unfortunately you changes are they stand, seem to be a breaking change because you're changing positional arguments is that right?

Hi @tiagolobocastro, thanks for taking a look at the pr. I am trying to make uuid as an optional field, so it is required to change the order of arguments. If we don't change the order of positional arguments, then it is not possible to make uuid as optional argument.

tiagolobocastro:
> Hi @tiagolobocastro, thanks for taking a look at the pr. I am trying to make uuid as an optional field, so it is required to change the order of arguments. If we don't change the order of positional arguments, then it is not possible to make uuid as optional argument.

This requires changing one doc and also we have to make sure our e2e tests don't use this command. You'd probably have the same issue again on create2 where we have now an extra parameter, name.
How about a compromise, if position parameter is `""` then autogenerate it?

simrankureel:
> > Hi @tiagolobocastro, thanks for taking a look at the pr. I am trying to make uuid as an optional field, so it is required to change the order of arguments. If we don't change the order of positional arguments, then it is not possible to make uuid as optional argument.
> 
> This requires changing one doc and also we have to make sure our e2e tests don't use this command. You'd probably have the same issue again on create2 where we have now an extra parameter, name. How about a compromise, if position parameter is `""` then autogenerate it?

Done
Thanks for the suggestion!

tiagolobocastro:
bors try

tiagolobocastro:
@simrankureel could you please squash your commits and have it s signed? Thanks

simrankureel:
> @simrankureel could you please squash your commits and have it s signed? Thanks

Done, please take a look.

tiagolobocastro (approved):

tiagolobocastro:
bors try

Abhinandan-Purkait (approved):

tiagolobocastro:
ah the cleanup script is broken.. raised a PR to try to fix it.. I've cleaned up the server manually for now.
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

tiagolobocastro:
@mtzaurus , seems we get persistent failures, not with the tests themselves, but seems mayastor does not exit gracefully but rather generates a coredump. It's always with the test: `persist_unexpected_restart`.

mtzaurus:
> * continuous-integration/jenkins/branch

Didn't this PR successfully pass 1 hour ago ?

mtzaurus:
Anyways, I will take a look at this failure.

tiagolobocastro:
@simrankureel please squash your commits again as I see a "Merge branch...". 
For future reference please update with rebase instead.

Sorry about the CI failures, after PR#1254 these failures should be resolved.

tiagolobocastro:
bors try

tiagolobocastro:
ok looks like CI is good, @simrankureel once you squash we should be good to merge :+1:

simrankureel:
> @simrankureel please squash your commits again as I see a "Merge branch...". For future reference please update with rebase instead.
> 
> Sorry about the CI failures, after PR#1254 these failures should be resolved.

Done!

tiagolobocastro:
bors merge

PR #1249, opened 2022-11-16
fix: re-enable ptpl and allow for not using reservations

tiagolobocastro:
chore: allow for not using reservations

This gives back control to the api client.
To do this, set the reservation parameters to "0".

---

fix(nvme-resv): setup ptpl from the caller side

Previously this was done incorrectly only when props was None.
It's better to leave it for the caller to decide.

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev.rs
@@ -1289,6 +1289,8 @@ async fn nexus_create_internal(
 
     match Nexus::register_instance(&mut nexus_bdev).await {
         Err(Error::NexusIncomplete {
+            name,
+            reason,
             ..
         }) => {
             // We still have code that waits for children to come online,
@@ -1302,7 +1304,8 @@ async fn nexus_create_internal(
                 let _ = device_destroy(child.uri()).await;
             }
             Err(Error::NexusCreate {
-                name: String::from(name),
+                name,
+                reason,
             })
         }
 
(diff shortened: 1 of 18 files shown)


io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -211,6 +211,12 @@ impl NexusNvmeParams {
     pub fn set_preempt_policy(&mut self, preempt_policy: NexusNvmePreemption) {
         self.preempt_policy = preempt_policy;
     }
+    /// Check if reservations are disabled.
+    pub fn disabled(&self) -> bool {
```
mtzaurus: Can we use here a name like 'reservations_disabled()' to denote that it deals with reservations ?
Also, as a nit, 'enabled()' looks a bit more natural to me than '.disabled()', but it just a suggestion here.

mtzaurus:
Can we use here a name like 'reservations_disabled()' to denote that it deals with reservations ?
Also, as a nit, 'enabled()' looks a bit more natural to me than '.disabled()', but it just a suggestion here.

mtzaurus (approved):

io-engine/src/bdev/nexus/nexus_bdev.rs
tiagolobocastro: Ah I thought the type name here was NvmeReservations which is why I left out reservations but looks like it's only NvmeParams, so I'll rename to reservations_enabled().

chriswldenyer (approved):
Looks reasonable

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1256, opened 2022-11-24
feat(reactor): support for destroying SPDK threads

dsavitskiy:
Reactors now remove and destroy exited SPDK threads. A check for long SPDK poll was added.

### DIFF
--- a/io-engine/src/core/reactor.rs
+++ b/io-engine/src/core/reactor.rs
@@ -180,10 +180,10 @@ impl Reactors {
             if unsafe { spdk_cpuset_get_cpu(mask, r.lcore) } {
                 let mt = spdk_rs::Thread::from_ptr(thread);
                 info!(
-                    "scheduled {} {:p} on core:{}",
+                    "Scheduled SPDK thread '{}' ({:p}) on core #{}",
                     mt.name(),
                     thread,
-                    r.lcore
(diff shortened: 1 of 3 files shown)


dsavitskiy:
bors try

io-engine/src/core/reactor.rs
```diff
@@ -497,19 +498,66 @@ impl Reactor {
...
         self.receive_futures();
         self.run_futures();
+
         drop(threads);
 
+        self.add_incoming();
+    }
+
+    /// Polls a thread, checking poll duration.
+    fn checked_poll(&self, t: &spdk_rs::Thread) {
```
tiagolobocastro: hmm do we really want to enable this by default? AFAIK getting the time will do a system call? If so this doesn't seem like something we want to do by default for every poll?

Looks like there's a tsc crate to get TSC, that might be better?

dsavitskiy: Ah, I forgot to remove it. Reactor freeze detector essentially does the same, but better.

dsavitskiy:
bors try

mtzaurus (approved):

tiagolobocastro (approved):

dsavitskiy:
bors merge

PR #1267, opened 2022-12-07
fix(io-engine): fixing heap issues

dsavitskiy:

### DIFF
--- a/io-engine/src/bdev/nexus/mod.rs
+++ b/io-engine/src/bdev/nexus/mod.rs
@@ -20,6 +20,7 @@ mod nexus_nbd;
 mod nexus_persistence;
 mod nexus_share;
 
+use crate::bdev::nexus::nexus_iter::NexusIterMut;
 pub(crate) use nexus_bdev::NEXUS_PRODUCT_ID;
 pub use nexus_bdev::{
     nexus_create,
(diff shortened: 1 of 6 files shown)


dsavitskiy:
bors try

io-engine/src/bdev/nexus/mod.rs
```diff
@@ -129,9 +130,14 @@ pub fn register_module() {
 
 /// called during shutdown so that all nexus children are in Destroying state
 /// so that a possible remove event from SPDK also results in bdev removal
+#[allow(clippy::needless_collect)]
 pub async fn shutdown_nexuses() {
     info!("Shutting down nexuses...");
-    for mut nexus in nexus_iter_mut() {
+
+    let nexuses: Vec<<NexusIterMut<'_> as Iterator>::Item> =
```
tiagolobocastro, line 143: Presumably you want to force collection ahead of the loop, rather than it being "lazy", is that why you've added the clip allow?

tiagolobocastro (approved):

io-engine/src/bdev/nexus/mod.rs
dsavitskiy: Yes.

dsavitskiy: Bdev are being deleted by SPDK asynchromously, in parallel, so the iterator becomes invalid.

mtzaurus (approved):

io-engine/src/bdev/nexus/mod.rs
tiagolobocastro: Might be worth adding a comment just in case someone tries to be more clippy friendly and doesn't know it's this way for a reason.

dsavitskiy: ok

dsavitskiy:
bors merge

PR #1281, opened 2023-01-10
refactor: shutdown grpc server and deregister on sigterm

Abhinandan-Purkait:
Signed-off-by: Abhinandan Purkait <purkaitabhinandan@gmail.com>

### DIFF
--- a/io-engine/src/core/env.rs
+++ b/io-engine/src/core/env.rs
@@ -49,6 +49,7 @@ use crate::{
         Mthread,
     },
     grpc,
+    grpc::MayastorGrpcServer,
     logger,
     persistent_store::PersistentStore,
     subsys::{
@@ -56,6 +57,7 @@ use crate::{
         registration::registration_grpc::ApiVersion,
         Config,
         PoolConfig,
+        Registration,
     },
 };
 
@@ -370,6 +372,14 @@ async fn do_shutdown(arg: *mut c_void) {
         warn!("Mayastor stopped non-zero: {}", rc);
     }
 
+    // Shutdown GRPC Server and Registration Client first, to not accept any
+    // more requests once we start shutting down, otherwise the control
+    // plane might schedule workloads on this instance while it's shutting
+    // down.
+    MayastorGrpcServer::get_or_init().fini();
+    if let Some(reg) = Registration::get() {
+        reg.fini();
+    }
     nexus::shutdown_nexuses().await;
     crate::lvs::Lvs::export_all().await;
     unsafe {
--- a/io-engine/src/subsys/registration/registration_grpc.rs
+++ b/io-engine/src/subsys/registration/registration_grpc.rs
@@ -145,8 +145,8 @@ impl Registration {
         GRPC_REGISTRATION.get()
     }
 
-    /// terminate and re-register
-    pub(super) fn fini(&self) {
+    /// Terminate the channel to deregister.
+    pub fn fini(&self) {
         self.fini_chan.close();
     }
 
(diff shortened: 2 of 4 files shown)


io-engine/src/subsys/registration/mod.rs
```diff
@@ -57,24 +54,13 @@ impl RegistrationSubsystem {
...
-    }
-
     fn new() -> Self {
         info!("creating Mayastor registration subsystem...");
         let mut ss = Box::new(spdk_subsystem::default());
         ss.name = b"mayastor_grpc_registration\x00" as *const u8
             as *const libc::c_char;
         ss.init = Some(Self::init);
-        ss.fini = Some(Self::fini);
+        ss.fini = None;
```
avishnu: Is this line needed?

avishnu (commented):
How did you test this?

io-engine/src/subsys/registration/registration_grpc.rs
```diff
@@ -66,10 +67,6 @@ pub struct Registration {
     config: Configuration,
     /// Registration client
     client: registration_client::RegistrationClient<tonic::transport::Channel>,
-    /// Receive channel for messages and termination
-    rcv_chan: async_channel::Receiver<()>,
-    /// Termination channel
-    fini_chan: async_channel::Sender<()>,
```
tiagolobocastro: Can we keep this and do the same for grpc?
Then everything can be triggered via the existing signal handler rather than spreading multiple signal handlers around which is looking a bit untidy.

Abhinandan-Purkait: We cannot control the sequence in that case. The fini seems to get called at the last, which we don't want.

Abhinandan-Purkait:
> How did you test this?

Checking the timing when the grpc server stops responding. Its same as the time of receiving of the sigterm, which wasn't the case earlier.

io-engine/src/subsys/registration/registration_grpc.rs
tiagolobocastro: We can simply call the inner of the ini directly from above

Abhinandan-Purkait: Yeah, we would need to do the same for grpc server then as well. Will do it.

Abhinandan-Purkait: Done.

io-engine/src/subsys/registration/mod.rs
Abhinandan-Purkait: Have reverted this, made the changes as discussed below.

io-engine/src/core/env.rs
```diff
@@ -370,6 +372,13 @@ async fn do_shutdown(arg: *mut c_void) {
         warn!("Mayastor stopped non-zero: {}", rc);
     }
 
+    // Shutdown GRPC Server first, to not accept any more requests once we
```
```suggestion
    // Shutdown GRPC Server and Registration Client first, to not accept any more requests once we
```

io-engine/src/core/env.rs
```diff
@@ -370,6 +372,13 @@ async fn do_shutdown(arg: *mut c_void) {
...
+    // start shutting down, otherwise the control plane might schedule workloads
+    // on this instance while its shutting down.
```
tiagolobocastro: s/its/it's

tiagolobocastro (approved):

avishnu (approved):
lgtm

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

PR #1291, opened 2023-01-23
feat: nexus: replica rebuild: Add replica rebuild history recording

dsharma-dc:
Added feature to record each rebuild job from all replicas of a nexus. The rebuild history is maintained till the lifetime of the nexus. For fetching the rebuild history, nexus uuid needs to be provided to the gRPC API.


io-engine/src/rebuild/rebuild_job.rs
```diff
@@ -74,6 +75,64 @@ impl std::fmt::Display for RebuildOperation {
     }
 }
 
+/// A rebuild record is a lightweight extract of rebuild job that is maintained
+/// for the statistics.
+pub struct RebuildRecord<'n> {
```
dsavitskiy: You don't need a lifetime here. Logically, rebuild record _can_ outlive a nexus.

io-engine/src/rebuild/rebuild_job.rs
```diff
@@ -74,6 +75,64 @@ impl std::fmt::Display for RebuildOperation {
...
+    /// End time of this rebuild
+    pub end: DateTime<Utc>,
+    // Unused lifetime parameter
+    _n: PhantomData<&'n ()>,
+}
+
+impl<'n> RebuildRecord<'n> {
+    /// Create a new rebuild history record when this child undergoes a rebuild
+    /// job.
+    pub fn new(rebuild_job: &RebuildJob) -> Result<Self, RebuildError> {
```
dsavitskiy: Why a Result return? Rebuild record creation should always succeed.

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -236,6 +243,14 @@ impl<'n> Nexus<'n> {
         Ok(rj.stats())
     }
 
+    /// Iterate over the replica rebuild history for this nexus
+    /// and return it as output.
+    pub async fn rebuild_history(
+        &self, // XXX: How to use Pin<&mut Self> and return vector reference?
+    ) -> Result<&Vec<RebuildRecord<'_>>, Error> {
```
dsavitskiy: No reason to have a Result return here. Nexus will always have a rebuild history, probably an empty one.

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -236,6 +243,14 @@ impl<'n> Nexus<'n> {
         Ok(rj.stats())
     }
 
+    /// Iterate over the replica rebuild history for this nexus
+    /// and return it as output.
+    pub async fn rebuild_history(
+        &self, // XXX: How to use Pin<&mut Self> and return vector reference?
```
dsavitskiy: You don't need a mut reference to Self (Pin<&mut Self>) for get-like methods.

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -349,6 +364,21 @@ impl<'n> Nexus<'n> {
         })
     }
 
+    /// Create a rebuild history record of current rebuild job and insert it
+    /// into Vector/Hashmap which is part of the Nexus. These history
+    /// records are kept as history data in-core, and not leveraged in data
+    /// path for rest of the rebuild logic.
+    fn create_rebuild_record(self: Pin<&mut Self>, job: &RebuildJob) {
```
dsavitskiy: I suggest to change the API so that RebuildJob returns a Record for itself. Since we create Record once in the end of rebuild, it makes sense to make a method that consumes a RebuildJob, returning a corresponding Record object.

dsavitskiy:
I would a history record to be created immediately when a Rebuild Job is started. Current status of such would be updated periodically, and finalized when the job finishes.

io-engine/src/grpc/v0/json_grpc.rs
```diff
@@ -36,9 +36,12 @@ impl JsonRpc for JsonRpcSvc {
     ) -> GrpcResult<JsonRpcReply> {
         let args = request.into_inner();
 
-        let result = self
+        let result = match self
             .spdk_jsonrpc_call(&args.method, empty_as_none(&args.params))
-            .await?;
+            .await {
+                Ok(val) => val,
+                Err(_) => todo!(),
```
dsharma-dc: @dsavitskiy @tiagolobocastro This line is modified to accommodate lack of implicit error conversion here coming up as a result of package version updates. But 'todo' doesn't look very correct here. Any suggestion to make it better?

dsharma-dc:
bors try

dsharma-dc:
bors try

io-engine/src/grpc/v0/json_grpc.rs
```diff
@@ -36,9 +36,13 @@ impl JsonRpc for JsonRpcSvc {
     ) -> GrpcResult<JsonRpcReply> {
         let args = request.into_inner();
 
-        let result = self
+        let result = match self
```
tiagolobocastro: I don't understand, why did error conversion stop working?

io-engine/src/grpc/v0/mayastor_grpc.rs
```diff
@@ -1665,6 +1692,38 @@ impl mayastor_server::Mayastor for MayastorSvc {
...
+    ) -> GrpcResult<RebuildHistoryReply> {
+        let ctx = GrpcClientContext::new(&request, function_name!());
+        let args = request.into_inner();
+
+        self.serialized(ctx, args.uuid.clone(), false, async move {
+            trace!("{:?}", args);
+            let rx = rpc_submit::<_, _, nexus::Error>(async move {
+                let records = nexus_lookup(&args.uuid)?
+                    .rebuild_history()
+                    .unwrap()
```
tiagolobocastro: please try to never unwrap unless it's required not to return error and you're absolutely it can never happen.

io-engine/src/grpc/v0/mayastor_grpc.rs
```diff
@@ -1665,6 +1692,38 @@ impl mayastor_server::Mayastor for MayastorSvc {
...
+                    .iter()
+                    .map(RebuildHistoryRecord::from)
+                    .collect();
+                Ok(RebuildHistoryReply {
+                    nexus: String::from(
+                        nexus_lookup(&args.uuid)?.uuid().to_string(),
```
tiagolobocastro: I'm confused, why not simply: `nexus: args.uuid.clone()` ?

io-engine/src/rebuild/rebuild_job.rs
```diff
@@ -292,6 +316,27 @@ impl<'n> RebuildJob<'n> {
             .collect()
     }
 
+    /// Create a RebuildRecord from RebuildJob
```
tiagolobocastro: In rust these are typically done using the From trait, so you'd impl From<RebuildJob> for RebuildRecord.

io-engine/src/grpc/v0/json_grpc.rs
dsharma-dc: Ok, so it was throwing build error about 'From' trait not being able to convert here. But I just tried it again in original form and it builds, so most likely I must have had some conflicting toml files when I ran into that error. I'll revert this file's change.

io-engine/src/grpc/v0/mayastor_grpc.rs
dsharma-dc: Yes I was conscious about that, but here it's ok as in the case of no records available the returned vector will be empty but valid anyway.

io-engine/src/grpc/v0/mayastor_grpc.rs
dsharma-dc: changed.

io-engine/src/rebuild/rebuild_job.rs
dsharma-dc: ok right, will modify this.

io-engine/src/grpc/v0/mayastor_grpc.rs
tiagolobocastro: Then why does `rebuild_history` return a Result at all, can simply return `&Vec<RebuildRecord>`

dsharma-dc: That's reasonable, removed Result as return type.

dsharma-dc:
Addressed review comments from @tiagolobocastro

dsharma-dc:
bors try

dsavitskiy (approved):

io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -257,6 +258,8 @@ pub struct Nexus<'n> {
     io_subsystem: Option<NexusIoSubsystem<'n>>,
     /// TODO
     event_sink: Option<DeviceEventSink>,
+    /// Rebuild history of all children of this nexus instance.
+    pub rebuild_history: Vec<RebuildRecord>,
```
tiagolobocastro: This doesn't need to be pub, and we've a getter anyway?

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -418,6 +430,14 @@ impl<'n> Nexus<'n> {
         }
 
         self.reconfigure(DrEvent::ChildRebuild).await;
+        // Translate the job into a new rebuild record and push into history.
+        trace!(
+            "Create Rebuild Record for child {} of nexus {}",
+            job.dst_uri,
+            job.nexus_name
+        );
+        let hist = unsafe { &mut self.get_unchecked_mut().rebuild_history };
```
tiagolobocastro: why is this needed?

dsavitskiy: That's because of how Pin<> works. You cannot that easily get write access to fields of a pinned object. If you could, you could easily moved out values from it. So a pinned object can only be written in an unsafe block, thus showing a potential place violating Pin<>'s restrictions.

dsavitskiy:
@dsharma-dc I'd also suggest squashing three commits into single one. Your commits are not independent changes, but rather one commit is a fix for a previous one.

tiagolobocastro (approved):

dsharma-dc:
bors try

dsharma-dc:
bors merge

PR #1292, opened 2023-01-23
refactor: remove unsafe constructs from rebuild

tiagolobocastro:
The previous rebuild code was making some assumptions about the way the rebuild job runs. The code is ever run in the master reactor, so there can never be two tasks concurrently touching the rebuild job - however with time seems in certain cases between awaits the rebuild job may be reused which is likely bad and could explain some weird issues we've seen with rebuild as of late.

This change removes the unsoundness of the rebuild at the cost of a few locks and channels. However these probably won't affect the rebuild performance much as they will be largely uncontended.
The rebuild job is kind of split into frontend and backend. The backend is where the rebuild actually happens and we have a channel used to communicate between frontend and backend.

This change also allows to, in the future, have dedicated cores just for rebuilds for example.

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
@@ -1,6 +1,6 @@
 use futures::channel::oneshot::Receiver;
 use snafu::ResultExt;
-use std::{marker::PhantomData, pin::Pin};
+use std::{marker::PhantomData, pin::Pin, sync::Arc};
 
 use super::{
     nexus_err,
@@ -75,11 +75,11 @@ impl<'n> Nexus<'n> {
     /// Starts a rebuild job and returns a receiver channel
     /// which can be used to await the rebuild completion
     pub async fn start_rebuild(
-        mut self: Pin<&mut Self>,
+        self: Pin<&mut Self>,
         child_uri: &str,
     ) -> Result<Receiver<RebuildState>, Error> {
         let name = self.name.clone();
-        trace!("{}: start rebuild request for {}", name, child_uri);
+        trace!("{name}: start rebuild request for {child_uri}");
 
         // Find a healthy child to rebuild from.
         let src_child_uri = match self
@@ -114,8 +114,7 @@ impl<'n> Nexus<'n> {
             }),
         }?;
 
-        self.as_mut()
-            .create_rebuild_job(&src_child_uri, &dst_child_uri)
+        self.create_rebuild_job(&src_child_uri, &dst_child_uri)
             .await?;
 
         // We're now rebuilding the `dst_child` which means it HAS to become an
@@ -138,7 +137,7 @@ impl<'n> Nexus<'n> {
 
     /// TODO
     async fn create_rebuild_job(
-        self: Pin<&mut Self>,
+        &self,
         src_child_uri: &str,
         dst_child_uri: &str,
     ) -> Result<(), Error> {
@@ -165,7 +164,7 @@ impl<'n> Nexus<'n> {
     }
 
     // Translate the job into a new rebuild record and push into history.
-    fn create_rebuild_record(self: Pin<&mut Self>, job: RebuildJob) {
+    fn create_rebuild_record(self: Pin<&mut Self>, job: Arc<RebuildJob>) {
         trace!(
             "Create Rebuild Record for child {} of nexus {}",
             job.dst_uri,
@@ -175,10 +174,10 @@ impl<'n> Nexus<'n> {
         hist.push(job.into());
     }
 
-    /// Terminates a rebuild in the background
-    /// used for shutdown operations and
+    /// Terminates a rebuild in the background.
+    /// Used for shutdown operations and
     /// unlike the client operation stop, this command does not fail
-    /// as it overrides the previous client operations
+    /// as it overrides the previous client operations.
     async fn terminate_rebuild(self: Pin<&mut Self>, child_uri: &str) {
         // If a rebuild job is not found that's ok
         // as we were just going to remove it anyway.
@@ -186,7 +185,7 @@ impl<'n> Nexus<'n> {
             let ch = rj.terminate();
             if let Err(e) = ch.await {
                 error!(
-                    "Failed to wait on rebuild job for child {} to terminate with error {}", child_uri,
+                    "Failed to wait on rebuild job for child {child_uri} to terminate with error {}",
                     e.verbose()
                 );
             }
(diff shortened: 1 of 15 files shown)


io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -131,10 +130,15 @@ impl<'n> Nexus<'n> {
...
-        self: Pin<&mut Self>,
+        &self,
         src_child_uri: &str,
         dst_child_uri: &str,
     ) -> Result<(), Error> {
+        let range = std::ops::Range::<u64> {
+            start: self.data_ent_offset,
+            end: self.num_blocks() + self.data_ent_offset,
+        };
+        tracing::info!("Nexus Range (in blocks): {:?}", range);
```
dsharma-dc: This trace may not be very helpful/relevant in case of partial rebuilds as that won't be dealing with ranges always.

io-engine/src/rebuild/rebuild_descriptor.rs
```diff
@@ -0,0 +1,65 @@
...
+use super::rebuild_error::*;
+
+pub(super) struct RebuildDescriptor {
+    /// Name of the nexus associated with the rebuild job.
+    pub(super) nexus_name: String,
+    /// The block size of the src and dst.
+    /// todo: allow for differences?
+    pub(super) block_size: u64,
+    /// The range of the entire rebuild.
+    pub(super) range: std::ops::Range<u64>,
```
dsharma-dc, line 13: RebuildDescriptor will also have to take in account that rebuild may not be dealing with a range, for partial rebuilds.

io-engine/src/rebuild/rebuild_job_backend.rs
```diff
@@ -0,0 +1,514 @@
...
+    /// current state of the rebuild job
+    pub(super) states: std::sync::Arc<parking_lot::RwLock<RebuildStates>>,
+    /// channel list which allows the await of the rebuild
+    pub(super) complete_chan:
+        std::sync::Arc<parking_lot::Mutex<Vec<oneshot::Sender<RebuildState>>>>,
+    /// rebuild copy error, if any
+    pub(super) error: Option<RebuildError>,
+    /// Channel to share information between frontend and backend.
+    pub(super) info_chan: RebuildClientChan,
+    /// All the rebuild related descriptors.
```
dsharma-dc, line 105: Isn't there a 1-1 mapping between frontend and backend? The comment says 'All'.

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -132,10 +131,15 @@ impl<'n> Nexus<'n> {
...
```
dsavitskiy: Could you add child or nexus info to the trace so one can understand which nexus is the message about?

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -314,22 +315,21 @@ impl<'n> Nexus<'n> {
...
     }
 
     /// Returns rebuild job associated with the destination child URI.
     /// Returns error if no rebuild job associated with it.
     pub fn rebuild_job_mut(
-        self: Pin<&mut Self>,
+        &self,
         dst_child_uri: &str,
-    ) -> Result<&mut RebuildJob<'n>, Error> {
+    ) -> Result<std::sync::Arc<RebuildJob>, Error> {
```
dsavitskiy: Probably just `Arc` as we do in other places, to reduct visual noise?

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -412,11 +414,18 @@ impl<'n> Nexus<'n> {
...
                     child_uri,
                     &self.name,
-                    job.state(),
+                    job_state,
                 );
             }
         }
 
+        let dst_child = self.as_mut().child_mut(child_uri)?;
+        // Should this be done only after reconfigure?
```
dsavitskiy: Could you add TODO marker, if this problem to be addressed?

io-engine/src/rebuild/mod.rs
```diff
@@ -1,10 +1,15 @@
...
 mod rebuild_job;
+mod rebuild_job_backend;
 mod rebuild_state;
 mod rebuild_task;
 
+use rebuild_descriptor::RebuildDescriptor;
 pub use rebuild_error::RebuildError;
 pub use rebuild_job::RebuildJob;
+use rebuild_job::RebuildOperation;
+use rebuild_job_backend::*;
```
dsavitskiy: Rather than *, better to import only those symbols which are actually used.

io-engine/src/rebuild/rebuild_error.rs
```diff
@@ -60,4 +61,10 @@ pub enum RebuildError {
     },
     #[snafu(display("Failed to get bdev name from URI {}", uri))]
     BdevInvalidUri { source: BdevError, uri: String },
+    #[snafu(display("The rebuild frontend has been dropped"))]
```
dsavitskiy, line 64: Do we want to expose these rather implementation-specific errors to user?

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
tiagolobocastro: Oh sorry this was for debugging, I'll remove it!

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
tiagolobocastro: Yup, done.

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
tiagolobocastro: Added todo as well with a more in-depth comment, ptal

io-engine/src/rebuild/mod.rs
tiagolobocastro: Yes indeed, done.

io-engine/src/rebuild/rebuild_error.rs
tiagolobocastro: Hmm, that error should not be exposed to the user but only to the nexus::Error which has its own rebuild errors? although it may be logged to stdout.

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
tiagolobocastro: That was a debugging, it's removed now.

io-engine/src/rebuild/rebuild_descriptor.rs
tiagolobocastro: Yes, that should be done as part of partial rebuild work.

io-engine/src/rebuild/rebuild_job_backend.rs
tiagolobocastro: All the rebuild descriptions yes, src, dst & nexus descriptors.

dsharma-dc (approved):

dsavitskiy (approved):

mtzaurus (approved):

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

PR #1293, opened 2023-01-23
Partial rebuild

dsavitskiy:


dsavitskiy:
bors try

mtzaurus:
Can we add some scenario-based BDDs ? That would allow us to better understand the behaviour of this feature.

io-engine/src/rebuild/rebuild_log.rs
```diff
@@ -0,0 +1,124 @@
...
+            self.num_blocks,
+            self.block_len
+        )
+    }
+}
+
+impl RebuildLog {
+    /// Creates a new rebuild log instance for the given device.
+    pub fn new(device_name: &str, num_blocks: u64, block_len: u64) -> Self {
+        let num_segments = div_ceil(num_blocks * block_len, SEGMENT_SIZE);
```
dsharma-dc, line 46: Can we avoid this calculation partly every time by simply using size_in_bytes from BlockDevice trait? Or perhaps keeping number of segments as a field in BlockDevice?

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -8,13 +8,17 @@ use std::{
...
+    core::{BlockDeviceHandle, CoreError, Cores},
+    rebuild::RebuildLogHandle,
+};
 
 /// io channel, per core
 #[repr(C)]
 pub struct NexusChannel<'n> {
     writers: Vec<Box<dyn BlockDeviceHandle>>,
     readers: Vec<Box<dyn BlockDeviceHandle>>,
+    rebuild_logs: Vec<RebuildLogHandle>,
```
dsharma-dc, line 21: How about making this a Hashmap instead of vector, so that attach logic doesn't need to iterate every time to match device_name?

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -218,13 +247,45 @@ impl<'n> NexusChannel<'n> {
             }
         }
 
+        // add all active rebuild logs.
```
dsharma-dc, line 255: Is it possible to cleanly merge this block with above block. The invariant being:
is_logging() implies is_rebuilding() but not other way round?

io-engine/src/rebuild/rebuild_log.rs
dsavitskiy: There's no problem calculating it here, as it is executed only once when a log is created. Even on I/O path, a single division won't hurt: our code is not CPU bound but rather I/O bound.
From the design perspective, "segment" is rebuild's job thing, block devices know nothing about how rebuild algorithm does its copy job.

io-engine/src/bdev/nexus/nexus_channel.rs
dsavitskiy: Probably yes, I'll check.

io-engine/src/bdev/nexus/nexus_channel.rs
dsavitskiy: We don't have that many children (1-3 mostly), so having a hash map or btree won't have any benefit. From calculation and memory allocation and access standpoint, HashMap and BTree would actually be worse here. For long-running applications, it's better to avoid having non-local data structures like linked list or trees (that's why Rust encourages Vec data structure in many cases).

Another problem with HashMap (as opposed to BTree) is that it doesn't not guarantee item order, which can change from run to run. That we'd like to avoid in I/O path.

io-engine/src/rebuild/rebuild_log.rs
tiagolobocastro: We're already cloning a string anyway, but I'm not sure it will always be I/O bound as we're talking about nvme and you could have multiple pools.
But anyway this is only for the creation so it won't happen very often.

io-engine/src/bdev/nexus/nexus_channel.rs
dsharma-dc: ok. What does "non-local data structures" mean here? Something in Rust parlance?

dsavitskiy:
bors try

io-engine/src/bdev/nexus/nexus_channel.rs
dsavitskiy: No, this term is sometimes used to refer to how a data structure is stored memory.

For example, a Vec<> or C++ std::vector<>, or C/C++ array allocate a large, continuous block of memory, storing all element one after another. When you iterate items, CPU reads from roughly the same region, thus with very high number of cache hits. Your next items is very likely to be already prefetched in CPU cache.

A B-tree or linked list allocate a small chuck of memory for each node (data item + pointers to other nodes). When you iterate such structure, CPU has to read from different regions of memory. This has two sonquences:
1) Each element must be accessed by a pointer dereference, which isn't quite friendly to CPU cache.
2) Every next item may lie in a different page, unpredictable by caching CPU logic. This pollutes the CPU cache with data you won't access very soon.
These two things lead to situation when CPU cache may become very ineffective, and performance of a CPU-bound program deteriorates.

That's one of the main contributing factors of relatively poor performance of GC languages like Java. Not the interpreted byte-code as some believe (it is actually compiled of-the-fly into CPU instructions), but the fact there's no memory-local data structures in Java. ArrayList is a vector of points to items, each allocate in a different memory location.

io-engine/src/rebuild/rebuild_log.rs
dsavitskiy: Division on modern CPUs is nothing, unless you have to process 8K 60fps video in realtime :-)

dsavitskiy:
@mtzaurus Yes we probably can add some BDD. Probably, it is better to have separate ticket for that?

io-engine/src/bdev/nexus/nexus_channel.rs
dsharma-dc: Alright, we are talking about 'locality of reference' here :)

io-engine/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -330,36 +332,44 @@ impl<'n> Nexus<'n> {
...
-            Ok(child) => {
-                child.offline().await;
-                self.reconfigure(DrEvent::ChildOffline).await;
-                Ok(self.status())
-            }
-            Err(e) => Err(e),
-        };
+        let dev_name = self.get_child_device_name(child_uri)?;
+
+        // TODO: check to prevent the last child from degrading (?).
```
tiagolobocastro: hmm Interesting, for sure we don't want to fault the last healthy child in etcd, though maybe that's handled elsewhere, but we should check.

io-engine/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -450,6 +433,11 @@ impl<'n> Nexus<'n> {
 
         let child = self.as_mut().child_mut(child_uri)?;
 
+        if child.state() == ChildState::Open {
+            warn!("{:?}: child already online", child);
```
tiagolobocastro: Should we check if we need to start a rebuild? Eg: what if `start_rebuild` below fails to start? The child is left as open but there's no rebuild in progress?

io-engine/src/bdev/nexus/nexus_io.rs
```diff
@@ -604,24 +654,21 @@ impl<'n> NexusBio<'n> {
...
-    fn retire_device(
+    /// Faults the device by its name, with the given I/O error.
+    /// The faulted device is scheduled to be retired.
+    fn fault_device(
         &mut self,
         child_device: &str,
         io_status: IoCompletionStatus,
     ) {
-        let reason = match io_status {
+        let s = match io_status {
```
tiagolobocastro: any reason for rename, s seems less descriptive than reason.

io-engine/src/grpc/v0/nexus_grpc.rs
```diff
@@ -38,14 +38,9 @@ impl From<&NexusChild<'_>> for rpc::ChildState {
...
-                Reason::NoSpace => rpc::ChildState::ChildDegraded,
-                Reason::TimedOut => rpc::ChildState::ChildDegraded,
-                Reason::Unknown => rpc::ChildState::ChildFaulted,
-                Reason::CantOpen => rpc::ChildState::ChildFaulted,
-                Reason::RebuildFailed => rpc::ChildState::ChildFaulted,
-                Reason::IoError => rpc::ChildState::ChildFaulted,
-                Reason::ByClient => rpc::ChildState::ChildFaulted,
-                Reason::AdminCommandFailed => rpc::ChildState::ChildFaulted,
+                FaultReason::IoError => rpc::ChildState::ChildFaulted,
+                s if s.is_recoverable() => rpc::ChildState::ChildDegraded,
```
tiagolobocastro: How does control-plane know when it can/has to online a Child in order to make it healthy again?
We should document this well, and in the protobuf too, otherwise we risk a child being degraded forever, because common control-plane logic is to leave degraded children alone.

io-engine/src/rebuild/rebuild_task.rs
```diff
@@ -124,8 +127,14 @@ impl RebuildTask {
             &mut copy_buffer
         };
 
-        source_hdl.set_read_mode(ReadMode::UnwrittenFail);
+        // Check if the block has to be copied.
+        if let Some(ref log) = self.rebuild_log {
```
tiagolobocastro: So seems like we still go through entire range and simply skip non-modified segments.
Would it be better to simply "walk the log" ?

io-engine/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -739,59 +749,53 @@ fn update_failfast_done(
...
+        if let Some(c) = self.as_mut().lookup_child_device_mut(child_device) {
+            if Ok(ChildState::Open)
+                == c.state.compare_exchange(
+                    ChildState::Open,
+                    ChildState::Faulted(reason),
+                )
+            {
+                warn!("{:?}: faulted with {}, will retire", c, reason);
+
+                if let Some(log) = c.start_rebuild_log() {
```
tiagolobocastro: Is there a race condition here? If we have >1 concurrent retires, the first retire "wins" and sets state to faulted. But before we attach the rebuild log the other retires would return. Could there be a situation where we have newer IO's before we attach the log?

dsavitskiy:
bors try

dsavitskiy:
bors try

io-engine/src/rebuild/rebuild_task.rs
dsavitskiy: My intention was to preserve the original algorithm, so we can say that it is valid. Just one "if" with obvious behavior won't brake it.

I don't see any performance issues here. The entire process is heavily I/O bound.

If I had to rewrite the logic to walk the log, we would end up with entire new code, with no real-world performance gain.

io-engine/src/grpc/v0/nexus_grpc.rs
dsavitskiy: Will do on this ticket: https://dcsw-bolt.atlassian.net/browse/GTM-469

io-engine/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -739,66 +773,64 @@ fn update_failfast_done(
...
-        if need_retire {
-            Reactors::master().send_future(Nexus::child_retire_routine(
-                self.name.clone(),
-                child_device.to_owned(),
-                retry,
-            ));
+        let name = self.name.clone();
+
+        if let Some(c) = self.as_mut().lookup_child_by_device_mut(child_device)
+        {
```
felix1209, line 786: just wondering, single a letter variable name is considered ok and RUST specific? Normally in C/C++, we force people to avoid single-letter vars since it makes it harder to maintain over some time

io-engine/src/rebuild/rebuild_log.rs
```diff
@@ -0,0 +1,135 @@
...
+            num_blocks,
+            block_len,
+        }
+    }
+
+    /// Logs the given operation, marking the corresponding segment as modified
+    /// in the case of a write operation.
+    fn log_op(&mut self, _io_type: IoType, block_offset: u64, num_blocks: u64) {
+        let start_seg = self.segment_index(block_offset);
+        let end_seg = self.segment_index(block_offset + num_blocks);
```
felix1209: should this be like this? Am I missing something?
        let end_seg = self.segment_index(block_offset + num_blocks*block_len);

io-engine/src/rebuild/rebuild_log.rs
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+
+    /// Determines if the given block was modified by a write operation.
+    fn is_modified(&self, block_offset: u64) -> bool {
+        let seg = self.segment_index(block_offset);
+        if seg < self.segments.len() {
+            self.segments[seg]
+        } else {
+            true
```
felix1209, line 89: why do we get this condition? nexus size increased after the log got created?

io-engine/src/rebuild/rebuild_log.rs
felix1209: ah, block_offset is not actually an offset in bytes. It is actually the block number. you may want to consider renaming this as lbn (logical block number)

io-engine/src/rebuild/rebuild_log.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    /// Counts the total number of modified blocks.
+    pub fn count_modified_blocks(&self) -> u64 {
+        let segs = self.segments.iter().filter(|i| *i).count() as u64;
+        segs * SEGMENT_SIZE / self.block_len
+    }
+}
+
+/// A handle object for a rebuild log instance.
+/// A rebuild log instance allows to have multiple handle to access it.
```
felix1209, line 120: This is part of my RUST learning. Can you elaborate on this comment? Why do we need multiple instances? Even if it needs multiple instances can't that be handled without having this wrapper?

io-engine/src/bdev/nexus/nexus_io.rs
```diff
@@ -498,15 +539,24 @@ impl<'n> NexusBio<'n> {
...
                 IoCompletionStatus::IoSubmissionError(
-                    IoSubmissionFailure::Read,
+                    IoSubmissionFailure::Write,
                 ),
             );
         }
 
-        // partial submission
+        // Log all write-like operation in the rebuild logs, if any exist.
+        self.channel().for_each_rebuild_log(|log| {
```
felix1209: looks like submit_all function is called for  the following IoType,
           IoType::Write
            | IoType::WriteZeros
            | IoType::Reset
            | IoType::Unmap
Hence we should log the IoType as well along with the updated or not? Even if we don't support unmap, for now, we should have the provision in the design.

io-engine/src/bdev/nexus/nexus_io.rs
dsavitskiy: will revert

io-engine/src/bdev/nexus/nexus_bdev_children.rs
dsavitskiy: That's matter of convention of a particular team or project. Oftentimes, even if a convention requires elaborated names, a single letter vars are okay for a scope of a "for" loop, like the index; or a similar kind of var for "if" scope like here.

But yes indeed, Rust conventions in general prefer shorter names than Java or C++. Rust does automatic type deduction in many cases when C/C++ does not. Thus it is impractical to encode type in var name like some C/C++ conventions do.

Also,  Rust IDEs are much more powerful than C/C++, and editors can be configured to show a small label showing data type for each var.

io-engine/src/rebuild/rebuild_log.rs
dsavitskiy: It's just a safety check. We're not supposed to get this condition normally, but rebuild copy logic is decoupled from rebuild log logic and I don't want a hard assert here because of that.

io-engine/src/bdev/nexus/nexus_io.rs
dsavitskiy: I do pass IoType to the rebuild log, if you're asking that.

io-engine/src/rebuild/rebuild_log.rs
dsavitskiy: Each rebuild log is referred from multiple places, that is, we have many writable pointers to the same object.

In C/C++ this is not considered an issue; in C++ we also have std::shared_ptr for ref counted pointers for automatic object delete.

But Rust does not allow multiple writable pointers to the same object by language design.

As C++, Rust has a shared_ptr called "Arc". But it is immutable. If you want to have many _writable_ pointers, you have to protect the object with a kind of a mutex, e.g. Arc<Mutex<Data>>, and lock/unlock mutex every time you access the object. It's not a recommended practice, it is required by Rust language: you cannot get a mutable reference from Arc without putting the object into a kind of mutable access "provider" like Mutex.

In this case, I want to hide how exactly the rebuild log is shared and protected, so I wrapped Arc<Mutex<RebuildLog>> with "RebuildLogHandle". I could just pass Arc<Mutex<RebuildLog>> around instead, but I want to have some flexibility in rebuild log internals.

io-engine/src/bdev/nexus/nexus_bdev_children.rs
dsavitskiy: That's why I traverse the channels and attach the log immediately.

io-engine/src/bdev/nexus/nexus_bdev_children.rs
dsavitskiy: Yes indeed, if start_rebuild fails, I must close the child. Will fix.

io-engine/src/bdev/nexus/nexus_channel.rs
dsavitskiy: This could be rewritten in a single "for" block in "imperative" style, but here we have kinda "functional" style which utilizes ideas of iterators, filters and maps. With functional style, you are often forced to do one task at a time.

Generally, for non-async code, functional style is preferred in Rust, similarly to Python comprehensions. It may require multiple re-iterations in certain cases like here, but for non-CPU critical code it is okay.

The benefit of functional style is clearer separation of logic and tasks (once you get used to it, initially it can be intimidating, especially if language does not provide special functional syntax, like in Python or Haskell).

Also, in certain situations, Rust's borrow checker prevents an easy use of "for" loops other mutable collections.

dsavitskiy:
bors try

io-engine/src/rebuild/rebuild_task.rs
tiagolobocastro: I take your point, but couldn't we simply change `send_segment_task` to walk the log instead of walking the range? This wouldn't be a change of the algorithm, but perhaps I'm oversimplifying it :)

io-engine/src/bdev/nexus/nexus_bdev_children.rs
tiagolobocastro: Just to clarify, so you mean there's no chance newer IO's will happen before we attach the log?

dsavitskiy:
bors try

io-engine/src/rebuild/rebuild_log.rs
dsharma-dc: nit: I'd also suggest renaming num_blocks parameter to something like op_blocks_cnt that would indicate number of blocks affected by this op, since there is a field num_blocks already in the RebuildLog that indicates device size as number of  blocks.

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -114,7 +115,11 @@ impl<'n> Nexus<'n> {
             }),
         }?;
 
-        self.create_rebuild_job(&src_child_uri, &dst_child_uri)
+        // Get child' rebuild log.
+        let log = self.lookup_child(child_uri).and_then(|c| c.rebuild_log());
```
dsharma-dc, line 119: If we are in start_rebuild path via add_child call, then I believe we shouldn't find the rebuild log here. If we come via online_child path we'd find it. Can we internally call `drop_rebuild_log` as soon as we return the log handle here? Instead of few lines down?

io-engine/src/rebuild/rebuild_task.rs
dsharma-dc: One thing I guess we need to ensure is that we don't "walk the log" without locking the lba range, which is handled by plugging into existing logic.

tiagolobocastro: that is handled by each copy task, so that would not have to change when "walking the log".

io-engine/src/bdev/nexus/nexus_bdev_children.rs
dsharma-dc: Sounds like a possible race. If I understand correctly:

```
  cpu0	                               cpu1
  ———	                               ———

child_retire(Child1)
                                     child_retire(Child1)
Success = set_state_faulted
                                     is_already_faulted - return
start_rebuild_log()
attach_log_to_iochannel_of_child1()
                                      new_write_op(W)
                                      attach_log_to_iochannel_of_child1()
```

The rebuild_log will miss logging `W`.

dsavitskiy:
bors try

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
dsavitskiy: Log must be dropped after "self.reconfigure(DrEvent::ChildRebuild).await" so that channels have a change to get the log. Otherwise, during reconfiguration, channels will have no log for the child, and some I/Os may potentially miss.

io-engine/src/rebuild/rebuild_log.rs
dsavitskiy: Ok, I renamed arg names for clarity.

io-engine-tests/src/file_io.rs
```diff
@@ -13,22 +13,41 @@ pub fn set_test_buf_rng_seed(seed: u64) {
...
+    Blocks(usize),
+    Mb(usize),
+    Kb(usize),
+}
+
+impl BufferSize {
+    pub fn size_bytes(&self) -> usize {
+        match &self {
+            BufferSize::Bytes(s) => *s,
+            BufferSize::Blocks(s) => s * 512,
```
mtzaurus: Just curious: can't we have devices with 4K blocks ?

io-engine/src/bdev/nexus/nexus_bdev_children.rs
dsavitskiy: I added a comment about this situation on the most recent version of my code.

(W) won't really miss. There two possibilities:
1. (W) goes okay. Data goes to the device, nothing missed, no need to log actually.
2. (W) fails. In this case, It goes down the retire path again but on the different channel, and gets the log (very same log created before), and logs the op.

Also, any failed write I/O is resubmitted if any child fails to write, until all children confirm the write. Failed children are removed in the process, so a write I/O succeeds (or all children are gone).
Every resubmission means the same op will be logged again.

In the most recent change, I eliminated two races I had before:
1) create a new /get an existing log on a child
2) Logging an op on a repeated retire.

io-engine/src/rebuild/rebuild_task.rs
dsavitskiy: We can address the way tasks are walked in as a separate task, if the current implementation (walk the range) turns out to have performance problems.

io-engine-tests/src/file_io.rs
tiagolobocastro: I think so yes, if not then I'd consider it a bug :)

io-engine/src/rebuild/rebuild_log.rs
```diff
@@ -0,0 +1,158 @@
...
+fn div_ceil(a: u64, b: u64) -> u64 {
+    (a + b - 1) / b
+}
+
+/// Rebuild log data structure.
+pub struct RebuildLog {
+    /// Name of the underlying device, stored for diagnostic purposes.
+    device_name: String,
+    /// Segment bitmap. Ones indicate modified segments.
+    segments: BitVec,
```
gila, line 21: Hi! @dsavitskiy I Hope you are well!

I was wondering if you have considered using a bitmap of dirty blocks rather than dirty segments as it avoids calculations for each write operation. When you want to replay the log or whatever, you can "pay the price" of blocks to segments then?

io-engine/Cargo.toml
```diff
@@ -8,6 +8,7 @@ default-run = "io-engine-client"
 build = "build.rs"
 
 [features]
+default = ["fault_injection"]
```
tiagolobocastro: Do we want this built by default?

io-engine/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -815,37 +816,70 @@ impl<'n> Nexus<'n> {
         child_device: &str,
         reason: FaultReason,
         retry: bool,
-    ) {
+    ) -> Option<RebuildLogHandle> {
         let name = self.name.clone();
 
-        if let Some(c) = self.as_mut().lookup_child_by_device_mut(child_device)
+        let Some(c) = self.as_mut().lookup_child_by_device_mut(child_device) else {
```
tiagolobocastro, line 766: Ah I thought let else was still unstable, cool then maybe we can now also use this in control-plane

io-engine/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -815,37 +816,70 @@ impl<'n> Nexus<'n> {
...
+                name,
+                child_device.to_owned(),
+                retry,
+            ));
         } else {
-            error!(
-                "{:?}: trying to retire a child device which \
-                        cannot be not found '{}'",
-                self, child_device
+            warn!("{c:?}: faulted with {reason}, already retired");
```
tiagolobocastro: If we've already retired, then shouldn't we return here?

io-engine/src/rebuild/rebuild_log.rs
```diff
@@ -0,0 +1,158 @@
...
+impl Debug for RebuildLogHandle {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        self.log.lock().fmt(f)
+    }
+}
+
+impl RebuildLogHandle {
+    /// Returns the device name of the rebuild log.
+    pub fn device_name(&self) -> String {
+        self.log.lock().device_name.clone()
```
gila: In effect, this puts a mutex on all IO which is something you may want to avoid straight from the start. Instead, it perhaps would be a better idea a bit vector per core?

io-engine/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -815,37 +816,70 @@ impl<'n> Nexus<'n> {
...
+        // If the parallel write operation fails, `retire_child_device` is
+        // called anyway, and that operation ends up logged.
+        // However, in this case, log attach will be called again,
+        // but that's okay since the channel checks if the same log is attached.
+        //
+        // Having a critical section covering log attachment would make the code
+        // validity easier to reason about, but that isn't practical as
+        // it would require a lock in I/O code path, and a complicated
+        // syncronization with SPDK-driven `traverse_io_channels`.
+        if let Some(ref log) = log {
```
tiagolobocastro, line 823: What if the retire happens before we run this piece of code, won't we miss IOs ?

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -127,6 +132,11 @@ impl<'n> Nexus<'n> {
         // rebuilt ranges in sync with the other children.
         self.reconfigure(DrEvent::ChildRebuild).await;
 
+        // Drop rebuild log from the child.
+        if let Ok(c) = self.as_mut().child_mut(&dst_child_uri) {
+            c.drop_rebuild_log()
```
tiagolobocastro, line 137: What if the rebuild fails, does this mean we lose the log and have to fully rebuild?

io-engine/src/bdev/nexus/nexus_child.rs
```diff
@@ -1232,4 +1237,37 @@ impl<'c> NexusChild<'c> {
             }
         }
     }
+
+    /// Creates a new rebuild log, or returns an existing one.
+    pub(super) fn get_or_init_rebuild_log(
+        &mut self,
+    ) -> Option<RebuildLogHandle> {
```
tiagolobocastro, line 1254: Shouldn't this return `RebuildLogHandle` rather than Option since it's `get_or_init`?

io-engine/src/rebuild/rebuild_log.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    /// Logs the given operation, marking the corresponding segment as modified
+    /// in the case of a write operation.
+    ///
+    /// # Arguments
+    ///
+    /// * `_io_type`: IoType of the operation to log.
+    /// * `lbn`: Logical block number.
+    /// * `lbn_cnt`: Number of logical blocks affected by the operation.
+    fn log_op(&mut self, _io_type: IoType, lbn: u64, lbn_cnt: u64) {
```
tiagolobocastro, line 66: Do we need `_io_type`, eg: we only need to set bit for write-like operations?

io-engine/src/rebuild/rebuild_log.rs
tiagolobocastro: Maybe should add a warn!() trace here?

io-engine/src/rebuild/rebuild_log.rs
tiagolobocastro: Hello mr mayastor :) 
And then we just collect the "partial-logs" from each core when we want to start rebuilding? (At that point the child is open for W so we don't need to keep the log anymore)

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -244,9 +283,25 @@ impl<'n> NexusChannel<'n> {
...
+    /// Attaches a new rebuild log to write to.
+    pub(super) fn attach_rebuild_log(&mut self, log: RebuildLogHandle) {
+        if self
+            .rebuild_logs
+            .iter()
+            .all(|i| i.device_name() != log.device_name())
+        {
+            debug!(
+                "{self:?}: attaching rebuild log for '{dev}'",
+                dev = log.device_name()
```
dsharma-dc, line 295: Can we just store the device name in a local string variable prior to comparison above.
We'd be able to avoid this call just for debug print as this takes lock.

io-engine/src/bdev/nexus/nexus_io.rs
```diff
@@ -562,15 +561,20 @@ impl<'n> NexusBio<'n> {
...
                 ),
-            );
+            ) {
+                self.log_operation(&log);
+            }
         }
 
-        // partial submission
+        self.channel()
+            .for_each_rebuild_log(|log| self.log_operation(log));
```
dsharma-dc, line 616: This would re-log the op on current channel in case of submission failure, but that's perhaps ok?

dsavitskiy:
bors try

io-engine/src/rebuild/rebuild_log.rs
dsavitskiy: It doesn't put mutex in I/O path unless a child is faulted and has a rebuild log created.
When all children are healthy, channels do not have any rebuild logs in their log lists, so channel's for_each_rebuild_log() is almost no-op.

When a channel has a rebuild log attached, locking the log isn't that much of problem:
* lock is held only for setting a single bit in a bit vector.
* a child is taken out of I/O path, so we save lots of CPU cycles due to that fact.
* parking_lot::Mutex is used which is spin-lock based. That means it is just a single cmp_exchange for the most likely case of a single child failed.
* Even for two or more children, I don't see a spinlock much of a problem if it's just a bit is set under the lock.

Having bit vector per core is also possible but I don't see much of improvement here performance-wise, but the memory footprint may increase considerably for larger devices. Contrary, we might even see a decrease of performance due to more CPU cache misses (a often overlooked case when optimizing for pure CPU cycles sacrificing memory use).

And what I want is to avoid adding additional complexity to this code path. Rebuild and fault/retire logic is quite subtle and sometimes hard to reason about, so less is more in this case.

io-engine/Cargo.toml
dsavitskiy: Not exactly, but we don't have an option to disable it for now: too much work required to modify testing pipeline to enable running this particular test with a special build.

I added a global boolean variable to check if any injections present. So when no injections, it is just an access to atomic bool per I/O operation. 

Once we have a support for a special builds for specific tests, we'll disable this by default.

io-engine/src/bdev/nexus/nexus_bdev_children.rs
dsavitskiy: Yes indeed. Thanks  for pointing out. Will fix.

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
dsavitskiy: Yes. I think it actually makes sense as if a (partial) rebuild fails, something fishy is already going on.

io-engine/src/bdev/nexus/nexus_child.rs
dsavitskiy: I would like to have rebuild log optional.
E.g. we can disable it completely by just returning None here, without modifying other places. Should probably add an env var or something to support that.
Also, potentially, we may add logic when a child refuses to create a rebuild log, e.g. if avail memory less than the log size would be; or for debugging purposes.

io-engine/src/rebuild/rebuild_log.rs
dsavitskiy: Currently, it's here for debugging purposes.

io-engine/src/bdev/nexus/nexus_io.rs
dsavitskiy: That's fine if we log the same op multiple times. Rather, what we don't want is to miss an op.

io-engine/src/bdev/nexus/nexus_channel.rs
dsavitskiy: Hmm, I think you're right. It also bothers me that I have to clone the device name String each time I needed to access it, because of the lock (Mutex won't allow to return &str). I'll change keep a copy of dev name in the handle itself.

(But again, isn't it a proverbial premature optimization? The code path where it is accessed isn't really sensitive, the parking_lot mutex is a spinlock.)

dsavitskiy:
bors try

io-engine/src/rebuild/rebuild_log.rs
dsavitskiy: Hi Jeffry!

Bitmap of dirty blocks (512 bytes) would be too large, and our rebuild is running at "segment" granularity anyway.
Calculating block to segment is just an integer multiplication and an integer division. Our code path is not that much computation bound.
And the faulty child is out of I/O path anyway, so we have few spare CPU cycles.
Again, bloating bitmap to avoid a division may be slower, not faster, because of more cache misses. Waiting for a line to fetch is certainly more than a int division.

io-engine/src/rebuild/rebuild_log.rs
gila: > Hello mr mayastor :) And then we just collect the "partial-logs" from each core when we want to start rebuilding? (At that point the child is open for W so we don't need to keep the log anymore)

@tiagolobocastro Gotcha.

gila: > It doesn't put mutex in I/O path unless a child is faulted and has a rebuild log created. When all children are healthy, channels do not have any rebuild logs in their log lists, so channel's for_each_rebuild_log() is almost no-op.
>

When are rebuild logs created? When we suspend/resume a rebuild? I cant seem to find a big theory statement and I can not access Jira. 

With regards to the locking it is exactly the cost of cache misses which *can* me detrimental for the performance overall. This problem typically is that you do not see the cost of this unless you run millions of IOPs (which we can and do!). 

Additionally, it is likely that when a replica has failed, more replicas have failed and out of the 100 volumes you may have 50% or more "rebuilding".  Much of little bits is still a lot =).

dsavitskiy:
bors try

io-engine/Cargo.toml
tiagolobocastro: Should we change the nix build to disable fault_injection when building release?

io-engine/src/rebuild/rebuild_log.rs
dsavitskiy: Giving it a second thought, I must admit that having a lock in I/O path is a bad idea indeed, even if it happens for the case of failed child(ren). Tiago and Michael agree on that too. So I'll change to a per-channel log; it's not that much more complicated.

As for the rebuild log's lifetime:

Rebuild log is created at the moment an I/O fails (either on submission or completion, or via API or timeout detection callback). Next, the log is "attached" to the I/O channels in the same way as "readers"/"writters" are attached. (See nexus_bdev_children.rs, retire_child_device().)

Then, the faulty child is detached from I/O channels (not immediately upon failure, but during child retire operation running asynchronously).

Once the rebuild starts, the child is re-attached into the I/O channels for write-only, and the log is detached (via reconfigure procedure). Rebuild job takes ownership of the log, and finally drops it when the rebuild finishes.

dsavitskiy:
@gila We came up with a decision to first merge this PR as is, and than I will proceed with lockless log as a separate ticket.

dsavitskiy:
bors try

tiagolobocastro (approved):

mtzaurus (approved):

dsavitskiy:
bors merge

dsavitskiy:
bors merge

PR #1295, opened 2023-01-25
fix(nexus): refactoring nexus child out-of-sync state

dsavitskiy:
If a nexus child is being rebuilt, and an I/O to that child fails, retire logic won't run because of incorrect handling of out-of-sync child state. This commit fixes how out-of-sync is managed.

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev.rs
@@ -23,7 +23,6 @@ use super::{
     nexus_err,
     nexus_injection::Injections,
     nexus_lookup_name_uuid,
-    ChildState,
     DrEvent,
     Error,
     NbdDisk,
@@ -971,14 +970,14 @@ impl<'n> Nexus<'n> {
                     .children
                     .iter()
                     // All children are online, so the Nexus is also online
-                    .all(|c| c.state() == ChildState::Open)
+                    .all(|c| c.is_healthy())
                 {
                     NexusStatus::Online
                 } else if self
                     .children
                     .iter()
                     // at least one child online, so the Nexus is also online
-                    .any(|c| c.state() == ChildState::Open)
+                    .any(|c| c.is_healthy())
                 {
                     NexusStatus::Degraded
                 } else {
@@ -1109,11 +1108,9 @@ impl<'n> BdevOps for Nexus<'n> {
         Reactor::block_on(async move {
             let self_ref = unsafe { &mut *self_ptr };
 
-            let n = self_ref
-                .children
-                .iter()
-                .filter(|c| c.state() == ChildState::Open)
-                .count();
+            // TODO: double-check interaction with rebuild job logic
+            // TODO: cancel rebuild jobs?
+            let n = self_ref.children.iter().filter(|c| c.is_healthy()).count();
 
             if n > 0 {
                 warn!(
@@ -1122,7 +1119,7 @@ impl<'n> BdevOps for Nexus<'n> {
                 );
 
                 for child in self_ref.children.iter_mut() {
-                    if child.state() == ChildState::Open {
+                    if child.is_healthy() {
                         if let Err(e) = child.close().await {
                             error!(
                                 "{:?}: child failed to close: {}",
--- a/io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
@@ -6,7 +6,7 @@ use super::{
     nexus_err,
     nexus_lookup_mut,
     nexus_persistence::PersistOp,
-    ChildState,
+    ChildSyncState,
     DrEvent,
     Error,
     Nexus,
@@ -75,9 +75,10 @@ impl<'n> Nexus<'n> {
         let name = self.name.clone();
         trace!("{}: start rebuild request for {}", name, child_uri);
 
+        // Find a healthy child to rebuild from.
         let src_child_uri = match self
             .children_iter()
-            .find(|c| c.state() == ChildState::Open && c.uri() != child_uri)
+            .find(|c| c.is_healthy() && c.uri() != child_uri)
         {
             Some(child) => Ok(child.uri().to_owned()),
             None => Err(Error::NoRebuildSource {
@@ -86,7 +87,7 @@ impl<'n> Nexus<'n> {
         }?;
 
         let dst_child_uri = match self.lookup_child(child_uri) {
-            Some(c) if c.state() == ChildState::Faulted(Reason::OutOfSync) => {
+            Some(c) if c.is_opened_unsync() => {
                 if c.rebuild_job().is_some() {
                     Err(Error::RebuildJobAlreadyExists {
                         child: child_uri.to_owned(),
@@ -370,13 +371,13 @@ impl<'n> Nexus<'n> {
 
         match job.state() {
             RebuildState::Completed => {
-                dst_child.set_state(ChildState::Open);
+                dst_child.sync_state = ChildSyncState::Synced;
                 info!("Child {} has been rebuilt successfully", child_uri);
                 let child_uri = child_uri.to_owned();
-                let child_state = dst_child.state();
+                let healthy = dst_child.is_healthy();
                 self.persist(PersistOp::Update {
                     child_uri,
-                    child_state,
+                    healthy,
                 })
                 .await;
             }
--- a/io-engine/src/bdev/nexus/nexus_child.rs
+++ b/io-engine/src/bdev/nexus/nexus_child.rs
@@ -116,8 +116,6 @@ pub enum Reason {
     /// No particular reason for the child to be in this state.
     /// This is typically the init state.
     Unknown,
-    /// Out of sync: child device is ok, but needs to be rebuilt.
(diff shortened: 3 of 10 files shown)


dsavitskiy:
bors try

io-engine/src/bdev/nexus/nexus_child.rs
```diff
@@ -757,17 +801,11 @@ impl<'c> NexusChild<'c> {
...
-            Reason::OutOfSync => {
-                self.set_state(ChildState::Faulted(reason));
-            }
-            _ => {
-                if let Err(e) = self.close().await {
-                    error!("{:?}: failed to close: {}", self, e.verbose());
-                }
-                self.set_state(ChildState::Faulted(reason));
-            }
+        if let Err(e) = self.close().await {
```
tiagolobocastro, line 803: These is changing the semantics for the doc comment, as it seems a fault(Reason::OutOfSync) would mark the child as Faulted(Reason::OutOfSync) without actually closing it?

io-engine/src/bdev/nexus/nexus_child.rs
```diff
@@ -909,7 +937,9 @@ impl<'c> NexusChild<'c> {
         }
 
         match state {
-            ChildState::Open | ChildState::Faulted(Reason::OutOfSync) => {
+            ChildState::Open => {
+                // TODO: double-check interaction with rebuild job logic
```
tiagolobocastro, line 940: Seems ok as looks like only a the ChildState::Open children will be part of the I/O path

io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -1109,11 +1108,9 @@ impl<'n> BdevOps for Nexus<'n> {
...
             let self_ref = unsafe { &mut *self_ptr };
 
-            let n = self_ref
-                .children
-                .iter()
-                .filter(|c| c.state() == ChildState::Open)
-                .count();
+            // TODO: double-check interaction with rebuild job logic
+            // TODO: cancel rebuild jobs?
+            let n = self_ref.children.iter().filter(|c| c.is_healthy()).count();
```
tiagolobocastro, line 1113: interesting, I guess we should keep the open here otherwise we won't close the children being rebuilt?

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -86,7 +87,7 @@ impl<'n> Nexus<'n> {
         }?;
 
         let dst_child_uri = match self.lookup_child(child_uri) {
-            Some(c) if c.state() == ChildState::Faulted(Reason::OutOfSync) => {
+            Some(c) if c.is_out_of_sync() => {
```
tiagolobocastro: Should we also check that it's open?

io-engine/src/bdev/nexus/nexus_child.rs
```diff
@@ -116,8 +116,6 @@ pub enum Reason {
     /// No particular reason for the child to be in this state.
     /// This is typically the init state.
     Unknown,
-    /// Out of sync: child device is ok, but needs to be rebuilt.
-    OutOfSync,
```
tiagolobocastro, line 120: Guess we need to update protobuf as well?

io-engine/src/bdev/nexus/nexus_child.rs
```diff
@@ -155,14 +152,15 @@ impl Display for Reason {
...
+/// State of a nexus child.
 #[derive(Debug, Clone, Copy, Serialize, Deserialize, Eq, PartialEq)]
 pub enum ChildState {
     /// Child has not been opened, but we are in the process of opening it.
     Init,
     /// Cannot add this block device to the parent as
     /// it iss incompatible property-wise.
     ConfigInvalid,
-    /// The child is open for R/W.
+    /// The child is open for I/O.
```
tiagolobocastro, line 163: Could this be `Open(SyncState)` ?

io-engine/src/grpc/v0/nexus_grpc.rs
```diff
@@ -25,16 +25,19 @@ use crate::{
 
 /// Map the internal child states into rpc child states (i.e. the states that
 /// the control plane sees)
-impl From<ChildState> for rpc::ChildState {
-    fn from(child: ChildState) -> Self {
-        match child {
+impl From<&NexusChild<'_>> for rpc::ChildState {
+    fn from(child: &NexusChild) -> Self {
+        if child.is_out_of_sync() {
```
tiagolobocastro: hmm so when a child is faulted while rebuilding we report degraded when should be faulted?
Might be better to do this if line 37 instead, only when child is open?

io-engine/src/bdev/nexus/nexus_child.rs
```diff
@@ -299,15 +320,38 @@ impl<'c> NexusChild<'c> {
...
+    pub fn state(&self) -> ChildState {
+        self.state.load()
+    }
+
+    /// Returns the sync state of the child.
+    pub fn sync_state(&self) -> ChildSyncState {
+        self.sync_state
+    }
+
+    /// Determines if the child is opened but out-of-sync.
```
tiagolobocastro: Ah I had missed this, it's already implied here, wonder if it's better not to be implied? 
Maybe it's clearer to have 2 functions, `is_out_of_sync` and `is_open_out_of_sync` ?

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
dsavitskiy: is_out_of_sync() implies Open.

io-engine/src/bdev/nexus/nexus_child.rs
dsavitskiy: No, not at this moment. I just remapped new child flags into old protobuf states.

io-engine/src/bdev/nexus/nexus_child.rs
dsavitskiy: Faulted can still be out-of-sync. It's independent of being open.

io-engine/src/grpc/v0/nexus_grpc.rs
dsavitskiy: is_out_of_sync() implies Open. Probably I should change the func name.

io-engine/src/bdev/nexus/nexus_child.rs
dsavitskiy: We don't really need a non-Open out-of-sync.

io-engine/src/bdev/nexus/nexus_bdev.rs
dsavitskiy: It was closing only Open before, so I don't want to change that at this moment. We can potentially bump into something unexpected if we change it.

io-engine/src/bdev/nexus/nexus_child.rs
dsavitskiy: Doc comment didn't make much sense, anyway. fault() is never called with OutOfSync. I'll remove the comment.

mtzaurus (approved):

io-engine/src/bdev/nexus/nexus_child.rs
tiagolobocastro: But faulted is always out-of-sync?

io-engine/src/bdev/nexus/nexus_bdev.rs
tiagolobocastro: Ok fair enough, maybe we can make this change later for develop branch only?

io-engine/src/bdev/nexus/nexus_child.rs
tiagolobocastro: Ok but if we don't use it might be worth removing otherwise could be misleading: https://github.com/openebs/mayastor-api/blob/6487977a5785510b9c5864bed561cd2c09170b79/protobuf/v1/nexus.proto#L96

io-engine/src/grpc/v0/nexus_grpc.rs
tiagolobocastro: Yeah I think better to rename then.

tiagolobocastro (approved):
passes all the tests I've been running :+1:

io-engine/src/bdev/nexus/nexus_child.rs
dsavitskiy: I've added a ticket to cover this and other things: https://dcsw-bolt.atlassian.net/browse/B2-842

io-engine/src/bdev/nexus/nexus_bdev.rs
dsavitskiy: I've added a ticket to cover this and other things: https://dcsw-bolt.atlassian.net/browse/B2-842

io-engine/src/bdev/nexus/nexus_child.rs
dsavitskiy: As for now, I'd rather think of SyncState as of another dimension of nexus states, independent from Open/Faulted.

dsavitskiy:
bors merge

PR #1300, opened 2023-01-27
B2-839 io-engine crashes when sharing bdev via v1 io-engine-client

felix1209:
io-engine-client sends the protocol field 'None' to io-engine when the following command is issued which results in io-engine crash

`io-engine-client bdev share abc0`

### DIFF
--- a/io-engine/src/bin/io-engine-client/v1/bdev_cli.rs
+++ b/io-engine/src/bin/io-engine-client/v1/bdev_cli.rs
@@ -44,11 +44,13 @@ pub fn subcommands<'a, 'b>() -> App<'a, 'b> {
         .arg(Arg::with_name("name").required(true).index(1))
         .arg(
             Arg::with_name("protocol")
+                .long("protocol")
                 .short("p")
                 .help("the protocol to used to share the given bdev.")
                 .required(false)
-                .possible_values(&["nvmf"])
-                .default_value("nvmf"),
+                .possible_values(&["Nvmf"])
+                .takes_value(true)
+                .default_value("Nvmf"),
         )
         .arg(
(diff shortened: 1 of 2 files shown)


dsavitskiy:
Could you squash these two commits into one, so you changes to commit hooks would disappear?

dsavitskiy:
Protocol parameter mapping logic in v1 API doesn't look solid, see io-engine/src/grpc/v1/bdev.rs:142. It seems like io-engine would panic if a bad value is passed.  Could you fix that in v1 API? (v0 API seems to be checking for invalid values properly.)

felix1209:
> Protocol parameter mapping logic in v1 API doesn't look solid, see io-engine/src/grpc/v1/bdev.rs:142. It seems like io-engine would panic if a bad value is passed. Could you fix that in v1 API? (v0 API seems to be checking for invalid values properly.)

With my changes, the CLI doesn't crash the io-engine. Will fix the io-engine as well.

dsavitskiy (approved):

mtzaurus (approved):

tiagolobocastro:
> io-engine-client sends the protocol field 'None' to io-engine when the following command is issued which results in io-engine crash
> 
> `io-engine-client bdev share abc0`

Should we also prevent the crash when "None" is sent?

felix1209:
> > io-engine-client sends the protocol field 'None' to io-engine when the following command is issued which results in io-engine crash
> > `io-engine-client bdev share abc0`
> 
> Should we also prevent the crash when "None" is sent?

Tiago, Added that change as well.

felix1209:
bors try

felix1209:
bors merge

PR #1306, opened 2023-01-31
fix(nexus): avoid duplicating replica records when readding replicas

mtzaurus:
In current implementation there is no check in persistent nexus info against presence of the record for the replica being readded, which leads to multiple records for the same replica being stored in ETCD. The fix adds extra checks to PersistOp::AddChild operation to update existing replica records in place when a replica with the same UUID is readded to nexus.


mtzaurus:
bors try

tiagolobocastro (approved):

blaisedias (approved):

mtzaurus:
bors merge

PR #1310, opened 2023-02-04
fix(nexus): replica reset

felix1209:
Whenever there is an initiator timeout with the nexus, the connection to all the replicas needs to be reset

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev.rs
@@ -6,6 +6,7 @@
 
 use std::{
     cmp::min,
+    collections::HashSet,
     convert::TryFrom,
     fmt::{Debug, Display, Formatter},
     marker::PhantomPinned,
@@ -268,6 +269,8 @@ pub struct Nexus<'n> {
     pub(crate) shutdown_requested: AtomicCell<bool>,
     /// Prevent auto-Unpin.
     _pin: PhantomPinned,
+    /// Initiators
+    initiators: parking_lot::Mutex<HashSet<String>>,
 }
 
 impl<'n> Debug for Nexus<'n> {
@@ -363,6 +366,7 @@ impl<'n> Nexus<'n> {
             nexus_target: None,
             nvme_params,
             has_io_device: false,
+            initiators: parking_lot::Mutex::new(HashSet::new()),
             nexus_info: futures::lock::Mutex::new(PersistentNexusInfo::new(
                 nexus_info_key,
             )),
@@ -450,6 +454,24 @@ impl<'n> Nexus<'n> {
     /// Returns the Nexus uuid.
     pub(crate) fn uuid(&self) -> Uuid {
         self.nexus_uuid
+    }
+
+    /// Add new initiator to the Nexus
+    pub(crate) unsafe fn add_initiator(self: Pin<&mut Self>, initiator: &str) {
+        self.unpin_mut()
+            .initiators
+            .lock()
+            .insert(initiator.to_string());
+    }
+
+    /// Remove initiator from the Nexus
+    pub(crate) unsafe fn rm_initiator(self: Pin<&mut Self>, initiator: &str) {
+        self.unpin_mut().initiators.lock().remove(initiator);
+    }
+
+    /// initiator count from the Nexus
+    pub(crate) fn initiator_cnt(&self) -> usize {
+        self.initiators.lock().len()
     }
 
     /// Sets the state of the Nexus.
@@ -819,6 +841,47 @@ impl<'n> Nexus<'n> {
     /// be called only from the master core.
     pub async fn resume(self: Pin<&mut Self>) -> Result<(), Error> {
         self.io_subsystem_mut().resume().await
+    }
+
+    /// Set the Nexus state to 'reset'
+    pub fn set_reset_state(self: Pin<&mut Self>) -> bool {
+        let mut state = self.state.lock();
+        match *state {
+            // Reset operation is allowed only when the Nexus is Open state
+            NexusState::Open => {
+                *state = NexusState::Reconfiguring;
+                true
+            }
+            _ => {
+                info!(
+                    nexus=%self.name,
+                    "Transition from {:?} is not permitted to NexusState::Reconfiguring",
+                    state
+                );
+                false
+            }
+        }
+    }
+
+    /// Set the Nexus state to 'open'
+    pub fn set_open_state(self: Pin<&mut Self>) -> bool {
+        let mut state = self.state.lock();
+        match *state {
+            // Open operation is allowed only when the Nexus is
+            // Init/Reconfiguring state
+            NexusState::Reconfiguring | NexusState::Init => {
+                *state = NexusState::Open;
+                true
+            }
+            _ => {
+                info!(
+                    nexus=%self.name,
+                    "Transition from {:?} is not permitted to NexusState::Open",
+                    state
+                );
+                false
+            }
+        }
     }
 
     pub async fn shutdown(mut self: Pin<&mut Self>) -> Result<(), Error> {
--- a/io-engine/src/bdev/nvmx/controller.rs
+++ b/io-engine/src/bdev/nvmx/controller.rs
@@ -25,7 +25,6 @@ use spdk_rs::{
         spdk_nvme_ctrlr_get_ns,
         spdk_nvme_ctrlr_is_active_ns,
         spdk_nvme_ctrlr_register_aer_callback,
-        spdk_nvme_ctrlr_reset,
         spdk_nvme_detach,
     },
     Poller,
@@ -666,31 +665,20 @@ impl<'a> NvmeController<'a> {
             return;
         }
 
-        let rc =
-            unsafe { spdk_nvme_ctrlr_reset(reset_ctx.spdk_handle.as_ptr()) };
-        if rc != 0 {
-            error!(
-                "{} failed to reset controller, rc = {}",
-                reset_ctx.name, rc
-            );
+        debug!(
+            "{} controller successfully reset, reinitializing I/O channels",
+            reset_ctx.name
+        );
 
-            NvmeController::_complete_reset(reset_ctx, rc);
-        } else {
-            debug!(
-                "{} controller successfully reset, reinitializing I/O channels",
-                reset_ctx.name
-            );
-
-            // Once controller is successfully reset, schedule another
-            //I/O channel traversal to restore all I/O channels.
-            let io_device = reset_ctx.io_device.clone();
-            io_device.traverse_io_channels(
-                NvmeController::_reset_create_channels,
-                NvmeController::_reset_create_channels_done,
-                NvmeIoChannel::inner_from_channel,
-                reset_ctx,
-            );
-        }
+        // Once controller is successfully reset, schedule another
+        //I/O channel traversal to restore all I/O channels.
+        let io_device = reset_ctx.io_device.clone();
+        io_device.traverse_io_channels(
+            NvmeController::_reset_create_channels,
+            NvmeController::_reset_create_channels_done,
+            NvmeIoChannel::inner_from_channel,
+            reset_ctx,
+        );
     }
 
     fn _reset_create_channels(
--- a/io-engine/src/subsys/nvmf/subsystem.rs
+++ b/io-engine/src/subsys/nvmf/subsystem.rs
@@ -8,11 +8,13 @@ use std::{
 use futures::channel::oneshot;
 use nix::errno::Errno;
 
+use crate::bdev::{nexus::nexus_lookup_mut, nvmx::NVME_CONTROLLERS};
 use spdk_rs::libspdk::{
     nvmf_subsystem_find_listener,
     nvmf_subsystem_set_ana_state,
     nvmf_subsystem_set_cntlid_range,
     spdk_bdev_nvme_opts,
+    spdk_nvmf_ctrlr,
     spdk_nvmf_ns_get_bdev,
     spdk_nvmf_ns_opts,
     spdk_nvmf_subsystem,
@@ -22,6 +24,7 @@ use spdk_rs::libspdk::{
     spdk_nvmf_subsystem_create,
     spdk_nvmf_subsystem_destroy,
     spdk_nvmf_subsystem_disconnect_host,
+    spdk_nvmf_subsystem_events,
     spdk_nvmf_subsystem_get_first,
     spdk_nvmf_subsystem_get_first_host,
     spdk_nvmf_subsystem_get_first_listener,
@@ -32,6 +35,7 @@ use spdk_rs::libspdk::{
     spdk_nvmf_subsystem_get_nqn,
     spdk_nvmf_subsystem_listener_get_trid,
     spdk_nvmf_subsystem_pause,
+    spdk_nvmf_subsystem_register_for_event,
     spdk_nvmf_subsystem_remove_host,
     spdk_nvmf_subsystem_resume,
     spdk_nvmf_subsystem_set_allow_any_host,
@@ -42,6 +46,9 @@ use spdk_rs::libspdk::{
     spdk_nvmf_subsystem_state_change_done,
     spdk_nvmf_subsystem_stop,
     spdk_nvmf_tgt,
+    SPDK_NVMF_SS_INIATOR_CONNECT,
+    SPDK_NVMF_SS_INIATOR_DISCONNECT,
+    SPDK_NVMF_SS_INIATOR_TIMEOUT,
     SPDK_NVMF_SUBTYPE_DISCOVERY,
     SPDK_NVMF_SUBTYPE_NVME,
 };
@@ -49,7 +56,7 @@ use spdk_rs::libspdk::{
 use crate::{
     constants::{NVME_CONTROLLER_MODEL_ID, NVME_NQN_PREFIX},
     core::{Bdev, Reactors, UntypedBdev},
-    ffihelper::{cb_arg, AsStr, FfiResult, IntoCString},
+    ffihelper::{cb_arg, done_cb, AsStr, FfiResult, IntoCString},
     subsys::{
         make_subsystem_serial,
         nvmf::{transport::TransportId, Error, NVMF_TGT},
@@ -74,6 +81,29 @@ impl Display for SubType {
 
 pub struct NvmfSubsystem(pub(crate) NonNull<spdk_nvmf_subsystem>);
 pub struct NvmfSubsystemIterator(*mut spdk_nvmf_subsystem);
+
+#[repr(C)]
+pub struct SpdkNvmfController(pub(crate) NonNull<spdk_nvmf_ctrlr>);
+
+impl SpdkNvmfController {
+    /// Get the hostnqn from the controller
+    fn hostnqn(&self) -> String {
+        unsafe { self.0.as_ref().hostnqn.as_str().to_string() }
+    }
+}
+impl From<*mut spdk_nvmf_ctrlr> for SpdkNvmfController {
+    fn from(s: *mut spdk_nvmf_ctrlr) -> Self {
+        SpdkNvmfController(NonNull::new(s).unwrap())
+    }
+}
+
+impl Debug for SpdkNvmfController {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
+        f.debug_struct("SpdkNvmfController")
+            .field("hostnqn", &self.hostnqn())
+            .finish()
+    }
+}
 
 impl Iterator for NvmfSubsystemIterator {
     type Item = NvmfSubsystem;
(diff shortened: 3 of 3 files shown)


io-engine/src/bdev/nexus/nexus_io.rs
```diff
@@ -432,6 +461,13 @@ impl<'n> NexusBio<'n> {
                 })
         });
 
+
+        let mut flag = false;
```
dsavitskiy: Please rename this to something more meaning-full like "is_failed"

io-engine/src/bdev/nvmx/mod.rs
```diff
@@ -19,7 +19,7 @@ use crate::{
 mod channel;
 mod controller;
 mod controller_inner;
-mod controller_state;
+pub mod controller_state;
```
dsavitskiy: It's better to add an accessor to NvmeController to get the state rather than exposing the state.

dsavitskiy (dismissed):

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -165,6 +232,132 @@ fn make_sn<T: AsRef<[u8]>>(uuid: T) -> CString {
...
+                }
+            };
+            let dev_name = dev.device_name();
+            debug!(
+                "nvmf_event_handler: device name {:?} - nexus subsys {:?}",
+                dev_name, subsys_nqn
+            );
+            let ctrlr = match NVME_CONTROLLERS.lookup_by_name(&dev_name) {
+                Some(ctrlr) => ctrlr,
+                None => {
```
mtzaurus: Such a situation can take place when nexus has local child bdev replicas. We should skip such devices.

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -165,6 +196,189 @@ fn make_sn<T: AsRef<[u8]>>(uuid: T) -> CString {
...
+                return;
+            }
+        };
+
+        //Pause the Nexus
+        if nexus.as_mut().pause().await.is_err() {
+            error!("Failed to pause Nexus {:?}", nexus);
+            return;
+        }
+        info!("Paused the nexus - {:?} ", nexus_name);
```
mtzaurus: We postpone nexus reset if there is at least one active initiator: 

```
extern "C" fn nvmf_event_handler()
   ...
   if nexus.initiator_cnt() > 0 {
        debug!("Reset operation not supported - initiator(s) present");
        return;
   }
```

Should we perform this check the final time before resetting the nexus after the subsystem has been paused ?
There can be a small chance that the first initiator is added again after 'reset_nexus()'  has been scheduled, but before the subsystem is paused.

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -165,6 +196,189 @@ fn make_sn<T: AsRef<[u8]>>(uuid: T) -> CString {
...
+            debug!(
+                "nvmf_event_handler: device name {:?} - nexus nexus_name {:?},
+                 driver {:?}, product {:?}",
+                dev_name,
+                nexus_name,
+                dev.driver_name(),
+                dev.product_name(),
+            );
+            let (sender, receiver) = oneshot::channel::<bool>();
+            Reactors::master().send_future(Self::reset_one_child(
```
mtzaurus: No need to schedule 'Self::reset_one_child' on Master reactor - 'reset_nexus()' is already running on Master reactor. So you can just call 'Self::reset_one_child' directly using '.await'.

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -165,6 +196,189 @@ fn make_sn<T: AsRef<[u8]>>(uuid: T) -> CString {
...
+                "nvmf_event_handler: controller successfully reset {:?}",
+                success
+            );
+        } else {
+            error!("nvmf_event_handler: failed to reset controller ");
+        }
+        done_cb(ctx, success);
+    }
+
+    pub async fn reset_one_child(dev_name: String, ctx: *mut c_void) {
```
mtzaurus: This function is marked as 'async' but it doesn't use any '.await', so it's synchronous.
I'd suggest to turn it into fully asynchronous by implementing asynchronous awaiting of the operation inside this function. So we could move this code from 'reset_nexus()':

```
let (sender, receiver) = oneshot::channel::<bool>();
```

to 'reset_one_child()' and have the waiting logic hidden from the callers also in 'reset_one_child()':

```
if !receiver.await.expect("Reset completed") {
```

So the callers just do '::reset_one_child().await'.

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -165,6 +196,189 @@ fn make_sn<T: AsRef<[u8]>>(uuid: T) -> CString {
...
+                    "Failed to reset {:?} for nexus_name {:?}",
+                    child, nexus_name
+                );
+            }
+        }
+
+        //Resume the Nexus
+        if nexus.as_mut().resume().await.is_err() {
+            error!("Failed to resume Nexus {:?}", nexus);
+            return;
```
mtzaurus: if we return here, we don't restore nexus state. Shall we move resume logic after code that restores the state ?

io-engine/src/subsys/nvmf/subsystem.rs
mtzaurus: We could even omit this function and perform nexus child reset directly on device I/O handle which every nexus child has (see 'io-engine/src/bdev/nexus/nexus_child.rs'):

```
impl<'c> NexusChild<'c> {
   ...
   pub fn get_io_handle(
        &self,
    ) -> Result<Box<dyn BlockDeviceHandle>, CoreError> {
   ...
   }
```

There is a 'reset()' function in this handle which hides low-level reset implementation

```
#[async_trait(?Send)]
pub trait BlockDeviceHandle {
   ...
       fn reset(&self, cb: IoCompletionCallback, cb_arg: IoCompletionCallbackArg) -> Result<(), CoreError>;
}
```

In case of NVMeOF replicas it performs exactly controller reset.

felix1209: looks like this reset is a type of IO from the initiator to reset a given channel, whereas we need to reset all the data channels. https://spdk.io/doc/group__bdev__io__submit__functions.html#gaca574d3e94be6150343823dd2e7ecf46

io-engine/src/subsys/nvmf/subsystem.rs
felix1209: In fact, we should not restore the state to open if the resume failed. At this point, it will need user intervention anyway. The user needs to unshare and share the nexus again

io-engine/src/subsys/nvmf/subsystem.rs
felix1209: I tried this way first, but it didn't work. It got stuck at reset_cb->done_cb

io-engine/src/subsys/nvmf/subsystem.rs
felix1209: I tried this way first, but it didn't work. It got stuck at reset_cb->done_cb

io-engine/src/subsys/nvmf/subsystem.rs
mtzaurus: IMHO, it should work. It's the same reactor -  Master core. If it doesn't, it worth  investigating, IMHO.

io-engine/src/subsys/nvmf/subsystem.rs
felix1209: Good thought! Probably, it is a narrow window, since by default initiator takes 10s to reconnect. Worst case it might time out again and reconnect. I have added this check for more future-looking, I am not clear on the impact of this reset in the multi-initiator scenarios where multiple initiators may be used for failover or load balancing.  If we allow multiple initiators in the future, I suspect that this code may result in all other initiators' timeout if one times out when we pause the IO until replicas are reconnected.

io-engine/src/subsys/nvmf/subsystem.rs
felix1209: Looks like I had a different issue when I tried last time, '::reset_one_child().await works. Thanks for the suggestion.

io-engine/src/bdev/nexus/nexus_io.rs
```diff
@@ -189,6 +190,12 @@ impl<'n> NexusBio<'n> {
                 status,
                 self.ctx()
             );
+            // seems like race condition, ignore the error if the controller is
+            // resetting
+            if NvmeController::reset_active(&child.device_name()) {
+                self.ok();
```
tiagolobocastro: is this right? We're oking the IO but it failed?

io-engine/src/bdev/nexus/nexus_io.rs
```diff
@@ -482,6 +489,12 @@ impl<'n> NexusBio<'n> {
                 })
         });
 
+        let mut reset_active = false;
+        if failed_device.is_some() {
+            let dev_name = failed_device.as_ref().unwrap();
+            reset_active = NvmeController::reset_active(dev_name);
+        }
```
tiagolobocastro:
```suggestion
        let reset_active = failed_device.map(|dev_name| NvmeController::reset_active(dev_name)).unwrap_or_default();
```

io-engine/src/bdev/nvmx/controller.rs
```diff
@@ -740,6 +728,32 @@ impl<'a> NvmeController<'a> {
...
+                    return true;
+                }
+            }
+            None => {
+                debug!(
+                    "nvmf_event_handler is not having valid ctrlr for device {:?}",
+                    dev_name
+                );
+            }
+        };
```
tiagolobocastro:
```suggestion
        }
```

io-engine/src/bdev/nvmx/controller.rs
```diff
@@ -740,6 +728,32 @@ impl<'a> NvmeController<'a> {
...
+                if ctrlr
+                    .lock()
+                    .state_machine
+                    .lookup_flag(ControllerFlag::ResetActive)
+                    .load()
+                {
+                    debug!(
+                        "nvmf_event_handler reset active for the controller {:?}",
+                        dev_name
+                    );
```
tiagolobocastro:
```suggestion
                    debug!(
                        "nvmf_event_handler reset active for the controller {dev_name}",
                    );
```

io-engine/src/bdev/nvmx/controller.rs
```diff
@@ -740,6 +728,32 @@ impl<'a> NvmeController<'a> {
...
+                    return true;
+                }
+            }
+            None => {
+                debug!(
+                    "nvmf_event_handler is not having valid ctrlr for device {:?}",
+                    dev_name
+                );
```
tiagolobocastro:
```suggestion
                debug!(
                    "nvmf_event_handler is not having valid ctrlr for device {dev_name}"
                );
```

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -75,6 +82,30 @@ impl Display for SubType {
...
+    }
+}
+
+impl Debug for SpdkNvmfController {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
+        unsafe {
+            f.debug_struct("SpdkNvmfController")
+                .field("hostnqn", &self.0.as_ref().hostnqn.as_str().to_string())
+                .finish()
+        }
```
tiagolobocastro: Reduce the unsafe code scope to what's necessary.
```suggestion
        let hostnqn = unsafe { self.get_hostnqn() };
            f.debug_struct("SpdkNvmfController")
                .field("hostnqn", &hostnqn)
                .finish()
```

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -75,6 +82,30 @@ impl Display for SubType {
 pub struct NvmfSubsystem(pub(crate) NonNull<spdk_nvmf_subsystem>);
 pub struct NvmfSubsystemIterator(*mut spdk_nvmf_subsystem);
 
+#[repr(C)]
+pub struct SpdkNvmfController(pub(crate) NonNull<spdk_nvmf_ctrlr>);
+
+impl SpdkNvmfController {
+    fn get_hostnqn(&self) -> String {
```
tiagolobocastro: In rust it's typical not to prefix getters with `get_`.

```suggestion
    fn hostnqn(&self) -> String {
```

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -165,6 +196,186 @@ fn make_sn<T: AsRef<[u8]>>(uuid: T) -> CString {
 }
 
 impl NvmfSubsystem {
+    pub fn extract_nexus_name(nqn: &str) -> Option<String> {
```
tiagolobocastro: Might be worth moving this function somewhere else, perhaps close to where we define how our nqn is defined.

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -165,6 +196,186 @@ fn make_sn<T: AsRef<[u8]>>(uuid: T) -> CString {
...
+        let vec: Vec<String> = nqn.split(':').map(|s| s.to_string()).collect();
+
+        if vec.len() != 2 {
+            return None;
+        }
+        Some(vec[1].clone())
```
tiagolobocastro:
```suggestion
        vec.get(1).cloned()
```

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -165,6 +196,186 @@ fn make_sn<T: AsRef<[u8]>>(uuid: T) -> CString {
...
+        }
+        done_cb(ctx, success);
+    }
+
+    pub async fn reset_one_child(dev_name: String, ctx: *mut c_void) {
+        let ctrlr = match NVME_CONTROLLERS.lookup_by_name(&dev_name) {
+            Some(ctrlr) => ctrlr,
+            None => {
+                debug!(
+                    "device name  {:?} not a valid NVMe controller",
```
tiagolobocastro:
```suggestion
                    "device name  {dev_name} not a valid NVMe controller"
```
Please replicate these changes across the traces. Btw strings don't need to use the debug trait (`:?`).

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -165,6 +196,186 @@ fn make_sn<T: AsRef<[u8]>>(uuid: T) -> CString {
...
+            );
+            Self::reset_cb(false, ctx);
+        }
+    }
+
+    pub async fn reset_nexus(nexus_name: String) {
+        let mut nexus = match nexus_lookup_mut(&nexus_name) {
+            Some(v) => v,
+            None => {
+                debug!("nqn {:?} is not having valid nexus", nexus_name);
```
tiagolobocastro:
```suggestion
                debug!("nqn {nexus_name} does not contain a valid nexus");
```

tiagolobocastro (commented):
Would it be good to also add tests for initiator connect/reconnect to test these code changes? @mtzaurus

io-engine/src/subsys/nvmf/subsystem.rs
felix1209: Looks like I had a different issue when I tried last time, '::reset_one_child().await works. Thanks for the suggestion.

io-engine/src/subsys/nvmf/subsystem.rs
felix1209: a good suggestion to use get_hostnqn, but it doesn't need the unsafe block.

io-engine/src/bdev/nexus/nexus_io.rs
felix1209: This code was added before I implement the nexus pause/resume. This is not needed now, will remove it

io-engine/src/bdev/nexus/nexus_io.rs
felix1209: This code was added before I implement the nexus pause/resume. This is not needed now, will remove it

io-engine/src/subsys/nvmf/subsystem.rs
felix1209: moved to util::uri

io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -820,6 +842,45 @@ impl<'n> Nexus<'n> {
         self.io_subsystem_mut().resume().await
     }
 
+    pub fn set_reset_state(self: Pin<&mut Self>) -> bool {
```
tiagolobocastro, line 847: Missing doc comments for these functions

io-engine/src/bdev/nvmx/controller.rs
```diff
@@ -666,31 +665,20 @@ impl<'a> NvmeController<'a> {
...
-                reset_ctx.name, rc
-            );
-
-            NvmeController::_complete_reset(reset_ctx, rc);
-        } else {
-            debug!(
-                "{} controller successfully reset, reinitializing I/O channels",
-                reset_ctx.name
-            );
+        debug!(
```
tiagolobocastro, line 668: So we don't need to issue `spdk_nvme_ctrlr_reset` now?

io-engine/src/bdev/nvmx/controller_state.rs
```diff
@@ -200,8 +200,7 @@ impl ControllerStateMachine {
         }
     }
 
-    // Private functions.
-    fn lookup_flag(&self, _flag: ControllerFlag) -> &AtomicCell<bool> {
+    pub fn lookup_flag(&self, _flag: ControllerFlag) -> &AtomicCell<bool> {
```
tiagolobocastro: Are you making use of this? Seems flag is not checked at all?

io-engine/src/bdev/util/uri.rs
```diff
@@ -46,3 +46,9 @@ pub(crate) fn uuid(
 ) -> Result<Option<uuid::Uuid>, uuid::Error> {
     value.map(|uuid| uuid::Uuid::parse_str(&uuid)).transpose()
 }
+
+pub(crate) fn extract_nexus_name(nqn: &str) -> Option<String> {
```
tiagolobocastro: Sorry I had led you astray before, seems the function that generates nqn was at the bottom of the file: https://github.com/openebs/mayastor/blob/92a23f609ee62477b60440ea225a4c481b4af853/io-engine/src/subsys/nvmf/subsystem.rs#L815
We should keep those together there then

io-engine/src/bdev/util/uri.rs
```diff
@@ -46,3 +46,9 @@ pub(crate) fn uuid(
...
+    let vec: Vec<String> = nqn.split(':').map(|s| s.to_string()).collect();
+
+    vec.get(1).cloned()
```
tiagolobocastro: Nit: don't need to clone two strings, just the "right" one:

```suggestion
    let vec: Vec<&str> = nqn.split(':').collect();

    vec.get(1).map(ToString::to_string)
```

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -165,6 +194,169 @@ fn make_sn<T: AsRef<[u8]>>(uuid: T) -> CString {
...
+                "nvmf_event_handler: controller successfully reset {:?}",
+                success
+            );
+        } else {
+            error!("nvmf_event_handler: failed to reset controller ");
+        }
+        done_cb(ctx, success);
+    }
+
+    pub async fn reset_one_child(dev_name: String, ctx: *mut c_void) {
```
tiagolobocastro: Don't need to use by value, reference should work here: `dev_name: &str`.
Also missing doc comment

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -165,6 +194,169 @@ fn make_sn<T: AsRef<[u8]>>(uuid: T) -> CString {
...
+        let ctrlr = match NVME_CONTROLLERS.lookup_by_name(&dev_name) {
+            Some(ctrlr) => ctrlr,
+            None => {
+                debug!("device name  {dev_name} not a valid NVMe controller");
```
tiagolobocastro:
```suggestion
                debug!("device name {dev_name} not a valid NVMe controller");
```

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -165,6 +194,169 @@ fn make_sn<T: AsRef<[u8]>>(uuid: T) -> CString {
...
+                Self::reset_cb(false, ctx);
+                return;
+            }
+        };
+        debug!("nvmf_event_handler: nvme controller {:?} ", ctrlr);
```
tiagolobocastro:
```suggestion
        debug!("nvmf_event_handler: nvme controller {:?}", ctrlr);
```

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -165,6 +194,169 @@ fn make_sn<T: AsRef<[u8]>>(uuid: T) -> CString {
...
+        let mut ctrlr = ctrlr.lock();
+        if ctrlr.reset(NvmfSubsystem::reset_cb, ctx, false).is_ok() {
+            debug!("nvmf_event_handler: Reset initiated for {:?} ", ctrlr);
+        } else {
+            error!(
+                "nvmf_event_handler: Failed to initiate reset for {:?} ",
+                ctrlr
+            );
+            Self::reset_cb(false, ctx);
```
tiagolobocastro: should we capture the actual error in the log? example:
```suggestion
        match ctrlr.reset(NvmfSubsystem::reset_cb, ctx, false) {
            Ok(_) => debug!("nvmf_event_handler: Reset initiated for {:?} ", ctrlr),
            Err(error) => {
                error!(
                    "nvmf_event_handler: Failed to initiate reset for {ctrlr:?} with error: {error:?}"
                );
                Self::reset_cb(false, ctx);
            }
```
Also, in case of error shouldn't the current function return an error?

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -165,6 +194,169 @@ fn make_sn<T: AsRef<[u8]>>(uuid: T) -> CString {
...
+        }
+    }
+
+    pub async fn reset_nexus(nexus_name: String) {
```
tiagolobocastro, line 237:
```suggestion
    pub async fn reset_nexus(nexus_name: &str) {
```

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -165,6 +194,169 @@ fn make_sn<T: AsRef<[u8]>>(uuid: T) -> CString {
...
+            Some(v) => v,
+            None => {
+                debug!("nqn {nexus_name} does not contain a valid nexus");
+                return;
+            }
+        };
+
+        //Pause the Nexus
+        if nexus.as_mut().pause().await.is_err() {
+            error!("Failed to pause Nexus {:?}", nexus);
```
tiagolobocastro: Should we log the actual error here as well?

io-engine/src/subsys/nvmf/subsystem.rs
tiagolobocastro: Doesn't this function also need to return an error?

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -165,6 +194,169 @@ fn make_sn<T: AsRef<[u8]>>(uuid: T) -> CString {
...
+            return;
+        }
+        info!("Paused the nexus - {:?} ", nexus_name);
+        for child in nexus.children_iter() {
+            debug!("Found a {:?} for nexus {nexus_name}", child);
+            if !child.is_healthy() {
```
tiagolobocastro: Do we need to check for OOS children as well since they are open for Writes?

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -165,6 +194,169 @@ fn make_sn<T: AsRef<[u8]>>(uuid: T) -> CString {
...
+                    error!(
+                        "Nexus {nexus_name} get_device failed for {:?} with Error {:?}",
+                         child, error
+                    );
+                    continue;
+                }
+            };
+            let dev_name = dev.device_name();
+            if dev.driver_name() != "nvme" {
+                debug!("Device {dev_name} is not NVMe-oF, reset is not needed");
```
tiagolobocastro, line 272: Why is reset not needed for local replicas?

io-engine/src/subsys/nvmf/subsystem.rs
tiagolobocastro: When can resume fail? Would it be worth retrying?

mtzaurus (approved):

io-engine/src/bdev/nvmx/controller.rs
felix1209: spdk_nvme_ctrlr_reset API includes the logic for both tearing down and re-establish connection. But, by this place, we had already torn down connections for all the channels and we have our own code for reestablishing the connection.

io-engine/src/bdev/nvmx/controller_state.rs
felix1209: good catch! I just changed the function from private to public. Looks like this code is incomplete. Anyway, I will change it back to private, it is not being used now

io-engine/src/bdev/util/uri.rs
felix1209: ah, I didn't notice this as well. It is actually URI parsing. Either way is fine, let me move it back to subsystem.rs near gen_nqn.

io-engine/src/subsys/nvmf/subsystem.rs
felix1209: The whole issue is related to NVMeOF protocol, the replica side responding to some old commands after the initiator restarted.

io-engine/src/subsys/nvmf/subsystem.rs
felix1209: The issue appears to be transient,  pronounced when there is a read followed by a write. However, it may not harm much to reset the unhealthy children participating in write

io-engine/src/subsys/nvmf/subsystem.rs
felix1209: reset_nexus is scheduled by the SPDK callback function and we shouldn't be holding the callback function until this completes.

io-engine/src/subsys/nvmf/subsystem.rs
felix1209: It can fail on nomem, the invalid state being passed, or busy. We have retry logic in case of a busy inside resume though It will unlikely be busy since it is already paused. Hence, it may not be useful to retry here.

io-engine/src/bdev/util/uri.rs
tiagolobocastro: It is URI parsing but it's also related to how we build our nqn, we could have built it slightly differently as the format is not very strict.

io-engine/src/subsys/nvmf/subsystem.rs
tiagolobocastro: Is there a test to confirm this PR fixes that issue?

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -165,6 +194,169 @@ fn make_sn<T: AsRef<[u8]>>(uuid: T) -> CString {
...
+            None => {
+                debug!("subsys {subsys_nqn} is not a valid nexus");
+                return;
+            }
+        };
+
+        let mut nexus = match nexus_lookup_mut(&nexus_name) {
+            Some(v) => v,
+            None => {
+                debug!("subsys {subsys_nqn} is not having valid nexus");
```
tiagolobocastro:
```suggestion
                debug!("subsys {subsys_nqn} does not contain a valid nexus name");
```

tiagolobocastro (approved):
Would be good to have a test that checks if the fio issue is resolved.
Also please squash the "stash" and "review fixes" commits.

io-engine/src/subsys/nvmf/subsystem.rs
felix1209: It is hard to reproduce the issue, hence tested it to make sure that it survives the timeout in random intervals. And, based on the tcpdump traces we collected when we hit the issue, this will fix it.

felix1209:
bors try

felix1209:
bors merge

PR #1313, opened 2023-02-07
fix: use same logic for all list_nexuses listings

tiagolobocastro:
We currently don't list nexuses when their state is Init. This is probably not good as we're hiding things from the control-plane and we should come up with a better strategy for handling this.
However, atm we're not consistent depending on the type of list nexuses so until we have a better fix we should make what we have consistent.


Abhinandan-Purkait (approved):

dsavitskiy (approved):

tiagolobocastro:
bors merge

PR #1316, opened 2023-02-09
chore(nexus/rebuild): add test assert for rebuild history

dsharma-dc:
Added a simple component test hook to assert that history is not empty.
Detailed test under pytest will be done separately.

### DIFF
--- a/test/python/v1/rebuild/features/rebuild_history.feature
+++ b/test/python/v1/rebuild/features/rebuild_history.feature
@@ -1,0 +1,20 @@
+Feature: Rebuild history
+
+    Background:
+        Given the io-engine is running and a volume is created with more than one replicas
+
+    Scenario: Record Full rebuild of a faulted replica
+        Given a volume with multiple replicas, undergoing IO
+        When one of the replica becomes faulted
+        And as a result, a new replica is hosted on a different pool
+        Then a full rebuild of the new replica starts
+        And the replica rebuild event is recorded upon rebuild completion
+        And this event history is retrievable
+
+    Scenario: Record Partial rebuild of a faulted replica
+        Given a volume with multiple replicas, undergoing IO
+        When one of the replica becomes faulted
+        And the replica comes back online on the same pool
+        Then a partial rebuild of the replica starts
+        And the replica rebuild event is recorded upon rebuild completion
+        And this event history is retrievable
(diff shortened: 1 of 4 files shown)


dsharma-dc:
bors try

test/python/v1/rebuild/features/rebuild_history.feature
```diff
@@ -0,0 +1,17 @@
+Feature: Rebuild history
+    Background:
+        Given the io-engine is running and a volume is created with more than one replicas
```
tiagolobocastro, line 4: why are there 2 different files that seem the same?

dsharma-dc: Presently I see under tests/ directory two paths where files are duplicated. One directory path is under v1, similar to how grpc code is laid out. I followed that.

tiagolobocastro: We don't need to update "v0" anymore. Only need to add new features for v1.

dsharma-dc: Updated accordingly, please take a look. The py file is marked as "to be implemented" which will be completed with GTM-352.

test/python/v1/rebuild/features/rebuild_history.feature
```diff
@@ -0,0 +1,18 @@
+Feature: Rebuild history
+
+    Background:
+        Given the io-engine is running and a volume is created with more than one replicas
+
+    Scenario: Record Full rebuild of a faulted replica
+        Given a volume with multiple replicas, undergoing IO
+        When one of the replica becomes faulted
+        And as a result, a new replica is hosted on a different pool
+        Then a full rebuild of the new replica happens
```
tiagolobocastro: Should we be explicit on when a rebuild event is added? eg:

```suggestion
        Then a full rebuild of the new replica starts
        And when the rebuild completes
```

tiagolobocastro (approved):

test/python/v1/rebuild/features/rebuild_history.feature
dsharma-dc: Modified

dsavitskiy (approved):

dsharma-dc:
bors merge

abhilashshetty04 (approved):

PR #1327, opened 2023-03-07
chore(nexus_child/fault): add fault timestamp for nexus child

dsharma-dc:


dsharma-dc:
Testing done:

Screenshot 2023-03-07 at 10 03 33 AM

dsharma-dc:
Testing done:

Screenshot 2023-03-07 at 10 08 06 AM

io-engine/src/bdev/nexus/nexus_child.rs
```diff
@@ -805,6 +808,7 @@ impl<'c> NexusChild<'c> {
         }
 
         self.set_state(ChildState::Faulted(reason));
+        self.faulted_at = Some(Utc::now());
```
tiagolobocastro, line 811: When the child is "unfaulted", should this be cleared?

dsharma-dc: We discussed that on GTM-464, and agreed that we can retain it as an indication of last faulted timestamp even though child may be currently unfaulted/healthy. Every new fault will overwrite this with latest timestamp.

tiagolobocastro: Hmm ok though not sure how we can use it, as when the nexus/process restarts well be losing this information anyway. If we need this sort of information we should persist it to etcd instead, perhaps something to consider in the future should we need to.

dsharma-dc: yeah I can record this as a future task. Moreover, when nexus restarts/moves we are not going to deal with partial rebuilds, and the purpose of adding timestamp field at this time is that this might help us take more informed decisions regarding partial rebuild.

mtzaurus (approved):

io-engine/src/bdev/nexus/nexus_child.rs
tiagolobocastro: In the future we might do partial rebuilds across restarts, and also this information might be useful for thin provisioning as well, so it's not just for partial rebuild.

tiagolobocastro (approved):

dsharma-dc:
bors try

dsharma-dc:
bors merge

PR #1328, opened 2023-03-07
feat(snapshot): common logical volume framework

hrudaya21:


hrudaya21:
bors try

hrudaya21:
bors try

io-engine/src/lvs/lvs_snapshot.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /// returns the pool uuid of the Snapshot
+    fn pool_uuid(&self) -> String {
+        self.logical_vol.lvs().uuid()
+    }
+
+    ///  wipe the first 8MB if unmap is not supported on failure the operation
+    /// needs to be repeated
+    async fn wipe_super(&self) -> Result<(), Error> {
```
mtzaurus: Snapshots are R/O entities, so they should not allow modifying underlying blocks.
It's better to move this method to Volume-specific operations.

io-engine/src/lvs/lvs_snapshot.rs
```diff
@@ -0,0 +1,74 @@
...
+        unsafe { spdk_blob_is_read_only(self.logical_vol.blob_checked()) }
+    }
+
+    /// return the size of the Snapshot in bytes
+    fn size(&self) -> u64 {
+        self.logical_vol.as_bdev().size_in_bytes()
+    }
+
+    /// Get Entity Id of the Snapshot
+    fn get_entity_id(&self) -> String {
```
mtzaurus: Could you please make this function return Option<String> ? There might be (potentially) situations when the blob property that represents entity_id can't be read/not present: default value doesn't reflect this situation properly.

io-engine/src/lvs/lvs_snapshot.rs
```diff
@@ -0,0 +1,74 @@
...
+        self.entity_id.clone().unwrap_or_default()
+    }
+
+    /// Get Group Id of the Snapshot
+    fn group_id(&self) -> String {
```
mtzaurus: The same as for 'get_entity_id()'. Thanks.

io-engine/src/lvs/lvs_snapshot.rs
```diff
@@ -0,0 +1,74 @@
...
+        self.group_id.clone().unwrap_or_default()
+    }
+
+    /// Get Parent Id of the Snapshot
+    fn get_parent_id(&self) -> String {
```
mtzaurus: It would also be nicer to make this function return Option<String>, as for the first snapshot there is no parent snapshot in the snapshot chain. Thanks.

hrudaya21: 3 Snapshot Related parameters functions return is chanegd to Option.

hrudaya21:
bors try

hrudaya21:
bors try

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -741,3 +774,7 @@ impl PtplFileOps for LvolPtpl {
             .with_extension("json")
     }
 }
+
+pub fn get_logical_volume(lvol: &Lvol) -> Box<&dyn LogicalVolume> {
```
mtzaurus: Boxing references looks a bit strange. Usually Box owns the value rather than a reference.

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -243,64 +507,117 @@ impl Share for Lvol {
...
+        &self,
+        io: *mut spdk_bdev_io,
+        snapshot_name: &str,
+    );
+    // /// Get a `PtplFileOps` from `&self`.
+    // fn ptpl(&self) -> impl PtplFileOps;
+}
+
+#[async_trait(?Send)]
+pub trait VolumeHandle {
```
mtzaurus: Why do we need this trait ?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -243,64 +507,117 @@ impl Share for Lvol {
...
+    /// Returns Lvol disk space usage.
+    fn usage(&self) -> LvolSpaceUsage;
+}
+
+/// It will give Interface Specific to Volume
+#[async_trait(? Send)]
+pub trait VolumeOps: LogicalVolume {
+    // wipe the first 8MB if unmap is not supported on failure the operation
+    // needs to be repeated
+    async fn wipe_super(&self) -> Result<(), Error>;
```
tiagolobocastro: Don't think we need this, this is used specifically for creation of new replicas?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -143,7 +143,272 @@ impl TryFrom<UntypedBdev> for Lvol {
...
+    fn subpath(&self) -> std::path::PathBuf {
+        self.lvs()
+            .subpath()
+            .join("replica/")
+            .join(self.uuid())
+            .with_extension("json")
+    }
+}
+#[derive(Copy, Clone)]
+pub struct Volume(Lvol);
```
tiagolobocastro: Why do we need this wrapper?
Also Volume is a bit misleading, this is not the Volume as we know it (from top-level).

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -317,35 +634,13 @@ impl Lvol {
...
-        self.lvs().name().to_string()
-    }
-
-    /// returns the pool uuid of the lvol
-    pub fn pool_uuid(&self) -> String {
-        self.lvs().uuid()
-    }
-
+#[async_trait(? Send)]
+impl VolumeOps for Volume {
```
tiagolobocastro: Why this separate trait?

io-engine/src/lvs/lvs_store.rs
```diff
@@ -542,23 +542,27 @@ impl Lvs {
...
-            for mut l in lvols {
-                let allowed_hosts = match l.get(PropName::AllowedHosts).await {
-                    Ok(PropValue::AllowedHosts(hosts)) => hosts,
-                    _ => vec![],
-                };
-
-                if let Ok(prop) = l.get(PropName::Shared).await {
+            for l in lvols {
+                let mut vol = Volume::from(l);
+                let vol_ops = vol.get_volume_ops_handle();
```
tiagolobocastro: Seems very verbose to always have to get `get_volume_ops_handle` when self already implements the trait, we can simply call the method right?
Btw for snapshots, if not already planned, we need a task to make sure we handle ptpl cleanup properly.

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -387,158 +682,13 @@ impl Lvol {
...
-                name: self.name(),
-            }
-        })?;
-
-        Ok(())
-    }
-
     /// get/read a property from this lvol from disk
-    pub async fn get(&self, prop: PropName) -> Result<PropValue, Error> {
+    async fn get(&self, prop: PropName) -> Result<PropValue, Error> {
```
tiagolobocastro, line 518:
```suggestion
    async fn read_property(&self, prop: PropName) -> Result<PropValue, Error>;
```
?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -243,64 +507,117 @@ impl Share for Lvol {
...
+    async fn wipe_super(&self) -> Result<(), Error>;
+
+    /// returns a boolean indicating if the lvol is a snapshot
+    fn is_snapshot(&self) -> bool;
+
+    /// get/read a property from this lvol from disk
+    async fn get(&self, prop: PropName) -> Result<PropValue, Error>;
+
+    /// Create a snapshot
+    async fn create_snapshot(&self, nvmf_req: &NvmfReq, snapshot_name: &str);
```
tiagolobocastro, line 432: We shouldn't box it unless we need to store this in some kind of struct for example? Maybe caller should be the one to determine.

You can also simply impl this against the actual type, otherwise seems a bit odd to add a trait so you can call a method to get self as another trait?

```suggestion
impl Lvol {
    fn as_lv(&self) -> impl LogicalVolume { self } 
}
```

But do we need this at all since self is already a `LogicalVolume` ?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -243,64 +507,117 @@ impl Share for Lvol {
...
+pub trait VolumeOps: LogicalVolume {
+    // wipe the first 8MB if unmap is not supported on failure the operation
+    // needs to be repeated
+    async fn wipe_super(&self) -> Result<(), Error>;
+
+    /// returns a boolean indicating if the lvol is a snapshot
+    fn is_snapshot(&self) -> bool;
+
+    /// get/read a property from this lvol from disk
+    async fn get(&self, prop: PropName) -> Result<PropValue, Error>;
```
tiagolobocastro, line 408:
```suggestion
    async fn read_property(&self, prop: PropName) -> Result<PropValue, Error>;
```
?

io-engine/src/lvs/lvs_lvol.rs
tiagolobocastro: You can simply call the functions is self already impls the trait, which is why you can call `get_logical_volume_handle` for example.
I think we can do without trait altogether.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: With the new changes, It is kept as it is and moved from direct call to SpdkLvol trait call. This is the common trait to provide interface to all the Lvol Common API

tiagolobocastro: hmm I don't think this makes sense to have in the common interface api because this is a very specific use case for the lvol when it's first created and it's only needed internally within the lvol creation path. Snapshots et all would not need this at all.

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -289,18 +289,184 @@ impl Lvol {
...
+
+    /// return the size of the Logical Volume in bytes
+    fn size(&self) -> u64;
+
+    /// Returns Lvol disk space usage.
+    fn usage(&self) -> LvolSpaceUsage;
+}
+
+#[async_trait(?Send)]
+pub trait SpdkLvol: LogicalVolume + Share {
```
tiagolobocastro: What's the difference between `SpdkLvol` and `LogicalVolume` ?

hrudaya21: There is a plan to implement a new test framwork from LogicalVolume trait. Hence certein fileds are exposed through LogicalVolume which will have different implementation in the test framwork.

io-engine/src/lvs/lvs_lvol.rs
mtzaurus: We will have to get rid of NVMe-specific arguments in generic volume operations during implementation of snapshots.

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -289,18 +289,184 @@ impl Lvol {
...
+
+    /// Format snapshot name
+    /// base_name is the nexus or replica UUID
+    fn format_snapshot_name(base_name: &str, snapshot_time: u64) -> String;
+
+    /// Create a snapshot
+    async fn create_snapshot(&self, nvmf_req: &NvmfReq, snapshot_name: &str);
+
+    /// Create snapshot for local replica
+    async fn create_snapshot_local(
```
mtzaurus, line 435: This one will also be removed later when implementing snapshots.

io-engine/src/bdev/nexus/nexus_bdev_snapshot.rs
```diff
@@ -3,7 +3,7 @@
 use mayastor_api::v0::CreateSnapshotReply;
 
 use super::{Error, Nexus};
-use crate::lvs::Lvol;
+use crate::lvs::{lvs_lvol::SpdkLvol, Lvol};
```
mtzaurus: Why should nexus know about this trait ?

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: Removed this API from the trait interface, kept it as direct Volume API as before.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: Currently not handled as part of this PR. As discussed, We will handle it when the Snapshot Trait will be added.

hrudaya21:
bors try

mtzaurus (approved):

io-engine/src/lvs/lvs_lvol.rs
tiagolobocastro: I get the need for the interface just confused here with the different between `SpdkLvol` and `LogicalVolume`.
Is it to keep spdk specific things out of LogicalVolume? 
If so then should this be moved outside of lvs? Also the lvs member function would have to be removed. Maybe I'd also suggest renaming `SpdkLvol` to `LvsLvol`.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: It is handled in the current changes. The older changes with Box based dyn trait is removed now.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: Handled as per the suggestion. Please check once.

hrudaya21:
bors try

tiagolobocastro (approved):

hrudaya21:
bors try

hrudaya21:
bors merge

PR #1333, opened 2023-03-16
feat(snapshot): implement snapshot for lvol

hrudaya21:
Implement Snapshot for Lvol

### DIFF
--- a/io-engine/src/core/snapshot.rs
+++ b/io-engine/src/core/snapshot.rs
@@ -1,0 +1,74 @@
+use async_trait::async_trait;
+
+/// Snapshot Captures all the Snapshot information for Lvol.
+#[derive(Debug)]
+pub struct SnapshotParams {
+    entity_id: Option<String>,
+    parent_id: Option<String>,
+    txn_id: Option<String>,
+    snap_name: Option<String>,
+}
+/// Implement Snapshot Common Function.
+impl SnapshotParams {
+    pub fn new(
+        entity_id: Option<String>,
+        parent_id: Option<String>,
+        txn_id: Option<String>,
+        snap_name: Option<String>,
+    ) -> SnapshotParams {
+        SnapshotParams {
+            entity_id,
+            parent_id,
+            txn_id,
+            snap_name,
+        }
+    }
+}
+
+///  Traits gives the common snapshot/clone interface for Local/Remote Lvol.
+#[async_trait(?Send)]
+pub trait SnapshotOps {
+    type Error;
+    /// Create Snapshot Common API.
+    async fn create_snapshot(
+        &self,
+        snap_param: SnapshotParams,
+    ) -> Result<(), Self::Error>;
+}
+/// Traits gives the Snapshots Related Parameters.
+pub trait SnapshotDescriptor {
+    /// Get Transaction Id of the Snapshot Create.
+    fn txn_id(&self) -> Option<String>;
+
+    /// Get Entity Id of the Snapshot.
+    fn entity_id(&self) -> Option<String>;
+
+    /// Get Parent Id of the Snapshot.
+    fn parent_id(&self) -> Option<String>;
+
+    /// Get Snapshot Name.
+    fn name(&self) -> Option<String>;
+}
+
+/// Trait to give interface for all Snapshot Parameters.
+impl SnapshotDescriptor for SnapshotParams {
+    /// Get Trasanction Id of the Snapshot.
+    fn txn_id(&self) -> Option<String> {
+        self.txn_id.clone()
+    }
+
+    /// Get Entity Id of the Snapshot.
+    fn entity_id(&self) -> Option<String> {
+        self.entity_id.clone()
+    }
+
+    /// Get Parent Id of the Snapshot.
+    fn parent_id(&self) -> Option<String> {
+        self.parent_id.clone()
+    }
+
+    /// Get Snapname of the Snapshot.
+    fn name(&self) -> Option<String> {
+        self.snap_name.clone()
+    }
+}
--- a/io-engine/src/lvs/lvs_error.rs
+++ b/io-engine/src/lvs/lvs_error.rs
@@ -108,4 +108,9 @@ pub enum Error {
     ReplicaShareProtocol {
         value: i32,
     },
+    #[snafu(display("Snapshot {} created with Resultcode {}", msg, source))]
+    SnapshotCreate {
+        source: Errno,
+        msg: String,
+    },
 }
--- a/io-engine/src/lvs/lvs_lvol.rs
+++ b/io-engine/src/lvs/lvs_lvol.rs
@@ -14,8 +14,6 @@ use nix::errno::Errno;
 use pin_utils::core_reexport::fmt::Formatter;
 
 use spdk_rs::libspdk::{
-    spdk_bdev_io,
-    spdk_bdev_io_get_thread,
     spdk_blob,
     spdk_blob_calc_used_clusters,
     spdk_blob_get_num_clusters,
@@ -26,7 +24,6 @@ use spdk_rs::libspdk::{
     spdk_blob_sync_md,
     spdk_bs_get_cluster_size,
     spdk_lvol,
-    spdk_nvmf_request_complete,
     vbdev_lvol_create_snapshot,
     vbdev_lvol_destroy,
     vbdev_lvol_get_from_bdev,
@@ -37,15 +34,17 @@ use spdk_rs::libspdk::{
 use super::{Error, Lvs};
 
 use crate::{
-    bdev::{nexus::Nexus, PtplFileOps},
+    bdev::PtplFileOps,
     core::{
         logical_volume::LogicalVolume,
+        snapshot::SnapshotDescriptor,
         Bdev,
-        Mthread,
         Protocol,
         PtplProps,
         Share,
         ShareProps,
+        SnapshotOps,
+        SnapshotParams,
         UntypedBdev,
         UpdateProps,
     },
@@ -59,6 +58,7 @@ use crate::{
     },
     subsys::NvmfReq,
 };
+use spdk_rs::libspdk::spdk_nvmf_request_complete;
 
 // Wipe `WIPE_SUPER_LEN` bytes if unmap is not supported.
 pub(crate) const WIPE_SUPER_LEN: u64 = (1 << 20) * 8;
@@ -428,13 +428,10 @@ pub trait LvsLvol: LogicalVolume + Share {
     /// Write the property prop on to the lvol which is stored on disk
     async fn sync_metadata(self: Pin<&mut Self>) -> Result<(), Error>;
 
-    /// Create a snapshot
-    async fn create_snapshot(&self, nvmf_req: &NvmfReq, snapshot_name: &str);
-
-    /// Create snapshot for local replica
-    async fn create_snapshot_local(
+    /// Create a snapshot in Remote
+    async fn create_snapshot_remote(
         &self,
-        io: *mut spdk_bdev_io,
+        nvmf_req: &NvmfReq,
         snapshot_name: &str,
     );
 }
@@ -715,9 +712,12 @@ impl LvsLvol for Lvol {
 
         Ok(())
     }
-
-    /// Create a snapshot
-    async fn create_snapshot(&self, nvmf_req: &NvmfReq, snapshot_name: &str) {
+    /// Create a snapshot in Remote
+    async fn create_snapshot_remote(
+        &self,
+        nvmf_req: &NvmfReq,
+        snapshot_name: &str,
+    ) {
         extern "C" fn snapshot_done_cb(
             nvmf_req_ptr: *mut c_void,
             _lvol_ptr: *mut spdk_lvol,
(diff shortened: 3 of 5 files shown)


io-engine/src/core/volume_snapshot.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    /// Create Snapshot (Readonly Copy) for Remote Volume
+    async fn create_snapshot_remote(
+        &self,
+        snap_name: &str,
+        nvmf_req: &NvmfReq,
+    ) -> Result<u64, CoreError>;
+
+    /// Return the list of Snapshots
+    fn list_snapshot(&self, source_vol: &str) -> Vec<Self::Snapshot>;
```
mtzaurus: What's the semantics of this method ? List snapshots for a snapshot ?

io-engine/src/core/volume_snapshot.rs
```diff
@@ -0,0 +1,36 @@
...
+use async_trait::async_trait;
+use spdk_rs::libspdk::spdk_bdev_io;
+
+///  Trait gives the interface for all Snapshot and Clone related interfaces
+#[async_trait(?Send)]
+pub trait VolumeSnapshot {
+    type Snapshot;
+
+    /// Create Snapshot (Readonly Copy) for Local Volume
+    async fn create_snapshot_local(
```
mtzaurus: IMHO, 'create_snapshot_....' methods should belong to a different trait - to something, that supports creation of snapshots (volumes in our case). Otherwise it means ability to create snapshots from snapshots. :-)

hrudaya21: Ok, it make sense to have a different trait for local/remote snapshot and from outside, it can be just one API. Let me try.

io-engine/src/core/volume_snapshot.rs
hrudaya21: It means, to list the snapshots of of a particular source volume, from which multiple snapshots are created.

felix1209: I think that list_snapshot should go under the volume trait

io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -0,0 +1,260 @@
...
+/// Trait to give all the common interface of LvolSnapshot.
+trait LvolSnapshotParam {
+    /// Get Group Id of the Snapshot.
+    fn group_id(&self) -> Option<String>;
+
+    /// Get Entity Id of the Snapshot.
+    fn entity_id(&self) -> Option<String>;
+
+    /// Get Parent Id of the Snapshot.
+    fn parent(&self) -> Option<String>;
```
felix1209: This parent id refers to entity_id or txn_id of the parent. you may wan to be specific.  BTW, you may have to search/replace all the group_id into txn_id

io-engine/src/core/volume_snapshot.rs
hrudaya21: Updated the Code, please check once.

io-engine/src/core/volume_snapshot.rs
```diff
@@ -0,0 +1,37 @@
+use super::CoreError;
+use crate::{lvs::LvolSnapshot, subsys::NvmfReq};
+use async_trait::async_trait;
+use spdk_rs::libspdk::spdk_bdev_io;
+
+///  Traits gives the common snapshot/clone interface for Local/Remote Lvol.
+#[async_trait(?Send)]
+pub trait VolumeSnapshotOps {
```
tiagolobocastro: Seems like we could split this trait into 2?
Example we want to provide the "getters" here for snapshot, and they don't need to be an option then?
And we don't wan the snapshot itself to impl create_snapshot?

io-engine/src/core/volume_snapshot.rs
```diff
@@ -0,0 +1,46 @@
...
+    /// Create Snapshot Common API.
+    async fn create_snapshot(&self) -> Result<u64, CoreError>;
+
+    /// Return the list of Snapshots for Local/remote Device.
+    fn list_snapshot(&self, source_vol: &str) -> Vec<LvolSnapshot>;
+}
+pub trait VolumeSnapshotParams {
+    /// Get Transaction Id of the Snapshot Create.
+    fn get_txn_id(&self) -> Option<String>;
```
tiagolobocastro: nit: in rust getters tend not to be prefixed with "get_", so this could be come simply `txn_id`.
Now that we have a separate trait, we don't need to return Option<String>?
Simply this would do?
`fn txn_id(&self) -> String;`

io-engine/src/core/volume_snapshot.rs
```diff
@@ -0,0 +1,46 @@
...
+use spdk_rs::libspdk::spdk_bdev_io;
+
+///  Traits gives the common snapshot/clone interface for Local/Remote Lvol.
+#[async_trait(?Send)]
+pub trait VolumeSnapshotOps {
+    /// Create Snapshot Common API.
+    async fn create_snapshot(&self) -> Result<u64, CoreError>;
+
+    /// Return the list of Snapshots for Local/remote Device.
+    fn list_snapshot(&self, source_vol: &str) -> Vec<LvolSnapshot>;
```
mtzaurus: No need to use 'source_vol', IMHO: we list snapshots for 'self' volume anyways.

io-engine/src/core/volume_snapshot.rs
```diff
@@ -0,0 +1,46 @@
...
+    fn entity_id(&self) -> Option<String>;
+
+    /// Get Parent Id of the Snapshot.
+    fn parent(&self) -> Option<String>;
+
+    /// Get Snapshot Name.
+    fn snap_name(&self) -> Option<String>;
+
+    /// Get Lvol.
+    fn lvol(&self) -> &Lvol;
```
mtzaurus: What kind of volume are we obtaining ? The parent of the snapshot ?

tiagolobocastro: How come we can get the parent directly, but not the parent id?

io-engine/src/core/volume_snapshot.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+pub trait VolumeSnapshotParams {
+    /// Get Transaction Id of the Snapshot Create.
+    fn txn_id(&self) -> Option<String>;
+
+    /// Get Entity Id of the Snapshot.
+    fn entity_id(&self) -> Option<String>;
+
+    /// Get Parent Id of the Snapshot.
+    fn parent(&self) -> Option<String>;
```
tiagolobocastro:
```suggestion
    fn parent_id(&self) -> Option<String>;
```

io-engine/src/core/volume_snapshot.rs
```diff
@@ -0,0 +1,46 @@
...
+
+    /// Get Snapshot Name.
+    fn snap_name(&self) -> Option<String>;
```
tiagolobocastro:
```suggestion
    fn name(&self) -> Option<String>;
```

io-engine/src/core/volume_snapshot.rs
tiagolobocastro: On another note, should this be in this trait? Seems we're constraining this snapshot to be a blobstor snapshot?
Maybe we could have another trait, for example:
pub trait LvolSnapshotParams: VolumeSnapshotParams { {
   /// Get Lvol.
    fn lvol(&self) -> &Lvol;
}

io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -0,0 +1,230 @@
...
+        LvolSnapshotBdev {
+            lvol_snapshot,
+            req_type,
+        }
+    }
+}
+
+/// Implement Snapshot functionality for Lvol.
+#[async_trait(?Send)]
+impl VolumeSnapshotOps for LvolSnapshotBdev {
```
tiagolobocastro: We're taking a snapshot out of a snapshot?

io-engine/src/core/volume_snapshot.rs
```diff
@@ -0,0 +1,46 @@
...
+///  Traits gives the common snapshot/clone interface for Local/Remote Lvol.
+#[async_trait(?Send)]
+pub trait VolumeSnapshotOps {
+    /// Create Snapshot Common API.
+    async fn create_snapshot(&self) -> Result<u64, CoreError>;
+
+    /// Return the list of Snapshots for Local/remote Device.
+    fn list_snapshot(&self, source_vol: &str) -> Vec<LvolSnapshot>;
+}
+pub trait VolumeSnapshotParams {
```
tiagolobocastro: nit: Do we want all of these prefixed by Volume? Could this cause some confusion since they're not really a mayastor volume as we know it.

io-engine/src/core/snapshot.rs
```diff
@@ -0,0 +1,76 @@
...
+#[derive(Debug)]
+pub struct SnapshotParams {
+    entity_id: Option<String>,
+    parent_id: Option<String>,
+    txn_id: Option<String>,
+    snap_name: Option<String>,
+}
+/// Implement Snapshot Common Function.
+impl SnapshotParams {
+    pub fn init_param(
```
mtzaurus: Canonical way is to use 'new()'.

io-engine/src/core/snapshot.rs
```diff
@@ -0,0 +1,76 @@
...
+#[async_trait(?Send)]
+pub trait SnapshotOps {
+    /// Create Snapshot Common API.
+    async fn create_snapshot(
+        &self,
+        snap_param: SnapshotParams,
+    ) -> Result<u64, CoreError>;
+
+    /// Return the list of Snapshots for Local/remote Device.
+    fn list_snapshot(&self) -> Vec<SnapshotParams>;
```
mtzaurus: Not sure we should return 'SnapshotParams' here rather than a list of snapshots.
I'd suggest removing this method for now and leave only 'create_snapshot()' - we will add snapshot listing later.

io-engine/src/subsys/nvmf/admin_cmd.rs
```diff
@@ -1,167 +1,168 @@
...
-/// Register custom NVMe admin command handler
-pub fn setup_create_snapshot_hdlr() {
-    unsafe {
-        spdk_nvmf_set_custom_admin_cmd_hdlr(
-            nvme_admin_opc::CREATE_SNAPSHOT,
-            Some(nvmf_create_snapshot_hdlr),
-        );
-    }
-}
+// //! Handlers for custom NVMe Admin commands
```
mtzaurus: Do we need to comment all the code out ? If it's not used we can mark it as accordingly to keep linter happy and just keep it as it is.

io-engine/src/subsys/nvmf/mod.rs
```diff
@@ -98,8 +97,8 @@ impl Nvmf {
 
         // this code only ever gets run on the first core
 
-        // set up custom NVMe Admin command handler
-        admin_cmd::setup_create_snapshot_hdlr();
+        // // set up custom NVMe Admin command handler
```
mtzaurus: Can you please add a comment here why we comment this code.

mtzaurus (approved):

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -715,77 +704,37 @@ impl LvsLvol for Lvol {
...
                 c_snapshot_name.as_ptr(),
-                Some(snapshot_done_cb),
-                io.cast(),
+                Some(snapshot_create_done_cb),
+                bio_ptr,
             )
         };
-
-        info!("{:?}: creating snapshot '{}'", self, snapshot_name);
+        info!("{:?}: creating snapshot '{}'", self, self.name());
```
mtzaurus: We should wait till the snapshot is created by 'vbdev_lvol_create_snapshot', shouldn't we ?

mtzaurus (changes requested):
Snapshot creation callback should notify the waiter via a channel like we do when call other SPDK functions,

io-engine/src/core/snapshot.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+///  Traits gives the common snapshot/clone interface for Local/Remote Lvol.
+#[async_trait(?Send)]
+pub trait SnapshotOps {
+    type Error;
+    /// Create Snapshot Common API.
+    async fn create_snapshot(
+        &self,
+        snap_param: SnapshotParams,
```
tiagolobocastro, line 35: Shouldn't creation itself require specific values for the id's? 
Seems like this api allows all of the none, which seems a bit odd.

io-engine/src/lvs/lvs_error.rs
```diff
@@ -108,4 +108,9 @@ pub enum Error {
     ReplicaShareProtocol {
         value: i32,
     },
+    #[snafu(display("Create Snapshot with ResultCode {}", source))]
```
tiagolobocastro:
```suggestion
    #[snafu(display("Failed to create snapshot with errno {} and message {}", source, msg))]
```
?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -715,77 +703,43 @@ impl LvsLvol for Lvol {
...
+                cb_arg(s),
             )
         };
-
-        info!("{:?}: creating snapshot '{}'", self, snapshot_name);
+        r.await
+            .expect("snapshot_create_done_cb")
+            .to_result(|error| Error::SnapshotCreate {
+                source: Errno::from_i32(error),
+                msg: "Create Snapshot with ResultCode ".to_string(),
```
tiagolobocastro: oh msg here seems kind of redundant?

io-engine/src/subsys/nvmf/admin_cmd.rs
```diff
@@ -9,60 +9,65 @@ use std::{
...
 #[derive(Clone)]
 pub struct NvmeCpl(pub(crate) NonNull<spdk_nvme_cpl>);
-
-impl NvmeCpl {
-    /// Returns the NVMe status
-    pub(crate) fn status(&mut self) -> &mut spdk_nvme_status {
-        unsafe { &mut *nvme_status_get(self.0.as_mut()) }
-    }
-}
+/// TO-DO: Currently Snapshot Creation in Remote for Lvol is not supported,
```
tiagolobocastro: Can't we just remove this if it's not used?

io-engine/src/subsys/nvmf/admin_cmd.rs
```diff
@@ -132,10 +137,21 @@ extern "C" fn nvmf_create_snapshot_hdlr(req: *mut spdk_nvmf_request) -> i32 {
         };
         let snapshot_name =
             Lvol::format_snapshot_name(&lvol.name(), snapshot_time);
-        let nvmf_req = NvmfReq(NonNull::new(req).unwrap());
+        let _nvmf_req = NvmfReq(NonNull::new(req).unwrap());
+        let snap_param = SnapshotParams::new(
+            Some(lvol.name()),
+            Some(lvol.name()),
+            Some(lvol.name()),
```
tiagolobocastro: everything is the lvol name?

io-engine/src/subsys/nvmf/admin_cmd.rs
mtzaurus: IMHO, we can keep it as potential code source for now, we can wipe out all unused code later once refactored.

mtzaurus (approved):

io-engine/src/subsys/nvmf/admin_cmd.rs
tiagolobocastro: It's in the git history, it won't go away :)

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -715,77 +703,43 @@ impl LvsLvol for Lvol {
...
+                cb_arg(s),
             )
         };
-
-        info!("{:?}: creating snapshot '{}'", self, snapshot_name);
+        r.await
+            .expect("snapshot_create_done_cb")
+            .to_result(|error| Error::SnapshotCreate {
+                source: Errno::from_i32(error),
+                msg: format!("Snapshot Created with ResultCode {error}"),
```
tiagolobocastro: IMHO better to remove msg from here and add it to the snafu display as that's the whole point of using such as library.

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -715,77 +703,43 @@ impl LvsLvol for Lvol {
...
-            // Must complete IO on thread IO was submitted from
-            Mthread::from_ptr(unsafe {
-                spdk_bdev_io_get_thread(bio_ptr.cast())
-            })
-            .with(|| Nexus::io_completion_local(errno == 0, bio_ptr));
+            s.send(errno).ok();
         }
 
-        let c_snapshot_name = snapshot_name.into_cstring();
+        let c_snapshot_name = snap_param.name().unwrap().into_cstring();
```
tiagolobocastro, line 779: need to handle when name is None to avoid panic.

io-engine/src/subsys/nvmf/admin_cmd.rs
```diff
@@ -144,15 +160,25 @@ extern "C" fn nvmf_create_snapshot_hdlr(req: *mut spdk_nvmf_request) -> i32 {
...
+    let snap_param = SnapshotParams::new(
+        Some(lvol.name()),
+        Some(lvol.name()),
+        Some(lvol.name()),
+        Some(snapshot_name),
+    );
     // Blobfs operations must be on md_thread
     Reactors::master().send_future(async move {
-        lvol.create_snapshot_local(io, &snapshot_name).await;
+        lvol.create_snapshot(snap_param).await.unwrap();
```
tiagolobocastro: should handle error rather than panic?

tiagolobocastro (approved):

hrudaya21:
bors try

hrudaya21:
bors try

hrudaya21:
bors merge

PR #1335, opened 2023-03-20
Fix rebuild channel hang

tiagolobocastro:
fix(rebuild): don't clone receiver channel

When creating a rebuild job we were reusing the same comms type which is used to do bi-dir
communication between the rebuild frontend and backend.
In some situations this could lead into blocking stats because the receiver channel was kept
alive by the frontend even after the frontend terminates.

Force closing the channel should in theory help but it seems the queue is not drained on close.
See: async-channel/issues/23

Instead, let's split the channels into two types meaning only the backend has the receiver side.

---

fix(rebuild): set final rebuild stats

When rebuild completes, right before the nexus completes the rebuild callback it's possible
that we get a stats request and in such case we have no stats to give.
To address this, then the rebuild backend completes it now sets the final rebuild stats in
the states, which will allow the stats to be retrieved.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,5 @@ ansible-hosts
 /test/python/venv/*
 /package.json
 /rust-toolchain.toml
+/test/python/registration_pb2.py
+/test/python/registration_pb2_grpc.py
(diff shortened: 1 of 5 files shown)


dsavitskiy (approved):

mtzaurus (approved):

tiagolobocastro:
bors merge

PR #1340, opened 2023-03-26
fix(child/fault): set fault timestamp in all possible places

dsharma-dc:
It would be ideal to set the timestamp from within set_state(), however that would required mutability changes for `self` in function parameters, which may propagate up. So simply setting the timestamp wherever we are currently setting child state as Faulted.


tiagolobocastro:
> It would be ideal to set the timestamp from within set_state(), however that would required mutability changes for `self` in function parameters, which may propagate up. So simply setting the timestamp wherever we are currently setting child state as Faulted.

Can we change mutability of `set_state` ? Or have some sort of a fault wrapper?

dsavitskiy:
@dsharma-dc I see your point about &self mutability in set_state. Still, a much better design decision would be to set "fault_at" in set_state, as otherwise it is very error prone.

You don't actually have to change &self to mut in set_state().

You can change "faulted_at" to be AtomicU64 (like "state" and "prev_state"), and store time since epoch there. To access it in form of Option<DateTime>, add an accessor method "get_fault_at()" that would convert u64 to Option<DateTime>.
You can use one the following conventions:
- faulted_at == 0 means None, and > 0 Some(DateTime).
- check if state is Faulted and if not , return None, otherwise Some(DateTime)

dsavitskiy:
> > It would be ideal to set the timestamp from within set_state(), however that would required mutability changes for `self` in function parameters, which may propagate up. So simply setting the timestamp wherever we are currently setting child state as Faulted.
> 
> Can we change mutability of `set_state` ? Or have some sort of a fault wrapper?

We don't need to change to &mut, see my comment above.

dsavitskiy (changes requested):
Please move setting fault_at to set_state().

dsharma-dc:
Testing:

```

NAME                                                    STATE   REASON    LAST_FAULTED_AT     
bdev:///repl1?uuid=da778218-7591-40ea-95fd-c891f8591c89 online  -         -                   
bdev:///repl2?uuid=5d271c87-42cc-46ff-8d34-18aa052d5cb2 faulted by client 2023-03-28T06:07:42Z

Removed bdev:///repl2?uuid=5d271c87-42cc-46ff-8d34-18aa052d5cb2 from specified nexus
f1350be1-75c4-40dd-90e5-a20bffcb619d
NAME                                                    STATE    REASON      LAST_FAULTED_AT
bdev:///repl1?uuid=da778218-7591-40ea-95fd-c891f8591c89 online   -           -              
bdev:///repl2?uuid=5d271c87-42cc-46ff-8d34-18aa052d5cb2 degraded out of sync -  
```

io-engine/src/bdev/nexus/nexus_child.rs
```diff
@@ -308,6 +310,16 @@ impl<'c> NexusChild<'c> {
...
         let prev_state = self.state.swap(state);
         self.prev_state.store(prev_state);
+
+        if let ChildState::Faulted(_) = state {
+            self.faulted_at.store(
+                SystemTime::now()
+                    .duration_since(UNIX_EPOCH)
+                    .unwrap_or_default()
+                    .as_secs(),
+                Ordering::Relaxed,
```
dsavitskiy, line 320: It makes to use AcqRel instead of Relaxed.

io-engine/src/bdev/nexus/nexus_child.rs
```diff
@@ -419,6 +431,17 @@ impl<'c> NexusChild<'c> {
         self.sync_state
     }
 
+    /// Return the last fault timestamp of the child.
+    pub fn fault_timestamp(&self) -> Option<DateTime<Utc>> {
+        if self.faulted_at.load(Ordering::Relaxed) == 0 {
+            None
+        } else {
+            let ts = self.faulted_at.load(Ordering::Relaxed);
```
dsavitskiy: Why loading two times?

io-engine/src/bdev/nexus/nexus_child.rs
```diff
@@ -419,6 +431,17 @@ impl<'c> NexusChild<'c> {
         self.sync_state
     }
 
+    /// Return the last fault timestamp of the child.
+    pub fn fault_timestamp(&self) -> Option<DateTime<Utc>> {
+        if self.faulted_at.load(Ordering::Relaxed) == 0 {
```
dsavitskiy: It makes to use AcqRel instead of Relaxed.

dsavitskiy (changes requested):
Please rebase on develop and squash your commits into single one before merging.

io-engine/src/bdev/nexus/nexus_child.rs
dsharma-dc: Looks redundant yes, changing to this:

 ```
   pub fn fault_timestamp(&self) -> Option<DateTime<Utc>> {
        let ts = self.faulted_at.load(Ordering::Relaxed);
        if ts == 0 {
            None
        }
        NaiveDateTime::from_timestamp_opt(ts as i64, 0)
            .map(|d| DateTime::from_utc(d, Utc))
    }
```

tiagolobocastro:
> @dsharma-dc I see your point about &self mutability in set_state. Still, a much better design decision would be to set "fault_at" in set_state, as otherwise it is very error prone.
> 
> You don't actually have to change &self to mut in set_state().
> 
> You can change "faulted_at" to be AtomicU64 (like "state" and "prev_state"), and store time since epoch there. To access it in form of Option, add an accessor method "get_fault_at()" that would convert u64 to Option. You can use one the following conventions:
> 
>     * faulted_at == 0 means None, and > 0 Some(DateTime).
> 
>     * check if state is Faulted and if not , return None, otherwise Some(DateTime)

A downside of this is that the timestamp does not fit into 64bits so we'll loose resolution.

dsharma-dc:
Something wrong happened with the rebase and unnecessary commits are showing up in this PR.  :( 
Fixing that.

PR #1341, opened 2023-03-27
feat(snapshot): getsnapshot api for lvol

hrudaya21:


dsavitskiy (approved):

io-engine/src/core/snapshot.rs
```diff
@@ -34,7 +35,11 @@ pub trait SnapshotOps {
         &self,
         snap_param: SnapshotParams,
     ) -> Result<(), Self::Error>;
+
+    // Get Snapshot Iterator
+    async fn get_snapshot(self) -> LvolSnapshotIter;
```
tiagolobocastro:
```suggestion
    /// Get a Snapshot Iterator.
    async fn snapshot_iter(self) -> LvolSnapshotIter;
```

tiagolobocastro: Should this be `LvolSnapshotIter` here? Seems like it's tying the trait to the LvolSnapshot iterator but it should be a generic snapshop iterator?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -754,6 +774,58 @@ impl LvsLvol for Lvol {
...
+        unsafe {
+            spdk_bs_iter_next(
+                self.lvs().blob_store(),
+                curr_blob,
+                Some(Self::blob_op_complete_cb),
+                cb_arg(s),
+            )
+        };
+        match r.await {
+            Ok((blob, _err)) => Some(blob),
```
tiagolobocastro, line 813: What is _err here, sounds like it could be an error, if so, should we handle it or we don't care?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -754,6 +774,58 @@ impl LvsLvol for Lvol {
...
+            Err(_) => None,
+        }
+    }
+
+    /// Build Snapshot Parameters from Blob.
+    fn build_snapshot_param(&self, _blob: *mut spdk_blob) -> SnapshotParams {
+        // TO-DO need to Integrate with Snapshot Property Enumeration
```
```suggestion
        // TODO: need to Integrate with Snapshot Property Enumeration
```

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: Currently the error is not handled directly. The blob will be NULL if there will be any error and next iterator will be exit.

io-engine/src/core/snapshot.rs
hrudaya21: Yes it should be generic. modified. Thanks

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: Changed.

io-engine/src/lvs/lvol_snapshot_iter.rs
```diff
@@ -0,0 +1,45 @@
...
+use spdk_rs::libspdk::spdk_blob;
+
+use crate::{
+    core::SnapshotParams,
+    lvs::{lvs_lvol::LvsLvol, Lvol},
+};
+#[async_trait(?Send)]
+trait AsyncIterator {
+    type Item;
+    async fn next(&mut self) -> Option<SnapshotParams>;
```
mtzaurus, line 11: Can't we iterate over snapshot objects here rather than SnapshotParams ?
IMHO, getting a sequence of snapshot objects which we can work with might bring us benefits.

mtzaurus (approved):

hrudaya21:
bors try

hrudaya21:
bors merge

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -754,12 +774,65 @@ impl LvsLvol for Lvol {
...
+    extern "C" fn blob_op_complete_cb(
+        arg: *mut c_void,
+        blob: *mut spdk_blob,
+        errno: i32,
+    ) {
+        let s = unsafe {
+            Box::from_raw(arg as *mut oneshot::Sender<(*mut spdk_blob, i32)>)
+        };
+        if errno != 0 {
+            error!("Blobstore Operation failed, errno {}", errno);
```
tiagolobocastro, line 788: would be good to include something about the blobstore, otherwise we don't know from the logs which blosbstor operation failed?

io-engine/src/lvs/lvol_snapshot_iter.rs
tiagolobocastro: how can we even get the params without accessing the snapshots themselves? Maybe I'm missing something.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: Yes.. I think it make sense. I will update the blobstor information as part of error message in the next PR.

io-engine/src/lvs/lvol_snapshot_iter.rs
hrudaya21: 1.  next() will get the next blob and use the prev blob information to fill the snapshot parameters.
2. build_snapshot_param(curr_blob) will be implemented by accessing the Snapshot Attributes for each `blob`.
3. A SPDK API is introduced to get the SnapshotAttr which is pending for integration with build_snapshot_param() function.

tiagolobocastro: Right, so if we get the blob, then we can just return the snapshot objects straight away rather than the params?

hrudaya21: Yes.. Returning of the Snapshot Object from blob, in place of SnapshotParams, will be part of another PR, along with new SPDK API integration. @mtzaurus is working on it.

PR #1342, opened 2023-03-27
fix(lvs/replica): destroy using pool identifier

tiagolobocastro:
Adding the pool to the request allows us determine whether a pool is loaded or not. Example: if a pool is not loaded then we cannot destroy the replica and we must return an adequate error code back as it's the control-plane job to import pools before attempting to destroy replicas.


tiagolobocastro:
bors try

dsharma-dc (approved):
nitpick: The commit title mentions "parent pool", which sounds like there is some hierarchical relation between 'pools' rather. Can we just say "destroy replica using pool identifier" ?

io-engine/src/grpc/v1/replica.rs
```diff
@@ -203,22 +203,57 @@ impl ReplicaRpc for ReplicaService {
...
+                let lvol = Bdev::lookup_by_uuid_str(&args.uuid)
+                    .and_then(|b| Lvol::try_from(b).ok())
+                    .ok_or(LvsError::RepDestroy {
+                        source: Errno::ENOENT,
+                        name: args.uuid.to_owned(),
+                        msg: "".into(),
+                    })?;
+
+                if let Some(lvs) = lvs {
+                    let same = lvs.name() != lvol.pool_name()
```
mtzaurus: I might be overlooking, but isn't the naming a bit confusing here ? Intuitively, 'same' is when something equals, however we check the inequality here. Also, 'same' will be true when name doesn't match the pool name, but in the error message below we inform about equality ('does match').

tiagolobocastro: Yes indeed, amended.

dsavitskiy (approved):

mtzaurus (approved):

tiagolobocastro:
bors merge

PR #1343, opened 2023-03-28
fix(child/fault): set fault timestamp in all possible places

dsharma-dc:


dsharma-dc:
Created this PR as a replacement of https://github.com/openebs/mayastor/pull/1340 because the branch got messed up there.

io-engine/src/bdev/nexus/nexus_child.rs
```diff
@@ -419,6 +431,16 @@ impl<'c> NexusChild<'c> {
...
 
+    /// Return the last fault timestamp of the child.
+    pub fn fault_timestamp(&self) -> Option<DateTime<Utc>> {
+        let ts = self.faulted_at.load(Ordering::Relaxed);
+        if ts == 0 {
+            return None;
+        }
+        NaiveDateTime::from_timestamp_opt(ts as i64, 0)
+            .map(|d| DateTime::from_utc(d, Utc))
+    }
```
dsavitskiy, line 431: Nitpicking: you could use more a idiomatic Rust:
`if res == 0 {
    None
} else {
  NaiveDatetime::...
}
`

dsavitskiy (approved):

tiagolobocastro:
Looks like we're "cramming" the timestamp into a 64bit number so we're losing some precision. Also we now have yet more atomics, can't we just add a parking lot mutex and store the appropriate type?

dsharma-dc:
> Looks like we're "cramming" the timestamp into a 64bit number so we're losing some precision. Also we now have yet more atomics, can't we just add a parking lot mutex and store the appropriate type?

I'm not sure if I understand the concern. Even with a signed-32 bit int we are good until we hit Y2038 problem, irrespective of platform  :) u64 is good enough till a reallly long time. We are just not using nanosecond level details from the epoch timestamp, which I don't think we care about for child fault timestamp.

io-engine/src/bdev/nexus/nexus_child.rs
dsharma-dc: Apparently, I had `else` earlier but IDE suggested me to unwrap the condition and do this way :)

tiagolobocastro:
> > Looks like we're "cramming" the timestamp into a 64bit number so we're losing some precision. Also we now have yet more atomics, can't we just add a parking lot mutex and store the appropriate type?
> 
> I'm not sure if I understand the concern. Even with a signed-32 bit int we are good until we hit Y2038 problem, irrespective of platform :) u64 is good enough till a reallly long time. We are just not using nanosecond level details from the epoch timestamp, which I don't think we care about for child fault timestamp.

Simply that we're making liberal use of atomics where it may not be necessary. If it is then it's fine.
Not everyone remembers all the atomic ordering flags and if a mutex would suffice whilst being simpler/clearer and allowing us to keep the "correct" type, then seems reasonable to me.

dsharma-dc:
> > > Looks like we're "cramming" the timestamp into a 64bit number so we're losing some precision. Also we now have yet more atomics, can't we just add a parking lot mutex and store the appropriate type?
> > 
> > 
> > I'm not sure if I understand the concern. Even with a signed-32 bit int we are good until we hit Y2038 problem, irrespective of platform :) u64 is good enough till a reallly long time. We are just not using nanosecond level details from the epoch timestamp, which I don't think we care about for child fault timestamp.
> 
> Simply that we're making liberal use of atomics where it may not be necessary. If it is then it's fine. Not everyone remembers all the atomic ordering flags and if a mutex would suffice whilst being simpler/clearer and allowing us to keep the "correct" type, then seems reasonable to me.

In my experience, atomic operations on entities upto 64bits have proven to be always better performant compared to spinlock or reader-writer locks, as revealed by x86 hwpmc stats. Agree that ordering flags are a slight annoyance to remember though.

tiagolobocastro:
> In my experience, atomic operations on entities upto 64bits have proven to be always better performant compared to spinlock or reader-writer locks, as revealed by x86 hwpmc stats. Agree that ordering flags are a slight annoyance to remember though.

For reference, the mutex implementation we use only requires a single atomic operation on the uncontended case but why do we care about performance here? We only read this field on grpc list on when faulting a child or am I missing something?

dsharma-dc:
> > In my experience, atomic operations on entities upto 64bits have proven to be always better performant compared to spinlock or reader-writer locks, as revealed by x86 hwpmc stats. Agree that ordering flags are a slight annoyance to remember though.
> 
> For reference, the mutex implementation we use only requires a single atomic operation on the uncontended case but why do we care about performance here? We only read this field on grpc list on when faulting a child or am I missing something?



> For reference, the mutex implementation we use only requires a single atomic operation on the uncontended case but why do we care about performance here? We only read this field on grpc list on when faulting a child or am I missing something?

For the use-case today, yes the performance isn't really a concern. grpc is going to be the only consumer of this for now.

dsavitskiy:
Seems like I overlooked the important thing there: state is not only changed by set_state(), but also retire_child_device() via compare_exchange. set_state() is not really the code path for faulting.
We need to set "fault_at" in retire_child_device().

dsavitskiy:
I'd suggest the following changes:
1. make `set_state` private to `NexusChild` (remove pub(crate)), and don't touch `faulted_at` there.
2. add a method to force setting a faulted state, like `pub(crate) set_faulted_state()` and set `faulted_at` there.
3. use `set_faulted_state()` instead of `set_state()` outside `nexus_child.rs`
4. update `faulted_at` in `retire_child_device()`, inside `c.state.compare_exchange` branch.

dsharma-dc:
> I'd suggest the following changes:
> 
> 1. make `set_state` private to `NexusChild` (remove pub(crate)), and don't touch `faulted_at` there.
> 2. add a method to force setting a faulted state, like `pub(crate) set_faulted_state()` and set `faulted_at` there.
> 3. use `set_faulted_state()` instead of `set_state()` outside `nexus_child.rs`
> 4. update `faulted_at` in `retire_child_device()`, inside `c.state.compare_exchange` branch.

Refactored the set_state paths, and used parking_lot as discussed with @tiagolobocastro above.
Please take a look.

mtzaurus (approved):

tiagolobocastro (approved):

dsharma-dc:
bors try

dsharma-dc:
bors merge

PR #1345, opened 2023-03-30
feat(snapshot): grpc implementation for create snapshot of replica

hrudaya21:
gRPC Implementation for the CreateSnapshot of Replica Volume

### DIFF
--- a/io-engine/src/bdev/device.rs
+++ b/io-engine/src/bdev/device.rs
@@ -9,13 +9,16 @@ use std::{
 };
 
 use async_trait::async_trait;
+use futures::channel::oneshot;
 use nix::errno::Errno;
 use once_cell::sync::{Lazy, OnceCell};
 
 use spdk_rs::{
     libspdk::{
         iovec,
+        spdk_bdev_flush,
         spdk_bdev_free_io,
+        spdk_bdev_has_write_cache,
         spdk_bdev_io,
         spdk_bdev_readv_blocks,
         spdk_bdev_reset,
@@ -30,27 +33,30 @@ use spdk_rs::{
     IoType,
 };
 
-use crate::core::{
-    mempool::MemoryPool,
-    Bdev,
-    BdevHandle,
-    BlockDevice,
-    BlockDeviceDescriptor,
-    BlockDeviceHandle,
-    BlockDeviceIoStats,
-    CoreError,
-    DeviceEventDispatcher,
-    DeviceEventSink,
-    DeviceEventType,
-    DeviceIoController,
-    IoCompletionCallback,
-    IoCompletionCallbackArg,
-    IoCompletionStatus,
-    NvmeStatus,
-    ReadMode,
-    UntypedBdev,
-    UntypedBdevHandle,
-    UntypedDescriptorGuard,
+use crate::{
+    core::{
+        mempool::MemoryPool,
+        Bdev,
+        BdevHandle,
+        BlockDevice,
+        BlockDeviceDescriptor,
+        BlockDeviceHandle,
+        BlockDeviceIoStats,
+        CoreError,
+        DeviceEventDispatcher,
+        DeviceEventSink,
+        DeviceEventType,
+        DeviceIoController,
+        IoCompletionCallback,
+        IoCompletionCallbackArg,
+        IoCompletionStatus,
+        NvmeStatus,
+        ReadMode,
+        UntypedBdev,
+        UntypedBdevHandle,
+        UntypedDescriptorGuard,
+    },
+    ffihelper::cb_arg,
 };
 
 /// TODO
@@ -504,6 +510,60 @@ impl BlockDeviceHandle for SpdkBlockDeviceHandle {
             source: Errno::ENXIO,
         })
     }
+    /// Flush the io in buffer to disk, for the Local Block Device.
+    async fn flush_io(&self) -> Result<u64, CoreError> {
+        let is_write_cache =
+            unsafe { spdk_bdev_has_write_cache(self.device.0.as_inner_ref()) };
+        if !is_write_cache {
+            debug!(
+                "No Write Cache, No need of Flush for bdev name: {}, uuid: {}",
+                self.device.device_name(),
+                self.device.uuid(),
+            );
+            return Ok(0);
+        }
+        let (desc, chan) = self.handle.io_tuple();
+        let bdev_size = self.device.size_in_bytes();
+        let (s, r) = oneshot::channel::<bool>();
+        let result = unsafe {
+            spdk_bdev_flush(
+                desc,
+                chan,
+                0,
+                bdev_size,
+                Some(bdev_flush_complete_cb),
+                cb_arg(s),
+            )
+        };
+        if result < 0 {
+            return Err(CoreError::DeviceFlush {
+                source: Errno::from_i32(result),
+                name: format!(
+                    "bdev_name: {}, uuid: {}",
+                    self.device.device_name(),
+                    self.device.uuid()
+                ),
+            });
+        }
+        match r.await {
+            Ok(result) => info!(
+                "Bdev Flush Success: uuid: {}, result: {}",
+                self.device.uuid(),
+                result,
+            ),
+            Err(_) => {
+                return Err(CoreError::DeviceFlush {
+                    source: Errno::from_i32(result),
+                    name: format!(
+                        "bdev_name: {}, uuid: {}",
+                        self.device.device_name(),
+                        self.device.uuid()
+                    ),
+                });
+            }
+        };
+        Ok(0)
+    }
 }
 
 /// TODO
@@ -643,4 +703,18 @@ pub fn bdev_event_callback<T: BdevOps>(
 /// Dispatches a special event for loopback device removal.
 pub fn dispatch_loopback_removed(name: &str) {
     dispatch_bdev_event(DeviceEventType::LoopbackRemoved, name);
+}
+
+/// Call back function for the Block Device Flush
+extern "C" fn bdev_flush_complete_cb(
+    bio: *mut spdk_bdev_io,
+    result: bool,
+    arg: *mut c_void,
+) {
+    let s = unsafe { Box::from_raw(arg as *mut oneshot::Sender<bool>) };
+    s.send(result).ok();
+    // Free replica's bio.
+    unsafe {
+        spdk_bdev_free_io(bio);
+    }
 }
--- a/io-engine/src/grpc/v1/replica.rs
+++ b/io-engine/src/grpc/v1/replica.rs
@@ -3,10 +3,12 @@ use crate::{
     bdev_api::BdevError,
     core::{
         logical_volume::LogicalVolume,
+        snapshot::SnapshotDescriptor,
         Bdev,
         Protocol,
         Share,
         ShareProps,
+        SnapshotOps,
         UntypedBdev,
         UpdateProps,
     },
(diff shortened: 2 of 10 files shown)


io-engine/src/bdev/device.rs
```diff
@@ -504,6 +510,40 @@ impl BlockDeviceHandle for SpdkBlockDeviceHandle {
             source: Errno::ENXIO,
         })
     }
+    /// Flush the io in buffer to disk, for the Local Block Device.
+    async fn flush_io(&self) -> Result<u64, CoreError> {
+        let is_write_cache =
+            unsafe { spdk_bdev_has_write_cache(self.device.0.as_inner_ref()) };
+        if !is_write_cache {
+            info!("No Write Cache, No need of Flush");
```
tiagolobocastro: Maybe this should be a debug?
Can we also add more information about who self?

io-engine/src/bdev/device.rs
```diff
@@ -504,6 +510,40 @@ impl BlockDeviceHandle for SpdkBlockDeviceHandle {
...
+                desc,
+                chan,
+                0,
+                bdev_size,
+                Some(bdev_flush_complete_cb),
+                cb_arg(s),
+            )
+        };
+        if result < 0 {
+            error!("flush failed for the spdk device, error: {}", result);
```
tiagolobocastro: information about self here as well

io-engine/src/bdev/device.rs
```diff
@@ -504,6 +510,40 @@ impl BlockDeviceHandle for SpdkBlockDeviceHandle {
...
+        }
+        match r.await {
```
tiagolobocastro, line 548: if result was <0 do we still need to await ?

io-engine/src/bdev/device.rs
```diff
@@ -644,3 +684,22 @@ pub fn bdev_event_callback<T: BdevOps>(
...
+
+/// Call back function for the Block Device Flush
+extern "C" fn bdev_flush_complete_cb(
+    bio: *mut spdk_bdev_io,
+    result: bool,
+    arg: *mut c_void,
+) {
+    let s = unsafe { Box::from_raw(arg as *mut oneshot::Sender<bool>) };
+    if !result {
+        error!("Bdev Flush failed");
```
tiagolobocastro: do we need both logs on callback and waiter?

io-engine/src/bdev/device.rs
```diff
@@ -644,3 +684,22 @@ pub fn bdev_event_callback<T: BdevOps>(
...
+        s.send(result).err();
```
tiagolobocastro: what's the point of using err and ok here?

io-engine/src/grpc/v1/replica.rs
```diff
@@ -427,4 +429,72 @@ impl ReplicaRpc for ReplicaService {
...
+                let args = request.into_inner();
+                info!("{:?}", args);
+                let rx = rpc_submit(async move {
+                    let lvol = match UntypedBdev::lookup_by_uuid_str(
+                        &args.replica_uuid,
+                    ) {
+                        Some(bdev) => Lvol::try_from(bdev)?,
+                        None => {
+                            return Err(LvsError::Invalid {
+                                source: Errno::ENOSYS,
```
```suggestion
                                source: Errno::ENOENT,
```

io-engine/src/grpc/v1/replica.rs
```diff
@@ -427,4 +429,72 @@ impl ReplicaRpc for ReplicaService {
...
+                    };
+                    // prepare snap config and flush IO before taking snapshot.
+                    let mut snap_config = match lvol
+                        .prepare_snap_config(&args.snapshot_name)
+                        .await
+                    {
+                        Some(snap_config) => snap_config,
+                        None => {
+                            return Err(LvsError::Invalid {
+                                source: Errno::ENOSYS,
```
tiagolobocastro: `ENOSYS` means not implemented, is this the right error code here? 
Please take care considering which error we return and how they map to grpc status so that the control plane may be use them.

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -875,4 +875,33 @@ impl SnapshotOps for Lvol {
     async fn snapshot_iter(self) -> LvolSnapshotIter {
         LvolSnapshotIter::new(self)
     }
+    /// Prepare Snapshot Config for Block/Nvmf Device, before snapshot create.
+    async fn prepare_snap_config(
+        &self,
+        snap_name: &str,
```
tiagolobocastro, line 892: IIRC should the txn_id be passed as argument as the control plane would want to set it?

io-engine/src/subsys/nvmf/admin_cmd.rs
```diff
@@ -158,15 +158,15 @@ pub fn create_snapshot(
...
+    let mut snap_param = SnapshotParams::new(
         Some(lvol.name()),
         Some(lvol.name()),
         Some(lvol.name()),
         Some(snapshot_name),
     );
     // Blobfs operations must be on md_thread
     Reactors::master().send_future(async move {
-        match lvol.create_snapshot(snap_param).await {
+        match lvol.create_snapshot(&mut snap_param).await {
```
tiagolobocastro: I'm curious, why does it need to be a reference rather than moved?

io-engine/src/bdev/device.rs
hrudaya21: > what's the point of using err and ok here?

if the callback function result is set to false, then need to consider as Flush failure else consider flush success. For that reason, both err and ok are handled.

io-engine/src/subsys/nvmf/admin_cmd.rs
hrudaya21: > I'm curious, why does it need to be a reference rather than moved?

Currently the idea is as part of create_snapshot, some of the `snapshotparams` will be set as part of SPDK call back function. To get the changed parameter information in gRPC code (replica.rs), I made it mutable reference.

io-engine/src/bdev/device.rs
hrudaya21: handled this scenario. If result < 0, no need to wait, the error should be return at that point.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: updated code. txn_id will be part of gRPC request.

io-engine/src/bdev/device.rs
hrudaya21: > Maybe this should be a debug? Can we also add more information about who self?

Accepted. Added extra logging information in all error scenario for better tracing.

io-engine/src/bdev/device.rs
hrudaya21: > information about self here as well

updated

io-engine/src/bdev/device.rs
hrudaya21: removed the duplicate logs.

io-engine/src/bdev/device.rs
mtzaurus: Looks like nobody is using Option which 'ok()' / 'err()' produces, so do we need it ?

io-engine/src/grpc/v1/replica.rs
hrudaya21: > `ENOSYS` means not implemented, is this the right error code here? Please take care considering which error we return and how they map to grpc status so that the control plane may be use them.

I have update the error code. Please review once.

io-engine/src/lvs/lvs_lvol.rs
mtzaurus: Yes, tx_id should be a mandatory gRPC argument.

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -875,4 +875,36 @@ impl SnapshotOps for Lvol {
...
+        // Entity Id will be same as lvol uuid for the replica snapshot.
+        let entity_id = Some(self.uuid());
+        // Current Lvol uuid is the parent for the snapshot.
+        let parent_id = Some(self.uuid());
+
+        let bdev_handle = device_open(self.as_bdev().name(), false)
+            .unwrap()
+            .into_handle()
+            .unwrap();
+        match bdev_handle.flush_io().await {
```
mtzaurus: Could we put I/O flush inside snapshot operation to make sure this is always done as part of the snapshot creation operation itself ? In such a case this also saves us one extra device open operation.

io-engine/src/subsys/nvmf/admin_cmd.rs
```diff
@@ -158,15 +158,15 @@ pub fn create_snapshot(
             | (nvme_cmd_cdw11_get_val(cmd) as u64) << 32
     };
     let snapshot_name = Lvol::format_snapshot_name(&lvol.name(), snapshot_time);
-    let snap_param = SnapshotParams::new(
+    let mut snap_param = SnapshotParams::new(
         Some(lvol.name()),
         Some(lvol.name()),
         Some(lvol.name()),
```
mtzaurus, line 164: Since we now support 'tx_id', can't we generate a GUID here to make transactions unique ?

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: > Could we put I/O flush inside snapshot operation to make sure this is always done as part of the snapshot creation operation itself ? In such a case this also saves us one extra device open operation.

Accepted and Done the modification.

io-engine/src/subsys/nvmf/admin_cmd.rs
hrudaya21: > Since we now support 'tx_id', can't we generate a GUID here to make transactions unique ?

Accepted, modified. Missed this part. Thanks for finding.

io-engine/src/bdev/device.rs
hrudaya21: > Looks like nobody is using Option which 'ok()' / 'err()' produces, so do we need it ?

Modification is done.

mtzaurus:
Could we also have a gRPC test for this call ? To fully test the 'create' functionality fully, we need a method to list snapshots, so kinda 'chicken-egg' problem. :-)
But i'd implement a simple test to make sure that gRPC method accepts the requested parameters and doesn't fail at least.

io-engine/src/subsys/nvmf/admin_cmd.rs
dsavitskiy: @hrudaya21 Sometimes it makes to pass by value (move) and return it. Like this:
```
fn foo(a: Data) -> Data { a.x = 123; a }

let a = Data::new();
let a = foo(a);
```

instead of 

```
fn foo(a: &mut Data) { a.x = 123; }

let a = Data::new();
foo(&mut a);
```

Mut references are very restrictive in Rust, for very good reasons. By moving data, you avoid having mut ref.

hrudaya21: understood. currently there is no requirement for changing the param though. But understood the rust best practice and modified code.

hrudaya21:
> Could we also have a gRPC test for this call ? To fully test the 'create' functionality fully, we need a method to list snapshots, so kinda 'chicken-egg' problem. :-) But i'd implement a simple test to make sure that gRPC method accepts the requested parameters and doesn't fail at least.

Added a basic gRPC create_snapshot case. No result is checked now. It will extend as part of list_snapshot requirement.

mtzaurus (approved):

io-engine/src/grpc/v1/replica.rs
```diff
@@ -427,4 +429,71 @@ impl ReplicaRpc for ReplicaService {
...
+                        }
+                    };
+                    // prepare snap config and flush IO before taking snapshot.
+                    let snap_config = match lvol
+                        .prepare_snap_config(&args.snapshot_name, &args.txn_id)
+                    {
+                        Some(snap_config) => snap_config,
+                        None => {
+                            return Err(LvsError::Invalid {
+                                source: Errno::EBUSY,
```
tiagolobocastro: is busy the correct error here? how can prepare fail?

io-engine/src/grpc/v1/replica.rs
```diff
@@ -427,4 +429,71 @@ impl ReplicaRpc for ReplicaService {
...
+                        }
+                    };
+                    // create snapshot
+                    match lvol.create_snapshot(snap_config.clone()).await {
+                        Ok(()) => {
+                            info!("Create Snapshot Success for lvol: {:?}", lvol);
+                            Ok(CreateReplicaSnapshotResponse {
+                                snapshot_uuid: snap_config
+                                    .txn_id()
+                                    .unwrap_or_default(),
```
tiagolobocastro, line 469: so txn_id is an option? If it's none we'll return an invalid id here with unwrap_or_default ?

io-engine/src/grpc/v1/replica.rs
```diff
@@ -427,4 +429,71 @@ impl ReplicaRpc for ReplicaService {
...
+                            })
+                        }
+                        Err(e) => {
+                            error!(
+                                "Create Snapshot Failed for lvol: {:?} with Error: {:?}",
+                                lvol, e
```
```suggestion
                                "Create Snapshot Failed for lvol: {lvol:?} with Error: {e:?}",
```

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -853,7 +853,19 @@ impl SnapshotOps for Lvol {
             }
             s.send(errno).ok();
         }
-
+        let bdev_handle = device_open(self.as_bdev().name(), false)
+            .unwrap()
+            .into_handle()
+            .unwrap();
```
tiagolobocastro, line 859: should we handle these unwrap and return error rather than panic?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -853,7 +853,19 @@ impl SnapshotOps for Lvol {
...
+        match bdev_handle.flush_io().await {
+            Ok(_) => info!("Flush is Success for lvol: {:?}", self),
```
tiagolobocastro, line 861:
```suggestion
            Ok(_) => info!("Flush is Successful for lvol: {self:?}"),
```

io-engine/src/subsys/nvmf/admin_cmd.rs
```diff
@@ -161,7 +162,7 @@ pub fn create_snapshot(
     let snap_param = SnapshotParams::new(
         Some(lvol.name()),
         Some(lvol.name()),
-        Some(lvol.name()),
+        Some(Uuid::generate().to_string()),
```
tiagolobocastro, line 165: I probably misunderstand but shouldn't we use the `txn_id` specified by grpc?

test/python/tests/replica/test_bdd_replica.py
```diff
@@ -353,3 +379,14 @@ def read_operation_should_succeed():
 @then("the write operation should succeed")
 def write_operation_should_succeed():
     raise NotImplementedError
+
+
+# @then('new snapshot with requested name shall be successfully created for replica')
```
tiagolobocastro: can we remove this code if commented out?

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: Currently it is a general practice in Nexus and other place, so kept the same behaviour.

io-engine/src/grpc/v1/replica.rs
hrudaya21: Done changes

io-engine/src/grpc/v1/replica.rs
hrudaya21: Accepted and changed. Prepare will not fail, as some logic moved out of this function.

PR #1347, opened 2023-03-31
improve tests

dsavitskiy:


dsavitskiy:
bors try

ci.nix
```diff
@@ -64,6 +64,7 @@ mkShell {
   PROTOC_INCLUDE = io-engine.PROTOC_INCLUDE;
   SPDK_PATH = if nospdk then null else "${libspdk-dev}";
   ETCD_BIN = "${etcd}/bin/etcd";
+  FIO_SPDK = "${libspdk-dev}/fio/spdk_nvme";
```
tiagolobocastro: I wonder if we need to check `if no spdk then null` like SPDK_PATH, in case we're using custom spdk?

tiagolobocastro (approved):

ci.nix
dsavitskiy: ah, yes. Will fix.

io-engine-tests/src/file_io.rs
```diff
@@ -93,3 +126,65 @@ pub async fn compute_file_checksum(
...
+            break;
+        }
+
+        for i in 0 .. na {
+            if buf_a[i] != buf_b[i] {
+                return Err(Error::new(
+                    ErrorKind::Other,
+                    format!(
+                        "Miscompare at {} ({} blk): {:?} {:#02x} != {:?} {:#02x}",
+                        pos, pos / 512, name_a, buf_a[i], name_b, buf_b[i],
```
dsharma-dc: This is assuming 512-bytes block size always? fio has configurable block size though.

io-engine-tests/src/fio.rs
```diff
@@ -0,0 +1,199 @@
...
+
+        let counter = JOB_COUNTER.fetch_add(1, Ordering::SeqCst);
+
+        Self {
+            counter,
+            name: format!("fio-{counter}"),
+            ioengine: "spdk".to_string(),
+            filename: String::new(),
+            direct: true,
+            blocksize: 4096,
```
dsharma-dc: Here it's hardcoded 4K

dsharma-dc (approved):

io-engine-tests/src/file_io.rs
dsavitskiy: I'll remove it, it's confusing indeed.

io-engine-tests/src/fio.rs
dsavitskiy: It's the same default as fio's. I'll make it optional to avoid confusion.

dsavitskiy:
bors merge

PR #1353, opened 2023-04-11
feat(snapshot): list, delete snapshot grpc implementation

hrudaya21:


io-engine/src/core/snapshot.rs
```diff
@@ -25,6 +26,35 @@ impl SnapshotParams {
     }
 }
 
+#[derive(Clone, Debug)]
+pub struct ListSnapshotParams {
```
mtzaurus: Could we rename 'ListSnapshotParams' to something like 'VolumeSnapshotDescriptor' or so ?
'Params' usually denote input parameters. :-)

io-engine/src/core/snapshot.rs
```diff
@@ -62,6 +95,27 @@ pub trait SnapshotDescriptor {
     fn name(&self) -> Option<String>;
 }
 
+/// Traits gives ListSnapshot Parameters.
+pub trait ListSnapshotParameters {
+    /// Get SnapshotParameters.
+    fn snapshot_params(&self) -> SnapshotParams;
+
+    /// Get snapshot_uuid.
+    fn snapshot_uuid(&self) -> String;
```
mtzaurus: You can avoid creating a copy of String by using '&str'.

io-engine/src/core/snapshot.rs
```diff
@@ -84,3 +138,34 @@ impl SnapshotDescriptor for SnapshotParams {
         self.snap_name.clone()
     }
 }
+
+impl ListSnapshotParameters for ListSnapshotParams {
+    /// Get SnapshotParameters.
+    fn snapshot_params(&self) -> SnapshotParams {
+        self.snap_param.clone()
```
mtzaurus: We could use &str here, IMHO, for String return values.

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -831,6 +837,27 @@ impl LvsLvol for Lvol {
...
         )
     }
+    /// Build SnapshotList Detail Parameters from Blob.
+    fn build_snap_list_param(
+        &self,
+        _blob: *mut spdk_blob,
+    ) -> ListSnapshotParams {
+        // TODO: need to Integrate with Snapshot Property Enumeration
+        // Currently it is stub.
+        ListSnapshotParams::new(
```
mtzaurus: Can't we setup here what's already known, i.e: snapshot UUID ?

io-engine/src/core/snapshot.rs
hrudaya21: > Could we rename 'ListSnapshotParams' to something like 'VolumeSnapshotDescriptor' or so ? 'Params' usually denote input parameters. :-)

Accepted. Code changed.

io-engine/src/core/snapshot.rs
hrudaya21: > You can avoid creating a copy of String by using '&str'.

Accepted, Code changed.

io-engine/src/core/snapshot.rs
hrudaya21: > We could use &str here, IMHO, for String return values.

Accepted, Code changed.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: > Can't we setup here what's already known, i.e: snapshot UUID ?

Accepted, Code changed.

io-engine/src/grpc/v1/replica.rs
```diff
@@ -106,6 +110,21 @@ impl Default for ReplicaService {
...
 
+impl From<VolumeSnapshotDescriptor> for ReplicaSnapshot {
+    fn from(s: VolumeSnapshotDescriptor) -> Self {
+        Self {
+            snapshot_uuid: s.snapshot_uuid().to_string(),
+            snapshot_name: s.snapshot_params().name().unwrap_or_default(),
+            replica_uuid: s.replica_uuid().to_string(),
+            bytes_referenced: s.bytes_referenced(),
+            num_clones: s.num_clones(),
+            timestamp: Some(s.timestamp().into()),
```
mtzaurus: Can timestamp here be None (it's an Option) ? If no, maybe we can make it a regular non-Option field ?

io-engine/src/grpc/v1/replica.rs
```diff
@@ -492,4 +511,99 @@ impl ReplicaRpc for ReplicaService {
...
+            async move {
+                let args = request.into_inner();
+                info!("{:?}", args);
+                let rx = rpc_submit(async move {
+                    let bdev = UntypedBdev::bdev_first().expect("Failed to enumerate devices");
+
+                    let mut devices = bdev
+                        .into_iter()
+                        .filter(|b| b.driver() == "lvol" && b.uuid_as_string() == args.snapshot_uuid)
+                        .map(|b| Lvol::try_from(b).expect("Can't create Lvol from device"))
```
mtzaurus: Could we relax check a bit and replace 'expect()' with error propagation to the caller ?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -464,6 +464,89 @@ impl Lvol {
...
+                unsafe {
+                    let r = spdk_blob_get_xattr_value(
+                        blob,
+                        attr_id.as_ptr(),
+                        &mut val as *mut *const c_void,
+                        &mut size as *mut u64,
+                    );
+                    // skip current lvol if any of the snapshot attribute not
+                    // found.
+                    if r != 0 {
```
mtzaurus, line 504: Could we add a warning message here to capture the fact that some attributes are missing ?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -464,6 +464,89 @@ impl Lvol {
...
+                    }
+                    curr_attr_val = String::from_raw_parts(
+                        val as *mut u8,
+                        size as usize,
+                        size as usize,
+                    );
+                }
+                match attr {
+                    SnapshotXattrs::ParentId => {
+                        // if lvol has a valid parent id then it is consider as
```
mtzaurus: You can use a dedicated method in Lvol to check if it's a snapshot:

```
    /// Returns a boolean indicating if the lvol is a snapshot
    fn is_snapshot(&self) -> bool;
```

mtzaurus:
Just a generic question: are we shipping gRPC BDD tests for newly introduced API along with this PR ?

hrudaya21:
> 

Currently I am adding rust unit test and gRPC case as well. Will add it part of this PR.

io-engine/src/grpc/v1/replica.rs
hrudaya21: > Can timestamp here be None (it's an Option) ? If no, maybe we can make it a regular non-Option field ?

We can keep it Option field to sync with other gRPC implementation. Now Will set to `None`.  Based on this field update, We will change the value in next PR.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: > You can use a dedicated method in Lvol to check if it's a snapshot:
> 
> ```
>     /// Returns a boolean indicating if the lvol is a snapshot
>     fn is_snapshot(&self) -> bool;
> ```

Accepted. Modified Code.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: > Could we add a warning message here to capture the fact that some attributes are missing ?

Accepted. Modified Code.

io-engine/src/grpc/v1/replica.rs
hrudaya21: > Could we relax check a bit and replace 'expect()' with error propagation to the caller ?

Modified the code.

PR #1358, opened 2023-04-14
feat(snapshot): lvol snapshot creation via generic block device handle

mtzaurus:
SPDK-backed block devices (SpdkBlockDevice) now support snapshot creation via block device I/O handle.


io-engine/src/bdev/device.rs
```diff
@@ -505,10 +508,31 @@ impl BlockDeviceHandle for SpdkBlockDeviceHandle {
...
-            source: Errno::ENXIO,
-        })
+    async fn create_snapshot(
+        &self,
+        snapshot: SnapshotParams,
+    ) -> Result<u64, CoreError> {
+        let bdev = self.handle.get_bdev();
+
+        // Snapshots are supported only for LVOLs.
+        if bdev.driver() != "lvol" {
```
dsavitskiy, line 518: Should we defined a global const for this to avoid magic constants?

io-engine/src/lvs/lvs_error.rs
```diff
@@ -119,8 +119,13 @@ pub enum Error {
     FlushFailed {
         name: String,
     },
-    #[snafu(display("Input Parameters for replica: {} is not correct", name))]
-    SnapConfigFailed {
+    #[snafu(display(
+        "Snapshot parameters for replica {} is not correct: {}",
```
dsavitskiy, line 123: "is" -> "are"

dsavitskiy (approved):

io-engine/src/bdev/device.rs
mtzaurus: Good point, will fix.

mtzaurus:
bors try

mtzaurus:
bors cancel

mtzaurus:
bors rty

mtzaurus:
bors try

mtzaurus:
bors cancel

blaisedias:
bors try-

blaisedias:
bors try

mtzaurus:
bors try

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -865,12 +924,29 @@ impl SnapshotOps for Lvol {
...
             }
         }
+
+        let mut attr_descrs: [spdk_xattr_descriptor; SnapshotXattrs::COUNT] =
+            [spdk_xattr_descriptor::default(); SnapshotXattrs::COUNT];
+
+        // Vector to keep allocated CStrings before snapshot  creation
+        // is complete to guarantee validity of attribute buffers
+        // stored inside CStrings.
+        let mut cstrs: Vec<CString> = Vec::new();
```
hrudaya21, line 934: `cstrs` is only for the buffer validation after createsnapshot or can be used as cache in future ? Can it be a hashmap in place of Vec<String> to fetch data in O(1) ?

mtzaurus:
bors try

io-engine/src/lvs/lvs_lvol.rs
mtzaurus: We don't need to access the content of 'cstrs': we just need to make sure the CStrings in 'cstrs' don't get freed till snapshot is created, because the string buffers inside CStrings are used to store attributes.
So it doesn't matter whether it's a vector or hash - we just need a container. Vector is simpler, IMHO.

hrudaya21 (approved):

mtzaurus:
bors merge

PR #1361, opened 2023-04-17
feat(nexus): refactoring nexus-related code

dsavitskiy:
* Removed unnecessary `&mut self` in nexus and nexus child code
* Cleaned up and improved log messages for rebuild and nexus code

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
@@ -1,6 +1,6 @@
 use futures::channel::oneshot::Receiver;
 use snafu::ResultExt;
-use std::{marker::PhantomData, pin::Pin, sync::Arc};
+use std::{marker::PhantomData, sync::Arc};
 
 use super::{
     nexus_err,
@@ -76,11 +76,11 @@ impl<'n> Nexus<'n> {
     /// Starts a rebuild job and returns a receiver channel
     /// which can be used to await the rebuild completion
     pub async fn start_rebuild(
-        mut self: Pin<&mut Self>,
+        &self,
         child_uri: &str,
     ) -> Result<Receiver<RebuildState>, Error> {
         let name = self.name.clone();
-        info!("{name}: start rebuild request for {child_uri}");
+        info!("{self:?}: start rebuild request for {child_uri}");
 
         // Find a healthy child to rebuild from.
         let src_child_uri = match self
@@ -133,7 +133,7 @@ impl<'n> Nexus<'n> {
         self.reconfigure(DrEvent::ChildRebuild).await;
 
         // Drop rebuild log from the child.
-        if let Ok(c) = self.as_mut().child_mut(&dst_child_uri) {
+        if let Ok(c) = self.child(&dst_child_uri) {
             c.drop_rebuild_log()
         };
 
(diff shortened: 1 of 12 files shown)


dsavitskiy:
bors try

mtzaurus (approved):

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -177,29 +177,27 @@ impl<'n> Nexus<'n> {
...
-            self.as_mut().get_unchecked_mut().rebuild_history.push(rec);
-        }
+        self.rebuild_history.lock().push(rec);
 
         debug!(
             "{self:?}: new rebuild history record for '{dst}'; \
             total {num} records",
             dst = job.dst_uri,
-            num = self.rebuild_history.len()
+            num = self.rebuild_history.lock().len()
```
tiagolobocastro, line 192: nit: should we get the len before the debug! to avoid taking the lock while writing to stdout? probably doesn't matter too much though

io-engine/src/rebuild/rebuild_job_backend.rs
```diff
@@ -389,8 +431,9 @@ impl RebuildJobBackend {
             });
 
         let progress = (blocks_recovered * 100) / blocks_total;
+        assert!(progress < 100 || blocks_remaining == 0);
```
tiagolobocastro, line 434:
```suggestion
        assert!(progress <= 100 || blocks_remaining == 0);
```
It may be called when it's complete.

io-engine/src/rebuild/rebuild_job_backend.rs
```diff
@@ -402,17 +445,7 @@ impl RebuildJobBackend {
...
-        info!(
-            "State: {state}, Nexus: {nex}, Src: {src}, Dst: {dst}: {res:?}",
-            state = self.state(),
-            nex = self.nexus_name,
-            src = self.src_uri,
-            dst = self.dst_uri,
-        );
-
-        res
+        }
```
tiagolobocastro, line 448: I see you removed the log message below, any reason why? I find it is useful to see these when going through the logs.

tiagolobocastro (approved):

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
dsavitskiy: isn't the arg expr calculated before calling a func?

io-engine/src/rebuild/rebuild_job_backend.rs
dsavitskiy: The stats are still printed in logs, I just wanted to avoid duplicate messages in  certain situations. I moved logging of the stats in other places from where get_stats() is called.

io-engine/src/rebuild/rebuild_job_backend.rs
dsavitskiy: There was a bug in rebuild log management, making blocks_remaining > 0 even after all blocks were copied. So I added this assert to check that.

- progress < 100: blocks_remaining can be > 0, or =0 if  no "dirty" blocks remain.
- progress == 100: blocks_remaining must be 0, as all blocks were copied.

If it is like assert!(progress <= 100 || EXPR), EXPR is never checked as progress is always <= 100 due to the way it is calculated.

dsavitskiy:
bors merge

io-engine/src/rebuild/rebuild_job_backend.rs
tiagolobocastro: Ah right sorry I had misread the statement previously, got it.

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
tiagolobocastro: It is, but as I understand it, the mutex guard temporary is only dropped at the end of the function call.

PR #1362, opened 2023-04-18
feat(nexus): nexus snapshot

mtzaurus:
Nexus object now supports snapshot creation on its replicas (currently only nexuses with 1 replica support such operation).


mtzaurus:
bors try

PR #1364, opened 2023-04-23
feat(test): improving test helpers for v1 api

dsavitskiy:
* Shared RPC helper can now be shared across thread/tokio tasks
* io-engine-dependencies updated


dsavitskiy:
bors try

PR #1367, opened 2023-04-25
Feature Lockless log for rebuild

dsavitskiy:


dsavitskiy:
bors try

io-engine/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -910,17 +919,26 @@ impl<'n> Nexus<'n> {
...
+        debug!("{self:?}: retire: pausing...");
         self.as_mut().pause().await?;
-        debug!("{self:?}: pausing ok");
+        debug!("{self:?}: retire: pausing ok");
 
         if let Some(child) = self.lookup_child_by_device(&device_name) {
+            // Cancel rebuild job for this child, if any.
+            if let Some(job) = child.rebuild_job() {
+                debug!("{self:?}: retire: stopping rebuild job...");
+                job.terminate().await.ok();
```
tiagolobocastro, line 930: looks like we're waiting for the job to complete here whilst still pausing, is this necessary?

tiagolobocastro: also what happens if rebuild succeeds (eg: let's say the last IO was sent before the replica got into a bad state)?
Conversely, what if this child was actually a rebuild source for another child? Probably that rebuild will fail, but can we somehow restart the rebuild from another child without getting rid of the log?

io-engine/src/bdev/nexus/nexus_child.rs
```diff
@@ -1262,37 +1279,46 @@ impl<'c> NexusChild<'c> {
...
     }
 
-    /// Creates a new rebuild log, or returns an existing one.
-    pub(super) fn get_or_init_rebuild_log(&self) -> Option<RebuildLogHandle> {
-        let mut lg = self.rebuild_log.lock();
+    /// Creates a new I/O log, if none existed.
+    /// Returns true if a log has been created or already exists, false if I/O
+    /// log is disabled for this child for whatever reason.
+    ///
+    /// I/O log is vever created if the child is not fully synced.
```
tiagolobocastro:
```suggestion
    /// I/O log is never created if the child is not fully synced.
```

io-engine/src/rebuild/rebuild_job.rs
```diff
@@ -164,9 +165,19 @@ impl RebuildJob {
 
     /// Schedules the job to start in a future and returns a complete channel
     /// which can be waited on.
-    pub fn start(
+    pub async fn start(
         &self,
+        map: Option<RebuildMap>,
```
tiagolobocastro, line 170: Is this because the rebuild cannot be created with the map already?

io-engine/src/rebuild/rebuild_job.rs
```diff
@@ -164,9 +165,19 @@ impl RebuildJob {
...
     ) -> Result<oneshot::Receiver<RebuildState>, RebuildError> {
+        let (s, r) = oneshot::channel();
```
tiagolobocastro: if map is None do we need to send at all?

io-engine/src/rebuild/rebuild_job.rs
```diff
@@ -164,9 +165,19 @@ impl RebuildJob {
...
+        self.comms
+            .send(RebuildJobRequest::SetRebuildMap((map, s)))
+            .await
+            .ok();
+        if let Err(e) = r.await {
+            error!("{uri}: failed to set rebuild map: {e}", uri = self.dst_uri);
```
tiagolobocastro: could be worth making it more explicit saying a full rebuild will be done? though this should never happen I guess..

io-engine/src/rebuild/rebuild_log.rs
```diff
@@ -109,43 +84,58 @@ impl RebuildLog {
...
-    /// and cloning the name when accessing it.
+/// Per-core I/O log channel.
+/// I/O log channel is enables lockless logging of I/O operations.
+pub struct IOLogChannelInner {
+    /// Channel's core.
+    core: u32,
+    /// Name of the underlying block device.
     device_name: String,
+    /// Map of device segments.
+    segments: UnsafeCell<Option<SegmentMap>>,
```
tiagolobocastro: Don't we have mutable access to the channel? Then we can make log_io take &mut self, and not have to use `UnsafeCell` ?

io-engine/src/bdev/nexus/nexus_child.rs
dsavitskiy: fixed

io-engine/src/rebuild/rebuild_job.rs
dsavitskiy: Rebuild map is optional, e.g. if a child wasn't sync, or if I/O logging is disabled.

io-engine/src/rebuild/rebuild_job.rs
dsavitskiy: Fixed

io-engine/src/rebuild/rebuild_job.rs
dsavitskiy: If it cannot communicate to backend, it's broken already.

io-engine/src/rebuild/rebuild_log.rs
dsavitskiy: IoLogChannel is accessed via Rc<>, which would require an UnsafeCell here or as Rc<UnsafeCell<>>. I prefer it here, as device_name is logically const.

dsavitskiy:
bors try

io-engine/src/bdev/nexus/mod.rs
```diff
@@ -163,3 +164,6 @@ pub async fn shutdown_nexuses() {
 
     info!("All nexus have been shutdown.");
 }
+
+/// Enables/disables partial rebuild.
+pub static ENABLE_PARTIAL_REBULD: AtomicBool = AtomicBool::new(true);
```
dsharma-dc: typo: REBUILD

io-engine/src/bin/io-engine.rs
```diff
@@ -46,6 +46,15 @@ fn start_tokio_runtime(args: &MayastorCliArgs) {
     let reactor_freeze_detection = args.reactor_freeze_detection;
     let reactor_freeze_timeout = args.reactor_freeze_timeout;
 
+    // Enable partial rebuild.
+    if let Ok(v) = std::env::var("NEXUS_PARTIAL_REBUILD") {
```
dsharma-dc, line 50: Why do we need this to be atomic? This is being set during startup only and can be a simple boolean perhaps that can be checked during rest of the io-engine lifecycle?

io-engine/src/rebuild/rebuild_log.rs
```diff
@@ -109,43 +84,58 @@ impl RebuildLog {
         (lbn * self.block_len / SEGMENT_SIZE) as usize
     }
 
-    /// Counts the total number of modified blocks.
-    pub fn count_modified_blocks(&self) -> u64 {
-        let segs = self.segments.iter().filter(|i| *i).count() as u64;
-        segs * SEGMENT_SIZE / self.block_len
+    /// Counts the total number of blocks set to one.
```
dsharma-dc: `Counts the total number of segments set to one.` ?

io-engine/src/rebuild/rebuild_descriptor.rs
```diff
@@ -69,4 +73,21 @@ impl RebuildDescriptor {
                 bdev: descriptor.get_device().device_name(),
             })
     }
+
+    /// Checks if the block has to be transferred.
+    /// If no rebuild map is present, all blocks are considered unsynced.
```
dsharma-dc, line 78: ` all blocks are considered synced.` ?

io-engine/src/rebuild/rebuild_job.rs
tiagolobocastro: Ah right...

io-engine/src/rebuild/rebuild_log.rs
```diff
@@ -109,43 +84,58 @@ impl RebuildLog {
...
-#[derive(Clone)]
-pub struct RebuildLogHandle {
-    /// Share log reference.
-    log: Arc<Mutex<RebuildLog>>,
-    /// Name of the underlying device.
-    /// Keep it here to avoid locking just to access the device name,
-    /// and cloning the name when accessing it.
+/// Per-core I/O log channel.
+/// I/O log channel is enables lockless logging of I/O operations.
+pub struct IOLogChannelInner {
```
tiagolobocastro: Should this part of the rebuild be inside rebuild_log module? The rebuild itself only seems to care about the combined segment rebuildMap?

io-engine/src/rebuild/rebuild_log.rs
```diff
@@ -158,35 +148,196 @@ impl RebuildLogHandle {
...
+        let segments = self
+            .channels
+            .lock()
+            .values_mut()
+            .map(|x| x.take_segments())
+            .reduce(|mut acc, e| {
+                acc.segments.or(&e.segments);
+                acc
+            })
+            .unwrap();
```
tiagolobocastro:
```suggestion
            .expect("Should have at least 1 core");
```
?

io-engine/src/rebuild/rebuild_task.rs
```diff
@@ -65,11 +62,8 @@ impl RebuildTask {
         blk: u64,
         descriptor: &RebuildDescriptor,
     ) -> Result<bool, RebuildError> {
-        // Check if the block has to be copied.
-        if let Some(ref log) = self.rebuild_log {
-            if !log.is_modified(blk) {
-                return Ok(false);
-            }
+        if descriptor.is_blk_sync(blk) {
```
tiagolobocastro, line 65: did we ever test if there's any perf difference between walking the device range and skipping IO versus walking the log itself?

tiagolobocastro (approved):

mtzaurus (approved):

dsavitskiy:
bors merge

pchandra19:
bors merge

dsavitskiy:
bors merge

io-engine/src/bdev/nexus/mod.rs
dsavitskiy: Fixed

io-engine/src/rebuild/rebuild_log.rs
dsavitskiy: Actually, "bits" and it must not be "pub".

io-engine/src/rebuild/rebuild_descriptor.rs
dsavitskiy: In this context:
sync: "this block has data known to be the same as on the other replicas; no need to copy this block"
unsync: "this block has invalid data / unknown data; must be copied from a valid replica".

When no rebuild map present, all blocks are considered invalid, or "unsync".

io-engine/src/rebuild/rebuild_log.rs
dsavitskiy: Fixed

io-engine/src/rebuild/rebuild_log.rs
dsavitskiy: Yes, true. That was actually my intent: I renamed non-rebuild part to "I/O log". Non-rebuild part has to be moved into nexus part, indeed. Will do.

io-engine/src/rebuild/rebuild_task.rs
dsavitskiy: No we didn't, as we don't have code walking the log in the first place :-)
I'd leave this optimization for the future, as it would change the rebuild logic (tasks, jobs) considerably.

dsavitskiy:
bors try

dsavitskiy:
bors merge

tiagolobocastro:
My bad, was changing settings...
bors merge

PR #1371, opened 2023-05-04
build(images): add tag argument

tiagolobocastro:
Allows one to explicitly build the binaries with a given tag, as opposed to the alias-tag which simply adds a docker image tag alias.

### DIFF
--- a/.github/workflows/pr-submodule-branch.yml
+++ b/.github/workflows/pr-submodule-branch.yml
@@ -1,0 +1,24 @@
+name: Submodule Branch Check
+on:
+  pull_request:
+    types: ['opened', 'edited', 'reopened', 'synchronize']
+  push:
+    branches:
+      - develop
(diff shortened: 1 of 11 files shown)


tiagolobocastro:
@dsavitskiy the second commit is an alternative fix for the first, which is much simpler as we simply tag the tag in the build derivation. However, the first one has the advantage that doesn't rely on the .git as a dependency for the cargo-build, so avoids unnecessary rebuilds by figuring out the tag in a separate derivation - if the tag doesn't change, the cargo build is never invoked.
Which one would you prefer here?

dsavitskiy (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1373, opened 2023-05-08
chore(rpc): updated dependencies for spdk and cargo packages

mtzaurus:
Updated dependency for the latest SPDK snapshot API and for other Cargo packages.


mtzaurus:
bors try

PR #1377, opened 2023-05-09
feat(snapshot): Create, List, Delete ReplicaSnapshot gRPC implementation

hrudaya21:
- gRPC parameters are aligned for CreateReplicaSnapshot
- gRPC implementation is done for ListReplicaSnapshot and DeleteReplicaSnapshot

### DIFF
--- a/io-engine/src/core/snapshot.rs
+++ b/io-engine/src/core/snapshot.rs
@@ -1,9 +1,10 @@
+use crate::{core::logical_volume::LogicalVolume, lvs::Lvol};
 use async_trait::async_trait;
+use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 use strum_macros::{EnumCount as EnumCountMacro, EnumIter};
-
 /// Snapshot Captures all the Snapshot information for Lvol.
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
 pub struct SnapshotParams {
     entity_id: Option<String>,
     parent_id: Option<String>,
@@ -28,6 +29,42 @@ impl SnapshotParams {
     }
 }
 
+/// Snapshot Descriptor to respond back as part of listsnapshot
+#[derive(Clone, Debug)]
+pub struct VolumeSnapshotDescriptor {
+    snapshot_uuid: String,
+    snapshot_size: u64,
+    num_clones: u64,
+    timestamp: DateTime<Utc>,
+    replica_uuid: String,
+    replica_size: u64,
+    snap_param: SnapshotParams,
+    // set to false, if any of the snapshotdescriptor is not filled properly
+    valid_snapshot: bool,
+}
+
+impl VolumeSnapshotDescriptor {
+    pub fn new(
+        snapshot: &Lvol,
+        num_clones: u64,
+        timestamp: DateTime<Utc>,
+        replica_uuid: String,
+        replica_size: u64,
+        snap_param: SnapshotParams,
+        valid_snapshot: bool,
+    ) -> Self {
+        Self {
+            snapshot_uuid: snapshot.uuid(),
+            snapshot_size: snapshot.size(),
+            num_clones,
+            timestamp,
+            replica_uuid,
+            replica_size,
+            snap_param,
+            valid_snapshot,
+        }
+    }
+}
 /// Snapshot attributes used to store its properties.
 #[derive(Debug, EnumCountMacro, EnumIter)]
 pub enum SnapshotXattrs {
@@ -55,7 +92,7 @@ pub trait SnapshotOps {
     async fn create_snapshot(
         &self,
         snap_param: SnapshotParams,
-    ) -> Result<(), Self::Error>;
+    ) -> Result<Lvol, Self::Error>;
 
     // Get a Snapshot Iterator.
     async fn snapshot_iter(self) -> Self::SnapshotIter;
@@ -64,8 +101,12 @@ pub trait SnapshotOps {
     fn prepare_snap_config(
         &self,
         snap_name: &str,
+        entity_id: &str,
         txn_id: &str,
     ) -> Option<SnapshotParams>;
+
+    /// List Snapshots.
+    fn list_snapshot(&self) -> Vec<VolumeSnapshotDescriptor>;
 }
 
 /// Traits gives the Snapshots Related Parameters.
@@ -73,14 +114,53 @@ pub trait SnapshotDescriptor {
     /// Get Transaction Id of the Snapshot Create.
     fn txn_id(&self) -> Option<String>;
 
+    /// Set Transaction Id.
+    fn set_txn_id(&mut self, txn_id: String);
+
     /// Get Entity Id of the Snapshot.
     fn entity_id(&self) -> Option<String>;
+
+    /// Set Entity Id.
+    fn set_entity_id(&mut self, entity_id: String);
 
     /// Get Parent Id of the Snapshot.
     fn parent_id(&self) -> Option<String>;
 
+    /// Set Parent id of the Snapshot.
+    fn set_parent_id(&mut self, parent_id: String);
+
     /// Get Snapshot Name.
     fn name(&self) -> Option<String>;
+
+    /// Set Snapshot Name.
+    fn set_name(&mut self, name: String);
+}
+
+/// Traits gives VolumeSnapshot Descriptor.
+pub trait VolumeSnapshotDescriptors {
+    /// Get snapshot_uuid.
+    fn snapshot_uuid(&self) -> &str;
+
+    /// Get snapshot size.
+    fn snapshot_size(&self) -> u64;
+
+    /// Give number of clones.
+    fn num_clones(&self) -> u64;
+
+    /// Give timestamp of Snapshot creation.
+    fn timestamp(&self) -> DateTime<Utc>;
+
+    /// Give replica uuid.
+    fn replica_uuid(&self) -> &str;
+
+    /// Give replica size.
+    fn replica_size(&self) -> u64;
+
+    /// Get SnapshotParameters.
+    fn snapshot_params(&self) -> &SnapshotParams;
+
+    /// Get ValidSnapshot value.
+    fn valid_snapshot(&self) -> bool;
 }
 
 /// Trait to give interface for all Snapshot Parameters.
@@ -90,9 +170,18 @@ impl SnapshotDescriptor for SnapshotParams {
         self.txn_id.clone()
     }
 
+    /// Set Transaction Id.
+    fn set_txn_id(&mut self, txn_id: String) {
+        self.txn_id = Some(txn_id);
+    }
     /// Get Entity Id of the Snapshot.
     fn entity_id(&self) -> Option<String> {
         self.entity_id.clone()
+    }
+
+    /// Set Entity Id.
+    fn set_entity_id(&mut self, entity_id: String) {
+        self.entity_id = Some(entity_id);
     }
 
     /// Get Parent Id of the Snapshot.
@@ -100,8 +189,58 @@ impl SnapshotDescriptor for SnapshotParams {
         self.parent_id.clone()
     }
 
+    /// Set Parent id of the Snapshot.
+    fn set_parent_id(&mut self, parent_id: String) {
+        self.parent_id = Some(parent_id)
+    }
     /// Get Snapname of the Snapshot.
     fn name(&self) -> Option<String> {
         self.snap_name.clone()
+    }
+    /// Set Snapshot Name.
+    fn set_name(&mut self, name: String) {
+        self.snap_name = Some(name);
+    }
+}
+
+impl VolumeSnapshotDescriptors for VolumeSnapshotDescriptor {
+    /// Get snapshot_uuid.
+    fn snapshot_uuid(&self) -> &str {
+        &self.snapshot_uuid
+    }
+
+    /// Give snapshot size.
+    fn snapshot_size(&self) -> u64 {
+        self.snapshot_size
+    }
+
+    /// Give number of clones.
+    fn num_clones(&self) -> u64 {
+        self.num_clones
+    }
+
+    /// Give timestamp of Snapshot creation.
+    fn timestamp(&self) -> DateTime<Utc> {
+        self.timestamp
+    }
+
+    /// Give replica uuid.
+    fn replica_uuid(&self) -> &str {
+        &self.replica_uuid
+    }
+
+    /// Give replica size.
+    fn replica_size(&self) -> u64 {
+        self.replica_size
+    }
+
+    /// Get SnapshotParameters.
+    fn snapshot_params(&self) -> &SnapshotParams {
+        &self.snap_param
+    }
+
+    /// Get ValidSnapshot value.
+    fn valid_snapshot(&self) -> bool {
+        self.valid_snapshot
     }
 }
--- a/io-engine/src/grpc/v1/replica.rs
+++ b/io-engine/src/grpc/v1/replica.rs
@@ -3,23 +3,33 @@ use crate::{
     bdev_api::BdevError,
     core::{
         logical_volume::LogicalVolume,
-        snapshot::SnapshotDescriptor,
+        snapshot::{
+            SnapshotDescriptor,
+            VolumeSnapshotDescriptor,
+            VolumeSnapshotDescriptors,
+        },
         Bdev,
         Protocol,
         Share,
         ShareProps,
         SnapshotOps,
+        SnapshotParams,
+        SnapshotXattrs,
         UntypedBdev,
         UpdateProps,
     },
     grpc::{rpc_submit, GrpcClientContext, GrpcResult, Serializer},
     lvs::{Error as LvsError, Lvol, LvolSpaceUsage, Lvs, LvsLvol},
+    spdk_rs::ffihelper::IntoCString,
 };
 use ::function_name::named;
+use core::ffi::c_void;
 use futures::FutureExt;
 use mayastor_api::v1::replica::*;
 use nix::errno::Errno;
+use spdk_rs::libspdk::spdk_blob_get_xattr_value;
 use std::{convert::TryFrom, panic::AssertUnwindSafe, pin::Pin};
+use strum::IntoEnumIterator;
 use tonic::{Request, Response, Status};
 
 #[derive(Debug)]
@@ -28,7 +38,25 @@ pub struct ReplicaService {
     name: String,
     client_context: tokio::sync::Mutex<Option<GrpcClientContext>>,
 }
-
+#[derive(Debug)]
+pub struct ReplicaSnapshotDescriptor {
+    pub snapshot_lvol: Lvol,
+    pub replica_uuid: String,
+    pub replica_size: u64,
+}
+impl ReplicaSnapshotDescriptor {
+    fn new(
+        snapshot_lvol: Lvol,
+        replica_uuid: String,
+        replica_size: u64,
+    ) -> Self {
+        Self {
+            snapshot_lvol,
+            replica_uuid,
+            replica_size,
+        }
+    }
+}
 #[async_trait::async_trait]
 impl<F, T> Serializer<F, T> for ReplicaService
 where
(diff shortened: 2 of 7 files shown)


io-engine/src/grpc/v1/replica.rs
```diff
@@ -106,6 +116,74 @@ impl Default for ReplicaService {
...
+                SnapshotXattrs::EntityId => {
+                    snapshot_param.set_entity_id(curr_attr_val);
+                }
+                SnapshotXattrs::TxId => {
+                    snapshot_param.set_txn_id(curr_attr_val);
+                }
+            }
+        }
+        let parent_lvol = Bdev::lookup_by_uuid_str(
+            snapshot_param.parent_id().unwrap().as_str(),
```
tiagolobocastro: What if parent_id is None? 
Should we handle that so we don't crash? Then we can use TryFrom instead of From trait

io-engine/src/grpc/v1/replica.rs
```diff
@@ -106,6 +116,74 @@ impl Default for ReplicaService {
...
+        )
+        .and_then(|b| Lvol::try_from(b).ok())
+        .expect("Invalid Parent Lvol");
```
tiagolobocastro: here as well, should probably return error instead of crashing

io-engine/src/grpc/v1/replica.rs
```diff
@@ -468,14 +547,16 @@ impl ReplicaRpc for ReplicaService {
...
+                            info!("Create Snapshot Success for {lvol:?}, {snap_lvol:?}");
                             Ok(CreateReplicaSnapshotResponse {
-                                snapshot_uuid: snap_config
-                                    .txn_id()
-                                    .unwrap_or_default(),
+                                replica_uuid: lvol.uuid(),
+                                snapshot: Some(ReplicaSnapshot::from(snap_lvol)),
                             })
                         }
+                        Ok(None) => {
```
tiagolobocastro: Why do we return an Option?

io-engine/src/grpc/v1/replica.rs
```diff
@@ -106,6 +116,74 @@ impl Default for ReplicaService {
...
+                    attr_id.as_ptr(),
+                    &mut val as *mut *const c_void,
+                    &mut size as *mut u64,
+                );
+                curr_attr_val = String::from_raw_parts(
+                    val as *mut u8,
+                    size as usize,
+                    size as usize,
+                );
+            }
```
```suggestion
            let curr_attr_val = unsafe {
                 let _r = spdk_blob_get_xattr_value(
                    blob,
                    attr_id.as_ptr(),
                    &mut val as *mut *const c_void,
                    &mut size as *mut u64,
                );
                String::from_raw_parts(
                    val as *mut u8,
                    size as usize,
                    size as usize,
                )
            };
```

io-engine/src/grpc/v1/replica.rs
```diff
@@ -492,4 +573,99 @@ impl ReplicaRpc for ReplicaService {
...
+            GrpcClientContext::new(&request, function_name!()),
+            async move {
+                let args = request.into_inner();
+                info!("{:?}", args);
+                let rx = rpc_submit(async move {
+                    let bdev = UntypedBdev::bdev_first().expect("Failed to enumerate devices");
+
+                    let mut devices = bdev
+                        .into_iter()
+                        .filter(|b| b.driver() == "lvol" && b.uuid_as_string() == args.snapshot_uuid)
```
tiagolobocastro: Rather than filter you could use .find which returns an Option rather than a list, then you don't have to handle multi instances found, which shouldn't really happen.

io-engine/src/grpc/v1/replica.rs
```diff
@@ -492,4 +573,99 @@ impl ReplicaRpc for ReplicaService {
...
+                    if devices.len() > 1 || devices.is_empty() {
+                        return Err(LvsError::Invalid {
+                                source: Errno::ENOENT,
+                                msg: format!(
+                                    "Snapshot {} not found or multiple instance of snapshot with same name present",
+                                    args.snapshot_uuid
+                                ),
+                        })
+                    }
+                    devices.remove(0).destroy().await?;
```
tiagolobocastro: On restart, do snapshots get loaded as soon as parent replica is loaded?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -404,14 +404,13 @@ impl Lvol {
...
     }
-
     async fn do_create_snapshot(
         &self,
         snap_param: SnapshotParams,
         done_cb: unsafe extern "C" fn(*mut c_void, *mut spdk_lvol, i32),
         done_cb_arg: *mut ::std::os::raw::c_void,
-        receiver: Option<oneshot::Receiver<i32>>,
-    ) -> Result<(), Error> {
+        receiver: Option<oneshot::Receiver<(i32, Lvol)>>,
```
tiagolobocastro: why is receiver an option?

io-engine/src/subsys/nvmf/admin_cmd.rs
```diff
@@ -227,8 +227,8 @@ pub fn create_snapshot(
     // Blobfs operations must be on md_thread
     Reactors::master().send_future(async move {
         match lvol.create_snapshot(snap_param).await {
-            Ok(()) => {
-                info!("Create Snapshot Success!!!");
+            Ok(lvol) => {
+                info!("Create Snapshot {:?} Success!!!", lvol);
```
tiagolobocastro:
```suggestion
                info!("Create Snapshot {:?} Success!", lvol);
```

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: > why is receiver an option?

it is used in create_snapshot_remote() function where no receiver parameter is used. create_snapshot() need receiver parameter as We need to get the lvol as part of SPDK callback. I guess similar things to be done in remote snapshot create (As part of NexusCreateSnapshot), which is not part of current PR scope.

tiagolobocastro: Ah, I see... in such case I suggest also having 2 functions, one with a receiver and another without a receiver at all, and the one with the receiver call the function without the receiver, if you see what I mean?

hrudaya21: ok.. let me check if create_snapshot_remote() also need a receiver to get the snapshot UUID in callback. In that case, We can use single function. Else will use different function for local and remote snapshot create.

io-engine/src/core/snapshot.rs
```diff
@@ -28,6 +30,39 @@ impl SnapshotParams {
     }
 }
 
+/// VolumeSnapshot Parameters to list the snapshots.
```
mtzaurus: It looks like it's not Parameters to list snapshots but rather a snapshot descriptor that represents existing snapshot.

io-engine/src/core/snapshot.rs
```diff
@@ -64,23 +99,63 @@ pub trait SnapshotOps {
     fn prepare_snap_config(
         &self,
         snap_name: &str,
+        entity_id: &str,
         txn_id: &str,
     ) -> Option<SnapshotParams>;
+
+    /// List Snapshots.
+    fn list_snapshot(self) -> Vec<VolumeSnapshotDescriptor>;
```
mtzaurus: Why do we consume Lvol object when listing its snapshots ?

hrudaya21: > Why do we consume Lvol object when listing its snapshots ?

While listing snapshot lvol, need to filter and show snapshot specific to input Lvol. (By matching parent uuid with input Lvol UUID)
There are 2 API implemented. 'list_snapahot' and 'list_all_snapshot'
First one is lvol specific snapshot, 2nd one to list all snapshot in the system.

io-engine/src/grpc/v1/replica.rs
hrudaya21: Taken a different approach to update the replica uuid and size. Fault case is handled now.

io-engine/src/grpc/v1/replica.rs
hrudaya21: Code changed.

io-engine/src/grpc/v1/replica.rs
hrudaya21: It will be handled based on decision about separte function for local and remote snapshot. Currently We wanted to keep it as a single function with Option as parameter.

io-engine/src/grpc/v1/replica.rs
hrudaya21: Updated Changes.

io-engine/src/grpc/v1/replica.rs
hrudaya21: Thank you for the suggestion.  Never observed difference between filter and find :) 
Changed the code.

io-engine/src/grpc/v1/replica.rs
hrudaya21: > On restart, do snapshots get loaded as soon as parent replica is loaded?

The bahaviour for snapshot must be same as basic replica. As post restart, all replica loaded back, same should happen for snapshot as well as both are lvol from SPDK point of view. Need a test verification to double sure on this.

tiagolobocastro: Thanks for confirming :+1:

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -980,28 +1077,124 @@ impl SnapshotOps for Lvol {
...
+            }
+            let blob = lvol.bs_iter_first();
+            let mut skip_lvol = false;
+            let mut snapshot_param: SnapshotParams = Default::default();
+            for attr in SnapshotXattrs::iter() {
+                let mut val = std::ptr::null();
+                let mut size: u64 = 0;
+                let attr_id = attr.name().to_string().into_cstring();
+                let curr_attr_val;
+                unsafe {
```
tiagolobocastro: Can do the same, and just "return" `String::from_raw_parts(` from the unsafe scope.
```suggestion
                let curr_attr_val = unsafe {
```

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -980,28 +1077,124 @@ impl SnapshotOps for Lvol {
...
+                    let r = spdk_blob_get_xattr_value(
+                        blob,
+                        attr_id.as_ptr(),
+                        &mut val as *mut *const c_void,
+                        &mut size as *mut u64,
+                    );
+                    // skip current lvol if any of the snapshot attribute not
+                    // found.
+                    if r != 0 {
```
tiagolobocastro: hmm if we skip, means we don't list snapshots, so we potentially leak them?

io-engine/src/grpc/v1/replica.rs
tiagolobocastro: IMHO with a single fn this api can be a bit confusing, as it's not obvious that this is the behaviour.

io-engine/src/grpc/v1/replica.rs
```diff
@@ -468,14 +560,18 @@ impl ReplicaRpc for ReplicaService {
...
                             Ok(CreateReplicaSnapshotResponse {
-                                snapshot_uuid: snap_config
-                                    .txn_id()
-                                    .unwrap_or_default(),
+                                replica_uuid: lvol.uuid(),
+                                snapshot: Some(ReplicaSnapshot::from(snapshot_descriptor)),
                             })
                         }
+                        Ok(None) => {
+                            todo!("Replica Snapshot will not hit this path");
```
tiagolobocastro: suggest changing from todo to something else, maybe a plain panic or assert

io-engine/src/grpc/v1/replica.rs
```diff
@@ -492,4 +588,103 @@ impl ReplicaRpc for ReplicaService {
...
+    ) -> GrpcResult<ListReplicaSnapshotsResponse> {
+        self.locked(
+            GrpcClientContext::new(&request, function_name!()),
+            async move {
+                let args = request.into_inner();
+                info!("{:?}", args);
+                let rx = rpc_submit(async move {
+                    // if replica_uuid is valid, filter snapshot based on
+                    // replica_uuid
+                    if !args.replica_uuid.is_empty() {
```
tiagolobocastro: hmm instead we should make replica_uuid optional to make it clear that not providing the uuid is a way to get all snapshots.

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -443,15 +442,108 @@ impl Lvol {
...
+            }
+            let blob = lvol.bs_iter_first();
+            let mut skip_lvol = false;
+            let mut snapshot_param: SnapshotParams = Default::default();
+            for attr in SnapshotXattrs::iter() {
+                let mut val = std::ptr::null();
+                let mut size: u64 = 0;
+                let attr_id = attr.name().to_string().into_cstring();
+                let curr_attr_val;
+                unsafe {
```
tiagolobocastro: this code seems a bit repeated, can we have a helper fn which can be called from multiple locations?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -443,15 +442,108 @@ impl Lvol {
...
+                }
+            }
+            // skip current lvol if any of the snapshot attribute not found.
+            if skip_lvol {
+                continue;
+            }
+            // set remaining snapshot parameters for snapshot list
+            snapshot_param.set_name(lvol.name());
+            let replica_lvol = Bdev::lookup_by_uuid_str(
+                snapshot_param.parent_id().unwrap().as_str(),
```
tiagolobocastro: should this unwrap be handled?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -443,15 +442,108 @@ impl Lvol {
...
+            )
+            .and_then(|b| Lvol::try_from(b).ok())
+            .expect("Invalid Parent Lvol");
```
tiagolobocastro: should this unwrap be handled?

io-engine/src/lvs/lvs_lvol.rs
tiagolobocastro: As this is the list... should we somehow return "broken" snapshots as well, with just uuid set so at least CP is aware of these "broken" snapshots?

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: Added a new flag to track the broken snapshots and return the flag to control plane. Please check once.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: Added a new flag to track the broken snapshots and return the flag to control plane. Please check once.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: Added an utility function.

io-engine/src/grpc/v1/replica.rs
hrudaya21: Accepted. Made the parameter Optional. Thanks for finding. Learn Optional parameter handling in gRPC. :)

io-engine/src/grpc/v1/replica.rs
hrudaya21: It is removed now, as function split is done for local and remote and `Option` is removed now.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: Added a new flag to track the broken snapshots and return the flag to control plane. Please check once.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: It is handled differently now by adding a common utility function.

io-engine/src/grpc/v1/replica.rs
hrudaya21: Changed to 2 separate function and Option parameter removed to simplify the return handling.

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -442,16 +441,154 @@ impl Lvol {
...
+        // stored inside CStrings.
+        let mut cstrs: Vec<CString> = Vec::new();
+
+        self.prepare_snapshot_xattrs(
+            &mut attr_descrs,
+            snap_param.clone(),
+            &mut cstrs,
+        )?;
+
+        let c_snapshot_name = snap_param.name().unwrap().into_cstring();
```
tiagolobocastro: should we handle unwrap and return error?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -442,16 +441,154 @@ impl Lvol {
...
+    pub fn get_blob_xattr(
+        snapshot: &Lvol,
+        attr: &SnapshotXattrs,
+    ) -> (String, bool) {
+        let mut valid_snapshot = true;
+        let mut val = std::ptr::null();
+        let mut size: u64 = 0;
+        let attr_id = attr.name().to_string().into_cstring();
+        let curr_attr_val;
+        unsafe {
```
```suggestion
        let curr_attr_val = unsafe {
```

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -442,16 +441,154 @@ impl Lvol {
...
+            // skip lvol if it is not snapshot.
+            if !lvol.is_snapshot() {
+                continue;
+            }
+            let mut valid_snapshot = true;
+            let mut snapshot_param: SnapshotParams = Default::default();
+            for attr in SnapshotXattrs::iter() {
+                let curr_attr_val;
+                (curr_attr_val, valid_snapshot) =
+                    Self::get_blob_xattr(&lvol, &attr);
```
tiagolobocastro: hmm valid_snapshot keeps getting overridden on each iteration of the for loop right?
Should it be "sticky" ?
Otherwise, rather than having it returned from get_xattr this could we simply have a fn in VolumeSnapshotDescriptor that checks if any param is none or empty and expose a self.valid() ?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -442,16 +441,154 @@ impl Lvol {
...
+        if lvol_devices.len() <= 1 {
+            return snapshot_list;
+        }
+        for lvol in lvol_devices {
+            // skip lvol if it is not snapshot.
+            if !lvol.is_snapshot() {
+                continue;
+            }
+            let mut valid_snapshot = true;
+            let mut snapshot_param: SnapshotParams = Default::default();
```
tiagolobocastro: all this code still seems very similar to the list_all, is there any way to reuse some code?

io-engine/src/core/snapshot.rs
mtzaurus: 'create_snapshot()' (and most of the other methods) uses '&self', but here we use 'self', i.e. the object is consumed. This is not fully aligned with Lvol API, IMHO.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: Accepted. Common Utility function is used to handle both list and list_all snapshot API.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: Accepted. Changed the logic.

io-engine/src/core/snapshot.rs
hrudaya21: Accepted. Changed the signature.

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -404,14 +403,13 @@ impl Lvol {
...
-
     async fn do_create_snapshot(
         &self,
         snap_param: SnapshotParams,
         done_cb: unsafe extern "C" fn(*mut c_void, *mut spdk_lvol, i32),
         done_cb_arg: *mut ::std::os::raw::c_void,
-        receiver: Option<oneshot::Receiver<i32>>,
-    ) -> Result<(), Error> {
+        receiver: oneshot::Receiver<(i32, Lvol)>,
+    ) -> Result<Lvol, Error> {
```
tiagolobocastro: Can this function call do_create_snapshot_remote ?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -442,17 +440,182 @@ impl Lvol {
...
+                Some(val) => val,
+                None => {
+                    valid_snapshot = false;
+                    continue;
+                }
+            };
+            match attr {
+                SnapshotXattrs::ParentId => {
+                    if replica_based_snapshot {
+                        let parent_lvol = parent.unwrap();
```
tiagolobocastro: might be clearer with if let some
```suggestion
                    if let Some(parent_lvol = parent {
```
and below as well

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -442,17 +440,182 @@ impl Lvol {
...
+                    attr_id, snapshot
+                );
+                return None;
+            }
+            String::from_raw_parts(val as *mut u8, size as usize, size as usize)
+        };
+        Some(curr_attr_val)
+    }
+    /// Common API to set SnapshotDescriptor for ListReplicaSnapshot.
+    pub fn set_snapshot_descriptor(
```
tiagolobocastro: name is a bit misleading, seems we're setting something
```suggestion
    pub fn snapshot_descriptor(
```

tiagolobocastro (approved):

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: no. Currently do_create_snapshot() and do_create_snapshot_remote() are 2 different function for local and remote snapshot operation respectively.

mtzaurus (approved):

hrudaya21:
bors try

hrudaya21:
bors try

hrudaya21:
bors merge

PR #1381, opened 2023-05-12
fix(nexus): fixing child remove logic

dsavitskiy:
* I/O is properly paused when a child is removed.
* ETCD is updated before removing a child.
* A test for child remove is added.

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev_children.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev_children.rs
@@ -262,37 +262,65 @@ impl<'n> Nexus<'n> {
 
         self.check_child_remove_operation(uri)?;
 
-        let paused = self.as_mut().pause_rebuild_jobs(uri).await;
+        if self.lookup_child(uri).is_none() {
+            return Ok(());
+        }
 
-        let idx = match self.children_iter().position(|c| c.uri() == uri) {
-            None => {
-                paused.resume().await;
-                return Ok(());
-            }
-            Some(val) => val,
-        };
+        // Pause subsystem and rebuild jobs.
+        debug!("{self:?}: remove child {uri}: pausing...");
+        let paused = self.pause_rebuild_jobs(uri).await;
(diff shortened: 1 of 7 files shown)


dsavitskiy:
bors try

tiagolobocastro (approved):

io-engine/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -262,37 +262,65 @@ impl<'n> Nexus<'n> {
...
-            Some(val) => val,
-        };
+        // Pause subsystem and rebuild jobs.
+        debug!("{self:?}: remove child {uri}: pausing...");
+        let paused = self.pause_rebuild_jobs(uri).await;
+        if let Err(e) = self.as_mut().pause().await {
+            error!(
+                "{self:?}: remove child {uri}: failed to pause subsystem: {e}"
+            );
+            paused.resume().await;
```
dsharma-dc, line 276: If say one of the the cancelled rebuilding job was rebuilding partially, do we ensure that the resume() here in a negative case leads to full rebuild ?

tiagolobocastro: hmm that's a good point, I suspect we'll have to full rebuild atm, but let's see what @dsavitskiy says.
I suspect the "saving grace" here is that CP should not usually do this, if decrementing replica count it would instead remove the degraded rebuild child.

dsavitskiy: It's a "pause", but rather cancellation. When "resumed", all "paused" jobs are actually restarted. Partial rebuild jobs will become full, as the rebuild map will be lost. And all jobs will be copying all the data again.
That's a known issues with our "pause/resume" behaviour. We'll need to deal with that at some point.

dsharma-dc (approved):

dsavitskiy:
bors merge

PR #1385, opened 2023-05-18
feat(nexus): adding support for disabling nexus reset logic

dsavitskiy:
In response to initiator timeout, nexus could be reset. This seemed to cause issues, so nexus reset logic is now temporarily disabled by default.
Reset logic can be re-enabled by setting env var `ENABLE_NEXUS_RESET=1`.

### DIFF
--- a/io-engine/src/subsys/nvmf/subsystem.rs
+++ b/io-engine/src/subsys/nvmf/subsystem.rs
@@ -3,12 +3,16 @@ use std::{
     fmt::{self, Debug, Display, Formatter},
     mem::size_of,
     ptr::{self, NonNull},
+    sync::atomic::Ordering,
 };
 
 use futures::channel::oneshot;
 use nix::errno::Errno;
 
-use crate::bdev::{nexus::nexus_lookup_mut, nvmx::NVME_CONTROLLERS};
+use crate::bdev::{
+    nexus::{nexus_lookup, nexus_lookup_mut},
+    nvmx::NVME_CONTROLLERS,
+};
 use spdk_rs::libspdk::{
     nvmf_subsystem_find_listener,
     nvmf_subsystem_set_ana_state,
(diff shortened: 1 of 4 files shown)


io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -308,71 +325,101 @@ impl NvmfSubsystem {
...
             }
         };
 
-        let mut nexus = match nexus_lookup_mut(&nexus_name) {
+        let nex = match nexus_lookup(&nexus_name) {
             Some(v) => v,
             None => {
-                debug!(
-                    "subsys {subsys_nqn} does not contain a valid nexus name"
+                warn!(
```
tiagolobocastro: This should not be a warn, probably also should not be logged, as this will show for replicas, which is a bit confusing.

tiagolobocastro (approved):
Btw your commit mentions `"ENABLE_NEXUS_RESET"` but the code has `"NEXUS_RESET"`

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -308,71 +325,101 @@ impl NvmfSubsystem {
...
-                unsafe {
-                    nexus.as_mut().rm_initiator(&hostnqn);
+                info!(
+                    "NVMF event handler: {nex:?}: \
+                    initiator timed out: '{hostnqn}'"
+                );
+
+                nex.rm_initiator(&hostnqn);
+
+                if !ENABLE_NEXUS_RESET.load(Ordering::SeqCst) {
```
dsharma-dc, line 367: This check needs to be better moved to the place where this event handler is registered via
`spdk_nvmf_subsystem_register_for_event()`, otherwise we'd keep calling add_initiator and rm_initiator for other events.

dsharma-dc (approved):

io-engine/src/subsys/nvmf/subsystem.rs
dsharma-dc: As discussed, we are intending to disable selectively only the reset event part here.

io-engine/src/subsys/nvmf/subsystem.rs
dsavitskiy: will fix

dsavitskiy:
bors merge

tiagolobocastro:
bors merge

PR #1393, opened 2023-05-29
feat(nexus): nexus snapshot

mtzaurus:
Draft pr to check BDD status / regressions.

### DIFF
--- a/io-engine/src/grpc/v0/mayastor_grpc.rs
+++ b/io-engine/src/grpc/v0/mayastor_grpc.rs
@@ -11,7 +11,11 @@
 use crate::{
     bdev::{
         nexus,
-        nexus::FaultReason,
+        nexus::{
+            Error as NexusError,
+            FaultReason,
+            NexusReplicaSnapshotDescriptor,
+        },
         NvmeControllerState as ControllerState,
         PtplFileOps,
     },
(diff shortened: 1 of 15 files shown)


mtzaurus:
bors try

mtzaurus:
bors try

mtzaurus:
bors try

io-engine/src/grpc/v0/mayastor_grpc.rs
```diff
@@ -1741,18 +1745,51 @@ impl mayastor_server::Mayastor for MayastorSvc {
                 let uuid = args.uuid.clone();
                 debug!("Creating snapshot on nexus {} ...", uuid);
```
hrudaya21, line 1749: Possibly We need to change the gRPC request name. As We are going to support create_snapshot for both replica and nexus. We may need to revisit the name ?
`create_replica_snapshot`
`create_nexus_snapshot`
Both of these operations can be put part of snapshot.proto and expose to Control plane.
https://github.com/openebs/mayastor-api/pull/63

mtzaurus: We aren't going to support replica snapshots in v0, only nexus one. In v1 we already have a good separation of API.

hrudaya21: > We aren't going to support replica snapshots in v0, only nexus one. In v1 we already have a good separation of API.

Ok got it. What about changing name in `v1/nexus.rs` ?

hrudaya21 (approved):

hrudaya21 (approved):

dsavitskiy (approved):

mtzaurus:
bors merge

PR #1394, opened 2023-05-29
fix(snapshot): io-engine panic while snapshot create with duplicate snapshot name

hrudaya21:


mtzaurus (commented):
I would also add checks against EEXIST at lvol level to handle this error properly inside I/O engine and avoid panics.

hrudaya21:
> I would also add checks against EEXIST at lvol level to handle this error properly inside I/O engine and avoid panics.

Avoid the panic scenario inside Lvol. Please check once.

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -1103,23 +1103,25 @@ impl SnapshotOps for Lvol {
     async fn create_snapshot(
         &self,
         snap_param: SnapshotParams,
-    ) -> Result<Lvol, Error> {
+    ) -> Result<Option<Lvol>, Error> {
```
tiagolobocastro: Why not return Error::Something instead of having an option in the return? makes it rather confusing.

io-engine/src/grpc/v1/replica.rs
```diff
@@ -555,6 +555,26 @@ impl ReplicaRpc for ReplicaService {
                             })
                         }
                     };
+                    // validate snapshot name
+                    if let Some(_r) = UntypedBdev::lookup_by_name(&args.snapshot_name) {
+                        return Err(LvsError::Invalid {
+                            source: Errno::EEXIST,
```
tiagolobocastro, line 564: Please map these source errors to grpc status code on the From<Lvs> to tonic::Status, I believe existing ones are already missing, and get converted to the wrong grpc status code.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: > Why not return Error::Something instead of having an option in the return? makes it rather confusing.

SPDK Callback function return spdk_lvol pointer which needs to be taken forward. In the failure scenario, this pointer will be NULL. To handle both Null and valid Lvol scenario, opted for Option<Lvol> to handle the return value of the callback.
s.send((errno, None)).err(); => During Error Flow
s.send((errno, Some(lvol))).ok(); => During Success flow.

If need to handle without Option, then need to use
s.send(error, lvol) (without ok or error) and receiver will handle the error flow. but here lvol itself can be null in error flow.

Possibly, I am missing some other better approach here.

hrudaya21: > Why not return Error::Something instead of having an option in the return? makes it rather confusing.

Simplified the changes to handle the Null Callback without Option.

io-engine/src/grpc/v1/replica.rs
hrudaya21: > Please map these source errors to grpc status code on the From to tonic::Status, I believe existing ones are already missing, and get converted to the wrong grpc status code.

Updated Code with new gRPC error code.

io-engine/src/lvs/lvs_lvol.rs
tiagolobocastro: Seems you've made some code changes.
What I was suggesting was not to modify that entirely, but to instead map the None to and error prior to returning, example:
```rust
match option_result {
Ok(None) => Err(Error::LvsInvalid { source: NoDev }),
Ok(Some(lvol)) => Ok(lvol),
Err(error) => Err(error)
}
```

io-engine/src/grpc/v1/replica.rs
```diff
@@ -555,6 +555,18 @@ impl ReplicaRpc for ReplicaService {
                             })
                         }
                     };
+                    // validate snapshot name
+                    if let Some(_r) = UntypedBdev::lookup_by_name(&args.snapshot_name) {
+                        return Err(LvsError::SnapshotNameExist {
```
tiagolobocastro: I think you misunderstood my comment. I meant a convertion from source was missing

tiagolobocastro: Here, we're missing the map from other Errnos, we're only mapping EIVAL and ENOMEDIUM.
```rust
LvsError::Invalid {
                source, ..
            } => match source {
                Errno::EINVAL => Status::invalid_argument(e.to_string()),
                Errno::ENOMEDIUM => Status::failed_precondition(e.to_string()),
                _ => Status::invalid_argument(e.to_string()),
            },
```

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: Got your idea. 
Currently the SPDK callback was returning Null Pointer in the error flow. `Lvol::from_inner_ptr(lvol_ptr)` This was panicing if handled in the callback. To simplify it, currently the Null pointer handling is kept in the error flow and Success flow handles the valid lvol_ptr.

io-engine/src/grpc/v1/replica.rs
hrudaya21: Added the mapping for missing Error code used.

io-engine/src/grpc/v1/replica.rs
```diff
@@ -563,21 +583,24 @@ impl ReplicaRpc for ReplicaService {
...
                     match lvol.create_snapshot(snap_config.clone()).await {
                         Ok(snap_lvol) => {
-                            info!("Create Snapshot Success for {lvol:?}, {snap_lvol:?}");
-                            let snapshot_descriptor =
-                                ReplicaSnapshotDescriptor::new(snap_lvol, lvol.uuid(), lvol.size());
-                            Ok(CreateReplicaSnapshotResponse {
-                                replica_uuid: lvol.uuid(),
-                                snapshot: Some(ReplicaSnapshot::from(snapshot_descriptor)),
-                            })
+                                info!("Create Snapshot Success for {lvol:?}, {snap_lvol:?}");
```
tiagolobocastro: Is there too much tabbing here?

io-engine/src/grpc/v1/replica.rs
```diff
@@ -563,21 +583,24 @@ impl ReplicaRpc for ReplicaService {
...
                             Some(snap_config) => snap_config,
-                            None => return Err(LvsError::SnapshotConfigFailed {
-                                name: args.replica_uuid,
-                                msg: "tx id / snapshot name not provided".to_string(),
+                            None => return Err(LvsError::Invalid {
+                                source: Errno::EINVAL,
+                                msg: format!(
+                                    "tx id / snapshot name not provided for replica {}",
+                                    args.replica_uuid
+                                ),
```
tiagolobocastro, line 594: nit: btw this could be replaced with let else:
```rust
                    let Some(snap_config) =
                        lvol.prepare_snap_config(
                            &args.snapshot_name,
                            &args.entity_id,
                            &args.txn_id
                        ) else {
                            return Err(LvsError::Invalid {
                                source: Errno::EINVAL,
                                msg: format!(
                                    "tx id / snapshot name not provided for replica {}",
                                    args.replica_uuid
                                ),
                            });
                        };
```

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

hrudaya21:
bors try

hrudaya21:
bors merge

hrudaya21:
bors merge

tiagolobocastro:
bors merge

PR #1395, opened 2023-05-30
Fix(snapshot): custom uuid to consider from control plane while creating snapshot.

hrudaya21:


io-engine/src/grpc/v1/replica.rs
```diff
@@ -560,20 +563,24 @@ impl ReplicaRpc for ReplicaService {
...
                         ) {
                             Some(snap_config) => snap_config,
                             None => return Err(LvsError::SnapshotConfigFailed {
                                 name: args.replica_uuid,
                                 msg: "tx id / snapshot name not provided".to_string(),
                             })
                         };
+                    let replica_uuid = lvol.uuid();
+                    let replica_size = lvol.size();
+                    info!("Prepare Snapshot for Replica {lvol:?}");
```
tiagolobocastro: Should this be debug?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -388,6 +388,15 @@ impl Lvol {
                         })
                     }
                 },
+                SnapshotXattrs::SnapshotUuid => match params.snapshot_uuid() {
```
tiagolobocastro, line 391: Why is uuid an xattr? This means it can potentially be nil if the attrs get corrupted for some reason?
Did we consider storing uuid in the uuid of the spdk snapshot (though not sure if that one is "persistent"?) Alternatively could we set the uuid as the spdk snapshot name, and store the name as an xattr? So this way we might loose the name, but never the uuid?

io-engine/src/grpc/v1/replica.rs
hrudaya21: Missed to remove it. It was added for my testing. Will remove now.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: > Why is uuid an xattr? This means it can potentially be nil if the attrs get corrupted for some reason? Did we consider storing uuid in the uuid of the spdk snapshot (though not sure if that one is "persistent"?) Alternatively could we set the uuid as the spdk snapshot name, and store the name as an xattr? So this way we might loose the name, but never the uuid?

Currently the newly added SPDK API take one xattr array where 2 parameters are reserved for `name` and `uuid` of the snapshot and other parameters are dynamically controlled by dataplane. This has simplified the interface and can be used for future expansion to add new attribute. These all attributes are persistent to blobstore so chances of loosing these data might be very less. As We are considering the `uuid` from xattr and save it to snapshot lvol uuid, there is always a copy in lvol.

tiagolobocastro: Ok sounds good

io-engine/src/core/snapshot.rs
```diff
@@ -79,6 +83,7 @@ impl SnapshotXattrs {
             Self::TxId => "mayastor.tx_id",
             Self::EntityId => "mayastor.entity_id",
             Self::ParentId => "mayastor.parent_id",
+            Self::SnapshotUuid => "uuid",
```
tiagolobocastro, line 86: any reason some are mayastor prefixed and others are not?

tiagolobocastro (approved):

io-engine/src/core/snapshot.rs
hrudaya21: > any reason some are mayastor prefixed and others are not?

I think We can uniform it. Last parameter is without `mayastor` prefix as in SPDK code the xattr name is without prefix.
Other 3 parameters with prefix can be changed back without prefix or last one to change to with prefix both in SPDK code and DP code. @mtzaurus What is your suggestion ?

tiagolobocastro: Discussed in a meeting, we could simply use io-engine as the prefix.

tiagolobocastro (approved):

mtzaurus (approved):

mtzaurus (approved):

hrudaya21:
bors try

mtzaurus:
bors try

hrudaya21:
bors merge

PR #1397, opened 2023-06-01
fix(snapshot): refactor and isloate snapshot common api to independent file

hrudaya21:
1. List and Delete Snapshot API moved to Snapshot Proto
2. CLI of List and Delete Snapshot is moved from replica section to common snapshot section.


io-engine/src/bin/io-engine-client/v1/snapshot_cli.rs
```diff
@@ -18,6 +18,8 @@ pub async fn handler(
 ) -> crate::Result<()> {
     match matches.subcommand() {
         ("create", Some(args)) => create(ctx, args).await,
```
hrudaya21: Need a suggestion here, if We need to keep separate `create` snapshot operation in CLI. as it will lead to confusion.
Currently create snapshot will be placed in `replica` and `nexus` section.

tiagolobocastro: Why not move creates to snapshot api then?
hmm alternatively starting to wonder if it's better to leave things as they are..

io-engine/src/core/snapshot.rs
```diff
@@ -113,6 +146,9 @@ pub trait SnapshotOps {
 
     /// List Snapshots.
     fn list_snapshot(&self) -> Vec<VolumeSnapshotDescriptor>;
+
+    /// List Single snapshot details based on UUID.
+    fn list_single_snapshot(&self) -> Vec<VolumeSnapshotDescriptor>;
```
tiagolobocastro: nit: consider renaming list_snapshot to list_snapshots, and list_single_snapshot to list_snapshot

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -1192,33 +1193,52 @@ impl SnapshotOps for Lvol {
...
+                })
+                .map(|b| Lvol::try_from(b).unwrap())
+            {
+                if let Some(snapshot_descriptor) =
+                    lvol.snapshot_descriptor(None)
+                {
+                    snapshot_list.push(snapshot_descriptor);
+                }
+            }
+        };
```
tiagolobocastro: `;` not needed AFAICT

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -1192,33 +1193,52 @@ impl SnapshotOps for Lvol {
...
+                    None => continue,
                 }
-                None => continue,
             }
-        }
+        };
+        snapshot_list
+    }
+    /// List Single snapshot details based on UUID.
+    fn list_single_snapshot(&self) -> Vec<VolumeSnapshotDescriptor> {
```
tiagolobocastro: hmm, seems like these list snapshots won't work if the base replica lvol is deleted?

io-engine/src/grpc/v1/snapshot.rs
```diff
@@ -0,0 +1,431 @@
...
+                    snapshot_param.set_snapshot_uuid(curr_attr_val);
+                }
+            }
+        }
+        Self {
+            snapshot_uuid: snap_lvol.uuid(),
+            snapshot_name: snap_lvol.name(),
+            snapshot_size: snap_lvol.size(),
+            num_clones: 0, //TODO: Need to implement along with clone
+            timestamp: None, //TODO: Need to update xAttr to track timestamp
```
tiagolobocastro, line 136: Ah, so this is why I'm getting None in the control-plane :)

tiagolobocastro (approved):

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: Yes. If replica is deleted, then this API will not return the snapshot result as the snapshot property stored in blobstore will be deleted as part of replica deletion. The other ListSnapshot API (list_all_snapshot) will return the snapshot with valid flag as `false`

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: fixed.

io-engine/src/core/snapshot.rs
hrudaya21: modified the name.

io-engine/src/grpc/v1/snapshot.rs
hrudaya21: Yes. Created a new task for it. Will work on this.
https://dcsw-bolt.atlassian.net/browse/GTM-939

io-engine/src/lvs/lvs_lvol.rs
tiagolobocastro: Why is valid false after deleting the replica?

hrudaya21: > Why is valid false after deleting the replica?

During listing snapshot, snapshot parent-id (xattr) is validated. As the xattr (parent_uuid -> source_uuid) is not valid, the snapshot is marked as invalid.
But I think I can remove this check, as snapshot can be used after replica deletion. So making `valid_flag` false is not correct. But these 2 attributes of snapshot (parent-uuid and parent-size) will be set to 0.

hrudaya21: > > Why is valid false after deleting the replica?
> 
> During listing snapshot, snapshot parent-id (xattr) is validated. As the xattr (parent_uuid -> source_uuid) is not valid, the snapshot is marked as invalid. But I think I can remove this check, as snapshot can be used after replica deletion. So making `valid_flag` false is not correct. But these 2 attributes of snapshot (parent-uuid and parent-size) will be set to 0.

Updated the code to return the snapshot `valid_flag` as true if replica is deleted. `source_uuid` will be return `Null` and `source_size` will be return as 0.

tiagolobocastro: Why is the xattr not valid? Are these not persistent in the metadata?

hrudaya21: xattr information in the blobstor is valid, as it is presistent. But as the replica is deleted already, this information will misslead. So We are returning the result as `null` and `0` for `source_uuid` and `source_size` as part of ListSnapshot response.

mtzaurus (approved):

mtzaurus:
bors try

io-engine/src/lvs/lvs_lvol.rs
tiagolobocastro: Why would it mislead? 
Anyways, this means we have to make these optional across the stack then, CC @Abhinandan-Purkait

hrudaya21: The assumption here is ListSnapshot should not response any parameter which is invalid in the system. As replica is already deleted, DP should not expose that UUID to CP.

Abhinandan-Purkait: Yeah, we should probably make these fields optional accross the stack. That would be cleaner imo.

hrudaya21:
bors try

hrudaya21:
bors try

hrudaya21:
bors try

hrudaya21:
bors try

hrudaya21:
bors try

hrudaya21:
bors merge

PR #1398, opened 2023-06-01
fix(nexus): fixing persistent store update issue during device retire

dsavitskiy:


dsavitskiy:
bors try

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -264,4 +278,35 @@ impl<'n> NexusChannel<'n> {
...
         self.core
     }
+
+    /// Sets the current I/O mode for this channel.
+    pub(super) fn set_io_mode(&mut self, io_mode: NexusIOMode) {
+        self.io_mode = io_mode;
+        debug!("{self:?}: setting I/O mode to {io_mode:?}");
+        if matches!(self.io_mode, NexusIOMode::Normal) {
+            let mut v = Vec::new();
+            std::mem::swap(&mut v, &mut self.suspended_ios);
```
tiagolobocastro: Do we need this intermediate var, could we for example:
```rust
self.suspended_ios.drain(..).for_each(|io| {
        debug!("{io:?}: resubmitting previously suspended I/O");
        io.submit_request();
    });
```

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -264,4 +278,35 @@ impl<'n> NexusChannel<'n> {
...
+
+            debug!(
+                "{self:?}: resubmitting {n} suspended I/Os ...",
+                n = v.len()
+            );
+
+            v.into_iter().for_each(|io| {
+                debug!("{io:?}: resubmitting previously suspended I/O");
```
tiagolobocastro: should this be trace? Could be a lot of IO if several nexus see faulted children, eg: remote pod is restarted

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -264,4 +278,35 @@ impl<'n> NexusChannel<'n> {
...
+    }
+
+    /// Determines if I/Os are suspended.
+    pub(super) fn is_suspended(&self) -> bool {
+        matches!(self.io_mode, NexusIOMode::SuspendSubmit)
+    }
+
+    /// Suspends submission of the given Nexus I/O.
+    pub(super) fn suspend_io_submission(&mut self, io: NexusBio<'n>) {
+        debug!("{io:?}: suspending I/O");
```
tiagolobocastro: Also, should this be trace?

io-engine/src/bdev/nexus/nexus_io.rs
```diff
@@ -154,6 +154,12 @@ impl<'n> From<*mut spdk_bdev_io> for NexusBio<'n> {
     }
 }
 
+impl Clone for NexusBio<'_> {
```
tiagolobocastro: the derive clone macro cannot do this?

tiagolobocastro (approved):

io-engine/src/bdev/nexus/nexus_io.rs
```diff
@@ -185,6 +191,12 @@ impl<'n> NexusBio<'n> {
 
     /// TODO
     pub(super) fn submit_request(mut self) {
+        if self.channel().is_suspended() {
```
tiagolobocastro: I wonder if we need to take a different approach for when etcd is down for some time, and initiator IO timeout, presumably we need to abort IOs that we've actually "paused" so we don't fire them up when we switch back to "normal mode" ?

io-engine/src/bdev/nexus/nexus_channel.rs
dsavitskiy: ah, yes, forgot about this idiom.

io-engine/src/bdev/nexus/nexus_channel.rs
dsavitskiy: will fix

io-engine/src/bdev/nexus/nexus_channel.rs
dsavitskiy: will fix

io-engine/src/bdev/nexus/nexus_io.rs
dsavitskiy: Yes, it makes sense.

dsavitskiy:
bors try

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -18,6 +18,9 @@ pub struct NexusChannel<'n> {
     io_logs: Vec<IOLogChannel>,
     previous_reader: UnsafeCell<usize>,
     fail_fast: u32,
+    io_mode: IOMode,
+    // suspended_ios: Vec<NonNull<spdk_bdev_io>>,
```
tiagolobocastro: what's suspended_ios? Not needed?

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -38,6 +41,15 @@ impl<'n> Debug for NexusChannel<'n> {
     }
 }
 
+/// Channel I/O disposition.
+#[derive(Debug, Copy, Clone)]
+pub enum IOMode {
```
tiagolobocastro:
```suggestion
pub enum IoMode {
```

we use "Io*" elsewhere, eg: IoType/IoChannel/IoType

io-engine/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -171,51 +173,45 @@ impl<'n> Nexus<'n> {
...
-                        );
-                        output_err = false;
-                    }
+            let Err(err) = PersistentStore::put(&key, &info.inner).await else {
+                trace!(?key, "{self:?}: the state was saved successfully");
+                return Ok(());
+            };
+
+            retry -= 1;
+            if retry == 0 {
```
tiagolobocastro, line 225: should this check be <= 0 ? What if retries starts with 0 and then we decrement it.

tiagolobocastro: Also making me wonder if this should be time based rather than retry based.
Example, what if endpoint rejecting immediately due to some temporary condition? Then we only sleep for 1s between retries, so with 5 retries for example, we'd give up after 5s which seems very small?

io-engine/src/core/env.rs
```diff
@@ -80,6 +80,31 @@ fn parse_mb(src: &str) -> Result<i32, String> {
     }
 }
 
+/// Parses a timeout specification.
+fn parse_timeout(src: &str) -> Result<Duration, String> {
```
tiagolobocastro: Ah, you can use human-time crate as well btw, very useful for this.

io-engine/src/bdev/nexus/nexus_persistence.rs
dsavitskiy: I made retries tunable with CLI args, so an admin can tune it to a particular setup/network.

dsavitskiy:
bors try

mtzaurus (approved):

dsavitskiy:
bors merge

io-engine/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -71,16 +73,22 @@ pub(crate) enum PersistOp<'a> {
...
-            return;
+            return Ok(());
         }
 
         let mut persistent_nexus_info = self.nexus_info.lock().await;
+
+        // We have to freeze I/O (re-)submissions while doing that, to prevent
+        // an uncontrollable storm of I/O resubmissions in the case
+        // the persistent store is slow to response, or has failed.
+        self.set_nexus_io_mode(IoMode::Freeze).await;
```
dsharma-dc, line 87: Should we freeze only when op is UpdateCond? Otherwise it seems not necessary perhaps?

io-engine/src/bdev/nexus/nexus_io.rs
```diff
@@ -185,6 +175,12 @@ impl<'n> NexusBio<'n> {
 
     /// TODO
     pub(super) fn submit_request(mut self) {
+        if self.channel().is_frozen() {
+            let s = self.clone();
+            self.channel_mut().freeze_io_submission(s);
```
dsharma-dc, line 180: Have we run any heavy workload to see how big we can grow. I'm thinking of a condition where the freeze list is huge and keeps holding mempool objects, leading to failure in bio/dma allocation for new IOs.

dsharma-dc: Nexus subsystem is perhaps paused at this point so the limit would be bound by how many IOs are already in the system.

tiagolobocastro: It's actually not paused, so we will be receiving IOs...
This may be a problem indeed... though on the usual case where etcd it shouldn't be an issue, and if etcd is down we now have timeouts to give up.
Ah @dsavitskiy, when we give up and shutdown the nexus, do we free the IO's?

io-engine/src/bdev/nexus/nexus_persistence.rs
dsavitskiy: It won't hurt freeze for all conditions. It will prevent a device failure/retire while ETCD is being written.

io-engine/src/bdev/nexus/nexus_io.rs
dsavitskiy: @tiagolobocastro  When we give up writing to ETCD, we fail and drop all frozen I/Os.

We should probably create a separate ticket to handle ENOMEM during I/O submission properly, as this is not the only place when ENOMEM can happen. SPDK itself does have some ENOMEM-during-submit processing logic.

dsavitskiy: @dsharma-dc @tiagolobocastro I create an epic to cover ENOMEM issues: https://dcsw-bolt.atlassian.net/browse/GTM-997.

dsavitskiy:
bors merge

io-engine/src/bdev/nexus/nexus_io.rs
dsharma-dc: I'd suggest we quantify our vulnerability for ENOMEM under this scenario, and if we don't happen to run into this under a minimally/moderately loaded system then it can be addressed under this new epic.

PR #1399, opened 2023-06-02
chore(lvs/lvol): add trace message before zeroing super

tiagolobocastro:
This is useful to trace how long zeroing takes.
Also use write_zeroes variant that is now available which is much simpler.

### DIFF
--- a/io-engine/src/lvs/lvs_store.rs
+++ b/io-engine/src/lvs/lvs_store.rs
@@ -756,6 +756,8 @@ impl Lvs {
             })
             .map(Lvol::from_inner_ptr)?;
 
+        info!("{:?}: wiping super", lvol);
+
         if let Err(error) = lvol.wipe_super().await {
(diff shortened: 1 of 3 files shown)


hrudaya21 (approved):

io-engine/src/lvs/lvs_store.rs
```diff
@@ -756,6 +756,8 @@ impl Lvs {
             })
             .map(Lvol::from_inner_ptr)?;
 
+        info!("{:?}: wipping super", lvol);
```
dsharma-dc: typo: `wiping`

dsharma-dc (approved):

io-engine/src/lvs/lvs_store.rs
tiagolobocastro: good catch!

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1400, opened 2023-06-06
fix(lvol): improper snapshot enumeration when no devices exist

mtzaurus:
This fix improves snapshot enumeration when no devices exist in I/O agent: in such a case device iterator properly returns None which should not be viewed as an error.


hrudaya21 (approved):

dsavitskiy (approved):

mtzaurus:
bors merge

PR #1404, opened 2023-06-08
feat(rebuild_history): add list api with record limit per nexus

abhilashshetty04:
This PR adds rebuild list Api which will allow control plane to get rebuild history of all nexus on io engine using single rpc. Caller can set maximum number of rebuild records per nexus as well.

### DIFF
--- a/io-engine/src/grpc/v1/nexus.rs
+++ b/io-engine/src/grpc/v1/nexus.rs
@@ -17,16 +17,17 @@ use crate::{
     grpc::{rpc_submit, GrpcClientContext, GrpcResult},
     rebuild::{HistoryRecord, RebuildState, RebuildStats},
 };
+use mayastor_api::v1::nexus::*;
+
 use futures::FutureExt;
 use std::{
+    collections::HashMap,
     convert::{From, TryFrom, TryInto},
     fmt::Debug,
     ops::Deref,
     pin::Pin,
 };
 use tonic::{Request, Response, Status};
-
-use mayastor_api::v1::nexus::*;
 
 #[derive(Debug)]
 struct UnixStream(tokio::net::UnixStream);
@@ -208,11 +209,9 @@ impl From<RebuildState> for mayastor_api::v1::nexus::RebuildJobState {
     }
 }
 
-impl From<&HistoryRecord> for RebuildHistoryRecord {
-    fn from(record: &HistoryRecord) -> Self {
+impl From<HistoryRecord> for RebuildHistoryRecord {
+    fn from(record: HistoryRecord) -> Self {
         RebuildHistoryRecord {
-            child_uri: record.child_uri.clone(),
-            src_uri: record.src_uri.clone(),
             state: RebuildJobState::from(record.state) as i32,
             blocks_total: record.blocks_total,
             blocks_recovered: record.blocks_recovered,
@@ -223,6 +222,8 @@ impl From<&HistoryRecord> for RebuildHistoryRecord {
             is_partial: record.is_partial,
             start_time: Some(record.start_time.into()),
             end_time: Some(record.end_time.into()),
+            child_uri: record.child_uri,
+            src_uri: record.src_uri,
         }
     }
 }
@@ -1126,13 +1127,12 @@ impl NexusRpc for NexusService {
     ) -> GrpcResult<RebuildHistoryResponse> {
         let ctx = GrpcClientContext::new(&request, function_name!());
         let args = request.into_inner();
-
         self.serialized(ctx, args.uuid.clone(), false, async move {
             trace!("{:?}", args);
             let rx = rpc_submit::<_, _, nexus::Error>(async move {
                 let records = nexus_lookup(&args.uuid)?
                     .rebuild_history()
-                    .iter()
+                    .into_iter()
                     .map(RebuildHistoryRecord::from)
                     .collect();
                 Ok(RebuildHistoryResponse {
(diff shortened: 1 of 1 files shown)


io-engine/src/grpc/v1/nexus.rs
```diff
@@ -1147,4 +1148,47 @@ impl NexusRpc for NexusService {
...
+        let count = args.count.unwrap_or_default();
+        let rx = rpc_submit::<_, _, nexus::Error>(async move {
+            let mut history_map: HashMap<String, RebuildHistoryResponse> =
+                HashMap::new();
+            for nexus in nexus::nexus_iter() {
+                let mut history_list: VecDeque<RebuildHistoryRecord> =
+                    VecDeque::new();
+                let nexus_uuid = nexus.uuid().to_string();
+                for record in
+                    nexus_lookup(&nexus_uuid)?.rebuild_history().iter()
```
dsavitskiy: you don't need to lookup a nexus here, as you already have it (from "for nexus in nexus::nexus_iter()")

io-engine/src/grpc/v1/nexus.rs
```diff
@@ -1147,4 +1148,47 @@ impl NexusRpc for NexusService {
...
+                {
+                    // We will return all of the rebuild records for nexus if
+                    // count is 0,Apply only non zero
+                    // limit.
+                    if history_list.len() == count.encoded_len() && count != 0 {
```
dsavitskiy: If you need just to limit list size, why don't you use vec slice or an iterator helper method that could this more idiomatically?

io-engine/src/grpc/v1/nexus.rs
```diff
@@ -1147,4 +1148,47 @@ impl NexusRpc for NexusService {
...
+                        history_list.pop_front();
+                    };
+                    history_list.push_back(RebuildHistoryRecord::from(record));
```
dsavitskiy: You don't need to have a mut hashmap, you can write all this in completely functional way:
```
let hist_map: HashMap<_, _> = nexus::nexus_iter()
    .map(|nex| {
        let uuid = nex.uuid().to_string();

        // get and modify history list here as needed
        let hist = nex.rebuild_history();
        let s = if count == 0 || count > hist.len() {
            0
        } else {
            hist.len() - count
        };
        let hist = hist[s..].to_owned();

        // return a tuple of (uuid, RebuildHistoryResponse)
        (
            uuid.clone(),
            RebuildHistoryResponse {
                nexus: uuid.clone(),
                records: hist.into(),
            },
        )
    })
    .collect();  // Collecting an iterator of tuples will work for a HashMap
```

dsavitskiy (changes requested):

io-engine/src/grpc/v1/nexus.rs
abhilashshetty04: I wanted to remove first element specifically when list len reaches limit. This is because we push from the back so first element will be the older rebuild job. Hence thought of using VecDeque. With slice i can can form the whole list as is, but while truncating will i be able to give preference to objects at higher indexes?

io-engine/src/grpc/v1/nexus.rs
abhilashshetty04: Incorporated

io-engine/src/grpc/v1/nexus.rs
abhilashshetty04: Incorporated

tiagolobocastro: hmm you can also use rev and take to simplify the limit:
and probably don't need to clone uuid the second time

```rust
        let hist = nex.rebuild_history();
        let hist = hist.into_iter().rev().take(count).collect::<Vec<_>>();

        // return a tuple of (uuid, RebuildHistoryResponse)
        (
            uuid.clone(),
            RebuildHistoryResponse {
                nexus: uuid,
                records: hist.into(),
            },
        )
```

abhilashshetty04: removed the clone already. Ill try rev

abhilashshetty04: Incorporated

dsavitskiy (approved):

io-engine/src/grpc/v1/nexus.rs
```diff
@@ -1147,4 +1147,49 @@ impl NexusRpc for NexusService {
...
+                        0
+                    } else {
+                        hist.len() - count
+                    };
+                    let hist = hist
+                        .into_iter()
+                        .rev()
+                        .take(s)
+                        .collect::<Vec<_>>()
+                        .iter()
```
tiagolobocastro: Why are you getting s again, it's not needed right?

```suggestion
                    let hist = hist
                        .into_iter()
                        .rev()
                        .take(count)
```

abhilashshetty04: Done

io-engine/src/grpc/v1/nexus.rs
```diff
@@ -1147,4 +1147,42 @@ impl NexusRpc for NexusService {
...
         .await
     }
+
+    async fn list_rebuild_history(
+        &self,
+        request: Request<ListRebuildHistoryRequest>,
+    ) -> GrpcResult<RebuildHistoryMapResponse> {
+        let args = request.into_inner();
+        let count = args.count.unwrap_or_default() as usize;
+        let rx = rpc_submit::<_, _, nexus::Error>(async move {
```
tiagolobocastro, line 1158:
```suggestion
        trace!("{:?}", args);
        let rx = rpc_submit::<_, _, nexus::Error>(async move {
```

tiagolobocastro (approved):

abhilashshetty04:
bors try

abhilashshetty04:
bors merge

abhilashshetty04:
bors merge

abhilashshetty04:
bors try

abhilashshetty04:
bors try

abhilashshetty04:
bors try

abhilashshetty04:
bors merge

abhilashshetty04:
bors merge

avishnu:
bors merge

avishnu:
bors try

avishnu:
bors merge

tiagolobocastro:
Replaced by https://github.com/openebs/mayastor/pull/1408

PR #1406, opened 2023-06-09
fix(snapshot): support snapshot create time attribute during create and list snapshot

hrudaya21:


io-engine/src/bdev/nexus/nexus_bdev_snapshot.rs
```diff
@@ -186,13 +185,13 @@ impl ReplicaSnapshotExecutor {
                 // snapshots.
                 let snapshot_params = SnapshotParams::new(
                     snapshot.entity_id(),
-                    snapshot.parent_id(),
+                    Some(ctx.replica_uuid.clone()),
```
tiagolobocastro, line 188: why isn't parent_id correct here? because it's the parent id for the "nexus snapshot"?

io-engine/src/core/snapshot.rs
```diff
@@ -108,6 +101,7 @@ pub enum SnapshotXattrs {
     EntityId,
     ParentId,
     SnapshotUuid,
+    CreateTime,
```
tiagolobocastro: suggestion: simply Timestamp?

io-engine/src/core/snapshot.rs
```diff
@@ -117,6 +111,7 @@ impl SnapshotXattrs {
             Self::EntityId => "io-engine.entity_id",
             Self::ParentId => "io-engine.parent_id",
             Self::SnapshotUuid => "uuid",
+            Self::CreateTime => "io-engine.create_time",
```
tiagolobocastro: btw, are these prefixed somewhere else with snapshot? If not should they be?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -602,23 +613,16 @@ impl Lvol {
             .and_then(|b| Lvol::try_from(b).ok())
             {
                 Some(parent) => (parent.uuid(), parent.size()),
-                None => {
-                    warn!("Snapshot{:?}, parent not proper", self);
-                    // valid_snapshot = false;
-                    (String::default(), 0)
-                }
+                None => (String::default(), 0),
```
tiagolobocastro, line 601: When can this happen, this we see this message in control-plane tests, this is not expected?

tiagolobocastro (approved):

io-engine/src/bdev/nexus/nexus_bdev_snapshot.rs
hrudaya21: Yes, here parent_id() is nexus id, but We need to update the individual replica_uuid for the each replica snapshot of the nexus.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: It is expected when the replica is deleted and We are trying to list the Snapshot, then We show the parent information with default value.

io-engine/src/core/snapshot.rs
hrudaya21: If We wanted to add some more attribute for snapshot like `lastclonetime`, then just keeping `timestamp` as name may be not explain the attribute clearly.

io-engine/src/core/snapshot.rs
hrudaya21: you mean the prefix `io-engine` ? No it is not used anywhere else other than here. currently the idea is all xattr in `io-engine` will be prefix with `io-engine`. The category will be distringuished based on which structure/enum it belongs to. Here all belongs to `SnapshotXattrs`

mtzaurus (approved):

hrudaya21:
bors try

io-engine/src/lvs/lvs_lvol.rs
tiagolobocastro: hmm in control-plane the replica is not deleted IIRC, we're doing a list all snapshots

hrudaya21: > hmm in control-plane the replica is not deleted IIRC, we're doing a list all snapshots

Currently control-plane is calling list all snapshots without any attribute. The issue what you have faced last time, because of wrong parent-id details saved for NexusSnapshot. 
Regarding Dataplane interface point of view, We support 3 different kind of ListSnapshot
1. List All Snapshot (Currently in use by CP)
2. List Snapshot based on Source_uuid
3. List Snapshot based on snapshot_uuid

io-engine/src/core/snapshot.rs
tiagolobocastro: Ah I see, hmm might createTime still be confusing, could be confused with createclonetime? Maybe createSnapshotTime?

io-engine/src/core/snapshot.rs
tiagolobocastro: Ah if they're in a category then it's fine :)

io-engine/src/lvs/lvs_lvol.rs
tiagolobocastro: We also use 3.
Oh I see, the wrong parent id caused the list to think it didn't have the parent, got it, thanks for explaining :+1:

io-engine/src/core/snapshot.rs
hrudaya21: While thinking of Clone implementation I realized that Clone attribute should be under another Attr umbrella. Hence name can be kept here as `timestamp` or `createtime` no need of snapshot keyword as it is under `snapshot` tag. Let me change it to `timestamp` for the better uniformity for snapshot operations.

hrudaya21:
bors try

hrudaya21:
bors cancel

hrudaya21:
bors try

hrudaya21:
bors try

hrudaya21:
bors merge

PR #1407, opened 2023-06-12
fix(fs/shutdown): pause nexus when faulted

tiagolobocastro:
When the last healthy child of a nexus is faulted, errors are propagated up the stack causing the filesystem to shutdown itself down.

Instead I propose that we simply leave the nexus in a paused state, allowing the initiator to start the re-connection cycle, giving us some “time” to recreate the nexus again when the replica is available again.

todo: we might want to add a time-limit after which we resort to failing IOs?

### DIFF
--- a/mayastor/src/bdev/nexus/nexus_bdev.rs
+++ b/mayastor/src/bdev/nexus/nexus_bdev.rs
@@ -983,9 +983,9 @@ impl<'n> Nexus<'n> {
     ) -> Result<(), Error> {
         self.child_retire_for_each_channel(Some(name.clone()))
             .await?;
-        debug!(?self, "PAUSE");
+        debug!(?self, "PAUSING");
         self.as_mut().pause().await?;
-        debug!(?self, "UNPAUSE");
+        debug!(?self, "PAUSED");
         if let Some(child) = self.lookup_child(&name) {
             let uri = child.name.clone();
             // schedule the deletion of the child eventhough etcd has not been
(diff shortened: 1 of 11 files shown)


mtzaurus (approved):

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -1030,6 +1030,15 @@ impl<'n> Nexus<'n> {
...
         }
+        // If we are faulted then rather than failing all IO back to the
+        // initiator we can instead leave the subsystem paused, and wait
+        // for the control-plane to do something about this.
+        // Meanwhile the initiator will begin it's reconnect loop and won't see
+        // a swarm of IO failures which could cause a fs to shutdown.
+        if self.status() == NexusStatus::Faulted {
+            tracing::warn!("Nexus {} Faulted: not resuming subsystem", name);
+            return Ok(());
+        }
```
dsharma-dc, line 1041: Shall we move this trace from above to here, just before resume:
`debug!(?self, "UNPAUSE");`

hrudaya21 (approved):

mayastor/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -1030,6 +1030,15 @@ impl<'n> Nexus<'n> {
                 })))
                 .await;
         }
+        // If we are faulted then rather than failing all IO back to the
+        // initiator we can instead leave the subsystem paused, and wait
+        // for the control-plane to do something about this.
+        // Meanwhile the initiator will begin it's reconnect loop and won't see
```
dsharma-dc, line 1036: Is this reconnect loop going to exponentially backoff till the nexus is paused?

dsharma-dc (approved):

mayastor/src/bdev/nexus/nexus_bdev.rs
tiagolobocastro: I adjusted the messages a bit, should be clearer now, PTAL

mayastor/src/bdev/nexus/nexus_bdev.rs
tiagolobocastro: No, it's every 10s after each connect failure, until a nexus is available again.

dsharma-dc (approved):

tiagolobocastro:
bors merge

PR #1409, opened 2023-06-13
fix(fs/shutdown): freeze nexus when faulted

tiagolobocastro:
When the last healthy child of a nexus is faulted, errors are propagated up the stack causing the filesystem to shutdown itself down.

Instead I propose that we simply leave the nexus in a paused state, allowing the initiator to start the re-connection cycle, giving us some “time” to recreate the nexus again when the replica is available again.

todo: we might want to add a time-limit after which we resort to failing IOs?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1789,15 +1789,6 @@ dependencies = [
 
 [[package]]
 name = "memoffset"
-version = "0.7.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5de893c32cde5f383baa4c04c5d6dbdd735cfd4a794b0debdb2bb1b421da5ff4"
-dependencies = [
- "autocfg",
-]
-
-[[package]]
-name = "memoffset"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


dsavitskiy (approved):

mittachaitu:
Hi @tiagolobocastro, I have general question about a scenario where FS goes into read only state due to errors returned back at the time of nexus being marked as faulted. Let's say if target accept bunch of IOs from application among them after submitting IOs to replica if there is failure on first IO, we immediately call pause in replica tear down workflow (which will not accept further IOs from client side). If inflight IOs are returned failure to initiator won't there be chance of marking FS into Read-Only?

tiagolobocastro:
> Hi @tiagolobocastro, I have general question about a scenario where FS goes into read only state due to errors returned back at the time of nexus being marked as faulted. Let's say if target accept bunch of IOs from application among them after submitting IOs to replica if there is failure on first IO, we immediately call pause in replica tear down workflow (which will not accept further IOs from client side). If inflight IOs are returned failure to initiator won't there be chance of marking FS into Read-Only?

hmm in this case we're leaving the nexus as paused until it is recreated, so we'll not be succeeded nor failing any IO back to the initiator.

mtzaurus (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1411, opened 2023-06-14
fix(snapshot): listsnapshot show zero result when replica is deleted

hrudaya21:
1. ListSnapshot response no result, in following scenario

- Replica has a single snapshot
- replica is deleted

ListSnapshot will show 0 result as part of all snapshot query

### DIFF
--- a/io-engine/src/lvs/lvs_lvol.rs
+++ b/io-engine/src/lvs/lvs_lvol.rs
@@ -627,10 +627,7 @@ impl Lvol {
             .filter(|b| b.driver() == "lvol")
             .map(|b| Lvol::try_from(b).unwrap())
             .collect::<Vec<Lvol>>();
-        // return empty snapshot parameter list, if no snapshot found.
-        if lvol_devices.len() <= 1 {
-            return snapshot_list;
-        }
+
         for snapshot_lvol in lvol_devices {
             // skip lvol if it is not snapshot.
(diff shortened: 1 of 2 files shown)


io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -1229,6 +1226,7 @@ impl SnapshotOps for Lvol {
                 .map(|b| Lvol::try_from(b).unwrap())
                 .collect::<Vec<Lvol>>();
             // return empty snapshot parameter list, if no snapshot found.
+            // 1 count is refer to the source lvol of snapshot
```
tiagolobocastro: tempted to suggest this should be removed, since this is just to avoid a single for loop entry AFAICT and may induce errors, case in point

tiagolobocastro (approved):

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: Yes.. that correct. by removing this check also We can achieve the same result. Will fix it.

hrudaya21:
bors try

tiagolobocastro:
I've pruned the network..
bors try

mtzaurus (approved):

hrudaya21:
bors merge

PR #1412, opened 2023-06-15
fix(snapshot): properly display snapshot allocated space

mtzaurus:
Currently snapshot descriptor provides total replica size rather than amount of storage allocated for snapshot data. This fix uses Lvol usage statistics to return allocated_bytes for snapshots to denote storage space used by snapshots.

### DIFF
--- a/io-engine/src/grpc/v1/snapshot.rs
+++ b/io-engine/src/grpc/v1/snapshot.rs
@@ -135,7 +135,7 @@ impl From<ReplicaSnapshotDescriptor> for SnapshotInfo {
         Self {
             snapshot_uuid: snap_lvol.uuid(),
             snapshot_name: snap_lvol.name(),
-            snapshot_size: snap_lvol.size(),
+            snapshot_size: snap_lvol.usage().allocated_bytes,
             num_clones: 0, //TODO: Need to implement along with clone
             timestamp: snapshot_param
                 .create_time()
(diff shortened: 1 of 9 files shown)


mtzaurus:
bors try

mtzaurus:
bors try

io-engine/src/grpc/v1/snapshot.rs
```diff
@@ -157,14 +157,14 @@ impl From<VolumeSnapshotDescriptor> for SnapshotInfo {
...
-            snapshot_size: s.snapshot_lvol().size(),
+            snapshot_size: s.snapshot_lvol().usage().allocated_bytes,
             num_clones: s.num_clones(),
             timestamp: s
                 .snapshot_params()
                 .create_time()
                 .map(|s| s.parse::<DateTime<Utc>>().unwrap_or_default().into()),
             source_uuid: s.source_uuid(),
-            source_size: s.source_size(),
+            source_size: s.snapshot_lvol().size(),
```
hrudaya21, line 167: `s.snapshot_lvol().size()` will return what size ? How it is different from replica lvol size ?
If it is not same, Do We need to add another parameter in gRPC to send back replica lvol size after snapshot is taken ?

mtzaurus: Snapshot and replica have the same size (not allocated space, but 'size', i.e. maximum allocated space).

hrudaya21 (approved):

dsavitskiy (approved):

tiagolobocastro (approved):

mtzaurus:
bors merge

PR #1415, opened 2023-06-19
fix(nexus/io): set nvmf subsystem as frozen

tiagolobocastro:
Multiple children may be retired at "the same time". In case the first retires "wins" and pauses the subsystem, if we simply leave it paused at the end, the other retires will remain stuck in the pause.
Instead, introduce a new state of Frozen, which allows other pauses to execute their paused code and at the end the subsystem remains frozen.

### DIFF
--- a/deploy/csi-daemonset.yaml
+++ b/deploy/csi-daemonset.yaml
@@ -30,7 +30,7 @@ spec:
       # the same.
       containers:
       - name: mayastor-csi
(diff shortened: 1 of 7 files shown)


mtzaurus (approved):

dsavitskiy (approved):

tiagolobocastro:
bors merge

PR #1417, opened 2023-06-20
fix(snapshot): return proper errorcode for createsnapshot

hrudaya21:
1. When snapshot_uuid is duplicate, return proper error code in gRPC response. => EEXIST => Status::already_exists
2. When Snapshot config parameters are not passed with proper value, return proper gRPC error response. ENOMEDIUM => Status => failed_precondition

### DIFF
--- a/io-engine/src/core/bdev.rs
+++ b/io-engine/src/core/bdev.rs
@@ -137,7 +137,7 @@ where
             Some(bdev) => {
                 let b: Vec<Self> = bdev
                     .into_iter()
-                    .filter(|b| b.uuid_as_string() == uuid)
+                    .filter(|b| b.uuid_as_string() == uuid.to_lowercase())
                     .collect();
 
                 b.first().map(|b| Self {
--- a/io-engine/src/grpc/v1/snapshot.rs
+++ b/io-engine/src/grpc/v1/snapshot.rs
@@ -375,7 +375,16 @@ impl SnapshotRpc for SnapshotService {
                             })
                         }
                     };
-                    // prepare snap config and flush IO before taking snapshot.
+                    if UntypedBdev::lookup_by_uuid_str(&args.snapshot_uuid).is_some() {
+                        return Err(LvsError::Invalid {
+                            source: Errno::EEXIST,
+                            msg: format!(
+                                "Snapshot {} already exist in the system",
+                                args.snapshot_uuid
+                            ),
+                        })
+                    }
+                    // prepare snap config before taking snapshot.
                     let snap_config =
                         match lvol.prepare_snap_config(
                             &args.snapshot_name,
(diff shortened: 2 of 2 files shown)


io-engine/src/core/bdev.rs
```diff
@@ -137,7 +138,7 @@ where
             Some(bdev) => {
                 let b: Vec<Self> = bdev
                     .into_iter()
-                    .filter(|b| b.uuid_as_string() == uuid)
+                    .filter(|b| b.uuid_as_string() == uuid.to_lowercase())
```
tiagolobocastro, line 140: should b.uuid_as_string() also be "lowercased" ?

io-engine/src/grpc/v1/snapshot.rs
```diff
@@ -375,7 +375,16 @@ impl SnapshotRpc for SnapshotService {
                             })
                         }
                     };
-                    // prepare snap config and flush IO before taking snapshot.
+                    if let Some(_b) = UntypedBdev::lookup_by_uuid_str(&args.snapshot_uuid) {
```
tiagolobocastro:
```suggestion
                    if UntypedBdev::lookup_by_uuid_str(&args.snapshot_uuid).is_some() {
```

io-engine/src/grpc/v1/snapshot.rs
```diff
@@ -384,9 +393,14 @@ impl SnapshotRpc for SnapshotService {
                             &args.snapshot_uuid
                         ) {
                             Some(snap_config) => snap_config,
-                            None => return Err(LvsError::SnapshotConfigFailed {
-                                name: args.replica_uuid,
-                                msg: "tx id / snapshot name not provided".to_string(),
+                            // if any of the prepare parameters not passed,
+                            // return failure as pre-condition failed.
```
tiagolobocastro: Should this be invalid argument rather than precondition failed?

io-engine/src/core/bdev.rs
hrudaya21: Based on testing what I have observed, bdev information stored in SPDK blob is always in lower case.

io-engine/src/grpc/v1/snapshot.rs
hrudaya21: > 

Accepted. Will update it. Learning Rust best practice :)
Thank you..

io-engine/src/grpc/v1/snapshot.rs
hrudaya21: Indeed. It will be more appropriate. Will change it.

tiagolobocastro (approved):

mtzaurus (approved):

hrudaya21:
bors try

hrudaya21:
bors try

hrudaya21:
bors try

hrudaya21:
bors merge

PR #1423, opened 2023-06-25
refactor(list-rebuild): add since timestamp and name

tiagolobocastro:
Allows using timestamp to avoid re-fetching all records everytime. Adds nexus name which can be useful, eg: to filter for volume uuid, which gets set as the nexus name by the control-plane.

### DIFF
--- a/io-engine/src/grpc/v1/nexus.rs
+++ b/io-engine/src/grpc/v1/nexus.rs
@@ -1131,13 +1131,16 @@ impl NexusRpc for NexusService {
         self.serialized(ctx, args.uuid.clone(), false, async move {
             trace!("{:?}", args);
             let rx = rpc_submit::<_, _, nexus::Error>(async move {
-                let records = nexus_lookup(&args.uuid)?
+                let nexus = nexus_lookup(&args.uuid)?;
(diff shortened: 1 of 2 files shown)


hrudaya21 (approved):

io-engine/src/grpc/v1/nexus.rs
```diff
@@ -1152,33 +1155,42 @@ impl NexusRpc for NexusService {
...
                 .map(|nexus| {
-                    let uuid = nexus.uuid().to_string();
-                    let hist = nexus.rebuild_history();
+                    let hist = nexus.rebuild_history_guard();
                     let hist = hist
-                        .into_iter()
+                        .iter()
+                        .filter(|record| {
+                            end_time
+                                .map(|t| record.end_time > t)
```
hrudaya21: Do We need to have `>=` check ? When the query is usually from the user, it is usually considered to include the time input by the user. Though the chances of such instance is very rare :)

chriswldenyer (approved):
Looks reasonable

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1424, opened 2023-06-26
fix(nexus): schedule nexus snapshot operation and completion handler …

mtzaurus:
…on master core

Nexus replica snapshot operation (as part of nexus snapshot operation) and NVMe admin completion handler for snapshot command are now both scheduled on master reactor.

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev_snapshot.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev_snapshot.rs
@@ -5,10 +5,11 @@ use futures::future::join_all;
 
 use super::{Error, Nexus, NexusOperation, NexusState};
 use crate::{
-    bdev::nexus::NexusChild,
+    bdev::nexus::{nexus_lookup, NexusChild},
     core::{
         snapshot::SnapshotDescriptor,
-        BlockDeviceHandle,
+        CoreError,
+        Reactor,
         SnapshotParams,
         ToErrno,
     },
@@ -41,6 +42,7 @@ pub struct NexusSnapshotStatus {
 /// Driver for performing snapshot creation on multiple nexus replicas in
 /// parallel.
 struct ReplicaSnapshotExecutor {
+    nexus_name: String,
     /// Replica UUID -> Snapshot UUID map.
     replica_ctx: Vec<SnapshotExecutorReplicaCtx>,
     skipped_replicas: Vec<String>,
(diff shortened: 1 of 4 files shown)


mtzaurus:
bors try

mtzaurus:
bors try

io-engine/src/bdev/nexus/nexus_bdev_snapshot.rs
```diff
@@ -191,19 +180,65 @@ impl ReplicaSnapshotExecutor {
...
+
                 let replica_uuid = ctx.replica_uuid.clone();
-                debug!(
-                    replica_uuid,
-                    ?snapshot_params,
-                    "Starting nexus replica snapshot operation",
-                );
+                let nexus_name = self.nexus_name.clone();
+
+                // Schedule replica snapshot operation on master core.
```
tiagolobocastro, line 187: Aren't we already on the master core here?
We seem to already be doing things like pausing the subsystem which has an assert for the master core.

io-engine/src/bdev/nexus/nexus_bdev_snapshot.rs
```diff
@@ -191,19 +180,65 @@ impl ReplicaSnapshotExecutor {
...
 
-                // Snapshot operation future shall be able to track back to the
-                // replica.
+                // Snapshot operation future shall be able to track back to the replica.
                 async move {
                     (
-                        replica_uuid,
-                        ctx.handle.create_snapshot(snapshot_params).await,
+                        ctx.replica_uuid.clone(),
+                        rx.await.expect("Snapshot sender disappeared"),
```
tiagolobocastro, line 241: this something that we might need to handle, if futures are dropped (eg: during shutdown), I found something similar during keep alive testing on device_monitor which also uses spawn_at_primary.

test/python/v1/nexus/docker-compose.yml
```diff
@@ -11,7 +11,7 @@ services:
         - MY_POD_IP=10.0.0.2
         - NEXUS_NVMF_ANA_ENABLE=1
         - NEXUS_NVMF_RESV_ENABLE=1
-    command: ${SRCDIR}/target/debug/io-engine -g 0.0.0.0 -l 1 -r /tmp/ms0.sock
+    command: ${SRCDIR}/target/debug/io-engine -g 0.0.0.0 -l 1,2 -r /tmp/ms0.sock
```
tiagolobocastro, line 14: you might want to use 5 or increment the other ones, otherwise this core 2 will be shared with ms1

tiagolobocastro (approved):

hrudaya21 (approved):

mtzaurus:
bors merge

PR #1430, opened 2023-06-27
fix(lvol): properly report provisioning mod for volume with snapshots

mtzaurus:
Currently a thick-provisioned volume is still reported as thick-provisioned after taking a snapshot, even if the volume has been turned into a thin-provisioned one. Now lvol code uses spdk_blob_is_thin_provisioned() API to dynamically query volume provisioning mode instead of relying on a static volume flag which is set only once upon volume creation.


mtzaurus:
bors try

tiagolobocastro (approved):

hrudaya21 (approved):

mtzaurus:
bors merge

PR #1449, opened 2023-07-10
fix(core): fixing passing read mode flags to block devices

dsavitskiy:


dsavitskiy:
bors try

io-engine/src/bdev/nvmx/handle.rs
```diff
@@ -576,6 +577,11 @@ impl BlockDeviceHandle for NvmeDeviceHandle {
             });
         }
 
+        let flags = mode.map_or(self.prchk_flags, |m| match m {
+            ReadMode::Normal => 0,
+            ReadMode::UnwrittenFail => SPDK_NVME_IO_FLAGS_UNWRITTEN_READ_FAIL,
+        });
```
tiagolobocastro: Now prchk_flags are not passed at all now, is this ok? Should prchk_flags be merged with mode?

dsavitskiy: Fixed

dsavitskiy:
In SPDK, a flag that is used to determine unallocated blocks for rebuild of thin-provisioned volumes, was passed to Blobstore internals incorrectly. This only happened for recursive code path, which works for snapshots only. It didn’t happen for thin provisioned vol code path, as it does not go recursively into underlying blobstore (it terminates in special zero-filled blob store instead).

Because of that, an invalid value could sometimes be read from that flag, resulting in read I/O failure with false DeallocatedOrUnwrittenBlock error. Potentially, as a pointer to an incompatible struct was dereferenced, io-engine could crash.

In io-engine, the I/O flag used to determine unallocated block was saved in device handle structures. This wasn't a good approach, as the flag had to be (re-)set before every I/O operation and potentially could lead to unexpected I/O failures, if a handle was shared. This was changed to passing the flag directly to read_at() method.

tiagolobocastro (approved):

hrudaya21 (approved):

dsavitskiy:
bors merge

PR #1452, opened 2023-07-11
fix(snapshot): allocated bytes for snapshot updated for thin provisioned replica

hrudaya21:


tiagolobocastro (commented):
I don't think this is correct, from my tests snapshot allocated_size is correct, rather it's the replica allocate size that is incorrect.
Let's add a few tests so we ensure these are correct and we're all on the same page.
EDIT: I think the correct fix is to clear this cache: `blob->num_used_clusters_cache`.

io-engine/src/bin/io-engine-client/v1/replica_cli.rs
```diff
@@ -300,8 +300,15 @@ async fn replica_list(
                     let size = ctx.units(Byte::from_bytes(r.size.into()));
                     let capacity = ctx
                         .units(Byte::from_bytes(usage.capacity_bytes.into()));
-                    let allocated = ctx
-                        .units(Byte::from_bytes(usage.allocated_bytes.into()));
+                    let allocated = if r.is_snapshot {
```
mtzaurus: This means that we display size of all predecessor snapshots and don't take into account how much space the snapshot itself occupies ?

hrudaya21: > This means that we display size of all predecessor snapshots and don't take into account how much space the snapshot itself occupies ?

This logic to be reverted back. By Clearing blob->num_used_clusters_cache for thin provisioned replica, the snapshot and replica allocated size is coming proper.

hrudaya21:
bors try

hrudaya21:
bors try

hrudaya21:
bors try

hrudaya21:
bors try

io-engine/tests/snapshot_lvol.rs
```diff
@@ -911,3 +911,178 @@ async fn test_snapshot_volume_provisioning_mode() {
...
+            lvol.usage().allocated_bytes,
+            0,
+            "Volume still has some space allocated after taking a snapshot"
+        );
+    })
+    .await;
+}
+
+#[tokio::test]
+async fn test_thik_provision_lvol_alloc_after_snapshot() {
```
mtzaurus, line 1006: Looks like 'test_thik_provision_lvol_alloc_after_snapshot()' and 'test_thin_provision_lvol_alloc_after_snapshot()' are fully identical (the only difference is flag that denotes thin/thick provisioned test volume) ?
If so, can we have just one function that performs test and just call it from these 2 testcases ? Simethong like this:

```
async fn test_lvol_alloc_after_snapshot(thin: bool) {
   // Do all test logic here.
}

#[tokio::test]
async fn test_thin_provision_lvol_alloc_after_snapshot() {
   test_lvol_alloc_after_snapshot(true).await;
}

#[tokio::test]
async fn test_thick_provision_lvol_alloc_after_snapshot() {
   test_lvol_alloc_after_snapshot(false).await;
}
```

hrudaya21: Fixed the changes.

tiagolobocastro (approved):

hrudaya21:
bors try

mtzaurus (approved):

hrudaya21:
bors try

hrudaya21:
bors try

hrudaya21:
bors try

hrudaya21:
bors merge

hrudaya21:
bors merge

PR #1459, opened 2023-07-19
feat(rebuild): adding rebuild I/O verification

dsavitskiy:


dsavitskiy:
bors try

dsavitskiy:
bors try

dsavitskiy:
bors try

Jenkinsfile
```diff
@@ -140,6 +140,7 @@ pipeline {
           anyOf {
             branch 'master'
             branch 'release/*'
+            branch 'trying'
```
tiagolobocastro: Did you mean to make this change?

io-engine/src/core/block_device.rs
```diff
@@ -170,30 +158,184 @@ pub trait BlockDeviceHandle {
...
+        offset_blocks: u64,
+        num_blocks: u64,
+    ) -> Result<bool, CoreError> {
+        let mut dst_buf = self.dma_malloc(src_buf.len()).map_err(|_| {
+            CoreError::DmaAllocationFailed {
+                size: src_buf.len(),
+            }
+        })?;
+
+        self.read_buf_blocks_async(
```
tiagolobocastro: Can't we use compare command? This way we don't need to receive any data and the target will compare it for us.

io-engine/src/rebuild/rebuild_task.rs
```diff
@@ -110,59 +111,137 @@ impl RebuildTask {
...
+        let num_blocks = desc.get_segment_size_blks(offset_blk);
 
-        let res = source_hdl
-            .read_at_ex(
-                blk * descriptor.block_size,
-                copy_buffer,
-                Some(ReadMode::UnwrittenFail),
-            )
-            .await;
+        let mut buf = src
```
tiagolobocastro: Why are we now allocating the buffer here? We have to allocate/deallocate every time, whilst before we ensured we had the buffers for the duration of the rebuild.

dsavitskiy:
bors try

io-engine/src/rebuild/rebuild_task.rs
dsavitskiy: Fixed and improved to prevent buffer allocation for the final segment.

dsavitskiy:
bors try

Jenkinsfile
dsavitskiy: Yes I did. Should I do a separate PR for it?

io-engine/src/core/block_device.rs
dsavitskiy: Yes we can. I try to use it.

dsavitskiy:
bors try

io-engine/src/core/block_device.rs
dsavitskiy: Done. I had to made a fix to SPDK to make compare APIs work properly: https://github.com/openebs/spdk/pull/35.

io-engine/src/rebuild/rebuild_task.rs
```diff
@@ -110,58 +123,21 @@ impl RebuildTask {
...
         &mut self,
-        blk: u64,
-        descriptor: &RebuildDescriptor,
+        offset_blk: u64,
+        desc: &RebuildDescriptor,
     ) -> Result<(), RebuildError> {
-        let mut copy_buffer: DmaBuf;
-        let source_hdl = descriptor.src_io_handle().await?;
-        let destination_hdl = descriptor.dst_io_handle().await?;
+        let buf_lock = self.buffer.lock().await;
```
tiagolobocastro: Why does the buffer need a lock? we have a mutable ref here so we can modify it with extra indirection?

io-engine/src/rebuild/rebuild_descriptor.rs
```diff
@@ -85,9 +100,168 @@ impl RebuildDescriptor {
...
+    /// size for the given offset. Given `DmaBuf` must be large enough.
+    #[inline(always)]
+    pub(super) fn adjusted_iov(
+        &self,
+        buffer: &DmaBuf,
+        offset_blk: u64,
+    ) -> IoVec {
+        let mut iov = buffer.to_io_vec();
+
+        let iov_len = self.get_segment_size_blks(offset_blk) * self.block_size;
```
tiagolobocastro, line 120: Do we need to do this if we now specify the num_blocks ? on read/writev_blocks_async?

Jenkinsfile
tiagolobocastro: Ah ok.
Should we skip just the submodule check rather than the entire lint? Reason for asking is that having the lint in place helps make sure code is correctly linted which makes core review easier as they won't be able to test changes on CI if code is not linted properly.

io-engine/src/core/block_device.rs
tiagolobocastro: Yes I saw, approved that one already.

io-engine/src/grpc/v1/test.rs
```diff
@@ -0,0 +1,172 @@
...
+    }
+}
+
+impl Default for TestService {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[cfg(feature = "fault-injection")]
```
tiagolobocastro: hmm this is not going to work very well if we want other things in testing in addition to fault injection, example: replica wiping and perhaps others in the future?

io-engine/src/rebuild/rebuild_descriptor.rs
dsavitskiy: It turned out that yes we have. I tried not doing that at first, and I/O failed in a strange way. I didn't look further but it looks like SPDK's APIs require num_blocks to be equal to iovecs size.

io-engine/src/rebuild/rebuild_task.rs
dsavitskiy: Ah, indeed we don't need a lock here. I didn't notice that each task is already locked when called from RebuildTasks. Will fix.

Jenkinsfile
dsavitskiy: I think, that's would be a better solution.

dsavitskiy:
bors try

io-engine/src/grpc/v1/test.rs
dsavitskiy: This block is for injection only. Any other kinds of APIs can be added outside this block, with their own feature config if they need to be enabled selectively.
There's another feature "io-engine-testing", which included "fault-injection". Another test sub-features can be added there.

tiagolobocastro: But traits have to be implemented as a whole right?

dsavitskiy: ah, yes indeed. I'll change to per-method cfg condition.

tiagolobocastro: You could add indirection by implementing injection as separate trait or just fn impls, and then in the TestRpc trait impl call into this injection fn's, example:
```rust
#[cfg(feature = injection)]
impl Injection { fn a { } fn b() { } } 
#[cfg(feature( no( injection))]
impl Injection { fn a { Not_impl } fn b() { Not_impl } }

impl TestRcp for TestService {
  fn a() { Injection::a() }
  fn b() { Injection::b() }
}
```

io-engine/src/rebuild/rebuild_descriptor.rs
tiagolobocastro: Ah I see, maybe that's the reason it was this way to begin with :(

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -151,6 +153,31 @@ impl<'n> Nexus<'n> {
         src_child_uri: &str,
         dst_child_uri: &str,
     ) -> Result<(), Error> {
+        let verify_mode = match std::env::var("NEXUS_REBUILD_VERIFY")
```
tiagolobocastro, line 156: Could this be a CLI argument?
Ex: sudo io-engine --rebuild-verify-ios

tiagolobocastro (approved):

io-engine/src/rebuild/rebuild_descriptor.rs
```diff
@@ -45,20 +54,23 @@ impl RebuildDescriptor {
     }
 
     /// Get a `BlockDeviceHandle` for the source.
+    #[inline(always)]
```
dsharma-dc, line 59: Is it a good idea to annotate `async fn` as `inline always` in Rust? or better leave it to compiler's wisdom.

io-engine-tests/src/test.rs
dsharma-dc, line 1: It'd be good to use a uniformly used `fault_injection` everywhere instead of simply `injection`. e.g `pub async fn add_fault_injection`

io-engine/src/rebuild/rebuild_descriptor.rs
```diff
@@ -183,4 +185,83 @@ impl RebuildDescriptor {
...
     }
+
+    /// Verify segment copy operation by reading destination, and comparing with
+    /// the source.
+    pub(super) async fn verify_segment(
+        &self,
+        offset_blk: u64,
+        iovs: &mut [IoVec],
+    ) -> Result<(), RebuildError> {
+        // Read the source again.
```
dsharma-dc, line 196: Why do we need to read source again? The LBA ranges are still locked I believe when `verify_segment` is called and the buffers that we used to write to destination can be still valid.

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
dsavitskiy: We would end up with too many CLI args. Also, we were planned to make this a parameter of rebuild start / child online API.

io-engine/src/rebuild/rebuild_descriptor.rs
dsavitskiy: There can be source read errors.

io-engine-tests/src/test.rs
dsavitskiy: Fixed

io-engine/src/rebuild/rebuild_descriptor.rs
dsavitskiy: I guess, it's the same as for C++ (Rust and C++ share the same LLVM backend): it may probably help to force the compiler to inline.

dsharma-dc: For C/C++ too unless we know function is getting called thousands of times per second and is self contained like a macro, we'd usually leave it to compiler as compilers today will do inlining even if we don't ask. Benchmarking would be the real measure of such impacts on instruction ordering and branching. I'd think it's superfluous here for these async functions.

io-engine/src/rebuild/rebuild_descriptor.rs
dsharma-dc: hmm, not sure if that should matter here. IIUC, the intent is to read from destination, and compare to the buffer from which we wrote.

dsharma-dc (approved):

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
tiagolobocastro: We can have a separate struct for Testing cli arguments, then it's clear any "user" would ignore it but it will make it visible for testing what's available.

io-engine/src/rebuild/rebuild_descriptor.rs
```diff
@@ -85,9 +100,168 @@ impl RebuildDescriptor {
...
+                ))
+            ) =>
+            {
+                self.verify_failure(offset_blk)
+            }
+            Err(err) => Err(RebuildError::VerifyIoFailed {
+                source: err,
+                bdev: self.dst_uri.clone(),
+            }),
+        }
```
tiagolobocastro, line 237: hmm I think this might yield false miscompares for thin volumes.
Let's say we have a thin volume and we're rebuilding a new replica.
1. want to rebuild 64k bytes which targets an unallocated cluster.
2. 64k range is locked
3. special read tells us we can skip the write
4. we want to now verify 
5. a user IO which targets a different range in the cluster means the cluster now is allocated
6. read from source
7. verify target might fail because allocating the cluster in 5 means now we have undefined data here

tiagolobocastro: All, please ignore this, we don't verify if read was skipped so we never end up hitting this scenario!

dsavitskiy: We don't verify if we don't write.

dsavitskiy:
bors merge

dsavitskiy:
bors merge

dsavitskiy:
bors merge

dsavitskiy:
bors merge

dsavitskiy:
bors merge

PR #1468, opened 2023-07-24
fix(clone): replace spdk default clone api with custom api to identify clone for a lvol

hrudaya21:


io-engine/src/core/snapshot.rs
```diff
@@ -320,6 +320,9 @@ pub trait SnapshotOps {
         &self,
         parent: Option<&Lvol>,
     ) -> Option<VolumeSnapshotDescriptor>;
+
+    /// Returns a boolean indicating if the lvol is a clone from snapshot.
+    fn is_snapshot_clone(&self) -> bool;
```
mtzaurus: Shall we also somehow provide the snapshot this clone was created from ? I presume for user it's also important to be able to answer the question "what's the parent of this clone ?".
Also, this clone-related function is part of 'pub trait SnapshotOps',  which is snapshot-related.
Isn't it better to have 'is_snapshot_clone()' be part of Lvol or so ?

hrudaya21: 1. Currently as part of list_clone gRPC, We are returning snapshot uuid from which clone is created. Do We need put it in any other gRPC response ?
2. For 2nd point, I will do the change. I was also little confuse if to put it in snapshot category or lvol. You are correct, as this will work for replica also, I will keep part of lvol trait.

hrudaya21: > 1. Currently as part of list_clone gRPC, We are returning snapshot uuid from which clone is created. Do We need put it in any other gRPC response ?
> 2. For 2nd point, I will do the change. I was also little confuse if to put it in snapshot category or lvol. You are correct, as this will work for replica also, I will keep part of lvol trait.

Got it for point: 1, I will optimize the is_snapshot_clone() API to optimize the query used in list_clone gRPC.

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -456,8 +456,9 @@ pub trait LvsLvol: LogicalVolume + Share {
     /// Returns a boolean indicating if the lvol is a snapshot.
     fn is_snapshot(&self) -> bool;
 
-    /// Returns a boolean indicating if the lvol is a clone.
-    fn is_clone(&self) -> bool;
+    /// Lvol is considered as clone if its sourceuuid attribute is a valid
+    /// snapshot. if it is clone, return the snapshot lvol.
+    fn is_snapshot_clone(&self) -> (bool, Option<Lvol>);
```
tiagolobocastro: Why not simply return Option<Lvol> ?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -601,10 +602,26 @@ impl LvsLvol for Lvol {
...
+        if let Some(source_uuid) =
+            Lvol::get_blob_xattr(self, CloneXattrs::SourceUuid.name())
+        {
+            let snap_lvol =
+                match UntypedBdev::lookup_by_uuid_str(source_uuid.as_str()) {
+                    Some(bdev) => match Lvol::try_from(bdev) {
+                        Ok(l) => l,
+                        _ => return (false, None),
+                    },
+                    None => return (false, None),
```
tiagolobocastro: Doesn't seem entirely correct here, self is indeed a clone, but the Snapshot Lvol is missing?

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: Yes, it can be done with single Option parameter as return type. Changed the code.

io-engine/src/lvs/lvs_lvol.rs
hrudaya21: let snap_lvol =
                match UntypedBdev::lookup_by_uuid_str(source_uuid.as_str()) {
                    Some(bdev) => match Lvol::try_from(bdev) {
                        Ok(l) => l,
                        **_ => return None,**
                    },
                    **None => return None,**
                };
Above 2 `return None` condition is to handle the failure scenario of `lookup_by_uuid_str` and `try_from` function. Am I missing any point here ?

tiagolobocastro (approved):

mtzaurus (approved):

hrudaya21:
bors try

hrudaya21:
bors merge

hrudaya21:
bors merge

PR #1469, opened 2023-07-24
chore(nexus/channel): reconfigure only device which is involved in DR…

dsharma-dc:
… event

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev.rs
@@ -602,29 +602,30 @@ impl<'n> Nexus<'n> {
 
     /// Reconfigures the child event handler.
     pub(crate) async fn reconfigure(&self, event: DrEvent) {
+        let evt = event.to_string();
         info!(
-            "{self:?}: dynamic reconfiguration event: {event}, \
+            "{self:?}: dynamic reconfiguration event: {evt}, \
             reconfiguring I/O channels...",
         );
 
         let (sender, recv) = oneshot::channel::<ChannelTraverseStatus>();
 
         self.traverse_io_channels(
-            sender,
-            |chan, _sender| -> ChannelTraverseStatus {
-                chan.reconnect_all();
+            (sender, event),
+            |chan, (_sender, dev_ctx)| -> ChannelTraverseStatus {
+                chan.reconfigure_channel(dev_ctx);
                 ChannelTraverseStatus::Ok
             },
-            |status, sender| {
+            |status, ctx| {
                 debug!("{self:?}: all I/O channels reconfigured");
-                sender.send(status).expect("reconfigure channel gone");
+                ctx.0.send(status).expect("reconfigure channel gone");
             },
         );
 
         let result = recv.await.expect("reconfigure sender already dropped");
 
         info!(
-            "{self:?}: dynamic reconfiguration event: {event}, \
+            "{self:?}: dynamic reconfiguration event: {evt}, \
             reconfiguring I/O channels completed with result: {result:?}",
         );
     }
--- a/io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
@@ -126,7 +126,11 @@ impl<'n> Nexus<'n> {
         // rebuilt would then need to be rebuilt again.
         // Ensuring that the dst child receives all frontend Write IO keeps all
         // rebuilt ranges in sync with the other children.
-        self.reconfigure(DrEvent::ChildRebuild).await;
+        self.reconfigure(DrEvent::ChildRebuildBegin(
+            // TODO: Free the rebuild job and backend perhaps?
+            self.get_child_device_name(dst_child_uri.as_str())?,
+        ))
+        .await;
 
         // Stop the I/O log and create a rebuild map from it.
         // As this is done after the reconfiguraion, any new write I/Os will
@@ -462,7 +466,11 @@ impl<'n> Nexus<'n> {
             }
         }
 
-        self.reconfigure(DrEvent::ChildRebuild).await;
+        self.reconfigure(DrEvent::ChildRebuildEnd(
+            // TODO: Free the rebuild job and backend perhaps?
+            self.get_child_device_name(child_uri)?,
+        ))
+        .await;
 
         Ok(())
     }
--- a/io-engine/src/bdev/nexus/nexus_channel.rs
+++ b/io-engine/src/bdev/nexus/nexus_channel.rs
@@ -49,14 +49,16 @@ pub enum IoMode {
     Freeze,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Eq, PartialEq)]
 #[allow(clippy::enum_variant_names)]
-/// Dynamic Reconfiguration Events occur when a child is added or removed
+/// Dynamic Reconfiguration Events occur when a child is added or removed.
 pub enum DrEvent {
-    /// Child unplug reconfiguration event.
-    ChildUnplug,
-    /// Child rebuild event.
-    ChildRebuild,
+    /// Child unplug reconfiguration event containing device name String.
+    ChildUnplug(String),
+    /// Child rebuild begin event containing device name String.
+    ChildRebuildBegin(String),
+    /// Child rebuild finish event containing device name String.
+    ChildRebuildEnd(String),
 }
 
 impl Display for DrEvent {
@@ -65,8 +67,9 @@ impl Display for DrEvent {
             f,
             "{}",
             match self {
-                Self::ChildUnplug => "unplug",
-                Self::ChildRebuild => "rebuild",
+                Self::ChildUnplug(_) => "unplug",
+                Self::ChildRebuildBegin(_) => "rebuild begin",
+                Self::ChildRebuildEnd(_) => "rebuild end",
             }
         )
     }
@@ -141,6 +144,20 @@ impl<'n> NexusChannel<'n> {
         self.readers.len()
     }
 
+    /// Check if the child with given device name is a writer in this channel.
+    pub fn is_writer(&self, dev_name: &str) -> bool {
+        self.writers
+            .iter()
+            .any(|h| h.get_device().device_name() == dev_name)
+    }
+
+    /// Check if the child with given device name is a reader in this channel.
+    pub fn is_reader(&self, dev_name: &str) -> bool {
+        self.readers
+            .iter()
+            .any(|h| h.get_device().device_name() == dev_name)
+    }
+
     /// Calls the given callback for each active writer.
     #[inline(always)]
     pub(super) fn for_each_writer<F>(&self, mut f: F) -> Result<(), CoreError>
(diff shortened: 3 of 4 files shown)


dsharma-dc:
bors try

dsharma-dc:
bors try

dsharma-dc:
Opening for review. The extended test run of "change volumes and replicas" which disrupts channels quite a bit has been running on this change for more than 12hours.

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -126,7 +126,10 @@ impl<'n> Nexus<'n> {
         // rebuilt would then need to be rebuilt again.
         // Ensuring that the dst child receives all frontend Write IO keeps all
         // rebuilt ranges in sync with the other children.
-        self.reconfigure(DrEvent::ChildRebuild).await;
+        self.reconfigure(DrEvent::ChildRebuildBegin(
+            self.get_child_device_name(dst_child_uri.as_str())?,
```
tiagolobocastro, line 131: In case of an error does it get cleaned up afterwards?

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -49,14 +49,16 @@ pub enum IoMode {
...
-#[derive(Debug)]
+#[derive(Debug, Eq, PartialEq)]
 #[allow(clippy::enum_variant_names)]
 /// Dynamic Reconfiguration Events occur when a child is added or removed
 pub enum DrEvent {
     /// Child unplug reconfiguration event.
-    ChildUnplug,
-    /// Child rebuild event.
-    ChildRebuild,
+    ChildUnplug(String),
```
tiagolobocastro, line 57: What is this String here, a device name or a uri?
Seems its the device name, any reason for using that instead of the uri?

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -193,67 +196,69 @@ impl<'n> NexusChannel<'n> {
...
+                            }
+                            Err(e) => {
+                                child.set_faulted_state(FaultReason::CantOpen);
+                                error!("{self:?}: failed to get I/O handle for {child:?}: {e}");
+                            }
+                        }
+                        // Make sure the child is not in readers. Most likely
+                        // the child won't be in readers
+                        // already here.
+                        self.readers
```
tiagolobocastro: Should this check be done out of the if let Some() ?

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -193,67 +196,69 @@ impl<'n> NexusChannel<'n> {
...
-                            "{self:?}: failed to get I/O handle \
-                                for {c:?}: {e}"
-                        );
+                }
+            }
+            DrEvent::ChildRebuildEnd(dev) => {
+                if let Some(child) =
+                    self.nexus().child_by_device_name(dev.as_str())
+                {
+                    if child.is_healthy() {
```
tiagolobocastro: Do we need to check if it's healthy here?

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -193,67 +196,69 @@ impl<'n> NexusChannel<'n> {
...
+                        // TODO: assert that the device is_writer and not
```
tiagolobocastro: Should we check that child is in the self.writers, and if not don't add it to the readers?

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -193,67 +196,69 @@ impl<'n> NexusChannel<'n> {
...
+                            }
+                            Err(e) => {
+                                child.set_faulted_state(FaultReason::CantOpen);
+                                error!("{self:?}: failed to get I/O handle for {child:?}: {e}");
+                            }
+                        }
                     }
-                });
+                }
+            }
```
tiagolobocastro, line 270: It seems that with this code now there's scope for duplicate entries in the lists, we should probably ensure that doesn't happen..

io-engine/src/bdev/nexus/nexus_child.rs
```diff
@@ -1104,6 +1104,7 @@ impl<'c> NexusChild<'c> {
 
         let state = self.state();
         let is_destroying = self.is_destroying();
+        let child_dev = self.get_device_name().unwrap_or_default();
```
tiagolobocastro: Seems we should leave it as Result/Option rather than default to empty string?

io-engine/src/bdev/nvmx/channel.rs
```diff
@@ -489,7 +489,6 @@ impl NvmeControllerIoChannel {
 
 impl Drop for NvmeControllerIoChannel {
     fn drop(&mut self) {
-        trace!("I/O channel {:p} dropped", self.0.as_ptr());
```
tiagolobocastro: Why is this removed, seems useful to have? Or is it too verbose?

io-engine/src/bdev/nvmx/qpair.rs
```diff
@@ -200,7 +200,6 @@ impl QPair {
         // multiple allocations of the same handle for the same thread, to make
         // sure we don't reconnect every time.
         if self.state() == QPairState::Connected {
-            trace!(?self, "I/O qpair already connected");
```
tiagolobocastro: Why is this removed, seems useful to have?

io-engine/src/bdev/nexus/nexus_channel.rs
dsharma-dc: Yes it's the device name. I guess no particular reason. For one, the traversal code is more closer to `device` side rather than `child`, and secondly, device name looks a bit compact/short compared to child uri.

io-engine/src/bdev/nexus/nexus_channel.rs
dsharma-dc: Yes because reconfigure can get called in case of rebuild job Completed/Failed/Stopped, all of these. This maps to the check that happens today as well here.

io-engine/src/bdev/nvmx/channel.rs
dsharma-dc: This looks like a mistake after I had modified these traces during testing. Will put back.

io-engine/src/bdev/nvmx/qpair.rs
dsharma-dc: This looks like a mistake after I had modified these traces during testing. Will put back.

io-engine/src/bdev/nexus/nexus_channel.rs
tiagolobocastro: Then shouldn't we remove the device from readers and writers in case the child is not healthy?

dsharma-dc: In case child is not healthy as a result of rebuild fail/stop, the unplug workflow will kick in anyway so we'll be back here again as `ChildUnplug`, which will do the needful.

tiagolobocastro: Isn't this different behaviour from what happens today, when rebuild fails the reconfigure removes the child from writers straight away?

io-engine/src/bdev/nexus/nexus_channel.rs
dsharma-dc: Yes, that can be done.

io-engine/src/bdev/nexus/nexus_channel.rs
dsharma-dc: we can check that but I think that should be an assertion (as TODO says), because it shouldn't be possible - that the rebuild finished and child is healthy but it's not a channel writer on this core?

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
dsharma-dc: you mean the String containing device name? hmm, It should be bound to lifetime of DrEvent I guess, so when the TraverseCtx is destroyed, the contained DrEvent(with String) should be destroyed too. Beyond `nexus::reconfigure`, there is no error path present today.

dsharma-dc: btw, today it's possible that if a rebuild fails and unplug is triggered, it'll reconfigure that channels and then later the `on_rebuild_update() `will again do the same which is redundant I believe. Now with this change, as a result of rebuild failure, the unplug will reconfigure the channels and the `on_rebuild_update() `will bail out early because `get_child_device_name() `will fail  with `ChildNotOpen` error:

> failed to process rebuild update notification with error: Child nvmf://65.21.0.104:8420/nqn.2019-05.io.openebs:233532e2-4fe0-436f-9eaa-c4652bc53f58?uuid=233532e2-4fe0-436f-9eaa-c4652bc53f58 of nexus f8079f24-6fe5-4532-ac88-fa0cbc31cbb4 is not open

io-engine/src/bdev/nexus/nexus_channel.rs
dsharma-dc: straight away in this case means via on_rebuild_update() callback from rebuild backend, but by that time unplug is much ahead and has already done the reconfigure. `on_rebuild_update`() calls `close_faulted` before reconfigure

io-engine/src/bdev/nexus/nexus_child.rs
dsharma-dc: hmm, we can. The handling would become more verbose I think. e.g. I'd need to skip calling `reconfigure()` if the child_dev is `None`? OR take an `Option` all the way from `reconfigure` and beyond.

io-engine/src/bdev/nexus/nexus_channel.rs
dsharma-dc: The check would probably be based on iterating comparing the device name between existing NvmeDeviceHandle object and newly acquired one? vector length based check may not be completely reliable.

io-engine/src/bdev/nexus/nexus_channel.rs
tiagolobocastro: It can happen because calling reconfigure is async, so things may change underneath you.

io-engine/src/bdev/nexus/nexus_channel.rs
tiagolobocastro: unplug IIRC is based on a callback from SPDK, so it may happen before or after depending on spdk and timing.
We should remove it on reconfigure and keep same behaviour.

io-engine/src/bdev/nexus/nexus_channel.rs
tiagolobocastro: I guess we could use the name from NvmeDeviceHandle?

io-engine/src/bdev/nexus/nexus_child.rs
tiagolobocastro: keep as Option and if None then skip the block_on call below.
Or should we force a full reconfigure in this case to be safe? (same as we have in current code)

io-engine/src/bdev/nexus/nexus_channel.rs
dsharma-dc: Added `is_reader() `and `is_writer()` helper functions in NexusChannel and used here.

dsharma-dc:
Another run of extended test framework's "Change Replicas and Volume" test has been running for more than 8 hours on this change.

dsharma-dc:
bors try

io-engine/src/bdev/nexus/nexus_channel.rs
tiagolobocastro: Let's document this here then, just a String could be anything.

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
tiagolobocastro: No, I mean does the rebuild job get cleaned up by upper layers if we fail here?

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
```diff
@@ -462,7 +465,10 @@ impl<'n> Nexus<'n> {
             }
         }
 
-        self.reconfigure(DrEvent::ChildRebuild).await;
+        self.reconfigure(DrEvent::ChildRebuildEnd(
+            self.get_child_device_name(child_uri)?,
```
tiagolobocastro, line 471: hmm what happens if we fail here, we don't reconfigure channels is this ok?

tiagolobocastro (approved):
lgtm though a lot more convoluted, probably should be reviewed by @mtzaurus and/or @dsavitskiy

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
dsharma-dc: Yes it should be ok because that would mean device is unplugged underneath us and reconfiguration would've happened. If the device is not found for any other reason, I think we would already have a broken system in other ways.

tiagolobocastro: Just wanted to clarify because now with this change, there's nothing to fixup the channels, if we somehow "leak" a handle in the lists, there's no more reconconfigure_all that would remove them.

io-engine/src/bdev/nexus/nexus_bdev_rebuild.rs
dsharma-dc: Looks like we need to call `child.remove_rebuild_job() `if at all this fails here. Will that drop the `RebuildJobBackEnd` as well or shall we explicitly do `self.comms.send(None)` before calling `child.remove_rebuild_job()`

io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -602,29 +602,30 @@ impl<'n> Nexus<'n> {
...
-            sender,
-            |chan, _sender| -> ChannelTraverseStatus {
-                chan.reconnect_all();
+            (sender, event),
+            |chan, (_sender, dev_ctx)| -> ChannelTraverseStatus {
+                chan.reconfigure_channel(dev_ctx);
                 ChannelTraverseStatus::Ok
             },
-            |status, sender| {
+            |status, ctx| {
```
dsavitskiy, line 619: you can destruct the tuple here as well: (sender, _).

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -193,67 +210,70 @@ impl<'n> NexusChannel<'n> {
...
-                .for_each(|c| match c.get_io_handle() {
-                    Ok(hdl) => {
-                        debug!(
-                            "{self:?}: connecting child device \
-                                in write-only mode: {c:?}"
-                        );
-                        writers.push(hdl);
+        match dev_ctx {
+            DrEvent::ChildRebuildBegin(dev) => {
+                if let Ok(child) = self.nexus().child_by_device(dev.as_str()) {
```
dsavitskiy, line 227: better to use &

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -193,67 +210,70 @@ impl<'n> NexusChannel<'n> {
...
+                    debug_assert!(!self.is_writer(dev.as_str()));
+                    if !self.readers.is_empty() && !self.is_writer(dev.as_str())
```
dsavitskiy, line 229: "!self.is_writer(dev.as_str()" is always true here, why adding it as a condition?

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -193,67 +210,70 @@ impl<'n> NexusChannel<'n> {
...
+                // the child won't be in readers already here.
+                self.readers
+                    .retain(|h| &h.get_device().device_name() != dev);
+            }
+            DrEvent::ChildRebuildEnd(dev) => {
+                if let Ok(child) = self.nexus().child_by_device(dev.as_str()) {
+                    debug_assert!(!self.is_reader(dev.as_str()));
+                    if child.is_healthy()
+                        && self.is_writer(dev.as_str())
+                        && !self.is_reader(dev.as_str())
```
dsavitskiy, line 253: Again, "!self.is_reader(dev.as_str()" is always true here.

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -141,6 +144,20 @@ impl<'n> NexusChannel<'n> {
         self.readers.len()
     }
 
+    /// Check if the child with given device name is a writer in this channel.
+    pub fn is_writer(&self, dev_name: &str) -> bool {
```
dsavitskiy, line 148: Is this needed to be "pub"? I cannot find it is used outside this module.

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -193,67 +210,70 @@ impl<'n> NexusChannel<'n> {
...
                     }
-                    Err(e) => {
-                        c.set_faulted_state(FaultReason::CantOpen);
-                        error!(
-                            "{self:?}: failed to get I/O handle \
-                                for {c:?}: {e}"
-                        );
+                }
+                // Also make sure the child is not in readers. Most likely
+                // the child won't be in readers already here.
```
dsavitskiy, line 244: If the child is most likely not to be find in reader, it should either be a warning or assert, if it is.

dsavitskiy (changes requested):

dsharma-dc:
Changes merged via #1506 to handle the situation differently.

PR #1475, opened 2023-07-26
feat(clone): special handling to delete snapshot when active clones are linked to it

hrudaya21:
1. When `DeleteSnapshot` gRPC is triggered where there are valid clones are linked to it, Snapshot will mark "future_delete" flag true. But don't delete the snapshot from SPDK.
2. When `ListSnapshot` gRPC is called, a new parameter is return which will tell the Snapshot is marked as "deleted"
3. If any new clones will be created from this Snapshot using `CreateClone` gRPC , request will be rejected from Dataplane.
4. When `DeleteReplica` gRPC is called, if the replica is identified as last clone linked to snapshot which is already marked as "deleted", delete the snapshot along with clone replica.
5. When `DeleteReplica`gPRC, if clone is deleted but before snapshot delete, there is any fault or io-engine crash, as part `pool_import` request from CP, clean the discarded snapshot which is not done in last request.

### DIFF
--- a/io-engine/src/core/snapshot.rs
+++ b/io-engine/src/core/snapshot.rs
@@ -9,6 +9,7 @@ use std::{
 };
 use strum::EnumCount;
 use strum_macros::{EnumCount as EnumCountMacro, EnumIter};
+
 /// Snapshot Captures all the Snapshot information for Lvol.
 #[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
 pub struct SnapshotParams {
@@ -18,6 +19,7 @@ pub struct SnapshotParams {
     snap_name: Option<String>,
     snapshot_uuid: Option<String>,
     create_time: Option<String>,
+    discarded_snapshot: bool,
 }
 
 /// Implement Snapshot Common Function.
@@ -29,6 +31,7 @@ impl SnapshotParams {
         snap_name: Option<String>,
         snapshot_uuid: Option<String>,
         create_time: Option<String>,
+        discarded_snapshot: bool,
     ) -> SnapshotParams {
         SnapshotParams {
             entity_id,
@@ -37,6 +40,7 @@ impl SnapshotParams {
             snap_name,
             snapshot_uuid,
             create_time,
+            discarded_snapshot,
         }
     }
 }
@@ -167,6 +171,11 @@ pub enum SnapshotXattrs {
     ParentId,
     SnapshotUuid,
     SnapshotCreateTime,
+    /// if any snapshot delete gRPC request came and there are valid clones
+    /// link to the snapshot, then snapshot can be marked as discarded,
+    /// delete of the last valid replica(clone) can delete the snapshot marked
+    /// as discarded.
+    DiscardedSnapshot,
 }
 
 impl SnapshotXattrs {
@@ -177,6 +186,7 @@ impl SnapshotXattrs {
             Self::ParentId => "io-engine.parent_id",
             Self::SnapshotUuid => "uuid",
             Self::SnapshotCreateTime => "io-engine.snapshot_create_time",
+            Self::DiscardedSnapshot => "io-engine.discarded_snapshot",
         }
     }
 }
@@ -214,6 +224,8 @@ pub trait SnapshotOps {
         nvmf_req: &NvmfReq,
         snapshot_params: SnapshotParams,
     );
+    /// Destroy snapshot.
+    async fn destroy_snapshot(mut self) -> Result<(), Self::Error>;
 
     /// List Snapshot details based on source UUID from which snapshot is
     /// created.
@@ -320,6 +332,16 @@ pub trait SnapshotOps {
         &self,
         parent: Option<&Lvol>,
     ) -> Option<VolumeSnapshotDescriptor>;
+
+    /// Return bool value to indicate, if the snapshot is marked as discarded.
+    fn is_discarded_snapshot(&self) -> bool;
+
+    /// During destroying the last linked cloned, if there is any fault
+    /// happened, it is possible that, last clone can be deleted, but linked
+    /// snapshot marked as discarded still present in the system. As part of
+    /// pool import, do the garbage collection to clean the discarded snapshots
+    /// leftout in the system.
+    async fn destroy_pending_discarded_snapshot();
 }
 
 /// Traits gives the Snapshots Related Parameters.
@@ -359,6 +381,12 @@ pub trait SnapshotDescriptor {
 
     /// Set snapshot create time.
     fn set_create_time(&mut self, time: String);
+
+    /// Check if the snapshot has been discarded.
+    fn discarded_snapshot(&self) -> bool;
+
+    /// Set discarded_snapshot
+    fn set_discarded_snapshot(&mut self, discarded: bool);
 }
 
 /// Trait to give interface for all Snapshot Parameters.
@@ -415,5 +443,15 @@ impl SnapshotDescriptor for SnapshotParams {
     /// Set snapshot create time.
     fn set_create_time(&mut self, time: String) {
         self.create_time = Some(time);
+    }
+
+    /// Check if the snapshot has been discarded.
+    fn discarded_snapshot(&self) -> bool {
+        self.discarded_snapshot
+    }
+
+    /// Set discarded_snapshot
+    fn set_discarded_snapshot(&mut self, discarded: bool) {
+        self.discarded_snapshot = discarded;
     }
 }
--- a/io-engine/src/grpc/v1/replica.rs
+++ b/io-engine/src/grpc/v1/replica.rs
@@ -245,7 +245,7 @@ impl ReplicaRpc for ReplicaService {
                     .ok_or(LvsError::RepDestroy {
                         source: Errno::ENOENT,
                         name: args.uuid.to_owned(),
-                        msg: "".into(),
+                        msg: "Replica not found".into(),
                     })?;
 
                 if let Some(lvs) = lvs {
@@ -263,7 +263,7 @@ impl ReplicaRpc for ReplicaService {
                         });
                     }
                 }
-                lvol.destroy().await?;
+                lvol.destroy_replica().await?;
                 Ok(())
             })?;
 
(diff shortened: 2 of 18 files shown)


io-engine/src/core/snapshot.rs
```diff
@@ -359,6 +374,12 @@ pub trait SnapshotDescriptor {
 
     /// Set snapshot create time.
     fn set_create_time(&mut self, time: String);
+
+    /// Get discarded_snapshot.
```
tiagolobocastro: nit:
```suggestion
    /// Check if the snapshot has been discarded.
```

io-engine/src/core/snapshot.rs
```diff
@@ -416,4 +437,14 @@ impl SnapshotDescriptor for SnapshotParams {
     fn set_create_time(&mut self, time: String) {
         self.create_time = Some(time);
     }
+
+    /// Get discarded_snapshot.
```
tiagolobocastro:
```suggestion
    /// Check if the snapshot has been discarded.
```

io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -691,4 +713,12 @@ impl SnapshotOps for Lvol {
             .filter(|b| b.is_snapshot_clone().is_some())
             .collect::<Vec<Lvol>>()
     }
+    /// Return bool value to indicate, if the snapshot is marked as future
+    /// delete.
```
tiagolobocastro:
```suggestion
    /// Check if the snapshot has been discarded.
```

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -947,6 +952,26 @@ impl LvsLvol for Lvol {
...
+    /// replica is identified as last clone.
+    async fn destroy_replica(mut self) -> Result<String, Error> {
+        let snapshot_lvol = self.is_snapshot_clone();
+        let name = self.name();
+        self.destroy().await?;
+
+        // If destroy replica is a snapshot clone and it is the last
+        // clone from the snapshot, destroy the snapshot
+        // if it is already marked as discardedSnapshot.
+        if let Some(snapshot_lvol) = snapshot_lvol {
```
tiagolobocastro, line 968: Let's say we crash after self.destroy and before we destroy snapshot_vol, then we "leak" it?
Should we also do this on pool import, so after the crash we "garbage collect"?
And can we also add a test to check this works?

hrudaya21: Yes, There will be leak if just clone.destroy() is success and there is a crash after that before snapshot destroy. Whether control plane will retry the replica destroy gRPC again after pool import ? 
If control plane trigger the replica destroy gRPC again, then "garbage collection" can be handled in replica destroy flow.

tiagolobocastro: Ah if destroy_replica call has enough information for us to do it then, I guess that could work too!
Not sure which way would be better, any thoughts @mtzaurus ?

hrudaya21: In destroy_replica request, We will get the replica uuid. If replica_uuid not found in the system, then rather failing the request, We can iterate through bdev to get lvol and match with `CloneXAttr.SourceUuid` with input replica_uuid to destroy it.

hrudaya21: Currently handled it as part of the destroy replica gRPC flow, when the retry of the destroy replica happens, a silent function (Without catching the snapshot destroy error) added to clean the discarded snapshot. The other way to send back the error code if destroy snapshot fails and CP will retry again. Not sure if We need to consider it @mtzaurus  @tiagolobocastro please suggest.

mtzaurus: > Let's say we crash after self.destroy and before we destroy snapshot_vol, then we "leak" it? Should we also do this on pool import, so after the crash we "garbage collect"? And can we also add a test to check this works?

We need to perform "garbage collection" also upon I/O engine start, to handle such a situation properly.

mtzaurus:
As further improvement, we could check whether we have only 1 clone of the snapshot and transparently transfer ownership of all snapshot's clusters to clone (via blob inflation) and then remove the snapshot.

io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -550,7 +559,20 @@ impl SnapshotOps for Lvol {
     async fn snapshot_iter(self) -> LvolSnapshotIter {
         LvolSnapshotIter::new(self)
     }
-
+    /// Destroy snapshot.
+    async fn destroy_snapshot(mut self) -> Result<(), Self::Error> {
+        if self.list_clones_by_snapshot_uuid().is_empty() {
```
mtzaurus, line 565: Here we can later add blob inflation for cases when there is exactly one clone exists.

io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -691,4 +713,44 @@ impl SnapshotOps for Lvol {
...
+    /// snapshot marked as discarded still present in the system. When
+    /// control plane does the redo of the destroy replica, clone uuid will
+    /// not be found in the system, but discarded snapshot can be searched
+    /// and destroyed.
+    async fn destroy_pending_discarded_snapshot(clone_uuid: &str) {
+        info!("HR: Pending destroy: {:?}", clone_uuid);
+        let bdev = match UntypedBdev::bdev_first() {
+            Some(b) => b,
+            None => {
+                info!("HR: No Device");
```
mtzaurus: Just wonder whether we need 'info' level here ? Isn't it too verbose?

io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -691,4 +713,44 @@ impl SnapshotOps for Lvol {
...
+            .into_iter()
+            .filter(|b| b.driver() == "lvol")
+            .map(|b| Lvol::try_from(b).unwrap())
+            .filter(|b| {
+                b.is_snapshot()
+                    && b.is_discarded_snapshot()
+                    && b.list_clones_by_snapshot_uuid().is_empty()
+            })
+            .collect::<Vec<Lvol>>();
+        for snapshot in discarded_snapshots {
```
mtzaurus: I'd perform parallel future execution for the sake of performance:
```
use futures::future::join_all;
```

io-engine/src/lvs/lvol_snapshot.rs
hrudaya21: > Just wonder whether we need 'info' level here ? Isn't it too verbose?

It was added for issue issue debugging. Will remove this info here. We have debug and error statement at the end of the function for success and failure case.

io-engine/src/lvs/lvol_snapshot.rs
hrudaya21: Sure. Thanks for the comment. Fix it now.

io-engine/src/lvs/lvol_snapshot.rs
hrudaya21: okay, I will explore this option to use of `spdk_bs_blob_decouple_parent` function to decouple snapshot from clone and blob inflation.
https://dcsw-bolt.atlassian.net/browse/GTM-1148

io-engine/src/grpc/v1/replica.rs
```diff
@@ -240,13 +241,22 @@ impl ReplicaRpc for ReplicaService {
...
+                    Some(l) => l,
+                    None => {
+                        // Try to destroy the pending snapshots without catching
+                        // the error when destroy_replica is tried again.
+                        Lvol::destroy_pending_discarded_snapshot(&args.uuid)
+                            .await;
+                        Err(LvsError::RepDestroy {
+                            source: Errno::ENOENT,
+                            name: args.uuid.to_owned(),
+                            msg: "".into(),
```
tiagolobocastro: nit:
```suggestion
                            msg: "Replica not found".into(),
```

io-engine/src/grpc/v1/replica.rs
```diff
@@ -240,13 +241,22 @@ impl ReplicaRpc for ReplicaService {
...
-                        name: args.uuid.to_owned(),
-                        msg: "".into(),
-                    })?;
+                {
+                    Some(l) => l,
+                    None => {
+                        // Try to destroy the pending snapshots without catching
+                        // the error when destroy_replica is tried again.
+                        Lvol::destroy_pending_discarded_snapshot(&args.uuid)
+                            .await;
```
tiagolobocastro: If we do it on import, then I guess we don't need it here?

io-engine/src/lvs/lvs_lvol.rs
tiagolobocastro: Yeah I also think it's better to do this on import rather than per-request later on.

io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -691,4 +714,42 @@ impl SnapshotOps for Lvol {
...
+    /// happened, it is possible that, last clone can be deleted, but linked
+    /// snapshot marked as discarded still present in the system. When
+    /// control plane does the redo of the destroy replica, clone uuid will
+    /// not be found in the system, but discarded snapshot can be searched
+    /// and destroyed.
+    async fn destroy_pending_discarded_snapshot(clone_uuid: &str) {
+        let bdev = match UntypedBdev::bdev_first() {
+            Some(b) => b,
+            None => return, /* No devices available */
+        };
```
tiagolobocastro: nit:
```suggestion
        let Some(bdev) = match UntypedBdev::bdev_first() else {
            return; /* No devices available */
        };
```

io-engine/src/grpc/v1/replica.rs
hrudaya21: Yes. Now it is moved to import_pool code flow.

io-engine/src/grpc/v1/replica.rs
hrudaya21: fixed.

mtzaurus (approved):

tiagolobocastro (approved):

hrudaya21:
bors try

hrudaya21:
bors cancel

hrudaya21:
bors try

sinhaashish:
bors cancel

hrudaya21:
bors merge

PR #1476, opened 2023-07-26
fix(lvs/bdev): don't show nil uuids

tiagolobocastro:
Recent spdk change made all bdev have nil uuid by default. This makes list-pools a bit confusing as shows device with nil uuid, and at the same time pool also has its own uuid.
So let's not add nil uuid to the URI.
todo: Should we set the bdev uuid to the lvs pool uuid?

### DIFF
--- a/io-engine/src/bdev_api.rs
+++ b/io-engine/src/bdev_api.rs
@@ -4,7 +4,10 @@ use snafu::Snafu;
 use std::{convert::TryFrom, num::ParseIntError, str::ParseBoolError};
 use url::ParseError;
 
-use crate::{bdev::uri, core::Bdev};
+use crate::{
+    bdev::uri,
+    core::{Bdev, Share},
+};
 
 // parse URI and bdev create/destroy errors common for all types of bdevs
 #[derive(Debug, Snafu, Clone)]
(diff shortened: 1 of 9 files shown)


dsharma-dc:
> This makes list-pools a bit confusing as shows device with nil uuid, and at the same time pool also has its own uuid.

@tiagolobocastro Could you please share this reported o/p here in comment?

io-engine/src/bdev_api.rs
```diff
@@ -152,21 +155,12 @@ where
...
-            if let Ok(mut uri) = url::Url::parse(alias) {
-                if bdev_uri_eq(&bdev, &uri) {
-                    if !uri.query_pairs().any(|e| e.0 == "uuid") {
-                        uri.query_pairs_mut()
-                            .append_pair("uuid", &bdev.uuid_as_string());
-                    }
-                    return Ok(uri);
-                }
-            }
+        match bdev.bdev_uri() {
```
dsharma-dc: Use `ok_or_else()` here ?

tiagolobocastro: Done, used `ok_or` instead

tiagolobocastro:
> > This makes list-pools a bit confusing as shows device with nil uuid, and at the same time pool also has its own uuid.
> 
> @tiagolobocastro Could you please share this reported o/p here in comment?

what's the o/p ?
The uri is: aio:///dev/sda&uuid=0000000...

dsharma-dc (approved):
This is the reason perhaps I was wondering few days back that the volume topology in support bundle shows like this for a pool:
  

>          "state": {
>             "capacity": 10724835328,
>             "disks": [
>               "aio:///dev/disk/by-id/scsi-0HC_Volume_34441339?uuid=00000000-0000-0000-0000-000000000000"
>             ],

tiagolobocastro:
> This is the reason perhaps I was wondering few days back that the volume topology in support bundle shows like this for a pool:
> 
> > ```
> >      "state": {
> >         "capacity": 10724835328,
> >         "disks": [
> >           "aio:///dev/disk/by-id/scsi-0HC_Volume_34441339?uuid=00000000-0000-0000-0000-000000000000"
> >         ],
> > ```

Yes seems silly :)

tiagolobocastro:
bors try

hrudaya21 (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1480, opened 2023-08-03
fix(snapshot/pool): get committed bytes from lvol committed

tiagolobocastro:
If we simply get committed from size then we don't take into account that snapshot lvol's will not grow to their capacity.
Instead, add a specific committed fn to accrue the committed bytes.

### DIFF
--- a/io-engine/src/lvs/lvs_lvol.rs
+++ b/io-engine/src/lvs/lvs_lvol.rs
@@ -637,9 +637,17 @@ impl LogicalVolume for Lvol {
         unsafe { spdk_blob_is_read_only(self.blob_checked()) }
     }
 
-    /// Return the size of the Snapshot in bytes.
+    /// Return the size of the Logical Volume in bytes.
     fn size(&self) -> u64 {
         self.as_bdev().size_in_bytes()
+    }
+
+    /// Return the committed size of the Logical Volume in bytes.
+    fn committed(&self) -> u64 {
+        match self.is_snapshot() {
+            true => self.usage().allocated_bytes,
+            false => self.size(),
+        }
     }
 
     /// Returns Lvol disk space usage.
--- a/io-engine/src/core/logical_volume.rs
+++ b/io-engine/src/core/logical_volume.rs
@@ -28,6 +28,9 @@ pub trait LogicalVolume {
     /// Return the size of the Logical Volume in bytes
     fn size(&self) -> u64;
 
+    /// Return the committed size of the Logical Volume in bytes.
+    fn committed(&self) -> u64;
+
     /// Returns Lvol disk space usage
     fn usage(&self) -> LvolSpaceUsage;
 }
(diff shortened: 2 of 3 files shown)


io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -637,11 +637,19 @@ impl LogicalVolume for Lvol {
...
 
-    /// Return the size of the Snapshot in bytes.
+    /// Return the size of the Logical Volume in bytes.
     fn size(&self) -> u64 {
         self.as_bdev().size_in_bytes()
     }
 
+    /// Return the committed size of the Logical Volume in bytes.
+    fn committed(&self) -> u64 {
+        match self.is_read_only() {
```
tiagolobocastro: Is this ok or should this be is_snapshot?

hrudaya21: If either of `data-ro` or `metadata-ro` is readonly in blob, We are marking read_only flag as true.
Now from SPDK not finding any instance where `metadata-ro` is true but `data-ro` is false. If such situation comes in future, then possibly `is_read_only` is not a better option. IMHO `is_snapshot` is safe for our scenario.

tiagolobocastro: Even if metadata is ro the snapshot can still allocate more clusters?
I've changed to is_snapshot as it seems safer then.

hrudaya21: From SPDK documentation, metadata section is fixed during blob initialization. In theory, I think We can add cluster if metadata is ro. But need to check once in the code.

tiagolobocastro: Then I think let's merge this as is for now, we can improve later with a more general read_only, if possible at all.

hrudaya21 (approved):

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -637,11 +637,19 @@ impl LogicalVolume for Lvol {
...
     }
 
-    /// Return the size of the Snapshot in bytes.
+    /// Return the size of the Logical Volume in bytes.
     fn size(&self) -> u64 {
         self.as_bdev().size_in_bytes()
     }
 
+    /// Return the committed size of the Logical Volume in bytes.
+    fn committed(&self) -> u64 {
```
mtzaurus, line 646: Shall we also handle thin-provisioned volumes which might have less allocated bytes than their size ?

tiagolobocastro: The way control-plane treats committed, is the amount that the pool usage can grow to, rather than how much more it can grow.
We could have a separate field of how much more we can grow, though we probably currently get this from pool.commited-pool.used

mtzaurus (approved):

tiagolobocastro:
bors merge

PR #1481, opened 2023-08-04
feat(clone): snapshot allocated bytes to be updated based on parent snapshot for clone

hrudaya21:


io-engine/src/core/logical_volume.rs
```diff
@@ -31,6 +32,9 @@ pub trait LogicalVolume {
     /// Return the committed size of the Logical Volume in bytes.
     fn committed(&self) -> u64;
 
+    /// Get BlobPtr from spdk_lvol.
```
tiagolobocastro, line 35:
```suggestion
    /// Get BlobPtr.
```

This is fine for now but I think in the future we'll need to refactor this as we're bringing already the second spdk specific method to an interface which is meant to be generic.

tiagolobocastro (approved):

io-engine/src/bin/io-engine-client/v1/replica_cli.rs
```diff
@@ -331,6 +336,8 @@ async fn replica_list(
                     "URI",
                     "IS_SNAPSHOT",
                     "IS_CLONE",
+                    "SNAP_ANCSTR_SIZE",
```
Abhinandan-Purkait: `SNAP_ANCESTOR_SIZE` Since only 2 letters maybe we can add?

Abhinandan-Purkait (approved):

io-engine/src/bin/io-engine-client/v1/replica_cli.rs
hrudaya21: It was just kept to save space in the console to add new fields. :)
Anyways, I have changed it. In future if any new fields comes, then We will realing all the fields accordingly.

hrudaya21:
bors try

hrudaya21:
bors try

hrudaya21:
bors merge

hrudaya21:
bors merge

PR #1482, opened 2023-08-08
feat(eventing): eventing integration and pool events

datacore-vvarakantham:
Eventing integration and pool create, delete events generation.

### DIFF
--- a/io-engine/src/lib.rs
+++ b/io-engine/src/lib.rs
@@ -17,6 +17,7 @@ pub mod delay;
 pub use spdk_rs::ffihelper;
 pub mod bdev_api;
 pub mod constants;
+pub mod eventing;
 pub mod grpc;
 pub mod host;
 pub mod jsonrpc;
--- a/io-engine/src/logger.rs
+++ b/io-engine/src/logger.rs
@@ -2,16 +2,24 @@ use ansi_term::{Colour, Style};
 use once_cell::sync::OnceCell;
 use std::{ffi::CStr, fmt::Write, os::raw::c_char, path::Path, str::FromStr};
 
+use crate::{
+    constants::{EVENTING_TARGET, SERVICE_NAME},
+    core::spawn,
+};
+use event_publisher::event_handler::EventHandle;
 use tracing_core::{event::Event, Level, Metadata};
 use tracing_log::{LogTracer, NormalizeEvent};
 use tracing_subscriber::{
+    filter::{filter_fn, Targets},
     fmt::{
-        format::{FmtSpan, FormatEvent, FormatFields},
+        format::{FmtSpan, FormatEvent, FormatFields, Writer},
         FmtContext,
         FormattedFields,
     },
+    layer::{Layer, SubscriberExt},
     registry::LookupSpan,
     EnvFilter,
+    Registry,
 };
 
 /// Returns hostname.
@@ -101,7 +109,7 @@ impl<'a> FormatLevel<'a> {
         }
     }
 
-    fn fmt_line(&self, f: &mut dyn Write, line: &str) -> std::fmt::Result {
+    fn fmt_line(&self, mut f: Writer<'_>, line: &str) -> std::fmt::Result {
         if self.ansi {
             write!(
                 f,
@@ -304,7 +312,7 @@ where
     fn format_event(
         &self,
         ctx: &FmtContext<'_, S, N>,
-        w: &mut dyn Write,
+        w: Writer<'_>,
         evt: &Event<'_>,
     ) -> std::fmt::Result {
         match self.style {
@@ -327,7 +335,7 @@ impl LogFormat {
     fn default_style<S, N>(
         &self,
         context: &FmtContext<'_, S, N>,
-        writer: &mut dyn Write,
+        mut writer: Writer<'_>,
         event: &Event<'_>,
     ) -> std::fmt::Result
     where
@@ -353,7 +361,7 @@ impl LogFormat {
             Location::new(meta)
         )?;
 
-        context.format_fields(writer, event)?;
+        context.format_fields(writer.by_ref(), event)?;
 
         writeln!(writer)
     }
@@ -362,7 +370,7 @@ impl LogFormat {
     fn compact_style<S, N>(
         &self,
         context: &FmtContext<'_, S, N>,
-        writer: &mut dyn Write,
+        mut writer: Writer<'_>,
         event: &Event<'_>,
     ) -> std::fmt::Result
     where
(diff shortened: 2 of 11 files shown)


io-engine/src/logger.rs
```diff
@@ -420,26 +425,45 @@ impl LogFormat {
...
     // Create a default subscriber.
-    let builder = tracing_subscriber::fmt::Subscriber::builder()
+    let builder = tracing_subscriber::fmt::layer()
         .with_span_events(FmtSpan::FULL)
         .event_format(format);
 
-    let subscriber = match EnvFilter::try_from_default_env() {
-        Ok(filter) => builder.with_env_filter(filter).finish(),
-        Err(_) => builder.with_env_filter(level).finish(),
+    let filter = match EnvFilter::try_from_default_env() {
```
tiagolobocastro, line 448: Should we filter out "mbus-events-target" from stdout?

protoc-3.15.8-linux-x86_64.zip
tiagolobocastro, line 1: this is not needed I'm assuming

chriswldenyer (approved):
Looks believable, but it looks like you need to remove the zip file.

protoc-3.15.8-linux-x86_64.zip
datacore-vvarakantham: Yes, it is not needed. I will remove it.

io-engine/src/logger.rs
```diff
@@ -420,26 +425,45 @@ impl LogFormat {
...
+        Ok(filter) => filter,
+        Err(_) => tracing_subscriber::EnvFilter::new(level),
+    };
+
+    // Get the optional eventing layer.
+    let events_layer = match events_url {
+        Some(url) => {
+            let events_filter =
+                Targets::new().with_target("mbus-events-target", Level::INFO);
```
sinhaashish: Better use constants for `mbus-events-target` and `io-engine`

io-engine/src/logger.rs
datacore-vvarakantham: Implemented excluding eventing traces now.

io-engine/src/lib.rs
```diff
@@ -45,3 +47,9 @@ pub extern "C" fn cps_init() {
     bdev::nexus::register_module();
     bdev::null_ng::register();
 }
+
+/// Event trait definition for creating events.
+pub(crate) trait Event {
+    /// Create event message.
+    fn event(&self, event_action: Option<EventAction>) -> EventMessage;
```
tiagolobocastro: Why is action optional?

tiagolobocastro (approved):

io-engine/src/lib.rs
datacore-vvarakantham: I've allowed for event_action to be optional in case it isn't required for any particular event generation. But there are no events with event_action of none for now. I can remove the optional if necessary.

tiagolobocastro: I see, it's fine then I was just curious.

io-engine/src/lvs/lvs_store.rs
```diff
@@ -848,3 +863,20 @@ impl PtplFileOps for LvsPtpl {
         std::path::PathBuf::from("pool/").join(self.uuid())
     }
 }
+
+// Pool event messages from Lvs data.
+impl Event for Lvs {
+    fn event(&self, event_action: Option<EventAction>) -> EventMessage {
+        let event_meta = EventMeta::new();
```
dsavitskiy: You should implement Default for EventMeta, if you have new() without args.

io-engine/src/lvs/lvs_store.rs
```diff
@@ -848,3 +863,20 @@ impl PtplFileOps for LvsPtpl {
...
+        let event_source = EventSource::new(node_name(&None));
+        EventMessage {
+            category: EventCategory::Pool as i32,
+            action: event_action.unwrap_or_default() as i32,
+            target: self.name().to_string(),
+            metadata: Some(EventMeta {
+                source: Some(event_source),
+                ..event_meta
```
dsavitskiy: If you implement Default for EventMeta, you can use `..Default::default()` instead of having a tmp variable "event_meta", making your code cleaner.

dsavitskiy: Another suggestion, you can add a method like "from_source()" to EventMeta, if creating from source is a frequent operation for meta.

io-engine/src/lvs/lvs_store.rs
```diff
@@ -52,6 +52,16 @@ use crate::{
     pool_backend::PoolArgs,
 };
 
+use events_api::event::{
```
dsavitskiy: Probably, we should consider moving mbus-specific code (dependent on generated code) to a separate module, like we do for gRPC-specific?

io-engine/src/logger.rs
```diff
@@ -420,26 +428,50 @@ impl LogFormat {
...
+
     LogTracer::init().expect("failed to initialise LogTracer");
 
     // Create a default subscriber.
-    let builder = tracing_subscriber::fmt::Subscriber::builder()
+    let builder = tracing_subscriber::fmt::layer()
         .with_span_events(FmtSpan::FULL)
-        .event_format(format);
+        .event_format(format)
+        .with_filter(filter_fn(|metadata| {
```
dsavitskiy, line 442: Instead of modifying logger, can we consider using existing functionality to selectively rise and lower level for specific target? Probably, changing RUST_LOG defaults to exclude "mbus-events-target" would work.

If changing RUST_LOG defaults proves to be no good for this particular case, could you move logger changes to separate commit?

dsavitskiy (changes requested):

io-engine/src/lvs/lvs_store.rs
datacore-vvarakantham: Done

io-engine/src/lvs/lvs_store.rs
datacore-vvarakantham: Added from_source() now in dependencies. Will pull in the changes here when the changes are merged in dependencies.

io-engine/src/logger.rs
datacore-vvarakantham: Changing the defaults to exclude "mbus-event-target" is removing the eventing traces from all the layers and the events are not generated.

Added eventing integration changes to separate commit now.

io-engine/src/lvs/lvs_store.rs
datacore-vvarakantham: I added from_source as suggested in dependencies which also removes the need for tmp variable.

dsavitskiy (approved):

datacore-vvarakantham:
bors merge

PR #1489, opened 2023-08-11
fix(snapshot): use customize function to judge snapshot inplace of spdk native function

hrudaya21:
To Judge if a Lvol is a snapshot, the custom xattr is checked now. Before it was using native SPDK function.

### DIFF
--- a/io-engine/src/lvs/lvol_snapshot.rs
+++ b/io-engine/src/lvs/lvol_snapshot.rs
@@ -429,7 +429,6 @@ impl SnapshotOps for Lvol {
                     if let Some(parent_lvol) = parent {
                         // Skip snapshots if it's parent is not matched.
                         if curr_attr_val != parent_lvol.uuid() {
-                            warn!("presisted parent ?curr_attr_val not matched to input parent ?parent_lvol.uuid()");
                             return None;
                         }
                     }
@@ -589,6 +588,7 @@ impl SnapshotOps for Lvol {
                 if !snapshot_lvol.is_snapshot() {
                     continue;
                 }
+
                 match snapshot_lvol.snapshot_descriptor(Some(self)) {
                     Some(snapshot_descriptor) => {
                         snapshot_list.push(snapshot_descriptor)
--- a/io-engine/src/lvs/lvs_lvol.rs
+++ b/io-engine/src/lvs/lvs_lvol.rs
@@ -21,7 +21,6 @@ use spdk_rs::libspdk::{
     spdk_blob_get_num_clusters_ancestors,
     spdk_blob_get_xattr_value,
     spdk_blob_is_read_only,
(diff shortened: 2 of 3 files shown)


io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -429,7 +429,7 @@ impl SnapshotOps for Lvol {
                     if let Some(parent_lvol) = parent {
                         // Skip snapshots if it's parent is not matched.
                         if curr_attr_val != parent_lvol.uuid() {
-                            warn!("presisted parent ?curr_attr_val not matched to input parent ?parent_lvol.uuid()");
+                            warn!("persisted parent: {curr_attr_val:?} not matched to input parent: {parent_lvol:?}");
```
tiagolobocastro: Suggest including self, and perhaps only showing parent uuid:

```suggestion
                            warn!("{self:?}, persisted parent {curr_attr_val:?} not matched to input parent {:?}", parent_lvol.uuid());
```

I've a question, won't we log this warn all the time when we're listing snapshots? I see this in the logs all over the place..

hrudaya21: I think now We can remove this warning statement from code.

1. When We listSnapshot with filter rule as source_uuid, this message prints for all the snapshots for which source_uuid is not matched. 
2. Currently this function is used by Clone as well to calculate the reference size. During this calculation, it is filtering other snapshots present in the system and giving warning, which is not needed.

tiagolobocastro: Right, so seems reasonable to remove this or at least make it optional if there's any scenario where we want to see it.

hrudaya21: This condition can hit while listing clone snapshot by uuid, when there is normal replica snapshot is available in the system or vice-versa. So better to remove this warning now.

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -717,7 +717,8 @@ impl LvsLvol for Lvol {
 
     /// Returns a boolean indicating if the lvol is a snapshot.
     fn is_snapshot(&self) -> bool {
-        unsafe { spdk_blob_is_snapshot(self.blob_checked()) }
+        Lvol::get_blob_xattr(self, SnapshotXattrs::SnapshotCreateTime.name())
```
tiagolobocastro, line 726: Please add a comment here stating why we're not using spdk native check

tiagolobocastro (approved):

sinhaashish (approved):

dsharma-dc (approved):

hrudaya21:
bors try

hrudaya21:
bors merge

tiagolobocastro:
bors merge

PR #1490, opened 2023-08-14
fix(qpair/connect): gracefully handle parallel running async and sync…

dsharma-dc:
… connect

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev.rs
@@ -607,8 +607,12 @@ impl<'n> Nexus<'n> {
         self.traverse_io_channels(
             sender,
             |chan, _sender| -> ChannelTraverseStatus {
-                chan.reconnect_all();
-                ChannelTraverseStatus::Ok
+                loop {
+                    match chan.reconnect_all() {
+                        11 => continue, // EAGAIN
+                        _ => return ChannelTraverseStatus::Ok,
+                    }
+                }
             },
             |status, sender| {
                 debug!("{self:?}: all I/O channels reconfigured");
--- a/io-engine/src/bdev/nexus/nexus_channel.rs
+++ b/io-engine/src/bdev/nexus/nexus_channel.rs
@@ -1,5 +1,6 @@
 //!
 //! IO is driven by means of so called channels.
+use nix::errno::Errno;
 use std::{
     cell::UnsafeCell,
     fmt::{Debug, Display, Formatter},
@@ -197,7 +198,8 @@ impl<'n> NexusChannel<'n> {
     /// online or offline. We don't know which child has gone, or was added, so
     /// we simply put back all the channels, and reopen the bdevs that are in
     /// the online state.
-    pub(crate) fn reconnect_all(&mut self) {
+    pub(crate) fn reconnect_all(&mut self) -> i32 {
+        let mut ret = 0;
         // clear the vector of channels and reset other internal values,
         // clearing the values will drop any existing handles in the
         // channel
@@ -220,6 +222,18 @@ impl<'n> NexusChannel<'n> {
                     writers.push(w);
                     readers.push(r);
                 }
+                (
+                    Err(CoreError::OpenBdev {
+                        source: _,
+                    }),
+                    _,
+                )
+                | (
+                    _,
+                    Err(CoreError::OpenBdev {
+                        source: _,
+                    }),
+                ) => ret = Errno::EAGAIN as i32,
                 _ => {
                     c.set_faulted_state(FaultReason::CantOpen);
                     error!("{self:?}: failed to get I/O handle for {c:?}");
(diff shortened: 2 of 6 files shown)


dsharma-dc:
bors try

dsharma-dc:
bors try

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -239,22 +253,35 @@ impl<'n> NexusChannel<'n> {
...
+                        }
+                    },
                 });
         }
 
+        if ret == Errno::EAGAIN as i32 {
+            // This would mean that we'll re-attempt to get the block device
+            // handles for all the child devices again, and not only
+            // the one for which we received EAGAIN.
+            return ret;
```
tiagolobocastro, line 275: I think there's an issue here, this might have succeeded for core 0 and not for other cores at which point we have a different view of children among cores which seems dangerous.

dsharma-dc: Agree, but it's no worse than what's possible today as well because the spdk callback to reconfigure for various spdk threads and cores are asynchronously and serially dispatched during channels iteration. Those callbacks can lag in time and land into similar situation.

tiagolobocastro: Well currently we'd assert and crash AIUI which is in a way safer.
I think we should attempt to pre-connect the handles before reconfigure, and then on reconnect we could just "acquire" those handles?

dsharma-dc: I'm trying out a change regarding this pre-connect approach.
btw, when I said the different view of channel from different cores is possible today also - I meant in a normal happy path also where no async connect is running in parallel with reconfigure, the spdk callbacks to different cores during iteration can lag in time and cause different views because spdk sends next callback only after we say `_iteration_continue_`.

dsharma-dc:
bors try

dsharma-dc:
bors try

io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -607,8 +607,12 @@ impl<'n> Nexus<'n> {
         self.traverse_io_channels(
             sender,
             |chan, _sender| -> ChannelTraverseStatus {
-                chan.reconnect_all();
-                ChannelTraverseStatus::Ok
+                loop {
+                    match chan.reconnect_all() {
```
tiagolobocastro, line 611: Whilst we're retrying here we're not yielding back to spdk, so how will the async connection connect in the background?

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -220,6 +222,18 @@ impl<'n> NexusChannel<'n> {
...
+                        source: _,
+                    }),
+                    _,
+                )
+                | (
+                    _,
+                    Err(CoreError::OpenBdev {
+                        source: _,
+                    }),
+                ) => ret = Errno::EAGAIN as i32,
```
tiagolobocastro, line 236: Could be be re-written in a more functional style collecting Results into vec, rather than having reads,writers and ret?

io-engine/src/bdev/nvmx/channel.rs
```diff
@@ -442,7 +442,7 @@ impl NvmeControllerIoChannel {
     /// Callback function to be invoked by SPDK to deinitialize I/O channel for
     /// NVMe controller.
     pub extern "C" fn destroy(device: *mut c_void, ctx: *mut c_void) {
-        trace!(
+        info!(
```
tiagolobocastro, line 445: don't think this was intended

io-engine/src/bdev/nvmx/handle.rs
```diff
@@ -197,15 +205,16 @@ impl NvmeDeviceHandle {
     }
 
     /// TODO:
-    fn connect_sync(&mut self) {
+    fn connect_sync(&mut self) -> i32 {
```
tiagolobocastro, line 208: could this return Errno?

tiagolobocastro (commented):
When do we have a parallel running async and sync connect?

dsharma-dc:
@tiagolobocastro 
> When do we have a parallel running async and sync connect?

It can happen under below scenario as seen from the reported instance. Before the poller completes, reconfigure happens on the core.

```
add_new_child(cN)
  get_async_io_handle
   set_qpair_state(Connecting)
    io_qpair_connect_poller running
retire_child(cX)
 unplug_cX
  reconfigure
   get_sync_io_handles()   <------- including cN 
    expect_qpair_state(Disconnected)
```

io-engine/src/bdev/nexus/nexus_bdev.rs
dsharma-dc: hmm, you're right. I'm able to reproduce this with a test intermittently and when it reproduces, I see the reconnect loop running forever. I guess I haven't understood poller behaviour completely yet. I'll see how to yield here and continue spdk iteration from same point.

tiagolobocastro:
> 

This further highlights to me that the current way of doing things per-core is dangerous. What if one core connects and the others don't?
IMHO we should instead follow the approach @dsavitskiy proposed before of connecting before hand, and just "pass" the handles on reconnect_all.

dsharma-dc:
Will create new PR.

PR #1491, opened 2023-08-16
feat(nvmf): adding support for configuring target CRD

dsavitskiy:
This PR adds support to allow to specify target CRDT (Command Retry Delay) via CLI/env, to prevent immediate initiator failure when a nexus fails when the last nexus child fails. This provides a period to allow control plane to recreate the nexus.  The default is set to 30 sec.

### DIFF
--- a/io-engine/src/core/env.rs
+++ b/io-engine/src/core/env.rs
@@ -164,6 +164,13 @@ pub struct MayastorCliArgs {
     #[structopt(short = "T", long = "tgt-iface", env = "NVMF_TGT_IFACE")]
     /// NVMF target interface (ip, mac, name or subnet).
     pub nvmf_tgt_interface: Option<String>,
+    /// NVMF target Command Retry Delay.
+    #[structopt(
+        long = "tgt-crdt",
+        env = "NVMF_TGT_CRDT",
+        default_value = "30"
+    )]
+    pub nvmf_tgt_crdt: u16,
     /// api Version
     #[structopt(
         long,
@@ -230,6 +237,7 @@ impl Default for MayastorCliArgs {
             nvme_ctl_io_ctx_pool_size: 65535,
             registration_endpoint: None,
             nvmf_tgt_interface: None,
+            nvmf_tgt_crdt: 30,
             api_versions: vec![ApiVersion::V0, ApiVersion::V1],
             diagnose_stack: None,
             reactor_freeze_detection: false,
@@ -328,6 +336,7 @@ pub struct MayastorEnvironment {
     bdev_io_ctx_pool_size: u64,
     nvme_ctl_io_ctx_pool_size: u64,
     nvmf_tgt_interface: Option<String>,
+    pub nvmf_tgt_crdt: u16,
     api_versions: Vec<ApiVersion>,
 }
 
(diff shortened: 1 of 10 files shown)


dsavitskiy:
bors try

io-engine/src/subsys/config/opts.rs
```diff
@@ -87,6 +87,13 @@ impl From<NvmfTgtConfig> for Box<spdk_nvmf_target_opts> {
         let mut out = Self::default();
         copy_str_with_null(&o.name, &mut out.name);
         out.max_subsystems = o.max_namespaces;
+        let c = try_from_env("NVMF_TGT_CRDT", 0);
```
tiagolobocastro: Should we enable it by default?
Also, can we make it a cli argument which would be enabled via helm?

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -537,6 +538,17 @@ impl NvmfSubsystem {
 
     /// destroy the subsystem
     pub fn destroy(&self) -> i32 {
+        unsafe {
+            if spdk_nvmf_subsystem_remove_ns(self.0.as_ptr(), 1) != 0 {
```
tiagolobocastro: Should we check that we're either stopped or not paused? Otherwise this will trigger an assert from spdk.

dsavitskiy: Probably, I better mark this method "unsafe", and make it pub(super). This method is only called from other method, after stopping the subsystem.

io-engine/src/subsys/config/opts.rs
dsavitskiy: Fixed

dsavitskiy:
bors try

io-engine/src/subsys/nvmf/subsystem.rs
tiagolobocastro: Yeah that's good :+1:

tiagolobocastro (approved):

dsharma-dc:
Could you please put some description in the description box of this PR about the rationale behind this change?

io-engine/src/subsys/config/opts.rs
```diff
@@ -87,15 +91,18 @@ impl From<NvmfTgtConfig> for Box<spdk_nvmf_target_opts> {
         let mut out = Self::default();
         copy_str_with_null(&o.name, &mut out.name);
         out.max_subsystems = o.max_namespaces;
+        out.crdt[0] = o.crdt;
```
dsharma-dc, line 94: Should we set o.crdt into all three values i.e. crdt[1] and crdt[2] also? or are we certain that spdk is always using the crdt[0] ?

io-engine/src/core/env.rs
```diff
@@ -372,6 +381,7 @@ impl Default for MayastorEnvironment {
             bdev_io_ctx_pool_size: 65535,
             nvme_ctl_io_ctx_pool_size: 65535,
             nvmf_tgt_interface: None,
+            nvmf_tgt_crdt: 30,
```
dsharma-dc, line 384: 30secs seems too long a wait. What's the reason to choose 30secs?

dsharma-dc (approved):

io-engine/src/subsys/config/opts.rs
dsavitskiy: We can consider that in future, but at the moment we don't have a way to use CRDT 2,3 because SPDK always responds with CRDT 1: https://github.com/openebs/spdk/blob/v23.01.x-mayastor/lib/nvmf/ctrlr.c#L4343

io-engine/src/core/env.rs
dsavitskiy: That's reasonable enough to allow control plane to recreate nexus.

dsavitskiy:
bors merge

PR #1493, opened 2023-08-18
fix(snapshot): reset usage cache for the parent lvol when snapshot is destroyed

hrudaya21:
**Change Details:**
1. When snapshot is destroyed, reset the usage cache for parent lvol.

2. When snapshot is destroyed, reset the usage cache of successor snapshots (same parent) and clones.
 
3. When the last clone is destroyed and there are any discarded snapshots leftout, as part of destroying snapshot, reset the parent Lvol usage details from cache.
 
4. If discarded Snapshot destroy happens as part of pool import, also reset the parent Lvol usage details from cache.

**Why Changes Required**

Volume usage information should show properly when chain of snapshots and clones are deleted. Thin provisioned lvol cache information to be cleared to show the actual allocated size when any linked lvol (snapshot/clone) is destroyed.

### DIFF
--- a/io-engine/src/lvs/lvol_snapshot.rs
+++ b/io-engine/src/lvs/lvol_snapshot.rs
@@ -19,6 +19,7 @@ use futures::channel::oneshot;
 use nix::errno::Errno;
 use spdk_rs::libspdk::{
     spdk_blob,
+    spdk_blob_reset_used_clusters_cache,
     spdk_lvol,
     spdk_xattr_descriptor,
     vbdev_lvol_create_clone_ext,
@@ -562,6 +563,7 @@ impl SnapshotOps for Lvol {
     /// Destroy snapshot.
     async fn destroy_snapshot(mut self) -> Result<(), Self::Error> {
         if self.list_clones_by_snapshot_uuid().is_empty() {
+            self.reset_snapshot_parent_successor_usage_cache();
             self.destroy().await?;
         } else {
             self.set_blob_attr(
@@ -731,7 +733,7 @@ impl SnapshotOps for Lvol {
         let Some(bdev) = UntypedBdev::bdev_first() else {
             return; /* No devices available */
         };
-        let futures = bdev
+        let snap_list = bdev
             .into_iter()
             .filter(|b| b.driver() == "lvol")
             .map(|b| Lvol::try_from(b).unwrap())
@@ -740,8 +742,11 @@ impl SnapshotOps for Lvol {
                     && b.is_discarded_snapshot()
                     && b.list_clones_by_snapshot_uuid().is_empty()
             })
-            .map(|s| s.destroy())
-            .collect::<Vec<_>>();
+            .collect::<Vec<Lvol>>();
+        snap_list
+            .iter()
+            .for_each(|s| s.reset_snapshot_parent_successor_usage_cache());
+        let futures = snap_list.into_iter().map(|s| s.destroy());
         let result = join_all(futures).await;
         for r in result {
             match r {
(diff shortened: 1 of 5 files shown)


io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -792,4 +797,75 @@ impl SnapshotOps for Lvol {
...
+    fn reset_snapshot_parent_successor_usage_cache(&self) {
+        if let Some(snapshot_parent_uuid) =
+            Lvol::get_blob_xattr(self, SnapshotXattrs::ParentId.name())
+        {
+            if let Some(parent_lvol) = if let Some(bdev) =
+                UntypedBdev::lookup_by_uuid_str(snapshot_parent_uuid.as_str())
+            {
+                Lvol::try_from(bdev).ok()
+            } else {
+                None
```
tiagolobocastro: What is this code doing?

io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -792,4 +797,75 @@ impl SnapshotOps for Lvol {
...
+            })
+            .collect::<Vec<Lvol>>();
+        let mut successor_clones: Vec<Lvol> = vec![];
+
+        loop {
+            if successor_snapshots.is_empty() && successor_clones.is_empty() {
+                break;
+            }
+            if !successor_snapshots.is_empty() {
+                let snapshot = successor_snapshots.remove(0);
```
tiagolobocastro:
```suggestion
            if let Some(snapshot) = successor_snapshots.pop() {
```

io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -792,4 +797,75 @@ impl SnapshotOps for Lvol {
...
+                unsafe {
+                    spdk_blob_reset_used_clusters_cache(
+                        snapshot.blob_checked(),
+                    );
+                }
+                let new_clone_list = snapshot.list_clones_by_snapshot_uuid();
+                successor_clones.extend(new_clone_list);
+            }
+            if !successor_clones.is_empty() {
+                let clone = successor_clones.remove(0);
```
tiagolobocastro:
```suggestion
            if let Some(clone) = successor_clones.pop() {
```

io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -792,4 +797,75 @@ impl SnapshotOps for Lvol {
...
+                    None
+                }
+            })
+            .collect::<Vec<Lvol>>();
+        let mut successor_clones: Vec<Lvol> = vec![];
+
+        loop {
+            if successor_snapshots.is_empty() && successor_clones.is_empty() {
+                break;
+            }
```
tiagolobocastro:
```suggestion
        while !successor_snapshots.is_empty() || !successor_clones.is_empty() {
```

io-engine/src/lvs/lvol_snapshot.rs
hrudaya21: > What is this code doing?

Taken different approch to clean the code.

io-engine/src/lvs/lvol_snapshot.rs
hrudaya21: fixed

io-engine/src/lvs/lvol_snapshot.rs
hrudaya21: fixed

io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -792,4 +797,69 @@ impl SnapshotOps for Lvol {
...
+                        );
+                    }
+                }
+            }
+            self.reset_successor_lvol_usage_cache(snapshot_parent_uuid);
+        }
+    }
+
+    /// When snapshot is destroyed, reset cache of successor snapshots and
+    /// clones based on snapshot parent uuid.
```
tiagolobocastro, line 822: Do we need to do this for the entire list or just the direct successors? Though I guess we don't have that kind of knowledge at this level?

io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -792,4 +797,69 @@ impl SnapshotOps for Lvol {
...
+    fn reset_successor_lvol_usage_cache(&self, snapshot_parent_uuid: String) {
+        let mut successor_snapshots = Lvol::list_all_snapshots()
+            .iter()
+            .map(|v| v.snapshot_lvol().clone())
```
tiagolobocastro: Should we clone only on 833 with Some(l.clone()) ? Otherwise we're cloning things that might get filtered out?

io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -792,4 +797,69 @@ impl SnapshotOps for Lvol {
...
+            .filter_map(|l| {
+                let uuid =
+                    Lvol::get_blob_xattr(self, SnapshotXattrs::ParentId.name());
+                if uuid.is_some()
+                    && uuid.unwrap_or_default() == snapshot_parent_uuid
+                {
+                    Some(l)
+                } else {
+                    None
+                }
```
tiagolobocastro:
```suggestion
                match uuid {
                    Some(uuid) if uuid == snapshot_parent_uuid => Some(l),
                    _ => None,
                }
```

tiagolobocastro (approved):

io-engine/src/lvs/lvol_snapshot.rs
hrudaya21: Yes. It will save unncessary cloning. I will fix it. Thank you.

io-engine/src/lvs/lvol_snapshot.rs
hrudaya21: I think entire downstream tree need to clean as clone inherit size from snapshot.

io-engine/src/lvs/lvol_snapshot.rs
hrudaya21: fixed. it is more simpler now. Thanks.

io-engine/src/lvs/lvol_snapshot.rs
hrudaya21: But I feel it is not efficient and performance is not good. Possibly need to have some enhancement in SPDK to identify the tree.

hrudaya21:
bors try

mtzaurus:
As a future optimisation, we could do all these actions inside SPDK, as part or post-removal logic. I

mtzaurus (approved):

hrudaya21:
bors merge

PR #1494, opened 2023-08-21
fix(casperf): invalid next offset calculation

tiagolobocastro:
Also add easy block size specification, eg: allows "1MiB". Disable cleanup of nexuses when they're not configured. Prevent double cleanup with duplicate handlers.

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_module.rs
+++ b/io-engine/src/bdev/nexus/nexus_module.rs
@@ -21,12 +21,17 @@ pub(crate) struct NexusModule {}
 
 impl NexusModule {
     /// Returns Nexus Bdev module instance.
+    /// # Warning
     /// Panics if the Nexus module was not registered.
     pub fn current() -> BdevModule {
         match BdevModule::find_by_name(NEXUS_MODULE_NAME) {
             Ok(m) => m,
             Err(err) => panic!("{}", err),
         }
+    }
+    /// Returns Nexus Bdev module instance, if registered.
+    pub fn current_opt() -> Option<BdevModule> {
+        BdevModule::find_by_name(NEXUS_MODULE_NAME).ok()
     }
 }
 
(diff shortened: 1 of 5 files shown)


io-engine/src/bin/casperf.rs
```diff
@@ -130,12 +130,12 @@ impl Job {
             return;
         }
 
-        let offset = (job.rng.gen::<u64>() % job.io_size) * job.io_blocks;
+        let offset = job.rng.gen_range(0 .. job.io_blocks) * job.io_size;
```
dsharma-dc, line 133: how do we ensure this doesn't span beyond device size?

dsharma-dc (approved):

io-engine/src/bdev/nexus/nexus_module.rs
```diff
@@ -21,13 +21,21 @@ pub(crate) struct NexusModule {}
...
     /// Panics if the Nexus module was not registered.
     pub fn current() -> BdevModule {
         match BdevModule::find_by_name(NEXUS_MODULE_NAME) {
             Ok(m) => m,
             Err(err) => panic!("{}", err),
         }
     }
+    /// Returns Nexus Bdev module instance, if registered.
+    pub fn current_opt() -> Option<BdevModule> {
+        match BdevModule::find_by_name(NEXUS_MODULE_NAME) {
```
mtzaurus: IMHO, you can use just '.ok()' method to get the resulting Option:
```
BdevModule::find_by_name(NEXUS_MODULE_NAME).ok()
```

mtzaurus (approved):

io-engine/src/bdev/nexus/nexus_module.rs
tiagolobocastro: yes, indeed

io-engine/src/bin/casperf.rs
tiagolobocastro: I believe that's ensured by the io_blocks calculation, which is the size of the device in io_size units.

tiagolobocastro:
bors merge

PR #1505, opened 2023-08-25
Lvs and PCIe fixes

tiagolobocastro:
fix(pcie): spdk requires nvme path id
    
    When destroying a pcie nvme bdev we must not specify the nvme path id.
    Also it was noticed that when destroy fails we end up with no alias which
    could cause some panics. For this particular case we now add the alias
    back if the bdev is still present though we need a more robust fix..
    
---

    fix(lvs/error): handle missing errors during import/create
    
    Some error codes not being properly mapped or handled making it impossible to create
    or import a pool in certain situations.
    This ensures we handle errors like already exists properly which allows the control-plane
    to ensure pool creation/import.
    

dsavitskiy (approved):

Abhinandan-Purkait (approved):

avishnu:
bors merge

PR #1506, opened 2023-09-08
fix(channel_reconfig): don't reconfigure the non-IO nexus channel

dsharma-dc:
This PR changes the way we handle nexus channel on init_thread spdk thread. Since the nexus channel on init_thread is not used for normal I/O, we now don’t attempt to reconfigure it as part of dynamic reconfiguration event. This avoids the race between some ongoing async qpair connect on init thread racing with a sync connect attempt from channel reconfigure path

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev.rs
@@ -608,7 +608,9 @@ impl<'n> Nexus<'n> {
         self.traverse_io_channels(
             sender,
             |chan, _sender| -> ChannelTraverseStatus {
-                chan.reconnect_all();
+                if chan.is_io_channel() {
+                    chan.reconnect_all();
+                }
                 ChannelTraverseStatus::Ok
             },
             |status, sender| {
--- a/io-engine/src/bdev/nexus/nexus_channel.rs
+++ b/io-engine/src/bdev/nexus/nexus_channel.rs
@@ -4,11 +4,13 @@ use std::{
     cell::UnsafeCell,
     fmt::{Debug, Display, Formatter},
     pin::Pin,
+    sync::atomic::Ordering,
 };
 
 use super::{FaultReason, IOLogChannel, Nexus, NexusBio};
 
 use crate::core::{BlockDeviceHandle, CoreError, Cores};
+use spdk_rs::Thread;
 
 /// I/O channel, per core.
 #[repr(C)]
@@ -22,6 +24,7 @@ pub struct NexusChannel<'n> {
     frozen_ios: Vec<NexusBio<'n>>,
     nexus: Pin<&'n mut Nexus<'n>>,
     core: u32,
+    is_io_chan: bool,
 }
 
 impl<'n> Debug for NexusChannel<'n> {
@@ -77,26 +80,45 @@ impl<'n> NexusChannel<'n> {
     pub(crate) fn new(nexus: Pin<&mut Nexus<'n>>) -> Self {
         debug!("{nexus:?}: new channel on core {c}", c = Cores::current());
 
+        let b_init_thrd_hdls =
+            super::ENABLE_IO_ALL_THRD_NX_CHAN.load(Ordering::SeqCst);
+        let is_io_chan =
+            Thread::current().unwrap() != Thread::primary() || b_init_thrd_hdls;
+
         let mut writers = Vec::new();
         let mut readers = Vec::new();
 
-        nexus
-            .children_iter()
-            .filter(|c| c.is_healthy())
-            .for_each(|c| match (c.get_io_handle(), c.get_io_handle()) {
-                (Ok(w), Ok(r)) => {
-                    writers.push(w);
-                    readers.push(r);
-                }
-                _ => {
-                    c.set_faulted_state(FaultReason::CantOpen);
-                    error!(
-                        "Failed to get I/O handle for {c}, \
-                        skipping block device",
-                        c = c.uri()
-                    )
-                }
-            });
+        if is_io_chan {
+            nexus.children_iter().filter(|c| c.is_healthy()).for_each(
+                |c| match (c.get_io_handle(), c.get_io_handle()) {
+                    (Ok(w), Ok(r)) => {
+                        writers.push(w);
+                        readers.push(r);
+                    }
+                    _ => {
+                        c.set_faulted_state(FaultReason::CantOpen);
+                        error!(
+                            "Failed to get I/O handle for {c}, \
+                            skipping block device",
+                            c = c.uri()
+                        )
+                    }
+                },
+            );
+        } else {
+            // If we are here, this means the nexus channel being created is not
+            // the one to be used for normal IOs. Such a channel is
+            // created in rebuild path today, and it's on the init
+            // thread. The channels that we use for normal nexus IO
+            // are not on init thread however, those are on spdk threads
+            // created for nvmf target during poll group init. Those are the
+            // spdk threads named mayastor_nvmf_tcp_pg_core_*. The
+            // channel on init thread is only used for rebuild IOs.
+            // And the rebuild IOs are dispatched by
+            // directly calling write API without going via writers abstraction.
+            // Refer GTM-1075 for the race condition details.
+            debug!("{nexus:?}: skip nexus channel setup({t:?}). is_io_channel: {is_io_chan}", t = Thread::current().unwrap());
+        }
 
         Self {
             writers,
@@ -108,6 +130,7 @@ impl<'n> NexusChannel<'n> {
             io_mode: IoMode::Normal,
             frozen_ios: Vec::new(),
             core: Cores::current(),
+            is_io_chan,
         }
     }
 
@@ -139,6 +162,11 @@ impl<'n> NexusChannel<'n> {
     /// Returns the total number of available readers in this channel.
     pub(crate) fn num_readers(&self) -> usize {
         self.readers.len()
+    }
+
+    // Returns a bool indicating whether this channel is setup for normal IOs.
+    pub(crate) fn is_io_channel(&self) -> bool {
+        self.is_io_chan
     }
 
     /// Calls the given callback for each active writer.
--- a/io-engine/src/core/env.rs
+++ b/io-engine/src/core/env.rs
@@ -192,6 +192,12 @@ pub struct MayastorCliArgs {
     /// termination.
     #[structopt(long, hidden = true)]
     pub skip_sig_handler: bool,
+    /// Whether the nexus channel should have readers/writers configured.
+    /// This must be set true ONLY from tests. This option can be removed once
+    /// dynamic reconfiguration of nexus channels can handle async-qpair
+    /// connect. Details in NexusChannel::new
+    #[structopt(long = "enable-io-all-thrd-nexus-channels", hidden = true)]
+    pub enable_io_all_thrd_nexus_channels: bool,
     /// Events message-bus endpoint url.
     #[structopt(long)]
     pub events_url: Option<url::Url>,
@@ -246,6 +252,7 @@ impl Default for MayastorCliArgs {
             reactor_freeze_detection: false,
             reactor_freeze_timeout: None,
             skip_sig_handler: false,
+            enable_io_all_thrd_nexus_channels: false,
             events_url: None,
         }
     }
@@ -344,6 +351,7 @@ pub struct MayastorEnvironment {
     pub nvmf_tgt_crdt: u16,
     api_versions: Vec<ApiVersion>,
     skip_sig_handler: bool,
+    enable_io_all_thrd_nexus_channels: bool,
 }
 
 impl Default for MayastorEnvironment {
@@ -390,6 +398,7 @@ impl Default for MayastorEnvironment {
             nvmf_tgt_crdt: 0,
             api_versions: vec![ApiVersion::V0, ApiVersion::V1],
             skip_sig_handler: false,
+            enable_io_all_thrd_nexus_channels: false,
         }
     }
 }
@@ -509,6 +518,8 @@ impl MayastorEnvironment {
             nvmf_tgt_crdt: args.nvmf_tgt_crdt,
             api_versions: args.api_versions,
             skip_sig_handler: args.skip_sig_handler,
+            enable_io_all_thrd_nexus_channels: args
+                .enable_io_all_thrd_nexus_channels,
             ..Default::default()
         }
         .setup_static()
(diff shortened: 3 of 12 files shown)


io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -606,9 +608,24 @@ impl<'n> Nexus<'n> {
...
+                    // we set a per-thread variable to store information about where were in the
+                    // channel iteration when this happened. As soon as the async connect completes, it
+                    // will consume this per-thread variable and resume the channel iteration.
+                    11 => {
+                        CHAN_ITER_MARK.with(|c| {
+                            let val = unsafe { &mut *c.get()};
+                            *val = Some((chan as *mut _ as *mut c_void, iter));
+                            warn!("EAGAIN: Putting marker and yielding iteration : {:?} - {}", chan, Mthread::current().unwrap().name());
+                        });
+                        return ChannelTraverseStatus::Cancel
```
tiagolobocastro: What does cancel do, leave the traversal "paused"?

io-engine/src/bdev/nvmx/qpair.rs
```diff
@@ -509,6 +525,21 @@ impl<'a> Connection<'a> {
...
                 .expect("Failed to notify a connection waiter");
         });
+
+        // Wake up the waiter who received EAGAIN because this async connect was ongoing.
+        CHAN_ITER_MARK.with(|c| {
+            if let Some(cstored) = unsafe { &mut *c.get()}.take() {
+                let chan = unsafe {&mut *(cstored.0 as *mut NexusChannel)};
+                let iter = unsafe {&mut *(cstored.1)};
+                warn!("Consume marker and continue traversing channels: {:?} - {}", chan, Mthread::current().unwrap().name());
+                match chan.reconnect_all() {
```
tiagolobocastro: this seems quite risky to run outside of the channel traversal..
Are we sure we're connecting the right channel here?

io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -606,9 +608,24 @@ impl<'n> Nexus<'n> {
...
+                    // The only path that can return EAGAIN here is the sync qpair connect via
+                    // get_io_handle() while doing reconnect_all the current thread has also sent
+                    // an async connect for the same qpair already. Upon receiving this error code,
+                    // we set a per-thread variable to store information about where were in the
+                    // channel iteration when this happened. As soon as the async connect completes, it
+                    // will consume this per-thread variable and resume the channel iteration.
+                    11 => {
+                        CHAN_ITER_MARK.with(|c| {
+                            let val = unsafe { &mut *c.get()};
+                            *val = Some((chan as *mut _ as *mut c_void, iter));
```
tiagolobocastro: what if we have an existing "mark" from another nexus, won't this overwrite it?

tiagolobocastro (commented):
Have you given any thought to my previous suggestion of splitting reconnect into 2, 1 to simply connect and setup handles, and another to "committing" the handles into the readers and writers vec? IMHO that may potentially be a cleaner approach.

dsharma-dc:
> Have you given any thought to my previous suggestion of splitting reconnect into 2, 1 to simply connect and setup handles, and another to "committing" the handles into the readers and writers vec? IMHO that may potentially be a cleaner approach.

I tried to write that way as well - i.e. iterate over children and dispatch a future to async connect, and wait for that future somehow. We **have** to wait for the future to finish otherwise we'd increase the chances of this race happening even more by leaving more async connects out there. On the other hand, when we send a future we can't be certain which `spdk_thread` will that future run against(which is correctly ensured by channel traversal way). On first core, we have two `spdk_thread`s, and we need to reconfigure channel for both. (We indeed create a nexus channel on `init_thread` when first rebuild happens).
I'll take a look at new jira that Mikhail created and see if another way is feasible to do generically. Then we don't need to handle only the race specifically.

io-engine/src/bdev/nexus/nexus_bdev.rs
dsharma-dc: This draft code is uploaded here to demo the idea. The per-thread variable need to be converted into a vector or map to handle multiple nexus children.

dsharma-dc:
This PR now contains only the test which reproduces the race. The merge of this change will follow the merge of the actual fix.

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -108,6 +110,7 @@ impl<'n> NexusChannel<'n> {
             io_mode: IoMode::Normal,
             frozen_ios: Vec::new(),
             core: Cores::current(),
+            is_io_chan: Thread::current().unwrap() != Thread::primary(),
```
dsharma-dc: Might be frowned upon, but this kind of unwrap usage exists in io-engine core code today as well, and it is guaranteed that this code is running in spdk thread context.

tiagolobocastro: I think we should be comparing it the other way around, it's an io_chan if it's an nvmf thread? This is worth commenting out why we're doing this as in the future we might have a need to do IO from nexus channels.

But what bothers me the most is that we still have sync connect nonetheless... above on L88. Perhaps we should also avoid setting up readers and writers?

dsharma-dc: > I think we should be comparing it the other way around, it's an io_chan if it's an nvmf thread? 

The current vs primary based check looks more finegrained. When you say nvmf thread - I guess it means there is a poll group attached to it for IO. nvmf target object itself doesn't seem to be spdk_thread bound. It's a thread_local (system thread local). So not sure if there is a clear way to identify this way.

> This is worth commenting out why we're doing this as in the future we might have a need to do IO from nexus channels.

Yeah I have a long code comment written, missed to paste here.

> But what bothers me the most is that we still have sync connect nonetheless... above on L88. Perhaps we should also avoid setting up readers and writers?

True. Things are in our favour here from the point that the probability of race is almost negligible because usually it's the reconfigurations that happen frequently and not this new channel construction. However, I agree that keeping the nexus channel dummy on init thread by not configuring `readers/writers` will be ok. The rebuild path anyway directly calls write api and not via `writers`. To achieve that, I'll move the check from L113 to very beginning, and do the connect loops only if bool is `true`.

dsharma-dc:
bors try

io-engine/src/bdev/nexus/nexus_channel.rs
tiagolobocastro: Isn't it this? https://github.com/openebs/mayastor/blob/558cfe7f798ce3de09552f224736133ea85024db/io-engine/src/subsys/nvmf/target.rs#L191

dsharma-dc:
bors try

io-engine/src/bdev/nexus/nexus_channel.rs
dsharma-dc: > Isn't it this?

Yes it's this one. The way I can think using this info is something like below pseudocode:


```
let is_io_channel = false;
 t = current_spdk_thread();

loop_through_all_poll_groups_in_NVMF_PGS{
  If t_is_found_linked_with_some_poll_group (
   is_io_channel = true;
  )
}

```
Anything better?

dsharma-dc:
bors try

dsharma-dc:
bors try

dsharma-dc:
bors cancel

Need to use new parameter in some more tests.

dsharma-dc:
bors cancel

dsharma-dc:
bors try-

dsharma-dc:
bors try

dsharma-dc:
bors try

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -77,26 +80,45 @@ impl<'n> NexusChannel<'n> {
...
+            // created in rebuild path today, and it's on the init
+            // thread. The channels that we use for normal nexus IO
+            // are not on init thread however, those are on spdk threads
+            // created for nvmf target during poll group init. Those are the
+            // spdk threads named mayastor_nvmf_tcp_pg_core_*. The
+            // channel on init thread is only used for rebuild IOs.
+            // And the rebuild IOs are dispatched by
+            // directly calling write API without going via writers abstraction.
+            // Refer GTM-1075 for the race condition details.
+            info!("{nexus:?}: skip nexus channel setup({}). is_io_channel: {is_io_chan}", Thread::current().unwrap().name());
```
tiagolobocastro: Debug enough?

```suggestion
            debug!("{nexus:?}: skip nexus channel setup({}). is_io_channel: {is_io_chan}", Thread::current().unwrap().name());
```

io-engine/src/bin/io-engine.rs
```diff
@@ -84,6 +93,12 @@ fn start_tokio_runtime(args: &MayastorCliArgs) {
         warn!("Nexus reset is disabled");
     }
 
+    // Whether the nexus channel should have readers/writers configured.
+    // This must be set true ONLY from tests.
+    if let Ok(v) = std::env::var("INIT_THRD_NX_CHAN_HANDLES") {
```
tiagolobocastro: Why do we need to set this again if we've set it on L75?

io-engine/src/core/env.rs
```diff
@@ -192,6 +192,14 @@ pub struct MayastorCliArgs {
...
     #[structopt(long, hidden = true)]
     pub skip_sig_handler: bool,
+    /// Whether the nexus channel should have readers/writers configured.
+    /// This must be set true ONLY from tests. Details in NexusChannel::new
+    #[structopt(
+        long = "init-thrd-nx-chan-hdls",
+        env = "INIT_THRD_NX_CHAN_HANDLES",
+        hidden = true
+    )]
+    pub init_thrd_nx_chan_hdls: bool,
```
tiagolobocastro: This name is a bit misleading, what the bool seems to do is for all channels to be io_channels.
Maybe something like
`enable_io_all_nexus_channels`

io-engine/src/bin/io-engine.rs
tiagolobocastro: Btw, is it not enough to do this on the init stage?

io-engine/tests/snapshot_lvol.rs
```diff
@@ -48,7 +48,12 @@ static POOL_DEVICE_NAME: &str = "aio:///tmp/disk1.img";
 static LVOL_SIZE: u64 = 24 * 1024 * 1024;
 /// Get the global Mayastor test suite instance.
 fn get_ms() -> &'static MayastorTest<'static> {
-    MAYASTOR.get_or_init(|| MayastorTest::new(MayastorCliArgs::default()))
+    MAYASTOR.get_or_init(|| {
+        MayastorTest::new(MayastorCliArgs {
+            init_thrd_nx_chan_hdls: true,
```
tiagolobocastro: Needs to be enabled for at least casperf binary.
Might want to consider making true the default, and false only for io_engine binary, would result in less changes overall.

io-engine/src/bdev/nexus/nexus_channel.rs
dsharma-dc: yes, `debug` will be enough. I keep `info` during testing only as that's the default.

io-engine/src/bin/io-engine.rs
dsharma-dc: > Btw, is it not enough to do this on the init stage?

It should be enough, yes. Earlier I was only setting from inside `start_tokio_runtime()` and it wasn't reflecting in tests, because test is seting up environment via MayastorTest object directly.

> Why do we need to set this again if we've set it on L75?

Removing the changes from this function. Don't we need to handle env variable and cli arg both? I can probably remove env variable dependency completely from `structopt` for this option.

io-engine/tests/snapshot_lvol.rs
dsharma-dc: Will enable for casperf.
I've tried to keep default that favours product code, rather than test code.

dsharma-dc:
bors try

io-engine/src/bin/io-engine.rs
tiagolobocastro: Ah nice if you can do this by only changing the `MayastorEnvironment` and not cli arg I think that's even better

dsharma-dc: Need to modify `MayastorTest` constructor like this. This will become a blanket setting `true` for all tests though, less flexible(should be ok? wdyt @tiagolobocastro ). Changing the signature of `MayastorTest` to take `MayastorEnvironment` won't be desirable as part of this change I'd say.

```            
@@ -75,7 +75,9 @@ impl<'a> MayastorTest<'a> {
         let thdl = std::thread::Builder::new()
             .name("mayastor_master".into()) 
           .spawn(move || {
-                MayastorEnvironment::new(args).init();
+                let mut ms_env = MayastorEnvironment::new(args);
+                ms_env.set_io_all_thrd_nexus_channels(true);
+                ms_env.init();
```

tiagolobocastro: Seems quite verbose though, then I'd suggest keep as is in cli args, hidden.

dsharma-dc:
bors try

io-engine/src/bdev/nexus/nexus_channel.rs
dsharma-dc: @tiagolobocastro  I'd say let's keep it as it's now in the PR, which is pretty reliable.

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -77,26 +80,45 @@ impl<'n> NexusChannel<'n> {
...
+            // created in rebuild path today, and it's on the init
+            // thread. The channels that we use for normal nexus IO
+            // are not on init thread however, those are on spdk threads
+            // created for nvmf target during poll group init. Those are the
+            // spdk threads named mayastor_nvmf_tcp_pg_core_*. The
+            // channel on init thread is only used for rebuild IOs.
+            // And the rebuild IOs are dispatched by
+            // directly calling write API without going via writers abstraction.
+            // Refer GTM-1075 for the race condition details.
+            debug!("{nexus:?}: skip nexus channel setup({}). is_io_channel: {is_io_chan}", Thread::current().unwrap().name());
```
dsavitskiy: You could use Thread's debug formatter instead of name(). It prints core and thread name.
`debug!("{nexus:?}: skip nexus channel setup({t:?}). is_io_channel: {is_io_chan}", t = Thread::current().unwrap());`

dsavitskiy (approved):

tiagolobocastro (approved):

io-engine/src/bdev/nexus/nexus_channel.rs
tiagolobocastro: Seems still not entirely clean, but if you're happy with the testing then it works for now.

dsharma-dc:
bors merge

PR #1512, opened 2023-09-22
feat(pool): blobstore cluster size to be set during pool creation through grpc parameter value

hrudaya21:
- The default cluster size considered by SPDK is 4MiB.
- During pool creation, cluster_size can be taken from the gRPC and consider that value.
- The cluster_size can be consider with multiple of 1MiB.
- If the cluster_size passed not multiple of 1MiB, then default 4MiB cluster size will be considered for the pool creation.

### DIFF
--- a/io-engine/src/bin/io-engine-client/v1/pool_cli.rs
+++ b/io-engine/src/bin/io-engine-client/v1/pool_cli.rs
@@ -1,5 +1,6 @@
 use crate::{
     context::{Context, OutputFormat},
+    parse_size,
     ClientError,
     GrpcStatus,
 };
@@ -25,6 +26,13 @@ pub fn subcommands<'a, 'b>() -> App<'a, 'b> {
                 .required(false)
                 .takes_value(true)
                 .help("Storage pool uuid"),
+        )
+        .arg(
+            Arg::with_name("cluster-size")
+                .long("cluster-size")
+                .required(false)
+                .takes_value(true)
+                .help("SPDK cluster size"),
         )
         .arg(
             Arg::with_name("disk")
@@ -124,7 +132,10 @@ async fn create(
         })?
         .map(|dev| dev.to_owned())
         .collect();
-
+    let cluster_size =
+        parse_size(matches.value_of("cluster-size").unwrap_or_default())
+            .map_err(|s| Status::invalid_argument(format!("Bad size '{s}'")))
+            .context(GrpcStatus)?;
     let response = ctx
         .v1
         .pool
@@ -133,6 +144,7 @@ async fn create(
             uuid: uuid.map(ToString::to_string),
             disks: disks_list,
             pooltype: v1rpc::pool::PoolType::Lvs as i32,
+            cluster_size: Some(cluster_size.get_bytes() as u32),
         })
         .await
         .context(GrpcStatus)?;
--- a/io-engine/src/lvs/lvs_store.rs
+++ b/io-engine/src/lvs/lvs_store.rs
@@ -56,6 +56,12 @@ use events_api::event::EventAction;
 
 use crate::eventing::Event;
 
+static ROUND_TO_MB: u32 = 1024 * 1024;
+/// Default spdk cluster size is 4MiB.
+static DEFAULT_CLUSTER_SIZE: u32 = 4 * 1024 * 1024;
+/// Maximum spdk cluster size can be considered as 1GiB.
+static MAX_CLUSTER_SIZE: u32 = 1024 * 1024 * 1024;
+
 impl Debug for Lvs {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         write!(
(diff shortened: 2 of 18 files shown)


hrudaya21:
Do We need to have any checks in the dataplane to limit the maximum cluster size, to ensure this field is not exploit with very high value in gRPC request ?

io-engine/src/lvs/lvs_store.rs
```diff
@@ -391,10 +393,19 @@ impl Lvs {
...
+        cluster_size: Option<u32>,
     ) -> Result<Lvs, Error> {
         let pool_name = name.into_cstring();
         let bdev_name = bdev.into_cstring();
-
+        let cluster_size = if let Some(cluster_size) = cluster_size {
+            if cluster_size % ROUND_TO_MB == 0 {
+                cluster_size
+            } else {
+                0
```
dsavitskiy: An error must be returned in this case, instead of silently ignoring an incorrect size.

dsavitskiy (changes requested):

dsavitskiy:
> Do We need to have any checks in the dataplane to limit the maximum cluster size, to ensure this field is not exploit with very high value in gRPC request ?

Yes indeed, we have to. Probably something around 64Mb~128Mb ? @tiagolobocastro ?

io-engine/src/lvs/lvs_store.rs
```diff
@@ -391,10 +393,19 @@ impl Lvs {
         name: &str,
         bdev: &str,
         uuid: Option<String>,
+        cluster_size: Option<u32>,
     ) -> Result<Lvs, Error> {
         let pool_name = name.into_cstring();
         let bdev_name = bdev.into_cstring();
-
+        let cluster_size = if let Some(cluster_size) = cluster_size {
```
dsavitskiy, line 412: @hrudaya21 as you suggested, please add max limit for cluster size.

dsavitskiy (changes requested):

tiagolobocastro:
> > Do We need to have any checks in the dataplane to limit the maximum cluster size, to ensure this field is not exploit with very high value in gRPC request ?
> 
> Yes indeed, we have to. Probably something around 64Mb~128Mb ? @tiagolobocastro ?

IIRC the default within the blobstore is 1GiB, maybe that could be the max, or is it probably too much?

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
```diff
@@ -26,6 +26,13 @@ pub fn subcommands<'a, 'b>() -> App<'a, 'b> {
                 .takes_value(true)
                 .help("Storage pool uuid"),
         )
+        .arg(
+            Arg::with_name("cluster_size")
+                .long("cluster_size")
```
tiagolobocastro:
```suggestion
            Arg::with_name("cluster-size")
                .long("cluster-size")
```

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
```diff
@@ -124,7 +131,9 @@ async fn create(
         })?
         .map(|dev| dev.to_owned())
         .collect();
-
+    let cluster_size = matches
+        .value_of("cluster_size")
+        .map(|x| x.parse::<u32>().unwrap_or_default());
```
tiagolobocastro:
```suggestion
    let cluster_size = parse_size(matches.value_of("cluster-size").unwrap_or_default())
        .map_err(|s| Status::invalid_argument(format!("Bad size '{s}'")))
        .context(GrpcStatus)?;
```

io-engine/src/lvs/lvs_store.rs
```diff
@@ -391,10 +393,19 @@ impl Lvs {
...
+            if cluster_size % ROUND_TO_MB == 0 {
+                cluster_size
+            } else {
+                0
+            }
+        } else {
+            0
```
tiagolobocastro: Rather than zero should we explicitly set 4MiB to make it clearer?

hrudaya21:
> > > Do We need to have any checks in the dataplane to limit the maximum cluster size, to ensure this field is not exploit with very high value in gRPC request ?
> > 
> > 
> > Yes indeed, we have to. Probably something around 64Mb~128Mb ? @tiagolobocastro ?
> 
> IIRC the default within the blobstore is 1GiB, maybe that could be the max, or is it probably too much?

I think We can keep it 1GiB. As per the SPDK documentation, If the BlobStore Size is multiple of GBs, then cluster size can be 1GiB.

io-engine/src/lvs/lvs_store.rs
hrudaya21: Code fixed.

io-engine/src/lvs/lvs_store.rs
hrudaya21: Kept the max cluster size limit to 1GiB. Based on future usage, We can change the macro value.

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
hrudaya21: fixed.

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
hrudaya21: fixed.

io-engine/src/lvs/lvs_store.rs
hrudaya21: code changed.

dsavitskiy (approved):

tiagolobocastro (approved):

hrudaya21:
bors try

hrudaya21:
bors try

hrudaya21:
bors merge

hrudaya21:
bors merge

PR #1513, opened 2023-09-25
feat(eventing): Replica events

datacore-vvarakantham:
Added replica creation and deletion events.
And Replica event specific data(pool name, pool uuid, replica name).

### DIFF
--- a/io-engine/src/lvs/lvs_lvol.rs
+++ b/io-engine/src/lvs/lvs_lvol.rs
@@ -1,6 +1,7 @@
 use async_trait::async_trait;
 use byte_unit::Byte;
 use chrono::Utc;
+use events_api::event::EventAction;
 use futures::channel::oneshot;
 use nix::errno::Errno;
 use pin_utils::core_reexport::fmt::Formatter;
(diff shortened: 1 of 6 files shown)


io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -1007,7 +1009,8 @@ impl LvsLvol for Lvol {
     async fn destroy_replica(mut self) -> Result<String, Error> {
         let snapshot_lvol = self.is_snapshot_clone();
         let name = self.name();
-        self.destroy().await?;
+        self.clone().destroy().await?;
```
tiagolobocastro: hmm not sure if it's safe to clone here as we're deleting it, we might then later access memory which is now freed?
Might be better to put event on var first, and then after delete here then just call generate?

datacore-vvarakantham: Followed this now.

Also changed replica creation and deletion events to when "lvol" is created and deleted.

tiagolobocastro (approved):

chriswldenyer (approved):
LGTM

datacore-vvarakantham:
bors merge

PR #1524, opened 2023-10-13
feat(nvmf): selection different CRD for replicas and certain nexus error

dsavitskiy:
NVMF subsystem will select different preconfigured CRDs for certain cases:
3 is selected for replica targets
2 is selected for ENOSPC and reservation errors on nexus targets
1 is select for all other error on nexus target


dsavitskiy:
bors try

dsavitskiy:
bors try

dsavitskiy:
bors try

dsavitskiy:
bors try

dsavitskiy:
bors try

dsavitskiy:
bors try

dsavitskiy:
bors try

dsavitskiy:
bors try

io-engine/src/core/env.rs
```diff
@@ -87,6 +88,38 @@ fn parse_ps_timeout(src: &str) -> Result<Duration, String> {
...
+        for i in &mut u {
+            *i = parse_val(items.next().unwrap())?;
+        }
+
+        Ok(u)
+    } else {
+        Err("Command Retry Delay argument must be a interger or \
+            a comma-separated list of three intergers"
+            .to_string())
+    }
```
tiagolobocastro: What about using slices here?

```suggestion
    match items.as_slice() {
        [ one ] => Ok([parse_val(one)?, 0, 0]),
        [ one, two, three ] =>  Ok([parse_val(one)?, parse_val(two)?, parse_val(three)? ]),
        _ => {
            Err("Command Retry Delay argument must be a interger or \
                a comma-separated list of three intergers"
                .to_string())
        }
```

io-engine/src/core/env.rs
```diff
@@ -163,9 +196,15 @@ pub struct MayastorCliArgs {
     #[clap(short = 'T', long = "tgt-iface", env = "NVMF_TGT_IFACE")]
     /// NVMF target interface (ip, mac, name or subnet).
     pub nvmf_tgt_interface: Option<String>,
-    /// NVMF target Command Retry Delay.
-    #[clap(long = "tgt-crdt", env = "NVMF_TGT_CRDT", default_value = "0")]
-    pub nvmf_tgt_crdt: u16,
+    /// NVMF target Command Retry Delay in x100 ms (single integer or three
```
tiagolobocastro, line 192: Ah interesting, is this not seconds? It's documented as seconds in the helm chart and defaults to 30.

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -259,6 +288,99 @@ impl NvmfSubsystem {
...
+    unsafe extern "C" fn nexus_cpl_error_cb(
+        req: *mut spdk_nvmf_request,
+        _cb_arg: *mut ::std::os::raw::c_void,
+    ) {
+        let req = &mut *req;
+        let cpl = req.nvme_cpl_mut();
+        let mut status = cpl.status();
+
+        if status.crd() == 0 {
+            return;
```
tiagolobocastro, line 301: This means we don't set crdt? So it's disabled?

io-engine/src/core/env.rs
```diff
@@ -163,9 +196,15 @@ pub struct MayastorCliArgs {
...
+    /// comma-separated integers).
+    #[clap(
+        long = "tgt-crdt",
+        env = "NVMF_TGT_CRDT",
+        default_value = "0",
+        value_parser = parse_crdt,
+    )]
+    pub nvmf_tgt_crdt: [u16; 3],
```
tiagolobocastro: So, if I understand this correctly, the first 2 are for the nexus and the last one for the replicas?
Would be good to document this here as well, including when they are used.
Also, could we have completely separate  crdt array for nexus and replica?

io-engine/src/bdev/nexus/nexus_io.rs
```diff
@@ -270,10 +281,37 @@ impl<'n> NexusBio<'n> {
...
+                self.nexus_mut().get_unchecked_mut().last_error = status;
+            }
+
             self.fail();
         }
     }
 
+    /// Fails the current I/O with a generic internal error. If the nexus
+    /// already had a last child error, it fails with it.
+    fn fail(&self) {
```
dsharma-dc, line 295: Could this be problematic for thin provisioning cases @tiagolobocastro @dsavitskiy . Say, first a replica fails with some nvme error, and then a replica fails the same IO with ENOSPC. This function will make the nexus IO fail with the nvme error because that happened first.

io-engine/src/core/env.rs
```diff
@@ -391,7 +430,7 @@ impl Default for MayastorEnvironment {
             bdev_io_ctx_pool_size: 65535,
             nvme_ctl_io_ctx_pool_size: 65535,
             nvmf_tgt_interface: None,
-            nvmf_tgt_crdt: 0,
+            nvmf_tgt_crdt: [0; 3],
```
dsharma-dc: For the sake of readability, can we #define the value 3 to a suitable macro name.

io-engine/src/core/env.rs
dsavitskiy: ah, it never occurred to me to use match. I'll try it.

io-engine/src/core/env.rs
dsavitskiy: It is not seconds, it is x100 ms. In Linux initiator code, it is multiplied by 100; and also I measured it to be sure.

io-engine/src/subsys/nvmf/subsystem.rs
dsavitskiy: Either it was not configured by us, or Do Not Retry flag was set, or the host did not enable ACRE when target was configured. So we just stick with the flags SPDK sets in this case.

io-engine/src/core/env.rs
dsavitskiy: So it is configured differently for two targets? Not at the moment, as we have one target, listening on two ports. CRD is configured per target, not per port.

io-engine/src/core/env.rs
dsavitskiy: Yes sure; I'll change to a const.

io-engine/src/bdev/nexus/nexus_io.rs
dsavitskiy: It's the last I/O error which is stored in Nexus, not the first. It is only stored when all children fails, and therefore the nexus is already in faulty / unrecoverable state.

io-engine/src/core/env.rs
dsavitskiy: Add documentation comments as well.

dsavitskiy:
bors try

io-engine/src/core/env.rs
```diff
@@ -87,6 +89,30 @@ fn parse_ps_timeout(src: &str) -> Result<Duration, String> {
...
+    }
+
+    let items = src.split(',').collect::<Vec<&str>>();
+    match items.as_slice() {
+        [one] => Ok([parse_val(one)?, 0, 0]),
+        [one, two, three] => {
+            Ok([parse_val(one)?, parse_val(two)?, parse_val(three)?])
+        }
+        _ => Err("Command Retry Delay argument must be a interger or \
+                  a comma-separated list of three intergers"
```
```suggestion
        _ => Err("Command Retry Delay argument must be an integer or \
                  a comma-separated list of three integers"
```

tiagolobocastro (approved):

dsharma-dc (approved):

dsavitskiy:
bors merge

PR #1548, opened 2023-11-23
nvme/opts: parse timeouts in humantime

tiagolobocastro:
ci(bors): add fake DCO for bors merge requests
    
    DCO is already validated on PR and bors itself does not need it as it's a bot..
    
---

    fix(nvme/opts): remove outdated async parameter
    
    Remove unused NVME_QPAIR_CONNECT_ASYNC opts parameter and add
    generated uuid parameter.
    
---

    fix(nvme/kato): adjust to spdk minimum
    
    SPDK minimum KATO is 10s, so setting 1s is pointless.
    Also the spec does suggest setting a sufficiently high KATO on a fabric, so perhaps
    we might want to consider increasing it in the future.
    
---

    fix(iobuf): increase large iobuf pool size
    
    Addresses warning from spdk itself:
    mayastor::spdk:transport.c:282]
    The num_shared_buffers value (2048) is larger than the available iobuf pool size (1024).
     Please increase the iobuf pool sizes.

    Also reduce size to power of 2 - 1 as recommended in spdk ticket:
    It’s always best to specify a value for –n that is one less than a power of two (2^x -1)
    because of the way rings work in DPDK. You end up allocating a ring twice the size for
    the one extra element between 511 and 512 for example.
    
---

    chore: silence flush not supported IO
    
    We don't keep any caching, so we don't actually need to make use of it anyway.
    TODO: do we need to even send flush?
    
---

    chore: show starting configuration as info
    
---

    feat(nvme/opts): parse timeouts in humantime
    
    This allows us to specify, for example, NVME_TIMEOUT=60s
    
### DIFF
--- a/io-engine/src/subsys/config/mod.rs
+++ b/io-engine/src/subsys/config/mod.rs
@@ -227,6 +227,6 @@ impl Config {
         assert!(self.socket_opts.set());
         assert!(self.iobuf_opts.set());
 
-        debug!("{:#?}", self);
+        info!("{:#?}", self);
     }
 }
--- a/io-engine/src/subsys/config/opts.rs
+++ b/io-engine/src/subsys/config/opts.rs
@@ -375,7 +375,7 @@ impl Default for NvmeBdevOpts {
             ),
             keep_alive_timeout_ms: time_try_from_env(
                 "NVME_KATO",
-                1_000,
+                10_000,
                 TimeUnit::MilliSeconds,
             ),
             transport_retry_count: try_from_env("NVME_RETRY_COUNT", 0),
--- a/io-engine/src/subsys/config/opts.rs
+++ b/io-engine/src/subsys/config/opts.rs
@@ -334,8 +334,11 @@ pub struct NvmeBdevOpts {
     pub fast_io_fail_timeout_sec: u32,
     /// TODO
     pub disable_auto_failback: bool,
-    /// enable creation of submission and completion queues asynchronously.
-    pub async_mode: bool,
+    /// Enable generation of unique identifiers for NVMe bdevs only if they
+    /// do not provide UUID themselves.
+    /// These strings are based on device serial number and namespace ID and
+    ///  will always be the same for that device.
+    pub generate_uuids: bool,
 }
 
 impl GetOpts for NvmeBdevOpts {
@@ -401,7 +404,7 @@ impl Default for NvmeBdevOpts {
             reconnect_delay_sec: 0,
             fast_io_fail_timeout_sec: 0,
             disable_auto_failback: false,
-            async_mode: try_from_env("NVME_QPAIR_CONNECT_ASYNC", false),
+            generate_uuids: try_from_env("NVME_GENERATE_UUIDS", true),
         }
     }
 }
@@ -428,7 +431,7 @@ impl From<spdk_bdev_nvme_opts> for NvmeBdevOpts {
             reconnect_delay_sec: o.reconnect_delay_sec,
             fast_io_fail_timeout_sec: o.fast_io_fail_timeout_sec,
             disable_auto_failback: o.disable_auto_failback,
-            async_mode: NvmeBdevOpts::default().async_mode,
+            generate_uuids: o.generate_uuids,
         }
     }
 }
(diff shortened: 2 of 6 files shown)


io-engine/src/subsys/config/mod.rs
```diff
@@ -227,6 +227,6 @@ impl Config {
         assert!(self.socket_opts.set());
         assert!(self.iobuf_opts.set());
 
-        debug!("{:#?}", self);
+        info!("{:#?}", self);
```
Abhinandan-Purkait, line 230: Does this need to be info?

Abhinandan-Purkait (approved):
LGTM

io-engine/src/subsys/config/opts.rs
```diff
@@ -174,6 +175,74 @@ where
...
+                }
+                Err(error) => Err(error.to_string()),
+            };
+            match result {
+                Ok(value) => {
+                    info!("Overriding {} value to '{}'", name, value);
+                    value
+                }
+                Err(e) => {
+                    error!("Invalid value: {} (error {}) specified for {}. Reverting to default ({}{})", value, e, name, default, unit.units());
```
Abhinandan-Purkait: `Reverting to default value` ? Since we said the word value after Invalid

io-engine/src/subsys/config/mod.rs
tiagolobocastro: Yes, otherwise debugging SB we don't see the full config by default

io-engine/src/subsys/config/opts.rs
tiagolobocastro: It follows existing message above (not shown on diff)
I'll change it for both

io-engine/src/subsys/config/opts.rs
```diff
@@ -668,7 +668,7 @@ impl Default for IoBufOpts {
     fn default() -> Self {
         Self {
             small_pool_count: try_from_env("IOBUF_SMALL_POOL_COUNT", 8192),
-            large_pool_count: try_from_env("IOBUF_LARGE_POOL_COUNT", 1024),
+            large_pool_count: try_from_env("IOBUF_LARGE_POOL_COUNT", 2048),
```
dsharma-dc, line 674: This config might be performance sensitive. I'd recommend doing a sanity perf run with and without this change. Also, sometime back I was reading spdk issue 809, where it says 
> It’s always best to specify a value for –n that is one less than a power of two (2^x -1) because of the way rings work in DPDK. You end up allocating a ring twice the size for the one extra element between 511 and 512 for example.

So along with this, we can consider changing NVMF_TCP_NUM_SHARED_BUF from 2048 to 2047 as well.

io-engine/src/subsys/config/opts.rs
```diff
@@ -668,7 +668,7 @@ impl Default for IoBufOpts {
     fn default() -> Self {
         Self {
             small_pool_count: try_from_env("IOBUF_SMALL_POOL_COUNT", 8192),
-            large_pool_count: try_from_env("IOBUF_LARGE_POOL_COUNT", 1024),
+            large_pool_count: try_from_env("IOBUF_LARGE_POOL_COUNT", 2047),
```
dsharma-dc: Sorry if it wasn't clear from my earlier comment, but the -1 is to be done only for num_shared_buffers. For iobuf pool count it can remain as it is, the spdk defaults are full value too:
> #define IOBUF_DEFAULT_SMALL_POOL_SIZE	8192
#define IOBUF_DEFAULT_LARGE_POOL_SIZE	1024

The transport shared buffers is what's needed to be at -1

> #define SPDK_NVMF_TCP_DEFAULT_NUM_SHARED_BUFFERS 511

tiagolobocastro: Right yes, done

io-engine/src/subsys/config/opts.rs
tiagolobocastro: Changed, and for my limited testing I see no perf impact.

dsharma-dc (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
Had to enable DCO from any source...
bors merge

PR #1549, opened 2023-11-27
Fixing I/O engine crashes

dsavitskiy:


dsavitskiy:
bors try

io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -1143,6 +1143,19 @@ impl<'n> Nexus<'n> {
         unsafe { Pin::new_unchecked(self.bdev_mut()) }
     }
 
+    /// Gets a nexus reference from an untyped bdev.
+    /// No checks are performed (e.g. bdev module name check), as it is assumed
```
tiagolobocastro, line 1148: Might be worth having this under a '/// # Warning: ' ?

io-engine/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -974,19 +975,29 @@ impl<'n> Nexus<'n> {
...
+        self.detach_device(&dev).await;
 
-        // Destroy (close) the device. The subsystem must be paused to do this
-        // properly.
+        // Disconnect, destroy and close the device. The subsystem must be
+        // paused to do this properly.
         {
             debug!("{self:?}: retire: pausing...");
-            self.as_mut().pause().await?;
+            let res = self.as_mut().pause().await;
```
tiagolobocastro, line 1000: need to change the log below because pausing was not ok?
match rest {
Ok(x) => pause ok,
Err(error) => pausing error
}

ci.nix
```diff
@@ -1,4 +1,4 @@
-{ nospdk ? false, norust ? false, spdk_rel ? false }:
+{ nospdk ? false, norust ? false, spdk_rel ? false, asan ? false }:
```
tiagolobocastro, line 1: missing equivalent changes on shell.nix?

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -233,57 +235,36 @@ impl NvmfSubsystem {
...
-                } else {
-                    subsystem.host_disconnect_replica(ctrlr);
-                }
-            }
-            NvmfSubsystemEvent::HostKeepAliveTimeout(ctrlr) => {
+        debug!("NVMF subsystem event {s:?}: {event:?}");
+
+        use NvmfSubsystemEvent::*;
+
+        match NqnTarget::lookup(&s.get_nqn()) {
```
tiagolobocastro: @datacore-vvarakantham I think Host events might be useful to add to the event bus?

nix/lib/rust.nix
```diff
@@ -6,4 +6,5 @@ in
 with pkgs; rec  {
   nightly = rust-bin.nightly."2023-08-25".default;
   stable = rust-bin.stable."1.72.0".default;
+  asan = rust-bin.nightly."2023-08-25".default;
```
tiagolobocastro: would it be worth keeping nightly and asan the same date, ex:
```
let
  pkgs =...
  nightly_version = "2023-08-25";
  stable_version = "1.72.0";
in ...
```

tiagolobocastro (approved):

io-engine/src/subsys/nvmf/subsystem.rs
datacore-vvarakantham: Yes. Should I add the category "Host" and actions "Connect", "Disconnect", "KeepAliveTimeout" ?

tiagolobocastro: Maybe `HostInitiator`? Otherwise sounds good to me

datacore-vvarakantham: Ok.

dsavitskiy:
bors try

ci.nix
dsavitskiy: shell.nix uses the system Rust, so it does not make sense to allow ASAN for it. ASAN forces a nightly Rust of a specific version.

dsavitskiy:
bors try

io-engine/src/bdev/nexus/nexus_bdev_children.rs
```diff
@@ -326,9 +326,10 @@ impl<'n> Nexus<'n> {
         // Close and remove the child.
         let res = match self.lookup_child(uri) {
             Some(child) => {
-                // Remove child from the I/O path.
+                // Detach the child from the I/O path, and close its handles.
                 if let Some(device) = child.get_device_name() {
-                    self.disconnect_device(&device).await;
+                    self.detach_device(&device).await;
+                    self.disconnect_all_detached_devices().await;
```
tiagolobocastro, line 332: Is there a race here, between pause and remove_child?

io-engine/src/bdev/nexus/nexus_io.rs
```diff
@@ -642,7 +642,8 @@ impl<'n> NexusBio<'n> {
             // set the IO as failed in the submission stage.
             self.ctx_mut().failed += 1;
 
-            self.channel_mut().disconnect_device(&device);
+            self.channel_mut().detach_device(&device);
+            self.channel_mut().disconnect_detached_devices(|_| true);
```
tiagolobocastro, line 646: hmm interesting, why do we disconnect here when we fail to submit?

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -209,16 +211,66 @@ impl<'n> NexusChannel<'n> {
...
+    }
+
+    /// Disconnects previously detached device handles by dropping them.
+    /// Devices to drop are filtered by the given predicate: true to drop
+    /// a device, false to keep it.
+    pub(super) fn disconnect_detached_devices<F>(&mut self, mut drop_pred: F)
+    where
+        F: FnMut(&dyn BlockDeviceHandle) -> bool,
+    {
+        let cnt = self
```
dsharma-dc: What's the reason we pass a closure here(always returning `true`), then filter to count? and not simply check vector's length? In other words, when do we want to disconnect selectively from `detached` list?

io-engine/src/bdev/nexus/nexus_io.rs
dsavitskiy: A subsequent I/O may succeed, e.g. a different I/O type or a failure is flickering.

io-engine/src/bdev/nexus/nexus_channel.rs
dsavitskiy: For debugging purposed only. I want to print a nice log messages here, with number of device to drop, devices remaining, etc.

io-engine/src/bdev/nexus/nexus_bdev_children.rs
dsavitskiy: IIRC we discussed with place: we are removing device which is not failed. Pause is done before disconnect, and will eventually succeed.

io-engine/src/bdev/nexus/nexus_channel.rs
dsharma-dc: Perhaps we should do minimal required here because this path is latency sensitive i.e. we'd want the device retire to finish as soon as possible and not  hold the pause longer than needed.

dsavitskiy: Ah, I just found a better way to report and avoid counting, I will print "before len" - "after len".

dsavitskiy:
bors try

dsharma-dc (approved):

io-engine/src/bdev/nexus/nexus_io.rs
tiagolobocastro: But isn't this bypassing the whole fault_child logic? We're removing the device before we persist to etcd?

dsavitskiy: It was like this before. I created a ticket to investigate and address this potential problem.

dsavitskiy:
bors merge

PR #1550, opened 2023-11-28
feat(replica/resize): add support for resizing a replica

dsharma-dc:
Added the support for resizing a replica. The replicas need to be expanded as part of the volume expand workflow.

### DIFF
--- a/io-engine/src/bin/io-engine-client/v1/replica_cli.rs
+++ b/io-engine/src/bin/io-engine-client/v1/replica_cli.rs
@@ -32,7 +32,6 @@ pub fn subcommands() -> Command {
             Arg::new("size")
                 .short('s')
                 .long("size")
-
                 .required(true)
                 .value_name("NUMBER")
                 .help("Size of the replica"))
@@ -107,6 +106,20 @@ pub fn subcommands() -> Command {
             .index(1)
             .help("Replica uuid"),
     );
+    let resize = Command::new("resize")
+        .about("Resize replica")
+        .arg(
+            Arg::new("uuid")
+                .required(true)
+                .index(1)
+                .help("Replica uuid"),
+        )
+        .arg(
+            Arg::new("size")
+                .required(true)
+                .index(2)
+                .help("Requested new size of the replica"),
+        );
     Command::new("replica")
         .subcommand_required(true)
         .arg_required_else_help(true)
@@ -115,6 +128,7 @@ pub fn subcommands() -> Command {
         .subcommand(destroy)
         .subcommand(share)
         .subcommand(unshare)
+        .subcommand(resize)
         .subcommand(Command::new("list").about("List replicas"))
         .subcommand(Command::new("stats").about("IO stats of replicas"))
 }
@@ -126,6 +140,7 @@ pub async fn handler(ctx: Context, matches: &ArgMatches) -> crate::Result<()> {
         ("list", args) => replica_list(ctx, args).await,
         ("share", args) => replica_share(ctx, args).await,
         ("unshare", args) => replica_unshare(ctx, args).await,
+        ("resize", args) => replica_resize(ctx, args).await,
         ("stats", args) => replica_stat(ctx, args).await,
         (cmd, _) => {
             Err(Status::not_found(format!("command {cmd} does not exist")))
@@ -408,6 +423,46 @@ async fn replica_unshare(
             println!("{}", &response.get_ref().uri);
         }
     };
+
+    Ok(())
+}
+
+async fn replica_resize(
+    mut ctx: Context,
+    matches: &ArgMatches,
+) -> crate::Result<()> {
+    let uuid = matches
+        .get_one::<String>("uuid")
+        .ok_or_else(|| ClientError::MissingValue {
+            field: "uuid".to_string(),
+        })?
+        .to_owned();
+
+    let requested_size =
+        parse_size(matches.get_one::<String>("size").ok_or_else(|| {
+            ClientError::MissingValue {
+                field: "size".to_string(),
+            }
+        })?)
+        .map_err(|s| Status::invalid_argument(format!("Bad size '{s}'")))
+        .context(GrpcStatus)?;
+
+    let _ = ctx
+        .v1
+        .replica
+        .resize_replica(v1_rpc::replica::ResizeReplicaRequest {
+            uuid: uuid.clone(),
+            requested_size: requested_size.get_bytes() as u64,
+        })
+        .await
+        .context(GrpcStatus)?;
+
+    match ctx.output {
+        OutputFormat::Json => {}
+        OutputFormat::Default => {
+            println!("replica {} is resized", &uuid);
+        }
+    }
 
     Ok(())
 }
--- a/io-engine/src/grpc/v0/mayastor_grpc.rs
+++ b/io-engine/src/grpc/v0/mayastor_grpc.rs
@@ -249,6 +249,18 @@ impl From<LvsError> for tonic::Status {
                 Errno::EMEDIUMTYPE => Status::aborted(e.to_string()),
                 _ => Status::internal(e.to_string()),
             },
+            LvsError::RepResize {
+                source, ..
+            } => match source {
+                Errno::ENOSPC | Errno::ENOMEM => {
+                    Status::resource_exhausted(e.to_string())
+                }
+                Errno::EPERM => Status::permission_denied(e.to_string()),
+                Errno::EINVAL | Errno::ENOENT => {
+                    Status::invalid_argument(e.to_string())
+                }
+                _ => Status::internal(e.to_string()),
+            },
             LvsError::RepExists {
                 ..
             } => Status::already_exists(e.to_string()),
--- a/io-engine/src/grpc/v1/replica.rs
+++ b/io-engine/src/grpc/v1/replica.rs
@@ -489,4 +489,36 @@ impl ReplicaRpc for ReplicaService {
         )
         .await
     }
+
+    #[named]
+    async fn resize_replica(
+        &self,
+        request: Request<ResizeReplicaRequest>,
+    ) -> GrpcResult<Replica> {
+        self.locked(
+            GrpcClientContext::new(&request, function_name!()),
+            async move {
+                let args = request.into_inner();
+                info!("{args:?}");
+                let rx = rpc_submit::<_, _, LvsError>(async move {
+                    let mut lvol = Bdev::lookup_by_uuid_str(&args.uuid)
+                        .and_then(|b| Lvol::try_from(b).ok())
+                        .ok_or(LvsError::RepResize {
+                            source: Errno::ENOENT,
+                            name: args.uuid.to_owned(),
+                        })?;
+                    let requested_size = args.requested_size;
+                    lvol.resize_replica(requested_size).await?;
+                    debug!("resized {:?}", lvol);
+                    Ok(Replica::from(lvol))
+                })?;
+
+                rx.await
+                    .map_err(|_| Status::cancelled("cancelled"))?
+                    .map_err(Status::from)
+                    .map(Response::new)
+            },
+        )
+        .await
+    }
 }
--- a/io-engine/src/lvs/lvs_error.rs
+++ b/io-engine/src/lvs/lvs_error.rs
@@ -86,6 +86,11 @@ pub enum Error {
         name: String,
         msg: String,
     },
+    #[snafu(display("failed to resize lvol {}", name))]
+    RepResize {
+        source: Errno,
+        name: String,
+    },
     #[snafu(display("bdev {} is not a lvol", name))]
     NotALvol {
         source: Errno,
@@ -216,6 +221,9 @@ impl ToErrno for Error {
                 source, ..
             } => source,
             Self::RepDestroy {
+                source, ..
+            } => source,
+            Self::RepResize {
                 source, ..
             } => source,
             Self::NotALvol {
--- a/io-engine/src/lvs/lvs_lvol.rs
+++ b/io-engine/src/lvs/lvs_lvol.rs
@@ -30,6 +30,7 @@ use spdk_rs::libspdk::{
     spdk_lvol,
     vbdev_lvol_destroy,
     vbdev_lvol_get_from_bdev,
+    vbdev_lvol_resize,
     LVS_CLEAR_WITH_UNMAP,
 };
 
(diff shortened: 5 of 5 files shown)


avishnu:
[like]  Vishnu Govind Attur reacted to your message:
________________________________
From: Diwakar Sharma ***@***.***>
Sent: Tuesday, November 28, 2023 7:35:12 AM
To: openebs/mayastor ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [openebs/mayastor] feat(replica/resize): add support for resizing a replica (PR #1550)


CAUTION: This email originated from outside of the organization. Do not click links or open attachments unless you recognize the sender and know the content is safe.


Added the support for resizing a replica. The replicas need to be expanded as part of the volume expand workflow.

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/openebs/mayastor/pull/1550

Commit Summary

  *   f4e48bf<https://github.com/openebs/mayastor/pull/1550/commits/f4e48bfa8e60f6031259b74bd8e65d86062a1a16> feat(replica/resize): add support for resizing a replica

File Changes

(4 files<https://github.com/openebs/mayastor/pull/1550/files>)

  *   M io-engine/src/bin/io-engine-client/v1/replica_cli.rs<https://github.com/openebs/mayastor/pull/1550/files#diff-8da37dd992f2a6259aa2fa5b85503fc139d65197968945ba2583f36caddb8e1a> (52)
  *   M io-engine/src/grpc/v1/replica.rs<https://github.com/openebs/mayastor/pull/1550/files#diff-4af8864bbb2f49ee3fd709d2343369f6116b2fbb7935f83d9f1508bb297f9a2e> (35)
  *   M io-engine/src/lvs/lvs_error.rs<https://github.com/openebs/mayastor/pull/1550/files#diff-6bb45dd51d417100dadde318aab207951252a4374558a37ee4ed6f74bb175bfe> (8)
  *   M io-engine/src/lvs/lvs_lvol.rs<https://github.com/openebs/mayastor/pull/1550/files#diff-6469ece4955c4578efcb3c20934d65fdd89740900e941e5fd9c986e9a0914720> (79)

Patch Links:

  *   https://github.com/openebs/mayastor/pull/1550.patch
  *   https://github.com/openebs/mayastor/pull/1550.diff

io-engine/src/grpc/v1/replica.rs
```diff
@@ -489,4 +489,37 @@ impl ReplicaRpc for ReplicaService {
...
+    #[named]
+    async fn resize_replica(
+        &self,
+        request: Request<ResizeReplicaRequest>,
+    ) -> GrpcResult<Replica> {
+        self.locked(
+            GrpcClientContext::new(&request, function_name!()),
+            async move {
+                let args = request.into_inner();
+                info!("{:?}", args);
```
dsavitskiy: As clippy suggests,  info!("{args:?}"); is more preferable.

io-engine/src/lvs/lvs_error.rs
```diff
@@ -86,6 +86,11 @@ pub enum Error {
         name: String,
         msg: String,
     },
+    #[snafu(display("failed to resize lvol {}", name))]
+    RepResize {
```
dsavitskiy, line 90: Rename to `ReplicaResize` for more clarity?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -1026,4 +1042,60 @@ impl LvsLvol for Lvol {
...
+                resize_to,
+                Some(lvol_resize_cb),
+                &mut ctx as *mut _ as *mut c_void,
+            );
+        }
+
+        let cb_ret = r.await.expect("lvol resize callback dropped");
+
+        if let Ok(_lv) = cb_ret {
+            info!("Resized {:?} successfully", self);
```
dsavitskiy: info!("Resized {self:?} successfully"); is now considered more preferable syntax.

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -114,6 +115,16 @@ impl Display for PropName {
     }
 }
 
+/// Resize context to be passed as callback args pointer to spdk.
```
dsavitskiy, line 118: We have `cb_arg` and `done_cb` generic helper functions in spdk-rs/src/ffihelper.rs to wrap/unwrap a sender alongside with any payload into an SPDK-style callback arg. We use them extensively. Probably they would work here as well.

dsavitskiy (approved):

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -594,6 +605,11 @@ pub trait LvsLvol: LogicalVolume + Share {
     /// Wrapper function to destroy replica and its associated snapshot if
     /// replica is identified as last clone.
     async fn destroy_replica(mut self) -> Result<String, Error>;
+
+    /// Resize a replica. The resize can be expand or shrink, depending
+    /// upon if required size is more or less than current size of
+    /// the replpica.
+    async fn resize_replica(mut self, resize_to: u64) -> Result<Lvol, Error>;
```
tiagolobocastro: Why move and return Lvol?

io-engine/src/lvs/lvs_error.rs
tiagolobocastro: You might have done it already but please ensure error codes are converted to grpc...

io-engine/src/bin/io-engine-client/v1/replica_cli.rs
```diff
@@ -412,6 +427,41 @@ async fn replica_unshare(
...
 }
 
+async fn replica_resize(
+    mut ctx: Context,
+    matches: &ArgMatches,
+) -> crate::Result<()> {
+    let uuid = matches.get_one::<String>("uuid").unwrap().to_owned();
+
+    let requested_size =
+        parse_size(matches.get_one::<String>("size").ok_or_else(|| {
```
tiagolobocastro, line 442: can probably unwrap here same as above since size is a required parameter?

io-engine/src/bin/io-engine-client/v1/replica_cli.rs
```diff
@@ -412,6 +427,41 @@ async fn replica_unshare(
...
+            uuid: uuid.clone(),
+            requested_size: requested_size.get_bytes() as u64,
+        })
+        .await
+        .context(GrpcStatus)?;
+
+    match ctx.output {
+        OutputFormat::Json => {}
+        OutputFormat::Default => {
+            println!("replica: {} is resized", &uuid);
```
tiagolobocastro: `:` seems a bit awkward?

```suggestion
            println!("replica {} is resized", &uuid);
```

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -1026,4 +1042,60 @@ impl LvsLvol for Lvol {
...
+            info!("Resized {:?} successfully", self);
+            Ok(Lvol::from_inner_ptr(self.as_inner_ptr()))
+        } else {
+            let status = cb_ret.unwrap_err();
+            error!("Resize {:?} failed, errno {}", self, status);
+            Err(Error::RepResize {
+                source: status,
+                name: self.name(),
+            })
+        }
```
tiagolobocastro: Use match instead of if else

```suggestion
        match cb_ret {
            Ok(_) => {
                info!("Resized {self:?} successfully");
                Ok(Lvol::from_inner_ptr(self.as_inner_ptr()))
            }
            Err(errno) => {
                error!("Resize {self:?} failed, errno {errno}");
                Err(Error::RepResize {
                    source: errno,
                    name: self.name(),
                })
            }
        }
```

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -1026,4 +1042,60 @@ impl LvsLvol for Lvol {
...
+    let sender = unsafe {
+        Box::from_raw(
+            (*ctx).sender as *mut oneshot::Sender<ErrnoResult<*mut spdk_lvol>>,
+        )
+    };
+
+    if errno == 0 {
+        // Make sure resize worked, and account for metadata while comparing
+        // i.e. the actual size will be a little more than requested.
+        assert!(lvol.size() >= req_size);
```
tiagolobocastro: hmm seems like we should return error rather than crashing?

io-engine/src/lvs/lvs_error.rs
dsharma-dc: > Rename to `ReplicaResize` for more clarity?

All the existing codes above this are using `Rep` so I've kept it consistent with that.

io-engine/src/lvs/lvs_lvol.rs
dsharma-dc: What would be a more conforming way? Perhaps in this function we can take a `&self` and return only `Result<(), Error)` ?

io-engine/src/lvs/lvs_lvol.rs
dsharma-dc: This kind of situation might have a potential for leading to data integrity issues, hence marked assert. What would be the most severe error that we can return back here? We'll have to overwrite errno.

io-engine/src/bin/io-engine-client/v1/replica_cli.rs
dsharma-dc: Followed from `create` so we can return a clear error :)

io-engine/src/lvs/lvs_lvol.rs
dsharma-dc: > We have `cb_arg` and `done_cb` generic helper functions in spdk-rs/src/ffihelper.rs to wrap/unwrap a sender alongside with any payload into an SPDK-style callback arg. We use them extensively. Probably they would work here as well.

I'm using `cb_arg` to wrap the sender in the resize_replica function. Couldn't see a clear way to use `done_cb` as err code conversion is required from `i32`. Let me take a look again if possible.

io-engine/src/bin/io-engine-client/v1/replica_cli.rs
tiagolobocastro: but it won't happen because the parameter is marked as required?

io-engine/src/lvs/lvs_lvol.rs
tiagolobocastro: Yeah we can overwrite errno to let control plane know

io-engine/src/bin/io-engine-client/v1/replica_cli.rs
dsharma-dc: So the point is what would one like to see let's say when issuing below wrong command:
> ./io-engine-client replica resize f3eb2bd1-e328-4955-8804-521cc28f738f 170MiiB

This:
> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: "170MiiB"', io-engine/src/bin/io-engine-client/v1/replica_cli.rs:444:81
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Aborted (core dumped)

OR
> Error: GrpcStatus { source: Status { code: InvalidArgument, message: "Bad size '170MiiB'", source: None }, backtrace: Backtrace(()) }

tiagolobocastro: AIUI the first cannot happen, because the argument is marked as required?
This is why we should be using derive api rather than the builder, then it'd be much clearer.

io-engine/src/lvs/lvs_lvol.rs
tiagolobocastro: Probably &mut and result Result<(), Error> I would say.

io-engine/src/lvs/lvs_lvol.rs
dsharma-dc: Will overwrite errno to `EAGAIN` here because it has to be retried, assuming failure.

io-engine/src/grpc/v0/mayastor_grpc.rs
```diff
@@ -249,6 +249,9 @@ impl From<LvsError> for tonic::Status {
                 Errno::EMEDIUMTYPE => Status::aborted(e.to_string()),
                 _ => Status::internal(e.to_string()),
             },
+            LvsError::RepResize {
+                ..
+            } => Status::internal("replica resize failed".to_string()),
```
tiagolobocastro: Shouldn't we inspect the errno code rather than harcode to internal?
Do we get different error codes from spdk depending on error condition, if new size is set to larger than pool etc?

dsharma-dc: > Shouldn't we inspect the errno code rather than harcode to internal? Do we get different error codes from spdk depending on error condition, if new size is set to larger than pool etc?

This seems like a reasonable mapping based on current possible spdk codes?
  
>   LvsError::RepResize {
                source, ..
            } => match source {
                Errno::ENOSPC | Errno::ENOMEM => Status::failed_precondition("Out of space/memory: replica resize failed".to_string()),
                Errno::EPERM => Status::permission_denied("operation not allowed: replica resize failed".to_string()),
                Errno::EINVAL => Status::invalid_argument("lvol missing? replica resize failed".to_string()),
                _ => Status::internal("internal error: replica resize failed".to_string())
            },

tiagolobocastro: I guess so, you may need to inspect the spdk code to check which error codes may be returned and how they should be converted to gRPC status codes

io-engine/src/grpc/v0/mayastor_grpc.rs
```diff
@@ -249,6 +249,22 @@ impl From<LvsError> for tonic::Status {
                 Errno::EMEDIUMTYPE => Status::aborted(e.to_string()),
                 _ => Status::internal(e.to_string()),
             },
+            LvsError::RepResize {
+                source, ..
+            } => match source {
+                Errno::ENOSPC | Errno::ENOMEM => Status::resource_exhausted(
+                    "Out of space/memory: replica resize failed".to_string(),
```
tiagolobocastro: You can use e.verbose() on each of these to include the error chain, rather than manually doing it as there's ambiguity here, example "Out of space or memory" ?

tiagolobocastro (approved):

io-engine/src/grpc/v0/mayastor_grpc.rs
dsharma-dc: Oh right, will use `e.verbose()` or `e.string()` not sure what's the difference. Other errors in this file are using e.string(), will take a look.

dsharma-dc:
bors try

dsharma-dc:
bors try

dsharma-dc:
bors merge

dsharma-dc:
bors cancel

dsharma-dc:
bors merge

dsharma-dc:
bors merge

PR #1551, opened 2023-11-28
feat(eventing): add host events

datacore-vvarakantham:
Added host events.

### DIFF
--- a/io-engine/src/eventing/host_events.rs
+++ b/io-engine/src/eventing/host_events.rs
@@ -1,0 +1,45 @@
+use events_api::event::{
+    EventAction,
+    EventCategory,
+    EventMessage,
+    EventMeta,
+    EventSource,
+};
+
+use crate::{
+    core::MayastorEnvironment,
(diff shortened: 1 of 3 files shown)


io-engine/src/eventing/host_events.rs
```diff
@@ -0,0 +1,46 @@
...
+    }
+}
+
+impl EventWithMeta for NvmfController {
+    fn event(
+        &self,
+        event_action: EventAction,
+        meta: EventMeta,
+    ) -> EventMessage {
+        let mut event_meta = meta.clone();
```
tiagolobocastro: Do we need to clone, can't we just mutate meta?

tiagolobocastro (approved):

io-engine/src/eventing/host_events.rs
datacore-vvarakantham: Yes we can mutate meta. Added the change now.

chriswldenyer (approved):
looks reasonable

datacore-vvarakantham:
bors merge

PR #1553, opened 2023-11-29
fix(device_remove): hot remove of spdk bdev should drop descriptor 

dsharma-dc:
If a bdev replica(non-nvmf) is destroyed, the device removal path doesn't close the device as in case of nvmf child. As a result, the child state isn't set as Destroyed, which results in bdev descriptor being kept opened and hence spdk internally  stucks into a ebusy state for that bdev.

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_child.rs
+++ b/io-engine/src/bdev/nexus/nexus_child.rs
@@ -1103,16 +1103,13 @@ impl<'c> NexusChild<'c> {
         info!("{self:?}: unplugging child...");
 
         let state = self.state();
-        let is_destroying = self.is_destroying();
 
-        // Only drop the device and the device descriptor if the child is being
-        // destroyed. For a hot remove event, keep the device and descriptor.
-        if is_destroying {
-            debug!("{self:?}: dropping block device");
-            self.device = None;
-        } else {
-            debug!("{self:?}: hot remove: keeping block device");
-        }
+        // Drop the device and the device descriptor if the child is being
+        // destroyed or hot removed. For hot remove, the DeviceRemoved event
+        // will send us here and keeping the device/descriptor isn't of
+        // any use.
+        debug!("{self:?}: dropping block device");
+        self.device = None;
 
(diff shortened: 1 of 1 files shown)


dsharma-dc:
bors try

dsharma-dc:
bors try

dsharma-dc:
bors try

io-engine/src/bdev/nexus/nexus_child.rs
```diff
@@ -1103,16 +1103,13 @@ impl<'c> NexusChild<'c> {
...
-            self.device = None;
-        } else {
-            debug!("{self:?}: hot remove: keeping block device");
-        }
+        // Drop the device and the device descriptor if the child is being
+        // destroyed or hot removed. For hot remove, the DeviceRemoved event
+        // will send us here and keeping the device/descriptor isn't of
+        // any use.
+        debug!("{self:?}: dropping block device");
+        self.device = None;
```
tiagolobocastro, line 1112: hmm if there's a hot removal this way, does this mean we don't go through the pause dance?
If so then we need to cater for that as well..

dsharma-dc: yeah we don't do that pause here because child retire isn't done. There are more nuances involved here than that make me think we are better maintain status quo with this code. I'm inclined to park this change for now.

1. A way to simulate hot remove is doing `bdev destroy` on pool disk rather than `pool destroy`.  In this case, both kind of child(nvmf and spdk bdev) behave similarly and are unplugged fine, without going via retire path and pause.
2. The `pool destroy` or `replica destroy` are to be rather considered as fault. nvmf child receives `AdminCommandCompletionFailure` for these, deems the child faulted, **retires it, and then triggers `DeviceRemoved` event.** For spdk bdev, we directly get `DeviceRemoved` in these cases without child having gone through retire.
3. Practically, 1 is more likely than 2 in actual system.

dsharma-dc:
not pursuing at the moment.

PR #1564, opened 2023-12-15
fix(pstor): increase persistence timeouts

tiagolobocastro:
The pstor might might unavailable for some time, for example during upgrade. We should have sufficient timeouts to cope with this. Also pstor can get slow at times as it's writing to disk, so we also should have a larger timeout on store.

todo: should we still carry on trying to persist on separate task after we fail the nexus?
todo: what happens when we try to shutdown nexus first and pstor fails?

We'll revisit this post-release.


avishnu (approved):
lgtm, maybe add a comment stating we want to revisit again.

hrudaya21 (approved):

tiagolobocastro:
> lgtm, maybe add a comment stating we want to revisit again.

Done.

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1570, opened 2024-01-04
feat(multireplicasnapshot): remove multi replica snapshot checks in dataplane

hrudaya21:


Abhinandan-Purkait (approved):

io-engine/src/bin/io-engine-client/v1/snapshot_cli.rs
```diff
@@ -202,24 +202,32 @@ async fn create_for_nexus(
         })?
         .to_owned();
     let replica_uuid = matches
-        .get_many::<String>("replica_uuid")
+        .get_one::<String>("replica_uuid")
```
tiagolobocastro, line 207: Any reason to move from many to one?
I guess this means all uuids must be provided with 1 string?

tiagolobocastro (commented):
Let's add some tests as well

io-engine/src/bin/io-engine-client/v1/snapshot_cli.rs
hrudaya21: It is changed to map each replica_uuid to snapshot_uuid. Both will be given as a single string as shown below.

run-io-engine snapshot create_for_nexus a4d9bbd8-ed98-4905-81cd-f2ff004eff06 e1 t1 snap1 "a0e7832c-6f46-461d-a8c5-74cdf4ddc53f,c2b46c53-2bbe-4286-949d-47f1859c9eac,6c9cdf49-7261-4fdf-81af-1e82949fc786" "15b41216-2fc6-4b4b-90e4-6d772d49bdb1,15b41216-2fc6-4b4b-90e4-6d772d49bdb2,15b41216-2fc6-4b4b-90e4-6d772d49bdb3"

tiagolobocastro: Oh, I see now, because the arguments are positional

hrudaya21:
> Let's add some tests as well

Added the test for multi-replica snapshot. Will more fault cases from control-plane BDD.

io-engine/tests/snapshot_nexus.rs
```diff
@@ -1282,3 +1249,135 @@ async fn test_replica_listing_with_query() {
...
+            skip: false,
+        },
+        NexusCreateSnapshotReplicaDescriptor {
+            replica_uuid: repl_3.uuid(),
+            snapshot_uuid: Some(Uuid::new_v4().to_string()),
+            skip: false,
+        },
+    ];
+    let snap_list = nex_0
+        .create_nexus_snapshot(snapshot_params, replicas)
```
tiagolobocastro: Do we have tests where we specify only 1 replica but the nexus has more? If not we should add, to ensure it won't create without skip true on the other replicas.

tiagolobocastro:
> > Let's add some tests as well
> 
> Added the test for multi-replica snapshot. Will more fault cases from control-plane BDD.

Depends on what type of fault cases, all paths should be tested in this repo only, it's not the control-plane's CI responsibility to find bugs in this repo..

tiagolobocastro (approved):

io-engine/tests/snapshot_nexus.rs
hrudaya21: Added few error case which can be checked directly in dataplane, including this condition.

hrudaya21:
> > > Let's add some tests as well
> > 
> > 
> > Added the test for multi-replica snapshot. Will more fault cases from control-plane BDD.
> 
> Depends on what type of fault cases, all paths should be tested in this repo only, it's not the control-plane's CI responsibility to find bugs in this repo..

I was indicating the cross node fault scenarios. As part of other review comments, cover error scenarios which can fail in dataplane explicitly.

hrudaya21:
bors try

hrudaya21:
bors merge

PR #1572, opened 2024-01-06
ci: use remote org for docker images

tiagolobocastro:
Uses the org of the remote git origin to set docker credentials and docker image org.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,33 +13,16 @@ def getLastNonAbortedBuild(build) {
   }
 }
 
-// Send out a slack message if branch got broken or has recovered
-def notifySlackUponStateChange(build) {
-  def cur = build.getResult()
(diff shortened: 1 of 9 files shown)


Abhinandan-Purkait (approved):

Jenkinsfile
```diff
@@ -13,33 +13,16 @@ def getLastNonAbortedBuild(build) {
   }
 }
 
-// Send out a slack message if branch got broken or has recovered
```
pchandra19, line 16: We need to replace alerting. If we are not doing that with this PR , then I will suggest to please create a task for this so that we don't miss it.

pchandra19 (approved):

Jenkinsfile
tiagolobocastro: There's no alerting today as we don't have our slack subscription anymore, I'll mention this in the standup.

tiagolobocastro:
bors merge

PR #1573, opened 2024-01-08
feat(eventing): add io-engine events

datacore-vvarakantham:
Added io-engine start, stop, shutdown events.
Reactor freeze, unfreeze events.
Nexus shutdown event.

### DIFF
--- a/io-engine/src/bdev/nexus/mod.rs
+++ b/io-engine/src/bdev/nexus/mod.rs
@@ -22,7 +22,10 @@ mod nexus_nbd;
 mod nexus_persistence;
 mod nexus_share;
 
-use crate::{bdev::nexus::nexus_iter::NexusIterMut, eventing::Event};
+use crate::{
+    bdev::nexus::nexus_iter::NexusIterMut,
+    eventing::{Event, EventMetaGen, EventWithMeta},
+};
 pub(crate) use nexus_bdev::NEXUS_PRODUCT_ID;
 pub use nexus_bdev::{
     nexus_create,
@@ -176,7 +179,7 @@ pub async fn shutdown_nexuses() {
         // Destroy nexus and persist its state in the ETCd.
         match nexus.as_mut().destroy_ext(true).await {
             Ok(_) => {
-                nexus.event(EventAction::Shutdown).generate();
+                Event::event(&(*nexus), EventAction::Shutdown).generate();
             }
             Err(error) => {
                 error!(
@@ -184,6 +187,12 @@ pub async fn shutdown_nexuses() {
                     error = error.verbose(),
                     "Failed to destroy nexus"
                 );
+                EventWithMeta::event(
+                    &(*nexus),
+                    EventAction::Shutdown,
+                    error.meta(),
+                )
+                .generate();
             }
         }
     }
--- a/io-engine/src/bdev/nexus/mod.rs
+++ b/io-engine/src/bdev/nexus/mod.rs
@@ -3,6 +3,7 @@
 use std::{pin::Pin, sync::atomic::AtomicBool};
 
 use crate::core::VerboseError;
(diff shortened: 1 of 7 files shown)


io-engine/src/core/env.rs
```diff
@@ -466,6 +468,9 @@ async fn do_shutdown(arg: *mut c_void) {
         spdk_rpc_finish();
         spdk_subsystem_fini(Some(reactors_stop), arg);
     }
+    MayastorEnvironment::global_or_default()
```
tiagolobocastro: Shouldn't this be generated at the start of shutdown?

io-engine/src/core/env.rs
```diff
@@ -477,6 +482,9 @@ pub fn mayastor_env_stop(rc: i32) {
             r.send_future(async move {
                 do_shutdown(rc as *const i32 as *mut c_void).await;
             });
+            MayastorEnvironment::global_or_default()
+                .event(EventAction::Stop)
+                .generate();
```
tiagolobocastro: Shouldn't this be generated right before we stop only?
And would it be worth it to include how long it took to stop?

datacore-vvarakantham: Yes. Should I add the time taken to stop field in nanoseconds time?

tiagolobocastro: Can just use `Duration` I guess?

datacore-vvarakantham: Yes, I can use the Duration.

datacore-vvarakantham: Fixed. Added the time taken to event specific details.

io-engine/src/core/env.rs
datacore-vvarakantham: Added the event at the start.

io-engine/src/eventing/mod.rs
```diff
@@ -1,12 +1,13 @@
...
+pub(crate) mod io_engine_events;
 mod nexus_child_events;
 pub(crate) mod nexus_events;
 mod pool_events;
 mod replica_events;
 use events_api::event::{EventAction, EventMessage, EventMeta};
 
 /// Event trait definition for creating events.
-pub(crate) trait Event {
+pub trait Event {
```
tiagolobocastro, line 10: Does it need to be pub?

tiagolobocastro (approved):

io-engine/src/bdev/nexus/mod.rs
```diff
@@ -173,15 +174,19 @@ pub async fn shutdown_nexuses() {
...
-                name = nexus.name,
-                error = error.verbose(),
-                "Failed to destroy nexus"
-            );
+        match nexus.as_mut().destroy_ext(true).await {
+            Ok(_) => {
+                nexus.event(EventAction::Shutdown).generate();
+            }
+            Err(error) => {
+                error!(
```
tiagolobocastro, line 185: Wonder if maybe shutdown can include a result?

io-engine/src/eventing/mod.rs
datacore-vvarakantham: It does not need to be pub here. pub(crate) works as all the events are in the same crate.

io-engine/src/bdev/nexus/mod.rs
datacore-vvarakantham: What would the result contain here? If the shutdown is failed or successful?

tiagolobocastro: Yes

datacore-vvarakantham: Added the error to event details.

sinhaashish (approved):

datacore-vvarakantham:
bors merge

PR #1577, opened 2024-01-11
feat(eventing): state change events for nexus and replica

datacore-vvarakantham:
Added state change events for nexus and replica.

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev.rs
@@ -10,6 +10,7 @@ use std::{
     convert::TryFrom,
     fmt::{Debug, Display, Formatter},
     marker::PhantomPinned,
+    ops::Deref,
     os::raw::c_void,
     pin::Pin,
     sync::atomic::Ordering,
@@ -56,7 +57,7 @@ use crate::{
         Share,
         VerboseError,
     },
-    eventing::Event,
+    eventing::{nexus_events::state_change_event_meta, Event, EventWithMeta},
     rebuild::HistoryRecord,
     subsys::NvmfSubsystem,
 };
@@ -508,8 +509,15 @@ impl<'n> Nexus<'n> {
 
     /// Sets the state of the Nexus.
     fn set_state(self: Pin<&mut Self>, state: NexusState) -> NexusState {
+        let previous = *self.state.lock();
         debug!("{:?}: changing state to '{}'", self, state);
         *self.state.lock() = state;
+        EventWithMeta::event(
+            self.deref(),
+            EventAction::StateChange,
+            state_change_event_meta(previous, state),
+        )
+        .generate();
         state
     }
 
@@ -859,7 +867,7 @@ impl<'n> Nexus<'n> {
             let name = self.name.clone();
 
             // After calling unregister_bdev_async(), Nexus is gone.
-            let evt = self.event(EventAction::Delete);
+            let evt = Event::event(&(*self), EventAction::Delete);
             match self.as_mut().bdev_mut().unregister_bdev_async().await {
                 Ok(_) => {
                     info!("Nexus '{name}': nexus destroyed ok");
@@ -960,6 +968,12 @@ impl<'n> Nexus<'n> {
                 // Save current state and mark nexus as being under shutdown.
                 t => {
                     *s = NexusState::ShuttingDown;
+                    EventWithMeta::event(
+                        self.deref(),
+                        EventAction::StateChange,
+                        state_change_event_meta(t, *s),
+                    )
+                    .generate();
                     t
                 }
             }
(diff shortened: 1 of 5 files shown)


io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -851,7 +858,7 @@ impl<'n> Nexus<'n> {
             let name = self.name.clone();
 
             // After calling unregister_bdev_async(), Nexus is gone.
-            let evt = self.event(EventAction::Delete);
+            let evt = Event::event(&(*self), EventAction::Delete);
```
tiagolobocastro: Why this change is needed?

io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -909,6 +916,12 @@ impl<'n> Nexus<'n> {
             // Reset operation is allowed only when the Nexus is Open state
             NexusState::Open => {
                 *state = NexusState::Reconfiguring;
+                EventWithMeta::event(
```
tiagolobocastro: Don't think we need such fine details, state changing to showdown should be enough?

io-engine/src/bdev/nexus/nexus_bdev.rs
datacore-vvarakantham: I had to use the event() function from two different traits Event and EventWithMeta in the same module. I added this to use event() fn from Event trait.

io-engine/src/bdev/nexus/nexus_bdev.rs
datacore-vvarakantham: Yes the events list is filled with many state change events as we are having event for every state change. Other state change events are not needed for nexus other than shutdown? Or should I just remove reconfiguring event?

datacore-vvarakantham: Removed nexus state change events for reconfiguring and open states.

io-engine/src/bdev/nexus/nexus_bdev.rs
tiagolobocastro: Perhaps, deref might be neater, but just a nit:
```suggestion
            let evt = Event::event(self.deref(), EventAction::Delete);
```

io-engine/src/eventing/nexus_events.rs
```diff
@@ -50,6 +53,16 @@ impl<'n> EventMetaGen for NexusChild<'n> {
     }
 }
 
+pub(crate) fn state_change_event_meta(
```
tiagolobocastro, line 59: pls add doc comment

io-engine/src/eventing/replica_events.rs
```diff
@@ -32,3 +36,28 @@ impl Event for Lvol {
         }
     }
 }
+
+pub(crate) fn state_change_event_meta(
```
tiagolobocastro, line 41: doc comment here

tiagolobocastro (approved):

io-engine/src/eventing/replica_events.rs
datacore-vvarakantham: Added the doc comment.

hrudaya21:
@tiagolobocastro Do We need to consider Subsystem pause/resume event and nexus init/reconfigure event ?

tiagolobocastro:
> @tiagolobocastro Do We need to consider Subsystem pause/resume event and nexus init/reconfigure event ?

You mean as events? Yes those would be very useful as well!

hrudaya21:
> > @tiagolobocastro Do We need to consider Subsystem pause/resume event and nexus init/reconfigure event ?
> 
> You mean as events? Yes those would be very useful as well!

I think, We can handle these scenario in this PR, as We already handling nexus events here. @datacore-vvarakantham

datacore-vvarakantham:
> > > @tiagolobocastro Do We need to consider Subsystem pause/resume event and nexus init/reconfigure event ?
> > 
> > 
> > You mean as events? Yes those would be very useful as well!
> 
> I think, We can handle these scenario in this PR, as We already handling nexus events here. @datacore-vvarakantham

Sure, I can add those as events.

io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -1017,8 +1052,42 @@ impl<'n> Nexus<'n> {
...
+        .generate();
+        let start_time = std::time::Instant::now();
+        let result = self.as_mut().io_subsystem_mut().suspend().await;
+        match result {
+            Ok(_) => {
+                EventWithMeta::event(
+                    self.deref(),
+                    EventAction::SubsystemPause,
+                    subsystem_pause_event_meta(
+                        "completed",
```
tiagolobocastro: "completed" or "failed" seems to be unnecessary the way it;s used here, since we can infer that from the error.
I think including the actual subsystem status would be better here.

datacore-vvarakantham: Yes. I added NexusPauseState now.

sinhaashish (approved):

datacore-vvarakantham:
bors merge

PR #1581, opened 2024-01-19
Rebuild refactoring

tiagolobocastro:
refactor(rebuild): make the rebuild more generic
    
    A rebuild is essentially a copy from one bdev to another.
    However, we can have different variations on this.
    This change aims to make the rebuild more generic, removing the nexus specific
    bits from the core rebuild logic and allowing us to compose different types of
    rebuild in the future, whilst being able to reuse the shared bits.
    
    To achieve this the rebuild backend is split into a rebuild manager which is the
    generic component responsible for running and managing the rebuild and its tasks.
    We can then implement different rebuild backends as we see fit, example a nexus
    rebuild which locks ranges or a regular bdev to bdev rebuild.
    
---

    fix(rebuild): send final rebuild state on drop
    
### DIFF
--- a/io-engine/src/rebuild/bdev_rebuild.rs
+++ b/io-engine/src/rebuild/bdev_rebuild.rs
@@ -1,0 +1,144 @@
+use std::{
+    ops::{Deref, Range},
+    rc::Rc,
+};
+
+use super::{
+    rebuild_descriptor::RebuildDescriptor,
+    rebuild_error::RebuildError,
+    rebuild_job_backend::RebuildBackend,
+    rebuild_task::{RebuildTasks, TaskResult},
+    RebuildJob,
+    RebuildJobOptions,
+    SEGMENT_TASKS,
+};
+
+use crate::gen_rebuild_instances;
+
+/// A Bdev rebuild job is responsible for managing a rebuild (copy) which reads
+/// from source_hdl and writes into destination_hdl from specified start to end.
+pub struct BdevRebuildJob(RebuildJob);
+
+impl std::fmt::Debug for BdevRebuildJob {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
+impl Deref for BdevRebuildJob {
+    type Target = RebuildJob;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl BdevRebuildJob {
+    /// Creates a new RebuildJob which rebuilds from source URI to target URI
+    /// from start to end (of the data partition); notify_fn callback is called
+    /// when the rebuild state is updated - with the source and destination
+    /// bdev URI's as arguments.
+    pub async fn new(
+        src_uri: &str,
+        dst_uri: &str,
+        range: Option<Range<u64>>,
+        options: RebuildJobOptions,
+        notify_fn: fn(&str, &str) -> (),
+    ) -> Result<Self, RebuildError> {
+        let descriptor =
+            RebuildDescriptor::new(src_uri, dst_uri, range, options).await?;
+        let tasks = RebuildTasks::new(SEGMENT_TASKS, &descriptor)?;
+        let backend =
+            BdevRebuildJobBackend::new(tasks, notify_fn, descriptor).await?;
+
+        RebuildJob::from_backend(backend).await.map(Self)
+    }
+}
+
+gen_rebuild_instances!(BdevRebuildJob);
+
+/// A rebuild job which is responsible for rebuilding from
+/// source to target of the `RebuildDescriptor`.
+pub(super) struct BdevRebuildJobBackend {
+    /// The next block to be rebuilt.
+    next: u64,
+    /// A pool of tasks which perform the actual data rebuild.
+    task_pool: RebuildTasks,
+    /// A generic rebuild descriptor.
+    descriptor: Rc<RebuildDescriptor>,
+    /// Notification callback with src and dst uri's.
+    notify_fn: fn(&str, &str) -> (),
+}
+
+#[async_trait::async_trait(?Send)]
+impl RebuildBackend for BdevRebuildJobBackend {
+    fn on_state_change(&mut self) {
+        (self.notify_fn)(&self.descriptor.src_uri, &self.descriptor.dst_uri);
+    }
+
+    fn common_desc(&self) -> &RebuildDescriptor {
+        &self.descriptor
+    }
+
+    fn task_pool(&self) -> &RebuildTasks {
+        &self.task_pool
+    }
+
+    fn schedule_task_by_id(&mut self, id: usize) -> bool {
+        if self.next >= self.descriptor.range.end {
+            false
+        } else {
+            let next = std::cmp::min(
+                self.next + self.descriptor.segment_size_blks,
+                self.descriptor.range.end,
+            );
+            self.task_pool.schedule_segment_rebuild(
+                id,
+                self.next,
+                self.descriptor.clone(),
+            );
+            self.task_pool.active += 1;
+            self.next = next;
+            true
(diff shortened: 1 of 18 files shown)


tiagolobocastro:
bors try

hrudaya21 (approved):
lgtm. 
We can have the entity relation diagram for bdev and nexus rebuild abstraction in LLD for future reference.

io-engine/src/rebuild/bdev_rebuild.rs
```diff
@@ -0,0 +1,144 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl BdevRebuildJob {
+    /// Creates a new RebuildJob which rebuilds from source URI to target URI
+    /// from start to end (of the data partition); notify_fn callback is called
+    /// when the rebuild state is updated - with the nexus and destination
+    /// URI as arguments.
```
dsharma-dc: "with the source and destination bdev URI as arguments"

io-engine/src/rebuild/bdev_rebuild.rs
```diff
@@ -0,0 +1,144 @@
...
+    notify_fn: fn(&str, &str) -> (),
+}
+
+#[async_trait::async_trait(?Send)]
+impl RebuildBackend for BdevRebuildJobBackend {
+    fn on_state_change(&mut self) {
+        (self.notify_fn)(&self.descriptor.src_uri, &self.descriptor.dst_uri);
+    }
+
+    fn cmn_descriptor(&self) -> &RebuildDescriptor {
```
dsharma-dc: if `cmn` is meant to say `common` then let's say `common_desc`

io-engine/src/rebuild/bdev_rebuild.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// when the rebuild state is updated - with the nexus and destination
+    /// URI as arguments.
+    pub async fn new(
+        src_uri: &str,
+        dst_uri: &str,
+        range: Option<Range<u64>>,
+        options: RebuildJobOptions,
+        notify_fn: fn(&str, &str) -> (),
+    ) -> Result<Self, RebuildError> {
+        let descriptor =
```
dsharma-dc, line 47: Normally it wouldn't happen - but shall we put a sanity check here that the `src_uri` and `dst_uri` are distinct, since here the selection won't be done by nexus as in Nexus rebuild?

io-engine/src/rebuild/bdev_rebuild.rs
```diff
@@ -0,0 +1,144 @@
...
+        f.debug_struct("BdevRebuildJob")
+            .field("next", &self.next)
+            .finish()
+    }
+}
+impl std::fmt::Display for BdevRebuildJobBackend {
+    fn fmt(&self, _f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        Ok(())
+    }
+}
```
dsharma-dc, line 121: This seems redundant, or do you intend to add some formatting logic later?

io-engine/src/rebuild/nexus_rebuild.rs
```diff
@@ -0,0 +1,279 @@
...
+
+/// Contains all descriptors and their associated information which allows the
+/// tasks to copy/rebuild data from source to destination.
+pub(super) struct NexusRebuildDescriptor {
+    /// Name of the nexus associated with the rebuild job.
+    pub nexus_name: String,
+    /// Nexus Descriptor so we can lock its ranges when rebuilding a segment.
+    pub(super) nexus: DescriptorGuard<()>,
+    /// The generic rebuild descriptor for copying from source to target.
+    pub(super) common: RebuildDescriptor,
```
dsharma-dc, line 90: nitpick: `common_desc` may be

io-engine/src/rebuild/nexus_rebuild.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    fn schedule_task_by_id(&mut self, id: usize) -> bool {
+        match self.send_segment_task(id) {
+            Some(next) => {
+                self.task_pool.active += 1;
+                self.next = next;
+                true
+            }
+            // we've already got enough tasks to rebuild the destination
+            None => false,
```
dsharma-dc, line 142: Current logic completes the job here if no more active tasks, how is that handled now?
```
                if self.task_pool.active == 0 {
                    self.complete();
                }
```

io-engine/src/rebuild/rebuild_instances.rs
```diff
@@ -0,0 +1,90 @@
...
+                static REBUILD_INSTANCES: once_cell::sync::OnceCell<
+                    parking_lot::Mutex<RebuildJobInstances>,
+                > = once_cell::sync::OnceCell::new();
+
+                REBUILD_INSTANCES
+                    .get_or_init(|| parking_lot::Mutex::new(std::collections::HashMap::new()))
+                    .lock()
+            }
+
+            /// Returns number of all nexus rebuild jobs on the system.
```
dsharma-dc: nitpick: Returns number of all rebuild jobs of type ty on the system.

io-engine/src/rebuild/bdev_rebuild.rs
tiagolobocastro: Yes it was :) Alright, modified.

io-engine/src/rebuild/bdev_rebuild.rs
tiagolobocastro: In general this is difficult to do with with just the strings, but probably can compare using the handles.
I wonder though if there could be any way we'd want to have src and dst the same?

io-engine/src/rebuild/bdev_rebuild.rs
tiagolobocastro: It's not redundant, it's required by the trait, so has to be implemented

io-engine/src/rebuild/nexus_rebuild.rs
tiagolobocastro: Same logic is still there, in the `RebuildJobBackendManager`

io-engine/src/rebuild/nexus_rebuild.rs
```diff
@@ -0,0 +1,279 @@
...
+        &self.0
+    }
+}
+
+impl NexusRebuildJob {
+    /// Creates a new RebuildJob which rebuilds from source URI to target URI
+    /// from start to end (of the data partition); notify_fn callback is called
+    /// when the rebuild state is updated - with the nexus and destination
+    /// URI as arguments.
+    /// todo: Should we use a builder? Example:
```
tiagolobocastro, line 53: @dsharma-dc @hrudaya21 @dsavitskiy 
What do you think, would this be good to have rather than a large new?

hrudaya21: As far I know, 5 parameters per function is industry standard. Mostly the future if there will be any enhancement here, We may extend more parameters to `RebuildOptions` which will keep the parameter list to 6.

My suggestion to keep 5 parameters rule, until We find some exceptions or complexity.

io-engine/src/rebuild/bdev_rebuild.rs
dsharma-dc: I don't see any use-case for keeping both handles same. If ever copy within device is needed, I guess the NVMe Copy command support would be a suitable candidate for that.

tiagolobocastro: Alright, added check for this.

dsharma-dc (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
Had forgotten to resolve comment :(
bors should complain right away, not sure if possible though!
bors merge

PR #1583, opened 2024-02-01
fix(nexus): fixing missing I/Os during nexus rebuild

dsavitskiy:
If a nexus was published after a rebuild started, the child being rebuilt
was not added to writers. This caused I/Os missing the child, leading to
a corrupted replica.


dsavitskiy:
bors try

scripts/rust-linter.sh
```diff
@@ -9,4 +9,5 @@ $CARGO clippy --all --all-targets -- -D warnings \
     -A clippy::option_env_unwrap \
     -A clippy::redundant-guards \
     -A clippy::suspicious-doc-comments \
+    -A clippy::useless-format \
```
tiagolobocastro, line 12: why do we need this?
Btw seems like we have a lot of whitelists here, any reason for this?

tiagolobocastro (approved):

dsavitskiy:
bors try

dsavitskiy:
bors try

scripts/rust-linter.sh
dsavitskiy: As we update rustc/clippy versions, new lints come in. Some of them are not easily fixable, and for some we just don't have enough time.

As for "useless-format", it is a newer one and a quite annoying.

E.g. you have a list like:
let args = [
    format!("--abc={a}"),
    format!("--def"),
    format!("--qwe={b}"),
];

It is nicely and uniformly formatted, and does not cause any troubles. Now, clippy started to complain about it, suggesting to replace format!("--def") with "--def".to_string(). I don't see much sense in this.

hrudaya21 (approved):

dsavitskiy:
bors merge

PR #1587, opened 2024-02-14
Feat: LVM as a pool backend

tiagolobocastro:
Add LVM as an experimental alternative backend to Mayastor:
1. it allows us to use a low latency single replica
local volume (app pinned to the same node)
2. robust backend for existing LVM users which may prefer it
3. combine local engines into mayastor

This differs from the existing lvm localpv as rather than importing
the existing VGs we create the LVM PV's and the LVM VG itself.
On destruction, if the VG has no non-mayastor LVs then we destroy it,
otherwise we leave it behind, untagged.

The existing pool/replica services are also refactored to support
adding other backends in the future as well.
This is achieved by implementing the service interface separately
and binding them at a higher level. This way it avoids mixing
the backend code, at least as much as possible.


io-engine/src/lvm/error.rs
```diff
@@ -0,0 +1,52 @@
...
+pub enum Error {
+    #[snafu(display("Failed to parse: {err}"))]
+    FailedParsing { err: String },
+    #[snafu(display("Failed to execute command: {err}"))]
+    FailedExec { err: String },
+    #[snafu(display("I/O error: {source}"))]
+    Io { source: std::io::Error },
+    #[snafu(display("Invalid PoolType: {value}"))]
+    InvalidPoolType { value: i32 },
+    #[snafu(display("Volume Group with {query} not found"))]
```
Abhinandan-Purkait: `Volume Group`  --> `LVM Volume Group` just to be clear it's not related to a Volume

io-engine/src/lvm/lv_replica.rs
```diff
@@ -0,0 +1,642 @@
...
+            Protocol::Off => {
+                Self::bdev_none_props(bdev).await?;
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Import the Logical Volume by loading it via an SPDK bdev using either
+    /// AIO or Uring.
```
Abhinandan-Purkait: Maybe specify only AIO for the time being?

io-engine/src/lvm/vg_pool.rs
```diff
@@ -0,0 +1,514 @@
...
+    free: u64,
+    /// Corresponds to the vg_tags field in json output, the tags set in the
+    /// volume group.
+    #[serde(
+        deserialize_with = "super::parsing::de_comma_separated",
+        rename = "vg_tags"
+    )]
+    tags: Vec<String>,
+    /// The physical vol disks used by the volume group.
+    #[serde(skip_deserializing)]
```
Abhinandan-Purkait: Do we not need the disks? How do we compare otherwise while import that it's same disk?

io-engine/src/pool_backend.rs
```diff
@@ -22,6 +26,7 @@ impl TryFrom<i32> for PoolBackend {
     fn try_from(value: i32) -> Result<Self, Self::Error> {
```
Abhinandan-Purkait: Can we not use concrete types here and convert from/to_i32 ?

io-engine/src/lvm/error.rs
tiagolobocastro: Done

io-engine/src/lvm/vg_pool.rs
tiagolobocastro: Done

io-engine/src/pool_backend.rs
tiagolobocastro: Done

io-engine/src/lvm/lv_replica.rs
tiagolobocastro: Maybe we can specify this somehow?

io-engine/src/bin/io-engine.rs
```diff
@@ -89,6 +89,11 @@ fn start_tokio_runtime(args: &MayastorCliArgs) {
     if !ENABLE_NEXUS_RESET.load(Ordering::SeqCst) {
         warn!("Nexus reset is disabled");
     }
+    if args.lvm {
+        env::set_var("LVM", "1");
+        env::set_var("LVM_SUPPRESS_FD_WARNINGS", "1");
+        warn!("LVM backend is enabled");
```
dsharma-dc: `LVM pool backend is enabled`

io-engine/src/lvm/cli.rs
```diff
@@ -0,0 +1,298 @@
...
+    #[strum(serialize = "vgs")]
+    VGList,
+    /// Change volume group attributes.
+    #[strum(serialize = "vgchange")]
+    VGChange,
+}
+
+/// LVM wrapper over `Command` with added qol such as error mapping and
+/// decoding of json output reports.
+pub(super) struct LvmCmd {
```
dsharma-dc, line 118: This struct and enum names appear like having singular and plural versions while they aren't. Shall we call the enum as `LvmCmdAction`

io-engine/src/lvm/cli.rs
```diff
@@ -0,0 +1,298 @@
...
+    /// See help for `Command::args`.
+    pub(super) fn args<I, S>(mut self, args: I) -> Self
+    where
+        I: IntoIterator<Item = S>,
+        S: AsRef<OsStr>,
+    {
+        self.cmder.args(args);
+        self
+    }
+    /// Runs the LVM command with the provided `Command` arguments et all.
```
dsharma-dc: typo    `et all` => `et al` , at multiple places.
Can even be removed.

io-engine/src/lvm/vg_pool.rs
```diff
@@ -0,0 +1,294 @@
...
+            .into_iter()
+            // todo: not needed as we did the select?
+            .filter(|vg| vg.matches(opts))
+            .collect::<Vec<_>>();
+
+        Ok(vgs)
+    }
+
+    /// Import a volume group with the name provided or create one with the name
+    /// and disks provided currently only import is supported.
```
dsharma-dc, line 141: How do we create Lvm backend pool? Using lvm utilities separately? The Lvs trait `create()` is defaulting to backend Lvs

io-engine/src/lvm/vg_pool.rs
```diff
@@ -0,0 +1,294 @@
...
+    size: u64,
+    /// Corresponds to the vg_free field in json output, the free space on
+    /// volume group in bytes.
+    #[serde(deserialize_with = "de::number_from_string", rename = "vg_free")]
+    free: u64,
+    /// Corresponds to the vg_tags field in json output, the tags set in the
+    /// volume group.
+    #[serde(deserialize_with = "de::comma_separated", rename = "vg_tags")]
+    tags: Vec<Property>,
+    /// The physical vol disks used by the volume group.
```
dsharma-dc, line 89: A general question - Is the basic expectation to use lvm backend with Mayastor going to remain that the disks used as physical volumes in a vg are unpartitioned?

io-engine/src/lvm/vg_pool.rs
```diff
@@ -0,0 +1,294 @@
...
+    /// Check if the volume group contains the given tag.
+    fn tagged(&self, tag: Option<&String>) -> bool {
+        let eq = tag
+            .map(|tag| self.tags.iter().any(|ttag| ttag.key() == tag.as_str()))
+            .unwrap_or(true);
+        tracing::trace!("{tag:?} == {:?} ? {eq}", self.tags);
+        eq
+    }
+    /// Check if the volume group uuid matches.
+    fn uuided(&self, uuid: Option<&String>) -> bool {
```
dsharma-dc, line 394: `uuided` goes rhyming with `named` or `tagged` but makes an invalid word :) Perhaps change to `has_uuid` or `uuid_match`

io-engine/src/lvm/lv_replica.rs
```diff
@@ -0,0 +1,739 @@
...
+    /// set name if we find the tag.
+    name: Option<String>,
+
+    /// SPDK Bdev parameters which are needed by LVM.
+    bdev: Option<BdevOpts>,
+}
+
+/// Runtime settings for the LogicalVolume.
+#[derive(Debug, Default, Clone)]
+pub struct BdevOpts {
```
dsharma-dc: Would this be better renamed and moved to common place where we already have `BdevOpts` and `NvmeBdevOpts` ?

io-engine/src/lvm/lv_replica.rs
```diff
@@ -0,0 +1,739 @@
...
+        select
+    }
+}
+
+/// A Logical Volume (LV) is a virtual block device that can be used by the
+/// system or applications.
+/// Each block of data in an LV is stored on one or more PV in the VG, according
+/// to algorithms implemented by Device Mapper (DM) in the kernel.
+#[derive(Debug, Clone, Deserialize)]
+pub struct LogicalVolume {
```
dsharma-dc, line 84: Are we going to impl the `LogicalVolume` trait for this lvm `LogicalVolume` type, like we have for spdk_lvol ? I'm wondering about the naming confusion here

io-engine/src/grpc/v1/lvm/pool.rs
```diff
@@ -0,0 +1,178 @@
...
+        // check if the disks are used by existing lvs pool
+        if Lvs::iter()
+            .map(|l| l.base_bdev().name().to_string())
+            .any(|d| args.disks.contains(&d))
+        {
+            return Err(Status::invalid_argument(
+                "an lvs pool already uses the disk",
+            ));
+        };
+        VolumeGroup::import_or_create(args)
```
hrudaya21: How the disk which is used for lvm pool will not be used in lvs pool?

io-engine/src/lvm/lv_replica.rs
```diff
@@ -0,0 +1,739 @@
...
+    /// Resize the logical volume AND the SPDK Bdev to the given size.
+    /// > Note: If the SPDK Bdev fails to be resized, then the LV size is
+    /// > reverted back.
+    /// # Warning: If the SPDK Bdev is not loaded, then it is also not resized
+    /// and the operation still succeeds.
+    pub(crate) async fn resize(&mut self, size: u64) -> Result<(), Error> {
+        let prev_size = self.size;
+        self.resize_lv(size).await?;
+
+        if let Err(error) = self.resize_bdev(size).await {
```
hrudaya21, line 293: How spdk resize is related to lvm resize here ? Both using different pool_type resource

tiagolobocastro: Not sure what you mean by both using different pool_type?
We put an spdk bdev on top of the lvm lv, so after resizing the lvm lv we have to resize the spdk bdev.

io-engine/src/grpc/v1/lvm/pool.rs
tiagolobocastro: there's a simple check above, though I suspect that is not sufficient...
I think we need to try to import first as lvs and only if it fails try to import as lvm, this as another PR I think, this is already too large

io-engine/src/lvm/cli.rs
tiagolobocastro: `LvmCmd` is inline with what it's doing, wrapper over `Command`, so I'd rather rename the enum, example: `LvmSubCmd`.

io-engine/src/lvm/vg_pool.rs
tiagolobocastro: Yeah separate with pvcreate and vgcreate, this follows existing behaviour from PR's I picked up.
Create can be added on another PR.

io-engine/src/lvm/vg_pool.rs
tiagolobocastro: What do you mean by unpartitioned?

io-engine/src/lvm/vg_pool.rs
tiagolobocastro: is it invalid though? uuid is a noun so we can add "ed" to it?

io-engine/src/lvm/lv_replica.rs
tiagolobocastro: These are different, this is a runtime "cache" of the bdev if you will. It shouldn't be pub either I'll fix that

io-engine/src/lvm/vg_pool.rs
dsharma-dc: With Mayastor, I guess we today expect the raw disk device to be provided(without any partitions) and the complete disk is used of creating a diskpool. Do we expect same here? In other words, can Mayastor use a logicalVolume created by collecting desired physical partitions from PVs?
Reference: https://en.wikipedia.org/wiki/Logical_Volume_Manager_(Linux)#/media/File:Lvm.svg

io-engine/src/lvm/lv_replica.rs
tiagolobocastro: I'm not sure, existing `LogicalVolume` trait has spdk specific things, so I'm not sure how to integrate it...
Also the trait seems to be used only inside Lvs so far.

io-engine/src/lvm/vg_pool.rs
tiagolobocastro: You can use Mayastor with partitions AFAIK, not sure how much of that is tested but should work.
Yes it can, atm we're just given a VG, doesn't matter how the VG is composed

io-engine/src/grpc/v1/lvm/pool.rs
hrudaya21: I mean do We need to have similar checks in `lvs::create_pool()` where if the lvm pool is already created with some disk, then same disk should not be used for `lvs pool` ?

tiagolobocastro: We'll use a common check for both cases, so it can be used on any pool_type

io-engine/src/grpc/v1/lvm/replica.rs
```diff
@@ -0,0 +1,296 @@
...
+                    .with_ptpl(lvol.ptpl().create().map_err(|source| {
+                        Error::BdevShare {
+                            source: CoreError::Ptpl {
+                                reason: source.to_string(),
+                            },
+                        }
+                    })?);
+
+                if let Err(error) = lvol.share_nvmf(Some(props)).await {
+                    error!("Failed to import lvol: {error}...");
```
dsharma-dc: `failed to share lvm lvol` ?

io-engine/src/lvm/vg_pool.rs
orville-wright: We should enable both modes (Raw and partitioned). That offers the most flexibility to users.
note: A while ago, @tiagolobocastro  and I discussed that we should enable GPT awareness in Mayastor so that we can consume GPT partitions instead of the full raw disk. - This adds a level of safety to the system where we can label a GPT partition as a Mayastor disk entity and it wont get grabbed by something by mistake because it has a GPT label, - I think that was the plan.

io-engine/src/lvm/vg_pool.rs
orville-wright: We need to automate and integrate the creation of the backend LVM pool and adding dish media to it into Mayastor so the the user only interacts with our system to do all the LVM work. We don't want the user experience to be that they have to do a bunch of work outside Mayastor.

io-engine/src/lvm/vg_pool.rs
```diff
@@ -0,0 +1,294 @@
...
+        let name = self.name().to_string();
+        self.export().await?;
+        info!("LVM pool '{name}' has been destroyed successfully");
+        self.ptpl().destroy().ok();
+        Ok(())
+    }
+
+    /// Exports the volume by removing our tag from it.
+    /// todo: export the LV's as well...
+    pub(crate) async fn export(&mut self) -> Result<(), Error> {
```
dsharma-dc, line 257: Is Property::Lvm the only property that needs to be cleared from VG, and other properties are LV specific?

dsharma-dc (approved):

io-engine/src/core/env.rs
```diff
@@ -245,18 +245,19 @@ pub struct MayastorCliArgs {
         hide = true
     )]
     pub enable_nexus_channel_debug: bool,
+    #[clap(long, env = "LVM")]
+    pub lvm: bool,
```
Abhinandan-Purkait, line 254: Should we call this `ENABLE_LVM` ?

io-engine/src/grpc/v1/lvm/pool.rs
```diff
@@ -0,0 +1,178 @@
...
+                "lvs pool with the same name already exists",
+            ));
+        };
+        // check if the disks are used by existing lvs pool
+        if Lvs::iter()
+            .map(|l| l.base_bdev().name().to_string())
+            .any(|d| args.disks.contains(&d))
+        {
+            return Err(Status::invalid_argument(
+                "an lvs pool already uses the disk",
```
Abhinandan-Purkait, line 98: nit: "a lvs" ?

io-engine/src/grpc/v1/pool.rs
```diff
@@ -337,38 +295,58 @@ impl PoolRpc for PoolService {
...
+
+                let mut pools = vec![];
+                if lvm {
+                    pools.extend(
+                        match LvmSvc::new().list_svc_pools(&args).await {
+                            Ok(pools) => Ok(pools),
+                            Err(mut status) => {
+                                status.metadata_mut().insert(
+                                    "lvm",
+                                    tonic::metadata::MetadataValue::from(0),
```
Abhinandan-Purkait, line 367: Hmm maybe we should separate list by types?

Abhinandan-Purkait: How do we use this metadata on caller side?

io-engine/src/lvm/lv_replica.rs
```diff
@@ -0,0 +1,739 @@
...
+/// Different list options for a logical volume.
+#[derive(Default, Debug)]
+pub(crate) struct QueryArgs {
+    /// Pertaining the volume group parent.
+    vg: super::vg_pool::QueryArgs,
+    /// Pertaining the logical volume itself.
+    lv: CmnQueryArgs,
+    /// Our lv's are "special", as in we set the uuid to the name, and add the
+    /// name as a lv_tag (LVM_TAG.lv.name=name)
+    regular_lv: bool,
```
Abhinandan-Purkait, line 26: What do we mean by add name as lv_tag? What is the name here? Volume uuid?

Abhinandan-Purkait: Also do we expect lv name to be uuidv4 always? If so can it be an issue while loading non mayastor created LVs?

io-engine/src/lvm/lv_replica.rs
```diff
@@ -0,0 +1,739 @@
...
+    }
+    pub(crate) fn vg_name(&self) -> &str {
+        &self.vg_name
+    }
+    pub(crate) fn vg_uuid(&self) -> &str {
+        &self.vg_uuid
+    }
+    pub(crate) fn extent_size(&self) -> u64 {
+        self.vg_extent_size
+    }
```
Abhinandan-Purkait: Doc comments needed for pubs ?

io-engine/src/lvm/vg_pool.rs
```diff
@@ -0,0 +1,370 @@
...
+        self.export().await?;
+        info!("LVM pool '{name}' has been destroyed successfully");
+        self.ptpl().destroy().ok();
+        Ok(())
+    }
+
+    /// Exports the volume group by unloading all logical volumes and finally
+    /// removing our tag from it
+    pub(crate) async fn export(&mut self) -> Result<(), Error> {
+        // todo: unload all logical volumes from this VG
```
Abhinandan-Purkait: We cannot delete a pool having replicas right, so should we fail? And even if we want to destroy in use pool, should we not delete the replicas(spdk bdevs) as we created them instead of export?

Question: Since we are using LVM as pool backend so I believe we don't need to deal with LVM LVs right? The only LV we need to consider is on top of VG on which we create spdk pool?

Abhinandan-Purkait (approved):

io-engine/src/lvm/vg_pool.rs
tiagolobocastro: Yeah I have spoken with @dsavitskiy about this and it should be possible, though we need to support existing non-gpt pools as well.
For LVM backends this is probably not necessary as disks will already be taken by LVM itself.

io-engine/src/lvm/vg_pool.rs
tiagolobocastro: Export usually means just "shutdown" the pool so it can be imported elsewhere. Probably we shouldn't even remove our tag from it?

dsharma-dc: Since we are doing export upon `destroy` as well here, the same disks can probably be used for an lvs pool as well after this, right? In that case it makes sense to cleanup whatever could be left behind from lvm pool, even though lvs pool will write it's own metadata.

tiagolobocastro: That should be as part of `destroy` and not export though. Currently we don't destroy because create is not creating, it's just importing because it's what the existing PR's did. Perhaps better to change that now to create and destroy, and don't support import existing until we have that figured out?

io-engine/src/lvm/vg_pool.rs
tiagolobocastro: As discussed, there is no spdk pool on top of the VG, lvm is a completely separate backend on this PR.

io-engine/src/grpc/v1/lvm/pool.rs
tiagolobocastro: hmm doesn't sound right though, I think an is correct here as it's pronounced “el-vee-ess”

io-engine/src/lvm/vg_pool.rs
dsharma-dc: > That should be as part of destroy and not export though.

Yeah, destroy = cleanup + export

Yes the most obvious logical next step should now be to integrate create and destroy. Until then, this should remain experimental(with a disclaimer). Do we plan to add some documentation/howto at this point w.r.t this feature?

io-engine/src/lvm/vg_pool.rs
tiagolobocastro: We now handle the lifecycle of the lvm backend pool (vg).

io-engine/src/lvm/vg_pool.rs
tiagolobocastro: We now do create and destroy

io-engine/src/grpc/v1/pool.rs
tiagolobocastro: We can separate, caller can just select the ones it wants.
In case caller selects all and an error is seen, caller side can check this metadata.

io-engine/src/lvm/lv_replica.rs
tiagolobocastro: As per the comment, we add the replica name as a tag, eg: mayastor.lv.name=replica1
If loading non-mayastor LVs then probably LV will have to be renamed; 
We might be able to do the opposite and create with name, setting up uuid as mayastor.lv.uuid=uuid but then the lv_path would have the name and not the uuid.

io-engine/src/lvm/lv_replica.rs
tiagolobocastro: Done

io-engine/src/lvm/vg_pool.rs
```diff
@@ -149,6 +178,14 @@ impl VolumeGroup {
             .with_vg(CmnQueryArgs::ours().uuid(self.uuid()).named(self.name()));
         LogicalVolume::list(&query).await
     }
+    async fn list_foreign_lvs(&self) -> Result<Vec<LogicalVolume>, Error> {
```
dsharma-dc, line 181: Doc comments are missing for many functions here.

dsharma-dc, line 181: Might consider naming as `list_foreign_lvm_lvols` to avoid confusion with `lvs` backend. Same with `list_lvs`

dsharma-dc (approved):

io-engine/src/lvm/vg_pool.rs
orville-wright: Here's how I think about this wrt our 2024 Roadmap thinking...
lvs = SPDK LVS, Lvol, Bdev backend
lvm = LVM backend
lzp = ZFS Zpool backend

io-engine/src/lvm/vg_pool.rs
orville-wright: The LVM VG is a new Backend disk media ownership + management subsystem. SPDK and Blobstor are not involved anywhere in the disk management architecture. The user are making an explicit decision on the type of pool and the functionality they are looking for.
- LVM = LVM exclusively owns + manages the backend disk media & provides RAID 0, 1, 4, 5, 6, and 10 with full volume aggregation but average performance
- LVS = SPDK exclusively owns + manages the backend disk media but provides no RAID (yet) or disk aggregation (yet), but delivers very fast deep NVMe integration
- LZP = ZFS Zpool exclusively owns + manages the backend disk media & provides RAID 0, 1, Z (5), Z2 (6), Z3, and 10 with full volume aggregation and superior Bit-rot protection, Snapshot send/recv/streaming, Dedupe, Compression, and Flash optimizations (ARC and L2ARC).

io-engine/src/lvm/vg_pool.rs
dsharma-dc: > Doc comments are missing for many functions here.

Ok, those are non pubs

io-engine/src/core/env.rs
dsavitskiy: Also, could you add a comment description?

io-engine/src/bin/io-engine.rs
```diff
@@ -89,6 +89,11 @@ fn start_tokio_runtime(args: &MayastorCliArgs) {
     if !ENABLE_NEXUS_RESET.load(Ordering::SeqCst) {
         warn!("Nexus reset is disabled");
     }
+    if args.lvm {
+        env::set_var("LVM", "1");
+        env::set_var("LVM_SUPPRESS_FD_WARNINGS", "1");
+        warn!("LVM pool backend is enabled");
```
dsavitskiy: Shouldn't this be an info?
The other messages, like "warn!("Nexus reset is disabled");" are warnings because they are about some functionality _disabled_.

io-engine/src/bin/io-engine.rs
```diff
@@ -89,6 +89,11 @@ fn start_tokio_runtime(args: &MayastorCliArgs) {
     if !ENABLE_NEXUS_RESET.load(Ordering::SeqCst) {
         warn!("Nexus reset is disabled");
     }
+    if args.lvm {
+        env::set_var("LVM", "1");
+        env::set_var("LVM_SUPPRESS_FD_WARNINGS", "1");
```
dsavitskiy: Probably, we shouldn't override this var, if it was set explicitly to "0".
Also, should we document somewhere that we override the environment?

ci.nix
```diff
@@ -72,6 +72,7 @@ mkShell {
   SPDK_PATH = if nospdk then null else "${spdk}";
   FIO_SPDK = if nospdk then null else "${spdk}/fio/spdk_nvme";
   ETCD_BIN = "${etcd}/bin/etcd";
+  LVM_BINS = "${lvm2.bin}/bin";
```
dsavitskiy, line 75: Probably we have to allow to disable LVM.

io-engine/src/core/share.rs
```diff
@@ -119,6 +120,10 @@ impl ShareProps {
     pub fn ptpl(&self) -> &Option<PtplProps> {
         &self.ptpl
     }
+    /// Convert to `UpdateProps`.
+    pub fn into_update_props(self) -> UpdateProps {
```
dsavitskiy: Consider using "idiomatic Rust" approach with From<> trait.

dsavitskiy (changes requested):

dsavitskiy:
This PR seems to be quite big and intrusive, and there are concerns about the implementation decisions, e.g. like communication with LVM with CLI tools, etc.
Shouldn't we go through our full feature design process as we do for other features? HLD, LLD, BDDs, and so on.

tiagolobocastro:
> This PR seems to be quite big and intrusive, and there are concerns about the implementation decisions, e.g. like communication with LVM with CLI tools, etc. Shouldn't we go through our full feature design process as we do for other features? HLD, LLD, BDDs, and so on.

This is a re-raise of existing PR's which did not get merged (not sure if there was a design for that or not, probably left in the old wiki), so we're finally merging them now (with a rebase and refactor).
This is an experimental feature and not enabled by default.

io-engine/src/bin/io-engine.rs
tiagolobocastro: The intent of the warn is because lvm support is still experimental, but I can change to info if you think it's better

io-engine/src/bin/io-engine.rs
tiagolobocastro: Done

io-engine/src/core/share.rs
tiagolobocastro: Done

ci.nix
tiagolobocastro: This doesn't control lvm enablement, only brings the binaries to the shell.

io-engine/src/bin/io-engine.rs
dsavitskiy: Ah, I see. In that case, shouldn't it explicitly state that it is experimental? "Experimental LVM pool backend is enabled".

dsavitskiy:
My point is that maybe it is worth it to have a couple of meetings to (re-)review the current design? Besides low level implementation considerations, there are a couple of topics that are potentially affected by LVM backend:
- Replica/volume expansion.
- Disk aggregation.

tiagolobocastro:
> My point is that maybe it is worth it to have a couple of meetings to (re-)review the current design? Besides low level implementation considerations, there are a couple of topics that are potentially affected by LVM backend:
> 
>     * Replica/volume expansion.
> 
>     * Disk aggregation.

Makes sense, I've sent an invite post-standup tomorrow

io-engine/src/core/env.rs
tiagolobocastro: Done

io-engine/src/bin/io-engine.rs
tiagolobocastro: Done

io-engine/src/lvm/property.rs
```diff
@@ -0,0 +1,190 @@
+//! Properties are attributes which are persisted in a given LVM resource and
+//! which can be used to identify or retrieve specific information from a
+//! resource, even across reboots.
+
+use crate::core::Protocol;
+use std::str::FromStr;
+
+#[macro_export]
+macro_rules! impl_properties {
```
dsavitskiy, line 9: Probably, a crate already exists for this.
Serde would do, at least for some stuff.
If there's no crate suitable for the purpose, this code is better to be moved to a separate module.

dsavitskiy (approved):

io-engine/src/lvm/property.rs
tiagolobocastro: I'll add a ticket to research this

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1590, opened 2024-02-19
feat(nexus/resize): add support for resizing a nexus

dsharma-dc:


dsharma-dc:
bors try

dsharma-dc:
bors try

dsharma-dc:
bors try

io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -725,7 +731,23 @@ impl<'n> Nexus<'n> {
             }
 
             match partition::calc_data_partition(self.req_size(), nb, bs) {
-                Some((start, end)) => {
+                Some((start, end, req_blocks)) => {
+                    // During expansion - if the requested number of blocks
+                    // aren't available on any child device,
+                    // then the operation need to fail in entirety. Hence make
```
tiagolobocastro:
```suggestion
                    // then the operation needs to fail in its entirety. Hence make
```

io-engine/src/bdev/nexus/nexus_bdev_error.rs
```diff
@@ -281,6 +286,12 @@ impl From<Error> for tonic::Status {
             Error::RebuildJobNotFound {
                 ..
             } => Status::not_found(e.to_string()),
+            Error::NexusIncomplete {
+                ..
+            } => Status::failed_precondition(e.to_string()),
+            Error::NexusResize {
+                ..
+            } => Status::failed_precondition(e.to_string()),
```
tiagolobocastro, line 294: Either use e.verbose() here or add source to snafu display above

io-engine/src/bin/io-engine-client/v1/nexus_cli.rs
```diff
@@ -564,6 +576,54 @@ async fn nexus_children_2(
...
+            println!(
+                "{}",
+                serde_json::to_string_pretty(&response.get_ref())
+                    .unwrap()
+                    .to_colored_json_auto()
+                    .unwrap()
+            );
+        }
+        OutputFormat::Default => {
+            println!("{}", &uuid,);
```
tiagolobocastro:
```suggestion
            println!("Resized nexus {uuid} to {size}");
```

io-engine/src/grpc/v1/nexus.rs
```diff
@@ -484,6 +484,34 @@ impl NexusRpc for NexusService {
...
 
+    #[named]
+    async fn resize_nexus(
+        &self,
+        request: Request<ResizeNexusRequest>,
+    ) -> GrpcResult<ResizeNexusResponse> {
+        let ctx = GrpcClientContext::new(&request, function_name!());
+        let args = request.into_inner();
+
+        self.serialized(ctx, args.uuid.clone(), true, async move {
```
tiagolobocastro: Does this need to be a global operation?

io-engine/tests/nexus_replica_resize.rs
```diff
@@ -0,0 +1,324 @@
...
+// Define an enum to represent the functions
+enum ResizeTest {
+    WithoutReplicaResize,
+    AfterReplicaResize,
+    WithRebuildingReplica,
+}
+
+// Define a trait for the test functions
+#[async_trait(?Send)]
+trait ResizeTestTrait {
```
tiagolobocastro, line 67: Is the trait needed, seems simple impl on ResizeTest would achieve the same?

tiagolobocastro (approved):

io-engine/tests/nexus_replica_resize.rs
```diff
@@ -0,0 +1,324 @@
...
+async fn compose_ms_nodes() -> io_engine_tests::compose::ComposeTest {
+    common::composer_init();
+
+    Builder::new()
+        .name("cargo-test")
+        .network("10.1.0.0/16")
+        .unwrap()
+        .add_container_bin(
+            "ms_nex_0",
+            Binary::from_dbg("io-engine").with_args(vec!["-l", "1,2"]),
```
niladrih, line 36: Doesn't need a vec I guess. Maybe an array would do?
```suggestion
            Binary::from_dbg("io-engine").with_args(&["-l", "1,2"]),
```

niladrih (approved):
LGMT. Left one nitpick comment.

io-engine/tests/nexus_replica_resize.rs
dsharma-dc: > Doesn't need a vec I guess. Maybe an array would do?

This is consistent in all the existing io-engine cargo tests.

io-engine/src/grpc/v1/nexus.rs
dsharma-dc: No. I referred the `create` or `destroy` methods perhaps. We don't need global lock in this case.

io-engine/src/bdev/nexus/nexus_bdev_error.rs
dsharma-dc: Added verbose for `NexusIncomplete`, for `NexusResize` source is present.

io-engine/tests/nexus_replica_resize.rs
dsharma-dc: > Is the trait needed, seems simple impl on ResizeTest would achieve the same?

In this case probably yes. However, It gives me an impression that we should standardize all the tests(at least new ones) in similar or even more generic manner by defining per test file trait and implement it for a bigger enum that encompasses all test files. Some of those chunks can be put in a common library file. I'll take it up on me, and keep this one as is for now.

io-engine/src/grpc/v1/nexus.rs
dsharma-dc: > Does this need to be a global operation?

No. I referred the create or destroy methods perhaps. We don't need global lock in this case. Changed to `false`.

dsharma-dc:
bors try

dsharma-dc:
bors merge

PR #1594, opened 2024-02-21
Rebuild Rangers

tiagolobocastro:
refactor(rebuild): use new rebuild rangers
    
    Make use of the rebuild rangers to configure rebuild types.
    This allows us to remove the setting of the rebuild map being done
    after the rebuild job is created for the nexus and removing it from
    the shared rebuild descriptor.
    
    The nexus still uses the partial but sequential rebuild to reduce
    the scope of changes.
    Once the fully partial rebuild is validated we can switch the nexus
    to it.
    
---

    feat(rebuild): add rebuild rangers
    
    Adds a set of rebuild rangers which can be used by different rebuilds:
    1. full rebuild - walk the entire device range and copy every segment
       (current nexus full rebuild behaviour).
    2. partial rebuild - walk the allocated segments only and copy them.
    3. partial eq rebuild - walk the entire device range and copy only allocated
       segments (current nexus partial rebuild behaviour).
    
### DIFF
--- a/io-engine/src/core/segment_map.rs
+++ b/io-engine/src/core/segment_map.rs
@@ -96,4 +96,15 @@ impl SegmentMap {
     pub(crate) fn count_dirty_blks(&self) -> u64 {
         self.count_ones() * self.segment_size / self.block_len
     }
+
+    /// Get the segment size in blocks.
+    pub(crate) fn segment_size_blks(&self) -> u64 {
+        self.segment_size / self.block_len
+    }
+}
+
+impl From<SegmentMap> for BitVec {
+    fn from(value: SegmentMap) -> Self {
+        value.segments
+    }
 }
--- a/io-engine/src/rebuild/nexus_rebuild.rs
+++ b/io-engine/src/rebuild/nexus_rebuild.rs
@@ -1,16 +1,23 @@
+use futures::channel::oneshot;
 use snafu::ResultExt;
 use spdk_rs::LbaRange;
-use std::{
-    ops::{Deref, Range},
-    rc::Rc,
-};
+use std::ops::{Deref, Range};
 
 use crate::{
     core::{DescriptorGuard, UntypedBdev},
     gen_rebuild_instances,
     rebuild::{
         rebuild_error::{RangeLockFailed, RangeUnlockFailed},
+        rebuild_job_backend::RebuildJobManager,
         rebuild_task::{RebuildTask, RebuildTaskCopier},
+        rebuilders::{
+            FullRebuild,
+            PartialSeqCopier,
+            PartialSeqRebuild,
+            RangeRebuilder,
+        },
+        RebuildMap,
+        RebuildState,
     },
 };
 
@@ -30,18 +37,37 @@ use super::{
 /// that there is no concurrent between the nexus and the rebuild.
 /// This is a frontend interface that communicates with a backend runner which
 /// is the one responsible for the read/writing of the data.
-pub struct NexusRebuildJob(RebuildJob);
+pub struct NexusRebuildJob {
+    job: RebuildJob,
+}
+
+/// Nexus supports both full and partial rebuilds. In case of a partial rebuild
+/// we have to provide the nexus rebuild job with a `RebuildMap`.
+/// However, this can only be provided after we've created the rebuild as taking
+/// the map from the nexus children is an operation which cannot be undone
+/// today.
+/// Therefore we use a rebuild job starter which creates the initial bits
+/// required for the rebuild job and which can be started later with or without
+/// the `RebuildMap`.
+pub struct NexusRebuildJobStarter {
+    /// The job itself is optional because it gets taken when we want to store.
+    /// After the job is taken, we then can schedule either a full or a partial
+    /// rebuild with the backend.
+    job: Option<NexusRebuildJob>,
+    manager: RebuildJobManager,
+    backend: NexusRebuildJobBackendStarter,
+}
 
 impl std::fmt::Debug for NexusRebuildJob {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        self.0.fmt(f)
+        self.job.fmt(f)
     }
 }
 impl Deref for NexusRebuildJob {
     type Target = RebuildJob;
 
     fn deref(&self) -> &Self::Target {
-        &self.0
+        &self.job
     }
 }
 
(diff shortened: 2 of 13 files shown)


io-engine/src/core/segment_map.rs
```diff
@@ -95,4 +95,9 @@ impl SegmentMap {
     pub(crate) fn count_dirty_blks(&self) -> u64 {
         self.count_ones() * self.segment_size / self.block_len
     }
+
+    /// Convert the segments into a bit map.
```
dsharma-dc: We are not converting per se, just returning `segments` which is already a BitVec

io-engine/src/rebuild/rebuilders.rs
```diff
@@ -0,0 +1,249 @@
...
+/// A rebuild may rebuild a device by walking it differently, for example:
+/// 1. full rebuild - walk the entire device range and copy every segment
+///    (current nexus full rebuild behaviour).
+/// 2. partial rebuild - walk the allocated segments only and copy them.
+/// 3. partial eq rebuild - walk the entire device range and copy only allocated
+///    segments (current nexus partial rebuild behaviour).
+pub(super) trait RangeRebuilder<T: RebuildTaskCopier> {
+    /// Get the next block to rebuild.
+    fn next(&mut self) -> Option<u64>;
+    /// Get the next block to rebuild.
```
dsharma-dc: `/// Peek the next block to rebuild.`

io-engine/src/rebuild/rebuilders.rs
```diff
@@ -0,0 +1,249 @@
...
+            }
+        }
+        None
+    }
+    fn peek_next(&self) -> Option<u64> {
+        // todo: should we add a wrapper to ensure we peek only set bits?
+        self.range.peek().map(|(blk, _)| *blk as u64)
+    }
+
+    fn blocks_remaining(&self) -> u64 {
```
dsharma-dc, line 115: This should account for only set bits while counting blocks remaining. But I'm not sure how that'll be achieved here. With peek we can only peek one step.

io-engine/src/rebuild/rebuilders.rs
```diff
@@ -0,0 +1,247 @@
...
+    fn next(&mut self) -> Option<u64> {
+        self.range.next()
+    }
+    fn peek_next(&self) -> Option<u64> {
+        self.range.peek().cloned()
+    }
+
+    fn blocks_remaining(&self) -> u64 {
+        self.peek_next()
+            .map(|r| self.desc().range.end.max(r) - r)
```
dsharma-dc, line 59: Also, this is going to return segments remaining, if I read that correctly. We need to multiply by `segment_size_blks`

dsharma-dc: Alright no, the step is set to `segment_size_blks`, so we're good.

io-engine/src/rebuild/rebuilders.rs
tiagolobocastro: hmm yeah, I guess we need to record the total bits set at the start, and decrement for every copy

io-engine/src/rebuild/nexus_rebuild.rs
```diff
@@ -30,18 +37,26 @@ use super::{
 /// that there is no concurrent between the nexus and the rebuild.
 /// This is a frontend interface that communicates with a backend runner which
 /// is the one responsible for the read/writing of the data.
-pub struct NexusRebuildJob(RebuildJob);
+pub struct NexusRebuildJob {
+    job: RebuildJob,
+}
+
+pub struct NexusRebuildJobStarter {
```
hrudaya21, line 52: For Snapshot rebuild, do We need to have similar structure ? Can We have generic rebuild starter ?

io-engine/src/rebuild/nexus_rebuild.rs
```diff
@@ -30,18 +37,26 @@ use super::{
...
+    job: Option<NexusRebuildJob>,
```
hrudaya21, line 56: Why job is  Option ?

io-engine/src/core/segment_map.rs
tiagolobocastro: Replaced with `From` trait

io-engine/src/rebuild/rebuilders.rs
tiagolobocastro: Indeed, done

io-engine/src/rebuild/nexus_rebuild.rs
tiagolobocastro: No, this is only needed for nexus because the rebuild map is added later, after we create the rebuild job.

io-engine/src/rebuild/nexus_rebuild.rs
tiagolobocastro: Because it's taken once we store it. I've added doc comments for this.

hrudaya21 (approved):

io-engine/src/rebuild/rebuilders.rs
```diff
@@ -0,0 +1,255 @@
...
+            segment_size,
+            copier: Rc::new(copier),
+        }
+    }
+}
+impl<T: RebuildTaskCopier> RangeRebuilder<T> for PartialRebuild<T> {
+    fn next(&mut self) -> Option<u64> {
+        for (blk, is_set) in self.range.by_ref() {
+            if is_set {
+                self.rebuilt_blks += self.segment_size;
```
dsharma-dc: this should be `self.rebuilt_blks += self.segment_size_blks`

io-engine/src/rebuild/rebuilders.rs
```diff
@@ -0,0 +1,255 @@
...
+        }
+        None
+    }
+    fn peek_next(&self) -> Option<u64> {
+        // todo: should we add a wrapper to ensure we peek only set bits?
+        self.range.peek().map(|(blk, _)| *blk as u64)
+    }
+
+    fn blocks_remaining(&self) -> u64 {
+        self.total_blks.max(self.rebuilt_blks) - self.rebuilt_blks
```
dsharma-dc: we don't need to use `max` here, instead simple `self.total_blks - self.rebuilt_blks` should work. If self.rebuilt_blks goes larger than total_blks, then something else is really wrong.

io-engine/src/rebuild/rebuilders.rs
tiagolobocastro: I've renamed it to _blks

io-engine/src/rebuild/rebuilders.rs
tiagolobocastro: Done

tiagolobocastro:
bors try

dsharma-dc (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
pytest CI also broken on dataplane :(

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1596, opened 2024-02-23
feat(replica_entity_id): adding implementation

abhilashshetty04:


io-engine/src/grpc/v1/replica.rs
```diff
@@ -137,6 +137,7 @@ impl From<Lvol> for Replica {
             is_snapshot: l.is_snapshot(),
             is_clone: l.is_snapshot_clone().is_some(),
             snapshot_uuid: source_uuid,
+            entity_id,
```
hrudaya21: entity_id parameter can be initialized here, as it is not used in any other place.

io-engine/src/grpc/v1/replica.rs
```diff
@@ -253,6 +254,9 @@ impl ReplicaRpc for ReplicaService {
                     }
                     Ok(lvol) => {
                         debug!("created lvol {:?}", lvol);
+                        if let Some(v) = &args.entity_id {
+                            let _ = lvol.set_blob_attr("entity_id", v.to_string(), true).await;
```
hrudaya21, line 256: If set_blob_attr is failed, how control-plane will know about this error. I think error should be return back from dataplane which can be used in control-plane to mark the replica creation fail.

io-engine/src/grpc/v1/replica.rs
```diff
@@ -532,10 +536,41 @@ impl ReplicaRpc for ReplicaService {
...
+            async move {
+                let args = request.into_inner();
+                info!("{args:?}");
+                let rx = rpc_submit::<_, _, LvsError>(async move {
+                    if let Some(bdev) =
+                        UntypedBdev::lookup_by_uuid_str(&args.uuid)
+                    {
+                        let lvol = Lvol::try_from(bdev)?;
+                        let _ = lvol
+                            .set_blob_attr("entity_id", args.entity_id, true)
```
hrudaya21, line 553: Here also need to catch the error return from `set_blob_attr` function and return back proper error code to control-plane. Possibly a new error code can be introduced.

io-engine/src/grpc/v1/replica.rs
abhilashshetty04: Do we want to fail create_replica if set_xattr fails? Even if we return error anyway, Im not sure what control plane can do? Should it call `set_replica_entity_id` if error type matches? 

 We will call `set_replica_entity_id` grpc anyway incase `entity_id` is None while `replica_share`

io-engine/src/grpc/v1/replica.rs
abhilashshetty04: It can hold off volume_publish call btw...

io-engine/src/grpc/v1/replica.rs
abhilashshetty04: @hrudaya21 , Sure

abhilashshetty04: Done, Made same change for snapshot_uuid as well.

abhilashshetty04:
bors try

io-engine/src/grpc/v1/replica.rs
hrudaya21: As We are sending back Option<entity_id> in the response, I think it is okay to mark the operation success but are We planning to do any action if entity_id is set to None in the create replica request or It will use the `set_replica_entity_id` grpc only during share replica flow ?

io-engine/src/grpc/v1/replica.rs
hrudaya21: Do We want to success the volume publish with this grpc failure ? Then how this attribute can be credible for the replica in control-plane, as in some scenario, volume can published without setting the attribute.

hrudaya21 (approved):

io-engine/src/grpc/v1/replica.rs
abhilashshetty04: As discussed, We are currently using entity_id for stats metadata. So better we handle error silently.

dsavitskiy (approved):

abhilashshetty04:
bors merge

io-engine/src/grpc/v1/replica.rs
tiagolobocastro: Seems a bit to not fail here when the one outcome we want to from this is to set the entity id!??

io-engine/src/grpc/v1/replica.rs
tiagolobocastro: I think we should fail the operation, because clearly the expectation that the replica has entity id is not met. What do you expect the control-plane to do?

PR #1597, opened 2024-02-28
fix(rebuild/partial): set bits for written blks only

tiagolobocastro:
The blk used to calculate the end segment was offset by 1 which can lead to rebuilding more data than the one which was actually written into. Example, if we write 1 blk to offset 4, then we only need to rebuild offset 4, and not offset 5 (4 + 1).


tiagolobocastro:
bors try

dsharma-dc (approved):

hrudaya21 (approved):

Abhinandan-Purkait (approved):

dsavitskiy (approved):

tiagolobocastro:
bors merge

PR #1598, opened 2024-02-28
feat(replica_stats): adding implementation

abhilashshetty04:


io-engine/src/bin/io-engine-client/v1/stats_cli.rs
```diff
@@ -24,6 +24,13 @@ pub fn subcommands() -> Command {
             .help("Volume target/nexus name"),
     );
 
+    let replica = Command::new("replica").about("Get Replica IO Stats").arg(
+        Arg::new("name")
```
dsharma-dc, line 28: Usually the replica CLI is based on replica uuids, why have we decided to use name here?

io-engine/src/grpc/v1/stats.rs
```diff
@@ -216,6 +216,51 @@ impl StatsRpc for StatsService {
...
+            let rx = rpc_submit::<_, _, CoreError>(async move {
+                let mut lvol = Vec::new();
+                if let Some(bdev) = UntypedBdev::bdev_first() {
+                    lvol = bdev
+                        .into_iter()
+                        .filter(|b| b.driver() == "lvol")
+                        .map(|b| Lvol::try_from(b).unwrap())
+                        .collect();
+                }
+                let replica_stats_future: Vec<_> = if let Some(name) = args.name
```
dsharma-dc, line 226: We should first check if `name` is present in args and search for the specific lvol. If `name` not provided, then take all lvol bdevs. Will save time.

io-engine/src/bin/io-engine-client/v1/stats_cli.rs
abhilashshetty04: repica.name and replica.uuid will have same uuid. This field is internal only.  In the current replica filtering logic also we match lvol.name == arg.name.

Moreover,  IoStats struct returns name a particualr device.

tiagolobocastro: that is not true, name==uuid is just an implementation detail of how control-plane names replicas today

tiagolobocastro: you can have replica with name replica1 and uuid b4865f3d-5c20-454d-949b-23369fdf94e9

io-engine/src/bin/io-engine-client/v1/stats_cli.rs
```diff
@@ -235,6 +244,116 @@ async fn nexus(mut ctx: Context, matches: &ArgMatches) -> crate::Result<()> {
...
+                    ctx.v1("No Replica IoStats found");
+                }
+                return Ok(());
+            }
+
+            let table = stats
+                .iter()
+                .map(|p| {
+                    let io_stat = p.stats.as_ref().unwrap();
+                    let read_latency = ticks_to_time(
```
tiagolobocastro: I wonder if this can be captured in a closure, and use it below on the vec as such:
vec![ticks_to_time(io_state.write_latency_ticks), ticks_to_time(io_state.read_latency_ticks), ...

io-engine/src/grpc/v1/stats.rs
abhilashshetty04: We dont have a way of directly getting a specific lvol. Like how we have lookup fn for Lvs and Nexus.

So we list all bdev, then filter of type lvol and get a specifc lvol if arg.name is present. 

Exporter will always list all replicas. We have gvien an option to specify name as it will be easier for us to filter a particular resource stat with io-engine-client.

io-engine/src/grpc/v1/stats.rs
```diff
@@ -278,6 +323,18 @@ async fn get_stats(
...
 }
 
+/// Returns IoStats for a given Lvol(Replica).
+async fn replica_stats(lvol: Lvol) -> Result<ReplicaIoStats, CoreError> {
+    let stats = lvol.as_bdev().stats_async().await?;
+    let io_stat =
+        IoStats::from(ExternalType((lvol.name(), lvol.uuid(), stats)));
+    let replica_stat = ReplicaIoStats {
+        stats: Some(io_stat),
+        entity_id: Lvol::get_blob_xattr(lvol.blob_checked(), "entity_id"),
```
tiagolobocastro: is this reading from the metadata in memory? I hope it's not causing a read IO..

abhilashshetty04: Ill check. You mean if this can alter read stats?

io-engine/src/bin/io-engine-client/v1/stats_cli.rs
abhilashshetty04: Right,

For this change, ListStatsOptions has name field,  Since we return IoStats also wiith stats.name . So thought of naming the cli arg as name.

io-engine/src/grpc/v1/stats.rs
abhilashshetty04: It doesnt change any ready stats for the replica when lvol xattr is read.

io-engine/src/grpc/v1/stats.rs
dsharma-dc: ok, but we can still optimise. If the args.name is provided and this lvol was the first registered bdev(`bdev_first()`) among say 1000s, then the current logic will first iterate over 1000, and then weed out 999 to keep this one. Instead if we run iteration to find this args.name with `.filter(|b| b.driver() == "lvol") && b.name() == args.name` , then we don't need to go all the way till 1000 always.

abhilashshetty04: Even if we filter with name also along with lvol . .`filter(|b| b.driver() == "lvol") && b.name() == args.name` . This doesnt break after getting a first match. It will proceed iterating to the end of elements right?

Ill check if we can use someother method inplace of filter

dsharma-dc: There is I think similar  `find` or`find_map` that can be used. They short-circuit upon finding the matching value.

abhilashshetty04: Done

io-engine/src/bin/io-engine-client/v1/stats_cli.rs
abhilashshetty04: Incorporated

io-engine/src/bin/io-engine-client/v1/stats_cli.rs
```diff
@@ -235,6 +244,93 @@ async fn nexus(mut ctx: Context, matches: &ArgMatches) -> crate::Result<()> {
...
+                        adjust_bytes(io_stat.bytes_written),
+                        io_stat.num_unmap_ops.to_string(),
+                        adjust_bytes(io_stat.bytes_unmapped),
+                        ticks_time(io_stat.read_latency_ticks, tick_rate),
+                        ticks_time(io_stat.write_latency_ticks, tick_rate),
+                        ticks_time(io_stat.unmap_latency_ticks, tick_rate),
+                        ticks_time(io_stat.max_read_latency_ticks, tick_rate),
+                        ticks_time(io_stat.min_read_latency_ticks, tick_rate),
+                        ticks_time(io_stat.max_write_latency_ticks, tick_rate),
+                        ticks_time(io_stat.min_write_latency_ticks, tick_rate),
```
tiagolobocastro: As I mentioned you don't need to pass tick_rate, you can capture it in the closure:

```suggestion
            let table = stats
                .iter()
                .map(|p| {
                    let io_stat = p.stats.as_ref().unwrap();
                    let tick_rate = io_stat.tick_rate;
                    let ticks_time = |ticks| -> String {
                        ticks_to_time(ticks, tick_rate).to_string()
                    };
                    vec![
                        io_stat.name.clone(),
                        io_stat.num_read_ops.to_string(),
                        adjust_bytes(io_stat.bytes_read),
                        io_stat.num_write_ops.to_string(),
                        adjust_bytes(io_stat.bytes_written),
                        io_stat.num_unmap_ops.to_string(),
                        adjust_bytes(io_stat.bytes_unmapped),
                        ticks_time(io_stat.read_latency_ticks),
                        ticks_time(io_stat.write_latency_ticks),
                        ticks_time(io_stat.unmap_latency_ticks),
                        ticks_time(io_stat.max_read_latency_ticks),
                        ticks_time(io_stat.min_read_latency_ticks),
                        ticks_time(io_stat.max_write_latency_ticks),
                        ticks_time(io_stat.min_write_latency_ticks),
```

abhilashshetty04: Incorporated

dsharma-dc (approved):

io-engine/src/bin/io-engine-client/v1/stats_cli.rs
tiagolobocastro: Probably we should have both uuid and name as arguments, but as different PR as we'd have to change the api, wdyt @dsharma-dc  @abhilashshetty04 ?

io-engine/src/grpc/v1/stats.rs
```diff
@@ -216,6 +216,60 @@ impl StatsRpc for StatsService {
...
+            let args = request.into_inner();
+            let rx = rpc_submit::<_, _, CoreError>(async move {
+                let replica_stats_future: Vec<_> = if let Some(name) = args.name
+                {
+                    if let Some(bdev) = UntypedBdev::bdev_first() {
+                        if let Some(lvol) = bdev
+                            .into_iter()
+                            .find(|b| {
+                                (b.driver() == "lvol")
+                                    && (*b.name().to_string() == name)
```
tiagolobocastro: name returns &str?
```suggestion
                                    && (b.name() == name)
```

io-engine/src/grpc/v1/stats.rs
```diff
@@ -216,6 +216,60 @@ impl StatsRpc for StatsService {
...
+                            })
+                            .and_then(|b| Lvol::try_from(b).ok())
+                        {
+                            vec![replica_stats(lvol.clone())]
```
tiagolobocastro:
```suggestion
                            vec![replica_stats(lvol)]
```

io-engine/src/grpc/v1/stats.rs
```diff
@@ -216,6 +216,60 @@ impl StatsRpc for StatsService {
...
+                        } else {
+                            vec![]
+                        }
+                    } else {
+                        vec![]
```
tiagolobocastro: can probably reduce all this nesting with map and unwrap_or_default

io-engine/src/grpc/v1/stats.rs
```diff
@@ -216,6 +216,60 @@ impl StatsRpc for StatsService {
...
+                    if let Some(bdev) = UntypedBdev::bdev_first() {
+                        lvols = bdev
+                            .into_iter()
+                            .filter(|b| b.driver() == "lvol")
+                            .map(|b| Lvol::try_from(b).unwrap())
+                            .collect();
+                    }
+                    lvols
+                        .iter()
+                        .map(|lvol| replica_stats(lvol.clone()))
```
tiagolobocastro:
```suggestion
                        .into_iter()
                        .map(|lvol| replica_stats(lvol))
```

io-engine/src/grpc/v1/stats.rs
```diff
@@ -278,6 +332,18 @@ async fn get_stats(
...
 }
 
+/// Returns IoStats for a given Lvol(Replica).
+async fn replica_stats(lvol: Lvol) -> Result<ReplicaIoStats, CoreError> {
+    let stats = lvol.as_bdev().stats_async().await?;
+    let io_stat =
+        IoStats::from(ExternalType((lvol.name(), lvol.uuid(), stats)));
+    let replica_stat = ReplicaIoStats {
+        stats: Some(io_stat),
+        entity_id: Lvol::get_blob_xattr(lvol.blob_checked(), "entity_id"),
```
tiagolobocastro: Isn't there a method entity_id on the lvol? If not, let's add one

tiagolobocastro (approved):

abhilashshetty04:
bors merge

PR #1601, opened 2024-03-06
feat(xattr_error): handle xattr set fail

abhilashshetty04:


io-engine/src/grpc/v1/replica.rs
```diff
@@ -253,7 +254,14 @@ impl ReplicaRpc for ReplicaService {
...
                         debug!("created lvol {:?}", lvol);
                         if let Some(v) = &args.entity_id {
-                            let _ = lvol.set_blob_attr("entity_id", v.to_string(), true).await;
+                            if let Err(e) = lvol.set_entity_id(v.to_string()).await {
+                                return Err(
+                                    LvsError::SetXAttr {
+                                        source: e.to_errno(),
+                                        attr: "entity_id".to_string(),
+                                        name: args.uuid
+                                    })
```
tiagolobocastro: can't set_entity_id return LvsError already?

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -572,6 +572,9 @@ pub trait LvsLvol: LogicalVolume + Share {
         prop: PropValue,
     ) -> Result<(), Error>;
 
+    /// Sets entity_id with given arg to Lvol xattr.
+    async fn set_entity_id(&self, id: String) -> Result<(), Error>;
```
tiagolobocastro: Ah looks like we have existing property below, I think you should add entityId as another enum variant to PropValue

io-engine/src/grpc/v1/replica.rs
tiagolobocastro: Also btw shouldn't this be done as part of the lvol creation itself?

abhilashshetty04: Sure will do.

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -842,6 +846,33 @@ impl LvsLvol for Lvol {
                     }),
                 }
             }
+
+            PropName::EntityId => {
+                let name = "entity_id".to_string().into_cstring();
+                let mut value: *const libc::c_char =
```
tiagolobocastro, line 852: note for later, seems a lot of duplicated code around this area (existing), maybe can improve later

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -943,6 +974,23 @@ impl LvsLvol for Lvol {
                     name: self.name(),
                 })?;
             }
+            PropValue::EntityId(id) => {
+                let name = "entity_id".to_string().into_cstring();
```
```suggestion
                let name = PropName::from(&prop).to_string().into_cstring();
```

tiagolobocastro (approved):

io-engine/src/lvs/lvs_lvol.rs
abhilashshetty04: Yes i noticed. The bit where we call spdk method for xattr ops right?

dsharma-dc (approved):

abhilashshetty04:
bors merge

tiagolobocastro:
bors cancel

io-engine/src/lvs/lvs_lvol.rs
tiagolobocastro: This is marked as resolved but not addressed..
Please do not mark comments as resolved without making the changes or discuss the change.

io-engine/src/lvs/lvs_lvol.rs
tiagolobocastro: Yes indeed, we can have a look at it at a later time

io-engine/src/lvs/lvs_lvol.rs
abhilashshetty04: Sorry missed it. Done now

abhilashshetty04:
bors merge

PR #1611, opened 2024-03-20
fix(lock): make import pool, create replica and share replica operations mutually exclusive

hrudaya21:
1. Current PR handles concurrent gRPC request of `import pool`, `create replica`, `share replica`. All are kept under resource mutex lock, where resource is `pool_name`.

### DIFF
--- a/io-engine/src/bin/io-engine.rs
+++ b/io-engine/src/bin/io-engine.rs
@@ -100,7 +100,9 @@ fn start_tokio_runtime(args: &MayastorCliArgs) {
 
     // Initialize Lock manager.
     let cfg = ResourceLockManagerConfig::default()
-        .with_subsystem(ProtectedSubsystems::NEXUS, 512);
+        .with_subsystem(ProtectedSubsystems::POOL, 32)
+        .with_subsystem(ProtectedSubsystems::NEXUS, 512)
+        .with_subsystem(ProtectedSubsystems::REPLICA, 1024);
     ResourceLockManager::initialize(cfg);
 
     Mthread::spawn_unaffinitized(move || {
--- a/io-engine/src/core/lock.rs
+++ b/io-engine/src/core/lock.rs
@@ -11,6 +11,8 @@ use once_cell::sync::OnceCell;
 pub struct ProtectedSubsystems;
 impl ProtectedSubsystems {
     pub const NEXUS: &'static str = "nexus";
+    pub const POOL: &'static str = "pool";
+    pub const REPLICA: &'static str = "replica";
 }
 
 /// Configuration parameters for initialization of the Lock manager.
@@ -41,6 +43,7 @@ impl ResourceLockManagerConfig {
 }
 
 /// Resource subsystem that holds locks for all resources withing this system.
+#[derive(Debug)]
 pub struct ResourceSubsystem {
     id: String,
     object_locks: Vec<Mutex<LockStats>>,
@@ -67,8 +70,9 @@ impl ResourceSubsystem {
     pub async fn lock(
         &self,
         wait_timeout: Option<Duration>,
+        try_lock: bool,
     ) -> Option<ResourceLockGuard<'_>> {
-        acquire_lock(&self.subsystem_lock, wait_timeout).await
+        acquire_lock(&self.subsystem_lock, wait_timeout, try_lock).await
     }
 
     /// Lock subsystem resource by its ID and obtain a lock guard.
@@ -76,13 +80,13 @@ impl ResourceSubsystem {
         &self,
         id: T,
         wait_timeout: Option<Duration>,
+        try_lock: bool,
     ) -> Option<ResourceLockGuard<'_>> {
         // Calculate hash of the object to get the mutex index.
         let mut hasher = DefaultHasher::new();
         id.as_ref().hash(&mut hasher);
         let mutex_id = hasher.finish() as usize % self.object_locks.len();
-
-        acquire_lock(&self.object_locks[mutex_id], wait_timeout).await
+        acquire_lock(&self.object_locks[mutex_id], wait_timeout, try_lock).await
     }
 }
 
@@ -122,14 +126,21 @@ static LOCK_MANAGER: OnceCell<ResourceLockManager> = OnceCell::new();
 async fn acquire_lock(
     lock: &Mutex<LockStats>,
     wait_timeout: Option<Duration>,
+    try_lock: bool,
 ) -> Option<ResourceLockGuard<'_>> {
     let mut lock_guard = if let Some(d) = wait_timeout {
         match tokio::time::timeout(d, lock.lock()).await {
             Err(_) => return None,
             Ok(g) => g,
         }
+    } else if try_lock {
+        // No timeout, try for taking lock immediately.
+        match lock.try_lock() {
+            Some(l) => l,
+            None => return None,
+        }
     } else {
-        // No timeout, wait for the lock indefinitely.
+        // No timeout, wait indefinitely.
         lock.lock().await
     };
 
@@ -162,8 +173,9 @@ impl ResourceLockManager {
     pub async fn lock(
         &self,
         wait_timeout: Option<Duration>,
+        try_lock: bool,
     ) -> Option<ResourceLockGuard<'_>> {
-        acquire_lock(&self.mgr_lock, wait_timeout).await
+        acquire_lock(&self.mgr_lock, wait_timeout, try_lock).await
     }
 
     /// Get resource subsystem by its id.
--- a/io-engine/src/grpc/mod.rs
+++ b/io-engine/src/grpc/mod.rs
@@ -1,19 +1,24 @@
+use futures::channel::oneshot::Receiver;
+use nix::errno::Errno;
+pub use server::MayastorGrpcServer;
 use std::{
     error::Error,
     fmt::{Debug, Display},
     future::Future,
     time::Duration,
 };
-
-use futures::channel::oneshot::Receiver;
-use nix::errno::Errno;
-pub use server::MayastorGrpcServer;
 use tokio::sync::RwLock;
 use tonic::{Request, Response, Status};
 
 use crate::{
     bdev_api::BdevError,
-    core::{CoreError, Reactor, VerboseError},
+    core::{
+        CoreError,
+        Reactor,
+        ResourceLockGuard,
+        ResourceSubsystem,
+        VerboseError,
+    },
 };
 
 impl From<BdevError> for tonic::Status {
@@ -166,6 +171,29 @@ where
 {
     Reactor::spawn_at_primary(future)
         .map_err(|_| Status::resource_exhausted("ENOMEM"))
+}
+
+/// Manage locks across multiple grpc services.
+pub async fn acquire_subsystem_lock<'a>(
+    subsystem: &'a ResourceSubsystem,
+    resource: Option<&str>,
+) -> Result<ResourceLockGuard<'a>, Status> {
+    if let Some(resource) = resource {
+        match subsystem.lock_resource(resource.to_string(), None, true).await {
+            Some(lock_guard) => Ok(lock_guard),
+            None => Err(Status::already_exists(format!(
+                "Failed to acquire lock for the resource: {resource}, lock already held"
+            ))),
+        }
+    } else {
+        match subsystem.lock(None, true).await {
+            Some(lock_guard) => Ok(lock_guard),
+            None => Err(Status::already_exists(format!(
+                "Failed to acquire subsystem lock: {:?}, lock already held",
+                subsystem
+            ))),
+        }
+    }
 }
 
 macro_rules! default_ip {
--- a/io-engine/src/grpc/v1/pool.rs
+++ b/io-engine/src/grpc/v1/pool.rs
@@ -1,6 +1,16 @@
 use crate::{
-    core::Share,
-    grpc::{rpc_submit, GrpcClientContext, GrpcResult, RWLock, RWSerializer},
+    core::{
+        lock::{ProtectedSubsystems, ResourceLockManager},
+        Share,
+    },
+    grpc::{
+        acquire_subsystem_lock,
+        rpc_submit,
+        GrpcClientContext,
+        GrpcResult,
+        RWLock,
+        RWSerializer,
+    },
     lvs::{Error as LvsError, Lvs},
     pool_backend::{PoolArgs, PoolBackend},
 };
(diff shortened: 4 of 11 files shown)


io-engine/src/grpc/v1/pool.rs
```diff
@@ -162,6 +165,124 @@ impl PoolService {
...
+        pool_name: Option<String>,
+        global_operation: bool,
+        write_lock: bool,
+        f: F,
+    ) -> Result<T, Status>
+    where
+        T: Send + 'static,
+        F: core::future::Future<Output = Result<T, Status>> + Send + 'static,
+    {
+        async fn acquire_lock(
```
tiagolobocastro: can this code be reused across services?

hrudaya21: Move to a trait. We can have different implementation for each service in future, if needed. Default implementation is kept in trait.

io-engine/src/bin/io-engine.rs
```diff
@@ -100,7 +100,9 @@ fn start_tokio_runtime(args: &MayastorCliArgs) {
 
     // Initialize Lock manager.
     let cfg = ResourceLockManagerConfig::default()
-        .with_subsystem(ProtectedSubsystems::NEXUS, 512);
+        .with_subsystem(ProtectedSubsystems::NEXUS, 512)
+        .with_subsystem(ProtectedSubsystems::REPLICA, 512)
```
dsharma-dc: The number of objects for replica should be proportionately more, isn't it? Assuming 3-way replication, 512*3, if that's the right way to calculate.

io-engine/src/core/lock.rs
```diff
@@ -107,35 +131,59 @@ pub struct ResourceLockManager {
...
-pub struct ResourceLockGuard<'a> {
-    _lock_guard: MutexGuard<'a, LockStats>,
+/// Automatically releases the read lock once dropped.
+pub struct ResourceLockReadGuard<'a> {
+    _lock_guard: RwLockReadGuard<'a, LockStats>,
+}
+
+/// Automatically releases the write lock once dropped.
+pub struct ResourceLockWriteGuard<'a> {
+    _lock_guard: RwLockWriteGuard<'a, LockStats>,
```
dsharma-dc: We don't really need to separate these as we never have to manipulate them? Perhaps it can be:
 ```
union ResourceLockGuard {
 _lock_guard_read: RwLockReadGuard,
_lock_guard_write: RwLockWriteGuard
}
```

File: io-engine/src/grpc/v1/pool.rs [hunk 4]
```diff
@@ -266,10 +298,16 @@ impl PoolRpc for PoolService {
...
-        self.locked(
-            GrpcClientContext::new(&request, function_name!()),
+        let ctx = GrpcClientContext::new(&request, function_name!());
+        let args = request.into_inner();
+
+        self.serialize_grpc(
+            ctx,
+            Some(args.clone().name),
+            self.init_subsystem().clone(),
+            true,
```
dsharma-dc: export shouldn't be global operation. @tiagolobocastro ?

io-engine/src/bin/io-engine.rs
tiagolobocastro: But there's only 1 replica per volume per node?
The pool below definitely won't reach 512 though

io-engine/src/core/lock.rs
hrudaya21: Handled with enum now.

io-engine/src/bin/io-engine.rs
hrudaya21: Yes, increased to 4 times.

io-engine/src/bin/io-engine.rs
```diff
@@ -100,7 +100,9 @@ fn start_tokio_runtime(args: &MayastorCliArgs) {
 
     // Initialize Lock manager.
     let cfg = ResourceLockManagerConfig::default()
-        .with_subsystem(ProtectedSubsystems::NEXUS, 512);
+        .with_subsystem(ProtectedSubsystems::NEXUS, 512)
+        .with_subsystem(ProtectedSubsystems::REPLICA, 2048)
```
tiagolobocastro: I don't think needed to be so large, because the x3 replica is only 1 per node?

io-engine/src/bin/io-engine.rs
```diff
@@ -100,7 +100,9 @@ fn start_tokio_runtime(args: &MayastorCliArgs) {
...
+        .with_subsystem(ProtectedSubsystems::POOL, 512);
```
tiagolobocastro: Probably won't have so many pools?

io-engine/src/core/lock.rs
```diff
@@ -81,15 +85,35 @@ impl ResourceSubsystem {
...
+        acquire_read_lock(&self.object_locks[mutex_id], wait_timeout).await
     }
 }
 
 /// Structure that holds per-lock statistics.
 #[derive(Debug, Default)]
-struct LockStats {
-    num_acquires: usize,
+pub struct LockStats {
+    pub num_acquires: usize,
```
tiagolobocastro: does this need to be pub?

io-engine/src/grpc/mod.rs
```diff
@@ -131,6 +138,127 @@ pub(crate) trait RWSerializer<F, T> {
...
+        subsystem: ResourceSubsystem,
+        global_operation: bool,
+        write_lock: bool,
+        f: F,
+    ) -> Result<T, Status>
+    where
+        T: Send + 'static,
+        F: core::future::Future<Output = Result<T, Status>> + Send + 'static,
+    {
+        async fn acquire_lock(
```
tiagolobocastro: This can be a separate function in this file to reduce nesting a bit

io-engine/src/grpc/v1/pool.rs
```diff
@@ -190,18 +212,22 @@ impl PoolRpc for PoolService {
...
     ) -> GrpcResult<Pool> {
-        self.locked(
-            GrpcClientContext::new(&request, function_name!()),
+        let ctx = GrpcClientContext::new(&request, function_name!());
+        let args = request.into_inner();
+        let pool_args = PoolArgs::try_from(args.clone())?;
+
+        self.serialize_grpc(
+            ctx,
+            Some(pool_args.clone().name),
```
tiagolobocastro: need not clone the args, just the name

io-engine/src/grpc/v1/pool.rs
tiagolobocastro: hmm not sure.
Which operations do we want to be global?

io-engine/src/grpc/v1/replica.rs
```diff
@@ -268,70 +290,83 @@ impl ReplicaRpc for ReplicaService {
...
-                    {
-                        let msg = format!(
+        let ctx = GrpcClientContext::new(&request, function_name!());
+        let args = request.into_inner();
+        self.serialize_grpc(
+            ctx,
+            Some(args.clone().uuid),
+            self.init_subsystem().clone(),
+            true,
+            true,
```
tiagolobocastro: So we're making the replica create a global write operation?
Is it not enough to just lock the pool which this replica references?

tiagolobocastro (commented):
Would be good to document what changes for each resource happened in the PR description

io-engine/src/core/lock.rs
hrudaya21: fixed

io-engine/src/grpc/v1/replica.rs
hrudaya21: Changed the code, made import pool and create replica under same resource lock (`pool_name`) in place of global lock.

io-engine/src/grpc/v1/replica.rs
```diff
@@ -219,6 +226,16 @@ impl ReplicaRpc for ReplicaService {
                         }
                     }
                 };
+                let pool_subsystem = ResourceLockManager::get_instance().get_subsystem(ProtectedSubsystems::POOL);
+                match acquire_subsystem_lock(pool_subsystem, Some(lvs.name().to_string()), false).await {
```
tiagolobocastro: if let Err(error)?

io-engine/src/grpc/v1/replica.rs
```diff
@@ -427,14 +465,19 @@ impl ReplicaRpc for ReplicaService {
...
+                                                crate::core::CoreError::Ptpl {
                                                     reason: source.to_string(),
                                                 },
-                                                name: lvol.name(),
-                                            },
-                                        )?);
+                                            name: lvol.name(),
+                                        }
+                                                },
+                                            )?,
```
```suggestion
                                            )?
```

io-engine/src/grpc/v1/replica.rs
```diff
@@ -219,6 +226,16 @@ impl ReplicaRpc for ReplicaService {
                         }
                     }
                 };
+                let pool_subsystem = ResourceLockManager::get_instance().get_subsystem(ProtectedSubsystems::POOL);
+                match acquire_subsystem_lock(pool_subsystem, Some(lvs.name().to_string()), false).await {
+                    Ok(_) => {}
+                    Err(_) => {
+                        return Err(LvsError::ResourceLockFailed {
+                            source: Errno::EBUSY,
+                            msg: format!("resource {}, for pooluuid {}", lvs.name().to_string(), args.pooluuid),
```
```suggestion
                            msg: format!("resource {}, for pooluuid {}", lvs.name(), args.pooluuid),
```

io-engine/src/grpc/v1/replica.rs
```diff
@@ -401,7 +418,28 @@ impl ReplicaRpc for ReplicaService {
...
+                            )
+                            .await
+                            {
+                                Ok(_) => {}
+                                Err(_) => {
+                                    return Err(LvsError::ResourceLockFailed {
+                                        source: Errno::EBUSY,
+                                        msg: format!(
+                                            "resource {}, for lvol {:?}",
+                                            lvol.lvs().name().to_string(),
```
```suggestion
                                            lvol.lvs().name(),
```

io-engine/src/lvs/lvs_error.rs
```diff
@@ -265,6 +270,9 @@ impl ToErrno for Error {
             Self::WipeFailed {
                 ..
             } => Errno::EINVAL,
+            Self::ResourceLockFailed {
```
tiagolobocastro, line 272: Missing conversion on mayastor_grpc v0 to tonic::Status

io-engine/src/grpc/mod.rs
```diff
@@ -168,6 +173,45 @@ where
...
+            None => {
+                return Err(Status::deadline_exceeded(
+                    "Failed to acquire global lock within given timeout"
+                        .to_string(),
+                ))
+            }
+        }
+    } else if let Some(resource) = resource {
+        match subsystem.lock_resource(resource, None).await {
+            Some(_) => (),
```
tiagolobocastro: this has to return a guard otherwise it's unlocked?

io-engine/src/grpc/mod.rs
```diff
@@ -168,6 +173,45 @@ where
         .map_err(|_| Status::resource_exhausted("ENOMEM"))
 }
 
+/// Manage locks accross multiple grpc services.
+pub async fn acquire_subsystem_lock(
+    subsystem: &ResourceSubsystem,
+    resource: Option<String>,
+    global: bool,
```
tiagolobocastro: If not being used let's remove it to keep it simple. We can add later if required.

hrudaya21: fixed

io-engine/src/grpc/mod.rs
hrudaya21: fixed

io-engine/src/lvs/lvs_error.rs
hrudaya21: fixed

io-engine/src/grpc/mod.rs
```diff
@@ -168,6 +173,33 @@ where
...
+            Some(l) => l,
+            None => {
+                return Err(Status::already_exists(
+                    "Failed to acquire subsystem lock, lock already held"
+                        .to_string(),
+                ))
+            }
+        }
+    };
+    Ok(lock_guard)
```
tiagolobocastro:
```suggestion
    if let Some(resource) = resource {
        match subsystem.lock_resource(resource, None).await {
            Some(lock_guard) => lock_guard,
            None => Err(Status::already_exists(
                "Failed to acquire lock for the resource, lock already held"
            )),
        }
    } else {
        match subsystem.lock(None).await {
            Some(lock_guard) => lock_guard,
            None => Err(Status::already_exists(
                    "Failed to acquire subsystem lock, lock already held"
            )),
        }
    }
```

io-engine/src/grpc/mod.rs
```diff
@@ -168,6 +173,33 @@ where
         .map_err(|_| Status::resource_exhausted("ENOMEM"))
 }
 
+/// Manage locks accross multiple grpc services.
+pub async fn acquire_subsystem_lock(
+    subsystem: &ResourceSubsystem,
+    resource: Option<String>,
```
tiagolobocastro: This would simplify calling side

```suggestion
    resource: Option<&str>,
    ....
    let resource = resource.map(|s| s.to_string());
```

tiagolobocastro (approved):

io-engine/src/grpc/mod.rs
hrudaya21: As We are returning Result type in this function, We need to return Ok(lock_guard) in both if and else branch.
I feel it is better to return Ok(lock_guard) in one place only.

io-engine/src/grpc/mod.rs
hrudaya21: With this change, We need to use lifetime for `acquire_subsystem_lock` function for all parameters, as lock_guard lifetime is different than &str lifetime.
With lifetime in place, caller also will have not uniform changes.

io-engine/src/grpc/mod.rs
```diff
@@ -168,6 +173,28 @@ where
         .map_err(|_| Status::resource_exhausted("ENOMEM"))
 }
 
+/// Manage locks accross multiple grpc services.
```
dsharma-dc: typo: across

io-engine/src/grpc/mod.rs
```diff
@@ -168,6 +173,28 @@ where
...
+pub async fn acquire_subsystem_lock<'a>(
+    subsystem: &'a ResourceSubsystem,
+    resource: Option<&str>,
+) -> Result<ResourceLockGuard<'a>, Status> {
+    if let Some(resource) = resource {
+        match subsystem.lock_resource(resource.to_string(), None).await {
+            Some(lock_guard) => Ok(lock_guard),
+            None => Err(Status::already_exists(
+                "Failed to acquire lock for the resource, lock already held",
```
dsharma-dc: Let's print resource id here

io-engine/src/grpc/mod.rs
```diff
@@ -168,6 +173,28 @@ where
...
+            )),
+        }
+    } else {
+        match subsystem.lock(None).await {
+            Some(lock_guard) => Ok(lock_guard),
+            None => Err(Status::already_exists(
+                "Failed to acquire subsystem lock, lock already held",
```
dsharma-dc: Let's print subsystem id here.

dsharma-dc (approved):

io-engine/src/grpc/mod.rs
hrudaya21: fixed

io-engine/src/grpc/mod.rs
hrudaya21: fixed

hrudaya21:
bors try

hrudaya21:
bors try

hrudaya21:
bors merge

PR #1621, opened 2024-04-04
fix(rebuild): connect the io log when detaching

tiagolobocastro:
When we detach a device, ensure that the io logs are connected. This usually happens on the fault path, however this change ensures that this happens during the detach itself and thus hardening it against races.


avishnu (approved):

dsavitskiy (approved):

dsharma-dc (approved):

tiagolobocastro:
bors merge

PR #1624, opened 2024-04-04
fix(rebuild): reconnect log on own channel when faulting

tiagolobocastro:
When an IO fails to submit the channel is removed right away. This means if another IO reaches the channel before the retire or the channel traversal then it might not get logged.
This change ensures it is logged by reconnecting the IO log straight away.


dsharma-dc (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #1626, opened 2024-04-05
feat(log_format): adding json stdout support

abhilashshetty04:
```
./target/debug/io-engine -Fjson
{"hostname":"","level":"INFO","timestamp":"Fri, 5 Apr 2024 13:06:18 +0000","fields":{"message":"Engine responsible for managing I/Os (debug) version 1.0.0, revision f221158ac43f (v1.0.0+815)"}}
{"hostname":"","level":"INFO","timestamp":"Fri, 5 Apr 2024 13:06:18 +0000","fields":{"message":"free_pages 2MB: 3072 nr_pages 2MB: 3072"}}
{"hostname":"","level":"INFO","timestamp":"Fri, 5 Apr 2024 13:06:18 +0000","fields":{"message":"free_pages 1GB: 0 nr_pages 1GB: 0"}}
{"hostname":"","level":"INFO","timestamp":"Fri, 5 Apr 2024 13:06:18 +0000","fields":{"message":"kernel io_uring support: yes"}}
{"hostname":"","level":"INFO","timestamp":"Fri, 5 Apr 2024 13:06:18 +0000","fields":{"message":"kernel nvme initiator multipath support: yes"}}
{"hostname":"","level":"INFO","timestamp":"Fri, 5 Apr 2024 13:06:18 +0000","fields":{"message":"Applying Mayastor configuration settings"}}
{"hostname":"","level":"INFO","timestamp":"Fri, 5 Apr 2024 13:06:18 +0000","fields":{"message":"NVMe Bdev options successfully applied"}}

```

### DIFF
--- a/io-engine/src/logger.rs
+++ b/io-engine/src/logger.rs
@@ -1,8 +1,10 @@
 use ansi_term::{Colour, Style};
 use once_cell::sync::OnceCell;
 use std::{
+    collections::HashMap,
     ffi::CStr,
-    fmt::Write,
+    fmt,
+    fmt::{Debug, Write},
     io::IsTerminal,
     os::raw::c_char,
     path::Path,
@@ -14,6 +16,7 @@ use crate::{
     core::spawn,
 };
 use event_publisher::event_handler::EventHandle;
+use tracing::field::{Field, Visit};
 use tracing_core::{event::Event, Level, Metadata};
 use tracing_log::{LogTracer, NormalizeEvent};
 use tracing_subscriber::{
@@ -112,6 +115,17 @@ impl<'a> FormatLevel<'a> {
             Level::WARN => "W",
             Level::ERROR => "E",
         }
+    }
+
+    fn long(&self) -> String {
+        match *self.level {
+            Level::TRACE => "TRACE",
+            Level::DEBUG => "DEBUG",
+            Level::INFO => "INFO",
+            Level::WARN => "WARN",
+            Level::ERROR => "ERROR",
+        }
+        .to_string()
     }
 
     fn fmt_line(&self, mut f: Writer<'_>, line: &str) -> std::fmt::Result {
@@ -262,6 +276,7 @@ impl std::fmt::Display for Location<'_> {
 pub enum LogStyle {
     Default,
     Compact,
+    Json,
 }
 
 // Custom struct used to format trace events.
@@ -294,6 +309,7 @@ impl FromStr for LogFormat {
             match p {
                 "default" => r.style = LogStyle::Default,
                 "compact" => r.style = LogStyle::Compact,
+                "json" => r.style = LogStyle::Json,
                 "color" => r.ansi = true,
                 "nocolor" => r.ansi = false,
                 "date" => r.show_date = true,
@@ -323,6 +339,7 @@ where
         match self.style {
             LogStyle::Default => self.default_style(ctx, w, evt),
             LogStyle::Compact => self.compact_style(ctx, w, evt),
+            LogStyle::Json => self.json_style(ctx, w, evt),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


abhilashshetty04:
bors try

dsharma-dc (approved):

Abhinandan-Purkait (approved):

io-engine/src/logger.rs
```diff
@@ -416,6 +461,44 @@ impl LogFormat {
...
+        let meta = normalized.as_ref().unwrap_or_else(|| event.metadata());
+        let fmt = FormatLevel::new(meta.level(), self.ansi);
+        let now = chrono::Local::now();
+
+        let mut output_string = String::new();
+        let mut visitor = StringVisitor::new(&mut output_string);
+        event.record(&mut visitor);
+        let output = visitor.string;
+        let output_split: Vec<&str> = output.split('=').collect();
+        let k = output_split.first().unwrap().trim().replace(';', "");
```
tiagolobocastro: what is the logic here, we're collecting the first key and value of what?

abhilashshetty04: This is the logger example if we dont modify msg which comes from Event object.

"msg":"message = Subsystem start completed: Ok; self = NvmfSubsystem { id: 0, subtype: \"Discovery\", subnqn: \"nqn.2014-08.org.nvmexpress.discovery\", sn: \"00000000000000000000\", mn: \"Mayastor NVMe controller\", allow_any_host: 1, ana_reporting: 0, listeners: Some([Transport ID { trtype: 3, trstring: \"TCP\", traddr: \"127.0.0.1\", trsvcid: \"8420\" }]) }; "

So we split at '=' and then trim (remove whitespace) to make json same as control plane logging.

tiagolobocastro: Is there no way of parsing the fields directly rather than generating the string and then parsing the string again?

io-engine/src/logger.rs
tiagolobocastro, line 1: Also pretty is missing...
Can we use the implementations from tracing itself?

io-engine/src/logger.rs
abhilashshetty04: Ill check

tiagolobocastro: We're not capturing the tracing fields here, only the message

abhilashshetty04: @tiagolobocastro , As discussed we will merge this as is. Then try to use better tracing package in this repo in future.

io-engine/src/logger.rs
abhilashshetty04: As discuss, We will refactor this to use different package as part of different PR.

io-engine/src/logger.rs
```diff
@@ -416,6 +461,44 @@ impl LogFormat {
...
+        let mut ch = HashMap::new();
+        ch.insert(k, v);
+
+        let json_log = JsonLogger {
+            hostname: self.hostname().to_string(),
+            level: fmt.long(),
+            timestamp: now.to_rfc2822(),
+            fields: ch,
+        };
+        let json_str = serde_json::to_string(&json_log).unwrap();
```
tiagolobocastro: map error if needed
```suggestion
        let json_str = serde_json::to_string(&json_log)?;
```

abhilashshetty04:
bors try

io-engine/src/logger.rs
```diff
@@ -416,6 +461,44 @@ impl LogFormat {
...
+        let normalized = event.normalized_metadata();
+        let meta = normalized.as_ref().unwrap_or_else(|| event.metadata());
+        let fmt = FormatLevel::new(meta.level(), self.ansi);
+        let now = chrono::Local::now();
+
+        let mut output_string = String::new();
+        let mut visitor = StringVisitor::new(&mut output_string);
+        event.record(&mut visitor);
+        let output = visitor.string;
+        let output_split: Vec<&str> = output.split('=').collect();
```
tiagolobocastro: hmm what if the message itself has '=' ?

abhilashshetty04: Doing strip_prefix now

abhilashshetty04:
bors try

tiagolobocastro (approved):

abhilashshetty04:
bors merge

PR #1632, opened 2024-04-09
feat(test): adding an extended test for miscompare during partial rebuild

dsavitskiy:


dsavitskiy:
bors test

ci.nix
```diff
@@ -13,6 +13,7 @@ let
   norust_moth =
     "You have requested environment without RUST, you should provide it!";
   norustc_msg = "no rustc, use rustup tool to install it";
+  nofio_msg = "FIO cannot be found";
```
tiagolobocastro: why is this needed? doesn't fio come from the shell?

io-engine/tests/nexus_rebuild_partial_loop.rs
```diff
@@ -0,0 +1,255 @@
...
+        }
+    });
+    tokio::pin!(j1);
+
+    println!("[C]");
+    println!("[C] Joining jobs ...");
+    let (io_res, _) = tokio::join!(j0, j1);
+    println!("[C] Joining jobs done");
+    let io_res = io_res.unwrap();
+    io_res.unwrap();
```
tiagolobocastro, line 262: how long does this run for if there's no miscompare?

io-engine-tests/src/fio.rs
```diff
@@ -265,6 +281,8 @@ impl Fio {
         self.push_err(&stderr);
         self.exit = exit;
 
+        // println!("{stdout}");
```
dsharma-dc: Remove this now?

io-engine-tests/src/nexus.rs
```diff
@@ -504,6 +504,50 @@ impl NexusBuilder {
         }
     }
 
+    pub async fn wait_children_online2(
```
dsharma-dc: Why not add the extra string printing logic into existing `wait_children_online` function itself?

io-engine/tests/nexus_rebuild_partial_loop.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    let j0 = tokio::spawn({
+        let nvmf = nex.nvmf_location();
+        async move {
+            println!("[A]");
+            let (_cg, path) = nvmf.open().unwrap();
+            println!("[A] nvme path: {path:?}");
+
+            let fio = FioBuilder::new()
+                .with_fio_binary(
```
dsharma-dc: Why do we have the need to provide fio binary explicitly, to rule out any fio bugs in default 3.33 ?

io-engine/tests/nexus_rebuild_partial_loop.rs
dsavitskiy: On my machine, around one hour.

ci.nix
dsavitskiy: Just in case it cannot be found, in order to prevent the version message from failing.

io-engine-tests/src/fio.rs
dsavitskiy: Yes.

io-engine-tests/src/nexus.rs
dsavitskiy: Sorry, I forgot to clean up tmp debug stuff.

io-engine/tests/nexus_rebuild_partial_loop.rs
dsavitskiy: Yes, Fio builder API for providing fio binary explicitly is needed for develpoment / debug purposes.
But this one is, again, I forgot to clean up.

ci.nix
tiagolobocastro: I don't think this can happen. If fio fails to build then nix-shell will exit

io-engine/tests/nexus_rebuild_partial_loop.rs
tiagolobocastro: Can we make this configurable somehow? If we create a curated list of tests 1h per each could be too much? If we run as a soak we may want to run for longer periods as well, ex: 1d or even 1w

dsavitskiy: Volume size, number of FIO loops and probably some FIO params can be made configurable for this test.
We'd need to discuss with Blaise how to do that better (via env var?).

Race conditions happen rarely, so a large number of loops is needed to be sure.

ci.nix
dsavitskiy: You're right, I'll remove this.

dsharma-dc (approved):

tiagolobocastro (approved):

tiagolobocastro:
bors merge

PR #1635, opened 2024-04-12
fix: update blob metadata during properties update

dsharma-dc:
When a nexus fails over and is moved to a new node, the nvmf replicas attached to it are re-shared. In the share path we don't update lvol blob metadata today, but only update in-core replica nvmf subsystem. Later on if this replica node reboots, the replica is imported with metadata that still contains nqn of old nexus node. This results into replica rejecting the connect attempts from nexus(on new node), e.g. during online child operation.
> [2024-04-11T10:41:09.290525051+00:00 ERROR mayastor::spdk:ctrlr.c:737] Subsystem 'nqn.2019-05.io.openebs:af36286f-2e22-4acc-afdb-bd306dd03056' does not allow host 'nqn.2019-05.io.openebs:node-name:io-engine-2'

### DIFF
--- a/io-engine/src/lvs/lvs_lvol.rs
+++ b/io-engine/src/lvs/lvs_lvol.rs
@@ -221,9 +221,17 @@ impl Share for Lvol {
     }
 
     async fn update_properties<P: Into<Option<UpdateProps>>>(
-        self: Pin<&mut Self>,
+        mut self: Pin<&mut Self>,
         props: P,
     ) -> Result<(), Self::Error> {
+        let props = UpdateProps::from(props.into());
+        let allowed_hosts = props.allowed_hosts().clone();
+        // Set but don't sync in lvol metadata. Sync will happen in case the
+        // node reboots e.g during upgrade.
+        self.as_mut()
+            .set_no_sync(PropValue::AllowedHosts(allowed_hosts))
+            .await?;
+
         Pin::new(&mut self.as_bdev())
             .update_properties(props)
             .await
@@ -564,7 +572,7 @@ pub trait LvsLvol: LogicalVolume + Share {
     /// snapshot. if it is clone, return the snapshot lvol.
     fn is_snapshot_clone(&self) -> Option<Lvol>;
 
-    /// Get/Read a property from this lvol from disk.
+    /// Get/Read a property of this lvol from the in-memory metadata copy.
     async fn get(&self, prop: PropName) -> Result<PropValue, Error>;
 
     /// Destroy the lvol.
@@ -781,7 +789,7 @@ impl LvsLvol for Lvol {
         None
     }
 
-    /// Get/Read a property from this lvol from disk.
+    /// Get/Read a property of this lvol from the in-memory metadata copy.
     async fn get(&self, prop: PropName) -> Result<PropValue, Error> {
         let blob = self.blob_checked();
 
@@ -943,6 +951,14 @@ impl LvsLvol for Lvol {
             PropValue::Shared(val) => {
                 let name = PropName::from(&prop).to_string().into_cstring();
                 let value = if val { "true" } else { "false" }.into_cstring();
+                // Return Ok early if the current value is already same as the
+                // new value.
+                if let Ok(pval) = self.get(PropName::Shared).await {
+                    match pval {
+                        PropValue::Shared(s) if s == val => return Ok(()),
+                        _ => (),
+                    }
+                };
                 unsafe {
                     spdk_blob_set_xattr(
                         blob,
@@ -957,9 +973,22 @@ impl LvsLvol for Lvol {
                     name: self.name(),
                 })?;
             }
-            PropValue::AllowedHosts(hosts) => {
+            PropValue::AllowedHosts(mut hosts) => {
                 let name = PropName::from(&prop).to_string().into_cstring();
                 let value = hosts.join(",").into_cstring();
+                // Return Ok early if the current value is already same as the
+                // new value.
+                if let Ok(PropValue::AllowedHosts(mut hlist)) =
+                    self.get(PropName::AllowedHosts).await
+                {
+                    if hlist.len() == hosts.len() {
+                        hosts.sort();
+                        hlist.sort();
+                        if hosts == hlist {
+                            return Ok(());
+                        }
+                    }
+                };
                 unsafe {
                     spdk_blob_set_xattr(
                         blob,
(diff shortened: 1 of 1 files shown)


dsharma-dc:
bors try

tiagolobocastro (commented):
We'd also need a `make_replica_accessible` call in control-plane for when this doesn't sync, ex: crash
But also to consider, if we have 1000 replicas, this could potentially be 1000 * meta_size during shutdown, could this become a problem?

dsharma-dc:
> We'd also need a `make_replica_accessible` call in control-plane for when this doesn't sync, ex: crash 

I'll create another ticket for that.

> But also to consider, if we have 1000 replicas, this could potentially be 1000 * meta_size during shutdown, could this become a problem?

Those 1000 replicas would belong to 1000 distinct nexuses, and unless all 1000 nexus are on same node it would be unlikely that all 1000 replicas will have dirty metadata. Moreover, we'd expect the blobstore unload to be able to handle this scale during clean restart. I did a simple test by creating about 900 replicas on a deployer node and shared them over nvmf **without syncing metadata**, restarted the node, reimported the pool, didn't notice any issue in io-engine during shutdown or re-import.

tiagolobocastro:
> Those 1000 replicas would belong to 1000 distinct nexuses, and unless all 1000 nexus are on same node it would be unlikely that all 1000 replicas will have dirty metadata. Moreover, we'd expect the blobstore unload to be able to handle this scale during clean restart. I did a simple test by creating about 900 replicas on a deployer node and shared them over nvmf **without syncing metadata**, restarted the node, reimported the pool, didn't notice any issue in io-engine during shutdown or re-import.

How much data is written per replica in this worst case scenario? Ex: if metadata sync is 4MiB then 4MiB x 900

dsharma-dc:
> How much data is written per replica in this worst case scenario? Ex: if metadata sync is 4MiB then 4MiB x 900

It seems spdk persists metadata sync in units of page size i.e. 4KiB depending on how many metadata pages are actively allocated for blob, so the actual amount of data to persist would be lesser than 4MiB x 900

tiagolobocastro:
> > How much data is written per replica in this worst case scenario? Ex: if metadata sync is 4MiB then 4MiB x 900
> 
> It seems spdk persists metadata sync in units of page size i.e. 4KiB depending on how many metadata pages are actively allocated for blob, so the actual amount of data to persist would be lesser than 4MiB x 900

Alright, this seems reasonable then :+1:

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -221,9 +221,17 @@ impl Share for Lvol {
...
+        mut self: Pin<&mut Self>,
         props: P,
     ) -> Result<(), Self::Error> {
+        let props = UpdateProps::from(props.into());
+        let allowed_hosts = props.allowed_hosts().clone();
+        // Set but don't sync in lvol metadata. Sync will happen in case the
+        // node reboots e.g during upgrade.
+        self.as_mut()
+            .set_no_sync(PropValue::AllowedHosts(allowed_hosts))
+            .await?;
```
tiagolobocastro, line 233: Can we check what the current stored value is? If it's the same then we don't need to call set?

dsharma-dc: > Can we check what the current stored value is? If it's the same then we don't need to call set?

get() call will read from the disk so will be expensive. If we **have** to go to disk we better do set(), rather than get() + conditional set_no_sync(). Looks unnecessary to me.

tiagolobocastro: IIRC the metadata is kept in memory

dsharma-dc: > IIRC the metadata is kept in memory

Yes a copy is kept in memory. I took this comment at face value :) 
```
/// Get/Read a property from this lvol from disk.
async fn get(&self, prop: PropName) -> Result<PropValue, Error> {
```
But yes, spdk API is giving the in-core value and not fetching from disk. Let me see, I can put a `get` check.

tiagolobocastro: Please fix the comment while you're at it :)

dsharma-dc: Though I'm still a bit wary about putting a `get` before `set_no_sync`. We don't do that anywhere today, so let's be consistent?

tiagolobocastro: Why don't we change the set_no_sync to do this internally? Check if current value is same, then don't do anything

dsharma-dc: > Why don't we change the set_no_sync to do this internally? Check if current value is same, then don't do anything

That can be done. For properties `Shared` and `EntityId` it'll be simpler, for `AllowedHosts` a bit verbose because it's a Vec<String>. Looks ok?

```
     fn set_no_sync()
           [...]
-            PropValue::AllowedHosts(hosts) => {
+            PropValue::AllowedHosts(mut hosts) => {
                 let name = PropName::from(&prop).to_string().into_cstring();
                 let value = hosts.join(",").into_cstring();
+                if let Ok(pval) = self.get(PropName::AllowedHosts).await {
+                    match pval {
+                        PropValue::AllowedHosts(mut hlist) => {
+                            if hlist.len() == hosts.len() {
+                                hosts.sort(); hlist.sort();
+                                if hosts == hlist {
+                                    return Ok(())
+                        },
+                        _ => (),
+                    }
+                };
```

tiagolobocastro: Seems like it yeah

dsharma-dc: @tiagolobocastro PTAL at latest diff.

dsavitskiy (approved):

io-engine/src/lvs/lvs_lvol.rs
```diff
@@ -976,6 +1005,14 @@ impl LvsLvol for Lvol {
...
             PropValue::EntityId(id) => {
                 let name = PropName::from(&prop).to_string().into_cstring();
+                // Return Ok early if the current value is already same as the
+                // new value.
+                if let Ok(pval) = self.get(PropName::EntityId).await {
+                    match pval {
+                        PropValue::EntityId(e) if e == id => return Ok(()),
+                        _ => (),
+                    }
+                };
```
tiagolobocastro, line 1015: wonder if a more generic check could be done by moving the duplicated calls to set_xattr to somewhere below and compare the raw values.

tiagolobocastro (approved):

io-engine/src/lvs/lvs_lvol.rs
dsharma-dc: For removing duplication, IIRC Abhilash created a ticket while working on EntityId property for stats.

dsharma-dc:
bors try

dsharma-dc:
bors merge

PR #1637, opened 2024-04-15
Snapshot Rebuild

tiagolobocastro:
feat(snap-rebuild/grpc): impl grpc backend
    
    Replaces the dummy impls with the actual implementation.
    Snapshot Rebuilds are now created, destroyed and listed.
    
---

    feat(snap-rebuild/cli): add snap rebuild cli commands
    
---

    feat(rebuild/snapshot): initial dummy snapshot rebuild
    
    Adds snapshot rebuild module, which only makes use of bdev
    rebuild internally.
    Adds rebuild option to copy only cluster allocated to the
    current blob. This is a simply way of copying the allocated
    snapshot without before the bitmap is in place.
    
    Adds tests for simple snapshot rebuilds.
    
### DIFF
--- a/io-engine/src/rebuild/rebuild_state.rs
+++ b/io-engine/src/rebuild/rebuild_state.rs
@@ -1,5 +1,6 @@
 use super::{RebuildError, RebuildOperation};
 use crate::rebuild::RebuildStats;
+use chrono::Utc;
 
 /// Allowed states for a rebuild job.
 #[derive(Default, Debug, PartialEq, Copy, Clone)]
@@ -72,7 +73,8 @@ impl RebuildStates {
         &self.final_stats
     }
     /// Set the final rebuild statistics.
-    pub(super) fn set_final_stats(&mut self, stats: RebuildStats) {
+    pub(super) fn set_final_stats(&mut self, mut stats: RebuildStats) {
+        stats.end_time = Some(Utc::now());
         self.final_stats = Some(stats);
     }
 
--- a/io-engine/src/rebuild/snapshot_rebuild.rs
+++ b/io-engine/src/rebuild/snapshot_rebuild.rs
@@ -1,0 +1,168 @@
+use snafu::ResultExt;
+use std::{convert::TryFrom, ops::Deref, sync::Arc};
+
+use super::{rebuild_error::RebuildError, RebuildJob, RebuildJobOptions};
+
+use crate::{
+    bdev::{device_create, device_destroy},
+    core::{Bdev, Reactors, ReadOptions, SegmentMap},
+    gen_rebuild_instances,
+    lvs::Lvol,
+    rebuild::{
+        bdev_rebuild::BdevRebuildJobBuilder,
+        rebuild_error::{SnapshotRebuildError, SourceUriBdev},
+        BdevRebuildJob,
+    },
+};
+
+/// A Snapshot rebuild job is responsible for managing a rebuild (copy) which
+/// reads from a source snapshot and writes into a local replica from specified
+/// start to end.
+pub struct SnapshotRebuildJob {
+    inner: BdevRebuildJob,
+    destroy_src_uri: bool,
+    name: String,
+}
+
+impl std::fmt::Debug for SnapshotRebuildJob {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("SnapshotRebuildJob")
+            .field("name", &self.name)
+            .field("inner", &self.inner)
+            .finish()
+    }
+}
+impl Deref for SnapshotRebuildJob {
+    type Target = RebuildJob;
+
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
+}
+
+/// Builder for the `SnapshotRebuildJob`.
+#[derive(Default)]
+pub struct SnapshotRebuildJobBuilder(BdevRebuildJobBuilder);
+impl SnapshotRebuildJobBuilder {
+    /// Specify the rebuild options.
+    pub fn with_option(self, options: RebuildJobOptions) -> Self {
+        Self(self.0.with_option(options))
+    }
+    /// Specify a notification function.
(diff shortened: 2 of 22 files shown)


io-engine/src/rebuild/snapshot_rebuild.rs
```diff
@@ -0,0 +1,124 @@
...
+        let _ = Self::remove(self.name());
+    }
+}
+
+impl Drop for SnapshotRebuildJob {
+    fn drop(&mut self) {
+        let src_uri = self.src_uri().to_owned();
+        Reactors::master().send_future(async move {
+            if let Err(error) = device_destroy(&src_uri).await {
+                // todo: how do we know it's safe to destroy?
```
tiagolobocastro, line 158: Reviewers, is there any way to do this currently?
I think fixing this would likely require changing this device create api to RAII.

In practice this is probably ok for now (nothing else would read the snapshot uri) though seems like we should handle this better in general?

dsavitskiy (approved):

io-engine/src/rebuild/snapshot_rebuild.rs
dsavitskiy: Probably, the device monitor that we're using to defer retiring, is what you're looking for.
Or, you're asking about reference counting for the device?

tiagolobocastro: Discussed offline.
FTWA, indeed I was referring to ref counting but for now we'll keep the devices as non-shared. (sharing was only intended for test cases anyway)

orville-wright (approved):
Looks Great.

io-engine/src/rebuild/rebuild_state.rs
```diff
@@ -72,7 +73,10 @@ impl RebuildStates {
         &self.final_stats
     }
     /// Set the final rebuild statistics.
-    pub(super) fn set_final_stats(&mut self, stats: RebuildStats) {
+    pub(super) fn set_final_stats(&mut self, mut stats: RebuildStats) {
+        if stats.end_time.is_none() {
+            stats.end_time = Some(Utc::now());
```
dsharma-dc: end_time can be set unconditionally.

io-engine/src/rebuild/snapshot_rebuild.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// Get the name of this rebuild job.
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+    /// Destroy this snapshot rebuild job itself.
+    pub fn destroy(self: std::sync::Arc<Self>) {
+        let _ = Self::remove(self.name());
+    }
+    /// Lookup a rebuild job by its name or target uri and return it.
+    pub fn lookup_any(name_or_uri: &str) -> Result<Arc<Self>, RebuildError> {
```
dsharma-dc, line 142: The `lookup` interface can probably be made simpler by just keeping a single `lookup()` call. That'll do the job what primarily `lookup_dst_uri()` is doing, additionally identifying whether input is uri or a name.

io-engine/tests/snapshot_rebuild.rs
```diff
@@ -0,0 +1,227 @@
...
+
+fn get_ms() -> &'static MayastorTest<'static> {
+    MAYASTOR.get_or_init(|| {
+        MayastorTest::new(MayastorCliArgs {
+            ..Default::default()
+        })
+    })
+}
+
+async fn create_pool() -> Result<Lvs, String> {
```
dsharma-dc, line 31: Can we not reuse these pool and replica create/destroy calls from other test modules?

io-engine/src/rebuild/rebuild_state.rs
tiagolobocastro: hmm but if something else sets it we'd be changing the end time again?

io-engine/src/rebuild/snapshot_rebuild.rs
tiagolobocastro: I don't think that's a good idea to have the current lookup do this because you'd overload the lookup and there'd be no way of opting out of the uri check.
We could modify the existing lookup to take an enum, example:
```rust
lookup(Name("bob"))
lookup(NameOrUri("bob"))
```
Though seems to complicate things a bit since most lookups are using name anyway.

io-engine/tests/snapshot_rebuild.rs
tiagolobocastro: I did not find any existing helper functions for this.
There's a builder for pools/replicas/nexuses but it operates over gRPC only.
I guess I could modify that one to also work on the local reactor?

hrudaya21 (approved):

io-engine/src/rebuild/rebuild_state.rs
dsharma-dc: > hmm but if something else sets it we'd be changing the end time again?

Do you foresee any scenario where it'd be already set on a running rebuild job''s stats? I can't see any.

io-engine/tests/snapshot_rebuild.rs
dsharma-dc: > I did not find any existing helper functions for this. There's a builder for pools/replicas/nexuses but it operates over gRPC only. I guess I could modify that one to also work on the local reactor?

That'd be cool bcz it's a pain at times to juggle between tests running local reactor vs gRPC. But that can be done as a separate work item.

io-engine/src/rebuild/rebuild_state.rs
tiagolobocastro: Ok, if you think this can never happen..

tiagolobocastro: Done

io-engine/tests/snapshot_rebuild.rs
tiagolobocastro: Ticket raised, I'll work on this next. We also need to abstract it to work with non-LVS, ie: LVM :)

dsharma-dc (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1639, opened 2024-04-18
fix(block/list): update deps to get safeMountIter fix

tiagolobocastro:
Dep update fixes deadlock on the new safe mount which basically looped forever trying to read mounts.
This would cause list_block_devices gRPC call to hang.


hrudaya21 (approved):

niladrih (approved):

tiagolobocastro:
bors merge

PR #1644, opened 2024-04-29
test: let replica resize events consolidate before nexus resize

dsharma-dc:
Add a small delay after replicas' resize and before nexus resize, to simulate real world workflow driven by control plane.

### DIFF
--- a/io-engine/tests/nexus_replica_resize.rs
+++ b/io-engine/tests/nexus_replica_resize.rs
@@ -119,6 +119,8 @@ async fn do_resize_after_replica_resize(
         assert!(ret.size >= EXPANDED_SIZE);
     }
 
+    // Slight wait to let replica resize events consolidate.
+    tokio::time::sleep(std::time::Duration::from_millis(500)).await;
     let nexus_obj = nexus
         .resize(EXPANDED_SIZE)
         .await


hrudaya21 (approved):

io-engine/tests/nexus_replica_resize.rs
```diff
@@ -119,6 +119,8 @@ async fn do_resize_after_replica_resize(
         assert!(ret.size >= EXPANDED_SIZE);
     }
 
+    // Slight wait to let replica resize events consolidate.
+    tokio::time::sleep(std::time::Duration::from_millis(500)).await;
```
tiagolobocastro, line 123: can we do retry instead?

dsharma-dc: > can we do retry instead?

I am having reservations about doing retries in product code at this time, when this situation is happening only due to the way this test works (having no natural delay unlike gRPC). This seems most harmless for now.

tiagolobocastro: Are you certain this won't happen even with gRPC when the system is under high load? 
Btw another suggestion might be to actually run this resize call using gRPC rather than direct call, just a thought.

tiagolobocastro (approved):

io-engine/tests/nexus_replica_resize.rs
dsharma-dc: It's rather more empirical than a guarantee. Assuming it happens in worst case, the next retry from top should succeed. I've a ticket for handling the retries too, that'll require some idempotency analysis for nexus setup call.

> Btw another suggestion might be to actually run this resize call using gRPC rather than direct call, just a thought.

It's actually via NexusBuilder::resize call which is taking rpc() handle. Isn't that what you mean?
```
    pub async fn resize(&self, req_size: u64) -> Result<Nexus, Status> {
        self.rpc()
            .lock()
            .await
            .nexus
            .resize_nexus(ResizeNexusRequest {
                uuid: self.uuid(),
                requested_size: req_size,
            })
            .await
            .map(|r| r.into_inner().nexus.unwrap())
    }
```

tiagolobocastro: oh I see this is already using gRPC... then most likely this can also happen on a regular "install", which also goes via gRPC

dsharma-dc: ok, I'll bring that ticket to upcoming work sprint.

dsharma-dc:
bors merge

PR #1648, opened 2024-05-03
Refactor backend pool

tiagolobocastro:
fix: replica destroy when pool not found
    
    Should return failed preconditions rather than NotFound.
    
---

    refactor(grpc): use new pool and replica traits
    
    Adds new GrpcPool and GrpcReplica types which make use of the Pool
    and Replica traits to ensure all pool/replica types are implemented the same.
    todo: add getters for the pools which will further simplify it.
    todo: how to handle creation?
    
---

    refactor: add shared traits for pools and replicas
    
    This helps define shared behaviour as we now have different types
    of backends (lvs and lvm) whilst hiding the internal details.
    todo: gRPC can use these new traits.
    
---

    refactor(lvs/lvol): tidy up property management
    
    Reuse property modification rather than duplicate code on each arm.
    Add new batch property setting.
    Don't modify property if metadata is unchanged.
    
---

    refactor(lvm): tidy up lvm share bits
    
    Reuse property setting code and also rename some sharing functions.
    
### DIFF
--- a/io-engine/src/grpc/v1/lvm/mod.rs
+++ b/io-engine/src/grpc/v1/lvm/mod.rs
@@ -1,0 +1,14 @@
+pub mod pool;
+pub mod replica;
+
+#[macro_export]
+macro_rules! lvm_run {
+    ($fut:expr) => {{
+        $crate::core::runtime::spawn_blocking(|| {
+            tokio::runtime::Handle::current().block_on($fut)
+        })
+        .await
+        .map_err(|_| Status::cancelled("cancelled"))?
+        .map(Response::new)
+    }};
+}
--- a/io-engine/src/grpc/v1/lvm/pool.rs
+++ b/io-engine/src/grpc/v1/lvm/pool.rs
@@ -3,12 +3,12 @@ use crate::{
     grpc::{
         acquire_subsystem_lock,
         lvm_enabled,
-        v1::pool::{PoolGrpc, PoolProbe},
+        v1::pool::{PoolGrpc, PoolIdProbe, PoolSvcRpc},
         GrpcResult,
     },
     lvm::{CmnQueryArgs, Error as LvmError, VolumeGroup},
     lvs::Lvs,
-    pool_backend::PoolArgs,
+    pool_backend::{PoolArgs, PoolBackend},
 };
 use io_engine_api::v1::pool::*;
 use std::{convert::TryFrom, fmt::Debug};
@@ -28,16 +28,16 @@ impl PoolService {
     }
     /// Probe the LVM Pool service for a pool.
     pub(crate) async fn probe(
-        probe: &PoolProbe,
+        probe: &PoolIdProbe,
     ) -> Result<bool, tonic::Status> {
         if !MayastorFeatures::get_features().lvm() {
             return Ok(false);
         }
 
         let query = match probe {
-            PoolProbe::Uuid(uuid) => CmnQueryArgs::ours().uuid(uuid),
-            PoolProbe::UuidOrName(uuid) => CmnQueryArgs::ours().uuid(uuid),
-            PoolProbe::NameUuid {
+            PoolIdProbe::Uuid(uuid) => CmnQueryArgs::ours().uuid(uuid),
+            PoolIdProbe::UuidOrName(uuid) => CmnQueryArgs::ours().uuid(uuid),
+            PoolIdProbe::NameUuid {
                 name,
                 uuid,
             } => CmnQueryArgs::ours().named(name).uuid_opt(uuid),
@@ -99,6 +99,13 @@ async fn find_pool(
         VolumeGroup::lookup(CmnQueryArgs::ours().named(name).uuid_opt(uuid))
             .await?;
     Ok(pool)
+}
+
+#[async_trait::async_trait]
+impl PoolSvcRpc for PoolService {
+    fn kind(&self) -> PoolBackend {
+        PoolBackend::Lvm
+    }
 }
 
 #[tonic::async_trait]
--- a/io-engine/src/grpc/v1/lvm/pool.rs
+++ b/io-engine/src/grpc/v1/lvm/pool.rs
@@ -3,7 +3,7 @@ use crate::{
     grpc::{
         acquire_subsystem_lock,
         lvm_enabled,
-        v1::pool::PoolProbe,
+        v1::pool::{PoolGrpc, PoolProbe},
         GrpcResult,
     },
     lvm::{CmnQueryArgs, Error as LvmError, VolumeGroup},
@@ -50,6 +50,7 @@ impl PoolService {
             Err(error) => Err(error.into()),
         }
     }
+
     pub(crate) async fn list_svc_pools(
         &self,
         args: &ListPoolOptions,
@@ -68,6 +69,38 @@ impl PoolService {
     }
 }
 
+async fn ensure_unique_pool(args: PoolArgs) -> Result<PoolArgs, tonic::Status> {
+    let args = crate::lvs_run!(async move {
+        // bail if an lvs pool already exists with the same name
+        if let Some(_pool) = Lvs::lookup(args.name.as_str()) {
+            return Err(Status::invalid_argument(
+                "lvs pool with the same name already exists",
+            ));
+        }
+        // check if the disks are used by existing lvs pool
+        if Lvs::iter()
+            .map(|l| l.base_bdev().name().to_string())
+            .any(|d| args.disks.contains(&d))
+        {
+            return Err(Status::invalid_argument(
+                "an lvs pool already uses the disk",
+            ));
+        }
+        Ok(args)
+    })?;
+    Ok(args.into_inner())
+}
+
+async fn find_pool(
+    name: &str,
+    uuid: &Option<String>,
+) -> Result<VolumeGroup, tonic::Status> {
+    let pool =
+        VolumeGroup::lookup(CmnQueryArgs::ours().named(name).uuid_opt(uuid))
+            .await?;
+    Ok(pool)
+}
+
 #[tonic::async_trait]
 impl PoolRpc for PoolService {
     async fn create_pool(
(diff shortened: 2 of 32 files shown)


tiagolobocastro:
bors try

tiagolobocastro:
bors try

io-engine/src/core/logical_volume.rs
```diff
@@ -52,3 +54,90 @@ pub struct LvolSpaceUsage {
...
+    pub(crate) thin: bool,
+    pub(crate) entity_id: Option<String>,
+}
+
+/// This interface defines the high level operations which can be done on a
+/// pool. Pool-Specific details should be hidden away in the implementation as
+/// much as possible, though we can allow for extra pool specific options
+/// to be passed as parameters.
+#[async_trait::async_trait(?Send)]
+pub trait Pool {
```
tiagolobocastro: This is probably not the best location?

io-engine/src/lvm/lv_replica.rs
```diff
@@ -388,52 +394,72 @@ impl LogicalVolume {
...
+        properties: Vec<Property>,
+    ) -> Result<(), Error> {
+        let args = self.build_set_properties_args(properties).await;
         if args.is_empty() {
             return Ok(());
         }
-        LvmCmd::lv_change().args(args).arg(&self.path).run().await
+        let result = LvmCmd::lv_change().args(args).arg(&self.path).run().await;
+        if result.is_err() {
+            self.tags_dirty = true;
```
dsharma-dc, line 465: What do we intend to do once marked dirty, some kind of retry? It's only getting cleared in `import_attrs` .

tiagolobocastro: Not at the moment, just making it visible if we debug print the self. Lv's get "reimported" on every list so it's then resynced. 
We actually have the same issue for the Blobstore, when a property sync fails (and therefore grpc call) fails we get potentially out of sync. Perhaps we need to expose this information up to the control-plane.

io-engine/src/core/logical_volume.rs
dsharma-dc: > This is probably not the best location?

Yeah, doesn't seem to fit well in here with all things volume. Also, we should name it `PoolOps` as this is defining the operations only, and not the pool itself. On a separate note, this file is named logical_volume.rs - I assume that encompasses both the spdk lvol and lvm logical volume?

io-engine/src/core/logical_volume.rs
```diff
@@ -52,3 +54,90 @@ pub struct LvolSpaceUsage {
     /// clone.
     pub allocated_bytes_snapshot_from_clone: Option<u64>,
 }
+
+/// Arguments for replica creation.
+pub struct ReplicaArgs {
```
dsharma-dc: `ReplicaArgs ` -> `CreateReplicaArgs` ?

io-engine/src/core/logical_volume.rs
```diff
@@ -52,3 +54,90 @@ pub struct LvolSpaceUsage {
...
+    async fn export(self) -> Result<(), Self::Error>;
+}
+
+/// This interface defines the high level operations which can be done on a
+/// `Pool` replica. Replica-Specific details should be hidden away in the
+/// implementation as much as possible, though we can allow for extra pool
+/// specific options to be passed as parameters.
+/// A `Replica` is also a `LogicalVolume` and also has `Share` traits.
+#[async_trait::async_trait(?Send)]
+pub trait Replica: Share + LogicalVolume {
```
dsharma-dc: Let's call this also something like `ReplicaOps` or `ReplicaActions`

io-engine/src/grpc/v1/lvm/mod.rs
```diff
@@ -0,0 +1,40 @@
+use crate::core::{runtime::Runtime, Mthread};
+use once_cell::sync::Lazy;
+use tokio::task::JoinHandle;
+
+pub mod pool;
+pub mod replica;
+
+static RUNTIME: Lazy<Runtime> = Lazy::new(|| {
```
dsharma-dc: What's the idea behind needing an all new runtime - is it to cater someone running two different backends in parallel without stepping on each other's toes?

io-engine/src/grpc/v1/lvm/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+    let rt = tokio::runtime::Builder::new_multi_thread()
+        .enable_all()
+        .worker_threads(2)
+        .max_blocking_threads(4)
```
dsharma-dc: Is it ok to have this number more than the worker threads( though they are not directly related) ? Also, since with lvm backend the type and number of operations that will need to be served will roughly be same as for lvs backend today, should we consider keeping worker threads to 4 same as the runtime today?

io-engine/src/core/logical_volume.rs
tiagolobocastro: I followed what we do for pool, which is `PoolArgs`. I guess we need to rename that too?

io-engine/src/core/logical_volume.rs
tiagolobocastro: Believe it or not I almost named it ReplicaOps previously :)
I left it as Replica simply because we also have `LogicalVolume` rather than `LogicalVolumeOps`.
So I think we if rename one we have to rename the other one as well?

io-engine/src/grpc/v1/lvm/mod.rs
tiagolobocastro: It's because of the ?Send which is added to the trait. We cannot call ?Send futures from the grpc handler which is a Send future.
We already have another runtime but I didn't want to step on its toes as you mentioned.
With latest Rust I think we might get away without this as the macro can generate 2 flavours, Send and ?Send.
I will revisit this once we update the compiler.

io-engine/src/grpc/v1/lvm/mod.rs
tiagolobocastro: We actually do not use use worker_threads, only blocking threads, but the api does not let us create less worker threads.

io-engine/src/core/logical_volume.rs
tiagolobocastro: what is your suggestion, move replicas&pool ops to another file (ex: pool.rs) or leave replica ops here and move pool only?
Initially it wasn't but I removed spdk specific bits a while back, so it can now be generic.

io-engine/src/core/logical_volume.rs
tiagolobocastro: Although `LogicalVolume` is more properties, so maybe it's ok then

dsharma-dc: I'd say let's rename now while you're at it, reason being it's already getting too confusing about names now with lvm in picture. `Lvol` implying `LvsLvol` vs `LogicalVolume` trait, `lvs` implying lvm logical volumes Vs `Lvs` backend, and so on !!
If it's too much to accommodate in this PR, then you may create a new ticket to address that as soon as it can be done.

io-engine/src/core/logical_volume.rs
dsharma-dc: Moving only pool ops sounds reasonable.

io-engine/src/core/logical_volume.rs
tiagolobocastro: I've renamed and moved, PTAL

tiagolobocastro:
bors try

io-engine/src/grpc/v1/lvm/pool.rs
```diff
@@ -68,6 +69,38 @@ impl PoolService {
     }
 }
 
+async fn ensure_unique_pool(args: PoolArgs) -> Result<PoolArgs, tonic::Status> {
+    let args = crate::lvs_run!(async move {
+        // bail if an lvs pool already exists with the same name
+        if let Some(_pool) = Lvs::lookup(args.name.as_str()) {
```
dsharma-dc, line 75: May be an overkill to handle for now(using some generics?), but we'll have to revisit this function once a third kind of backend(zfs) comes in the mix.

io-engine/src/grpc/v1/lvs/pool.rs
```diff
@@ -82,99 +82,70 @@ impl PoolService {
...
 }
 
+async fn find_pool(
+    name: &str,
+    uuid: &Option<String>,
+) -> Result<Lvs, tonic::Status> {
+    let Some(pool) = Lvs::lookup(name) else {
+        return Err(LvsError::PoolNotFound {
+            source: Errno::ENOMEDIUM,
+            msg: format!("Destroy failed as pool {name} was not found"),
```
dsharma-dc: We wouldn't want to mention `Destroy` here as it's a common function.

tiagolobocastro:
bors try

io-engine/src/grpc/v1/lvm/mod.rs
dsavitskiy: I don't like the idea of multiple runtimes as well.
Since async traits are already there, can update the compiler first, and make the necessary changes in this PR?

dsavitskiy (approved):

io-engine/src/grpc/v1/pool.rs
```diff
@@ -39,6 +39,28 @@ pub enum PoolProbe {
...
+        Self::NameUuid {
+            name: name.to_owned(),
+            uuid: uuid.to_owned(),
+        }
+    }
+}
+pub(crate) enum Probe {
+    Kind(PoolBackend),
+    Probe(PoolProbe),
+}
```
dsharma-dc: let's keep enum name and the enum variant name different. Perhaps call `enum ProbeType`

dsharma-dc (approved):

io-engine/src/grpc/v1/lvm/pool.rs
tiagolobocastro: Yeah, we need to add finder functions which return Box<PoolOps> and then add some others which ensure there's no conflict between different pool types.
For the latter, if we keep a global list of the loaded pools I think it will simplify things as we won't need to go through runtimes, we just check the list?

io-engine/src/grpc/v1/lvs/pool.rs
tiagolobocastro: Oops, done

io-engine/src/grpc/v1/pool.rs
tiagolobocastro: Yes that's a better name :)

io-engine/src/grpc/v1/lvm/mod.rs
tiagolobocastro: I've removed this runtime altogether and increased the max blocking tasks on the existing one.

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

PR #1654, opened 2024-05-14
refactor(pool/replica): add factories for pool&replica

tiagolobocastro:
This allows us to replace the existing reuse of the gRPC trait and allows for a easier addition of another backend type.
New backends should impl at least PoolOps, PoolProps, ReplicaOps and LogicalVolume. todo: snapshot is a separate service, which would still require refactoring.


io-engine/src/core/logical_volume.rs
```diff
@@ -1,34 +1,53 @@
-///  LogicalVolume Trait Provide all the Generic Interface for Volume
-pub trait LogicalVolume {
-    /// Returns the name of the Logical Volume
+use crate::{core::Protocol, pool_backend::PoolBackend};
+
+///  LogicalVolume Trait Provide all the Generic Interface for Volume.
```
dsharma-dc:
```suggestion
///  LogicalVolume Trait Provide all the Generic Interface for a Logical Volume on any backend type.
```

io-engine/src/core/logical_volume.rs
```diff
@@ -1,34 +1,53 @@
...
+    /// Returns a boolean indication if the Logical Volume is a snapshot.
+    fn is_snapshot(&self) -> bool;
+    /// Returns a boolean indication if the Logical Volume is a clone.
+    fn is_clone(&self) -> bool;
+
+    /// Returns its parent snapshot uuid if the Logical Volume is a snapshot.
+    fn snapshot_uuid(&self) -> Option<String>;
+
+    /// Return the share protocol of this Logical Volume.
+    fn share(&self) -> Protocol;
```
dsharma-dc: Let's call this `share_proto` or `shared_over` , otherwise the name conflicts with the `share` verb function.

io-engine/src/grpc/v1/lvm/pool.rs
```diff
@@ -211,17 +31,8 @@ impl From<LvmError> for Status {
...
-            state: PoolState::PoolOnline.into(),
-            capacity: v.capacity(),
-            used: v.used(),
-            pooltype: PoolType::Lvm as i32,
-            committed: v.committed(),
-            cluster_size: v.cluster_size() as u32,
-        }
+    fn from(vg: VolumeGroup) -> Self {
+        let lvs = &vg as &dyn PoolProps;
+        lvs.into()
```
dsharma-dc:
```suggestion
        let lvm_pool = &vg as &dyn PoolProps;
        lvm_pool.into()
```

io-engine/src/grpc/v1/pool.rs
```diff
@@ -327,31 +352,152 @@ impl PoolService {
...
+    ) -> Result<Option<PoolGrpc>, tonic::Status> {
+        let pool = self.as_pool_ops().find(args).await?;
+        match pool {
+            Some(pool) => {
+                let pool_subsystem = ResourceLockManager::get_instance()
+                    .get_subsystem(ProtectedSubsystems::POOL);
+                let lock_guard =
+                    acquire_subsystem_lock(pool_subsystem, Some(pool.name()))
+                        .await?;
+                Ok(Some(PoolGrpc::new(pool, lock_guard)))
```
dsharma-dc, line 431: What is the lifetime of PoolGrpc object ? The guard is tied to that.

dsharma-dc (commented):
halfway...

io-engine/src/grpc/v1/pool.rs
tiagolobocastro: Can think of it as : 'static, which is, until `PoolGrpc` is dropped, example:
```rust
crate::spdk_submit!(async move '1: {
      let pool: PoolGrpc  = ...;
      pool.create_replica(args).await    // <- when create_replica completes and scope '1 is complete then PoolGrpc is dropped
  })
```

dsavitskiy (approved):

io-engine/src/core/logical_volume.rs
```diff
@@ -1,34 +1,54 @@
...
     fn usage(&self) -> LvolSpaceUsage;
+    /// Returns the backend type which owns this Logical Volume.
+    fn backend(&self) -> PoolBackend;
+
+    /// Returns a boolean indication if the Logical Volume is a snapshot.
+    fn is_snapshot(&self) -> bool;
+    /// Returns a boolean indication if the Logical Volume is a clone.
+    fn is_clone(&self) -> bool;
+
+    /// Returns its parent snapshot uuid if the Logical Volume is a snapshot.
```
dsharma-dc:
```suggestion
    /// Returns its parent snapshot uuid if the Logical Volume is a clone.
```

io-engine/src/grpc/v1/pool.rs
```diff
@@ -447,59 +575,46 @@ impl PoolRpc for PoolService {
...
-                    Some(pool_type) => {
-                        Some(PoolType::try_from(pool_type).map_err(|_| {
-                            Status::invalid_argument("Unknown pool type")
-                        })?)
+                crate::spdk_submit!(async move {
+                    let args = request.into_inner();
+
+                    // todo: what is the intent here when None, to only return
+                    // pools  of Lvs?
+                    // todo: Also, what todo when we hit an error listing any of
```
dsharma-dc, line 597: My suggestion is, for listing, let's not be concerned about pool type specific behaviour. Consider to list pools of all types, and if there is any error listing something then let it be reported as such.

io-engine/src/grpc/v1/replica.rs
```diff
@@ -277,17 +337,34 @@ where
...
         Ok(())
     }
+    fn verify_pool(&self, pool: &PoolGrpc) -> Result<(), Status> {
+        let pool = pool.as_props();
+        let replica = &self.replica;
+        if pool.name() != replica.pool_name()
+            || pool.uuid() != replica.pool_uuid()
+        {
+            let msg = format!(
+                "Specified pool: {pool:?} does match the target replica: {replica:?}!"
```
dsharma-dc:
```suggestion
                "Specified pool: {pool:?} does not match the target replica's pool: {replica:?}!"
```

io-engine/src/lvm/lv_replica.rs
```diff
@@ -869,6 +866,34 @@ impl crate::core::LogicalVolume for LogicalVolume {
             allocated_bytes_snapshot_from_clone: None,
         }
     }
+
+    fn is_snapshot(&self) -> bool {
```
dsharma-dc, line 870: Do we somehow disallow snapshots on lvm LVs today? (I'm guessing yes because we haven't implemented `--snap` support yet?). If yes then returning this blanket `false` looks ok, otherwise we need some logic here.

io-engine/src/replica_backend.rs
```diff
@@ -18,39 +17,121 @@ pub trait ReplicaOps: Share + LogicalVolume {
...
+        &self,
+        args: &FindReplicaArgs,
+    ) -> Result<Option<Replica>, Error>;
+    async fn list(&self, args: &ListReplicaArgs)
+        -> Result<Vec<Replica>, Error>;
+    fn backend(&self) -> PoolBackend;
+}
+
+#[macro_export]
+macro_rules! repl_factory {
```
dsharma-dc: I am confused here! Why are we implementing ReplicaFactory via macro here, and we have implemented it already as:
> impl ReplicaFactory<LogicalVolume, Error> for ReplLvmFactory
and
> impl ReplicaFactory<Lvol, LvsError> for ReplLvsFactory

dsharma-dc (approved):
This refactoring is worth documenting(the class/interface details) on confluence for benefit and clarity of everyone.

io-engine/src/grpc/v1/pool.rs
tiagolobocastro: how do we report an error, output to stdout is enough?

io-engine/src/lvm/lv_replica.rs
tiagolobocastro: Correct, not implemented :)
Also the snapshot code is atm all lvs specific, would have to modify traits for snapshots too :(
Can be taken as separate effort

io-engine/src/replica_backend.rs
tiagolobocastro: This was to automatically generate a factory with the grpc types, example: `PoolFactory<Pool, tonic::Status>`
I have now made the trait more generic, and grpc conversion implemented without a trait so these macros will all be removed.

io-engine/src/core/logical_volume.rs
```diff
@@ -1,34 +1,54 @@
...
+    /// Returns a boolean indication if the Logical Volume is a snapshot.
+    fn is_snapshot(&self) -> bool;
+    /// Returns a boolean indication if the Logical Volume is a clone.
+    fn is_clone(&self) -> bool;
+
+    /// Returns its parent snapshot uuid if the Logical Volume is a clone.
+    fn snapshot_uuid(&self) -> Option<String>;
+
+    /// Return the share protocol of this Logical Volume.
+    fn share_proto(&self) -> Protocol;
```
Abhinandan-Purkait: `share_proto` --> `share_protocol` proto is a different entity even though it might not be related to this.

io-engine/src/lvm/lv_replica.rs
```diff
@@ -16,6 +16,7 @@ use crate::{
     },
 };
 
+use crate::pool_backend::PoolBackend;
```
Abhinandan-Purkait: Move with line 2?

io-engine/src/grpc/v1/pool.rs
dsharma-dc: > how do we report an error, output to stdout is enough?

Yeah, perhaps add a log or an event as well? I wouldn't think it's possible  to send a partial list and and error code too indicating that it's incomplete.

dsharma-dc (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

PR #1655, opened 2024-05-14
fix: crash when there are no bdevs and lvm export

tiagolobocastro:
fix: crash when there are no bdevs
    
    When there are no bdevs the iterator will return nothing.
    In this case we should not crash.
    
---

    fix(lvm): don't export if not enabled
    
    There's no point trying and erroring with message..
    
### DIFF
--- a/io-engine/src/grpc/v1/snapshot.rs
+++ b/io-engine/src/grpc/v1/snapshot.rs
@@ -569,27 +569,24 @@ impl SnapshotRpc for SnapshotService {
                             Ok(None)
                         }
                     }?;
-                    let bdev = UntypedBdev::bdev_first()
-                        .expect("Failed to enumerate devices");
-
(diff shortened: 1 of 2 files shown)


chriswldenyer (approved):
Looks reasonable

io-engine/src/grpc/v1/snapshot.rs
dsharma-dc, line 1: The change in this file is unrelated to the crash w.r.t lvm. You may want to split the commit.

dsharma-dc (approved):

io-engine/src/grpc/v1/snapshot.rs
tiagolobocastro: Indeed, done

tiagolobocastro:
bors merge

PR #1663, opened 2024-05-21
Upgrading to SPDK 24.01 and refactoring Nix scripts

dsavitskiy:


dsavitskiy:
bors try

dsavitskiy:
bors try

io-engine/src/subsys/config/opts.rs
```diff
@@ -525,7 +527,8 @@ impl From<&BdevOpts> for spdk_bdev_opts {
             bdev_auto_examine: false,
             reserved9: Default::default(),
             opts_size: std::mem::size_of::<spdk_bdev_opts>() as u64,
-            reserved: Default::default(),
+            iobuf_small_cache_size: Default::default(),
```
tiagolobocastro: Are these something we may want to set to specific value through env variable? Similar to the `bdev_io_*_size` ?

nix/lib/rust.nix
tiagolobocastro, line 1: Where's this set now?

dsavitskiy:
bors try

shell.nix
```diff
@@ -1,92 +1,5 @@
...
-    export SRCDIR=`pwd`
-    if [ -z "$CI" ]; then
-      echo
-      pre-commit install
-      pre-commit install --hook commit-msg
-    fi
-    export PATH=$PATH:$(pwd)/scripts/nix-sudo
-  '';
-}
+{ rust ? "stable"         # "stable" "nightly" "asan" "none"
```
tiagolobocastro: What does this actually do? Don't we now get rust from spdk-rs?
What's the difference now between shell.nix and ci.nix?
IIRC previously difference was ci.nix carries stable compiler and nightly the nightly compiler or allowed for rustup?

dsavitskiy: This allows to choose Rust channel/config, e.g. "nightly" for experimenting with new features/flags. (ASAN must always use nightly, so it goes here as well). The actual Nix script that provides Rust packages is in spdk-rs. This arg is forwarded to spdk-rs.

Previously, ci.nix provided the "correct" Rust. shell.nix used system Rust which was no good, as we build images and test with ci.nix. Now, to get system Rust for experimenting, use "--argstr rust none".

Now ci.nix and shell.nix are identical and kept for compatibility with other scripts.

io-engine/src/subsys/config/opts.rs
dsavitskiy: Fixed

nix/lib/rust.nix
dsavitskiy: In spdk-rs. Io-engine now includes Rust nix pckgs from spdk-rs.

shell.nix
tiagolobocastro: Does this make it more difficult for UIs? As they won't work properly with rust from the nix-shell? At least I had problems before, and had to use rust from the rustup.
Otherwise let's delete this file completely and let folks keep their own shell.nix

tiagolobocastro (commented):
lgtm but but let's wait until we branch 2.7

shell.nix
dsavitskiy: I see your point. I didn't know it affects IDEs. I can change shell.nix to use system rust by default, as before.

dsharma-dc (approved):

tiagolobocastro (approved):
We've branched whilst I was on PTO, so we can now merge this?

shell.nix
tiagolobocastro: That would be good yes

dsavitskiy: As per our discussion, I'll add "ide.nix" file that would use the rustup's system Rust. "shell.nix" will be the same as ci.nix, to keep the default configuration the same as the one used for building images.

dsavitskiy:
bors merge

dsavitskiy:
bors merge

PR #1665, opened 2024-05-24
Several Fixes

tiagolobocastro:
chore(wiper): temporarily reduce replica checksum size
    
    This change is required because otherwise e2e checksum tests incorrectly
    report data mismatch.
    To better explain this, the nexus used to format replicas with a gpt
    partition table. This is no longer the case, but for back-compat the nexus
    ignores the blocks which were used for this.
    As such the gpt backup is never read or written to by an application and
    it is also not rebuilt! So this data may not be the same among the nexus
    replicas.
    
    As a short term solution, we simply discount these last 33 sectors from
    the replica checksum.
    
---

    fix(replica): don't use disabled backends
    
    Check if backend is enabled before using its function.
    
---

    fix(grpc/replica): keep behaviour for replica destroy
    
    On replica destruction, if the pool doesn't match, then return aborted, rather
    than invalid argument.
    In case of replica not found, also set "gtm-602" in the error header.
    Both these are for back-compat.
    
### DIFF
--- a/io-engine-tests/src/replica.rs
+++ b/io-engine-tests/src/replica.rs
@@ -73,6 +73,11 @@ impl ReplicaBuilder {
         self
     }
 
+    pub fn with_size_b(mut self, size_b: u64) -> Self {
+        self.size = Some(size_b);
+        self
+    }
+
     pub fn with_size_mb(mut self, size_mb: u64) -> Self {
         self.size = Some(size_mb * 1024 * 1024);
         self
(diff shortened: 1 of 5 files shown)


dsavitskiy (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

abhilashshetty04 (approved):

dsharma-dc (approved):

tiagolobocastro:
bors merge

PR #1668, opened 2024-06-03
refactor(snapshot/rebuild): support push/pull model

tiagolobocastro:
Allows doing the snapshot rebuild from either the snapshot or replica node. This is done by allowing specification of either URI's. If a URI is not specified, then the uuid must be specified.

### DIFF
--- a/io-engine/src/rebuild/rebuild_error.rs
+++ b/io-engine/src/rebuild/rebuild_error.rs
@@ -93,14 +93,14 @@ pub enum RebuildError {
 #[snafu(visibility(pub(crate)), context(suffix(false)))]
 #[allow(missing_docs)]
 pub enum SnapshotRebuildError {
-    #[snafu(display("Destination replica bdev not found"))]
-    ReplicaBdevNotFound {},
-    #[snafu(display("Destination replica bdev uri is missing"))]
-    ReplicaNoUri {},
-    #[snafu(display("Given destination uuid is not a replica"))]
+    #[snafu(display("Local bdev not found"))]
+    LocalBdevNotFound {},
+    #[snafu(display("Remote bdev uri is missing"))]
+    RemoteNoUri {},
+    #[snafu(display("Local bdev is not a replica"))]
     NotAReplica {},
-    #[snafu(display("Failed to open the source uri as a bdev: {source}"))]
-    SourceUriBdev { source: BdevError },
+    #[snafu(display("Failed to open {uri} as a bdev: {source}"))]
+    UriBdevOpen { uri: String, source: BdevError },
 }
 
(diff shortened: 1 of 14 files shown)


dsavitskiy (approved):

io-engine/src/rebuild/rebuild_error.rs
```diff
@@ -93,14 +93,14 @@ pub enum RebuildError {
...
+    #[snafu(display("Local bdev not found"))]
+    LocalBdevNotFound {},
+    #[snafu(display("Remote bdev uri is missing"))]
+    RemoteNoUri {},
+    #[snafu(display("Local bdev is not a replica"))]
     NotAReplica {},
-    #[snafu(display("Failed to open the source uri as a bdev: {source}"))]
-    SourceUriBdev { source: BdevError },
+    #[snafu(display("Failed to open {uri} as a bdev: {source}"))]
+    OpenUri { uri: String, source: BdevError },
```
dsharma-dc: Prefer a clearer naming perhaps. `UriBdevOpen`, `BdevUriOpen` or `BdevOpen, since we essntially open the bdev and not uri(which is just a string identifier).

io-engine/src/rebuild/snapshot_rebuild.rs
```diff
@@ -40,71 +42,201 @@ impl Deref for SnapshotRebuildJob {
     }
 }
 
+#[derive(Default)]
+struct Uri {
+    create: bool,
+    delete: bool,
+    uri: String,
+}
+impl Uri {
```
dsharma-dc, line 51: This type seems to fit the existing `CreateDestroy` trait. Have you considered implementing that trait or is there a catch?

io-engine/src/rebuild/rebuild_error.rs
tiagolobocastro: Done

io-engine/src/rebuild/snapshot_rebuild.rs
tiagolobocastro: This actually makes use of `CreateDestroy` internally but exposes slightly different semantics.
We could make Uri implement the trait but since this is private there's probably no need for that at the moment?

dsharma-dc: Would look consistent if we implement now. However it seems we'll have to add a `close` method in the trait and implement that for all existing impls as well?
A separate ticket can be used for that work.

tiagolobocastro: Actually this cannot be done because the trait has ?Send limitation

dsharma-dc (approved):

tiagolobocastro:
bors merge

PR #1672, opened 2024-06-13
feat(rebuild): prefer a local replica as a rebuild source

dsavitskiy:
This would reduce network traffic when rebuilding a remote replica and a local one is healthy.

### DIFF
--- a/io-engine-tests/src/nexus.rs
+++ b/io-engine-tests/src/nexus.rs
@@ -130,6 +130,11 @@ impl NexusBuilder {
         self.with_bdev(&bdev)
     }
 
+    pub fn with_replicas(self, replicas: &[ReplicaBuilder]) -> Self {
+        let cc = replicas.iter().map(|r| self.replica_uri(r)).collect();
+        self.with_children(cc)
+    }
+
     pub fn with_local_replica(self, r: &ReplicaBuilder) -> Self {
         if r.rpc() != self.rpc() {
             panic!("Replica is not local");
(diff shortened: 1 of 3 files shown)


dsavitskiy:
bors try

tiagolobocastro (commented):
Should there be some limits? Example let's say we have 200 nexuses on a node and they're all doing rebuilds with the local replica, perhaps in such case we should not only choose the local replica

dsavitskiy:
> Should there be some limits? Example let's say we have 200 nexuses on a node and they're all doing rebuilds with the local replica, perhaps in such case we should not only choose the local replica

This statement equally applies to rebuilding from a remote replica. My change is to reduce overall load on the network: previously, we were always blindly selecting the first healthy replica.

To implement a smarter way of source replica selection, we'd need control plane to tell which replica is preferable at the given moment, as the data plane cannot know what is better to cluster as a whole. Data plane can probably tell if its own node is highly loaded, but it cannot tell if another node is actually so much less loaded, and increasing network traffic by rebuilding from that remote node would actually improve the situation for the entire cluster.

We could add a preferred source replica as an optional argument for "add_child", "online_child" and "rebuild_child" APIs. Data plane would take that replica if it is healthy at the moment the rebuild starts. Control plane would need to have some logic implemented to determine which replica to prefer.

tiagolobocastro (approved):

dsharma-dc (approved):

dsavitskiy:
bors merge

PR #1687, opened 2024-07-05
refactor(snapshots): use generic snapshot ops interface

tiagolobocastro:
The previous SnapshotOps interface was not generic, and rather dependent on Lvol.
Add a new SnapshotOps which attempts to be more generic. A single reference to Lvol remains which should still be removed.
The previous SnapshotOps is renamed to LvolSnapshotOps.

This also fixes a bug where destroy_snapshot was actually also destroying regular replicas! And vice versa!

### DIFF
--- a/io-engine/src/lvs/lvol_snapshot.rs
+++ b/io-engine/src/lvs/lvol_snapshot.rs
@@ -2,11 +2,11 @@ use std::{
     convert::TryFrom,
     ffi::{c_ushort, c_void, CString},
     mem::zeroed,
+    ops::Deref,
     os::raw::c_char,
 };
 
 use async_trait::async_trait;
-use chrono::Utc;
 use futures::{channel::oneshot, future::join_all};
 use nix::errno::Errno;
 use strum::{EnumCount, IntoEnumIterator};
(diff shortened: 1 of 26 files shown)


tiagolobocastro:
bors try

tiagolobocastro:
bors try

abhilashshetty04 (approved):

abhilashshetty04:
bors try

io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -44,6 +38,163 @@ use crate::{
...
+        &self,
+        nvmf_req: &NvmfReq,
+        snapshot_params: SnapshotParams,
+    );
+    /// Destroy snapshot.
+    async fn destroy_snapshot(mut self) -> Result<(), Self::Error>;
+
+    /// List Snapshot details based on source UUID from which snapshot is
+    /// created.
+    fn list_snapshot_by_source_uuid(&self) -> Vec<VolumeSnapshotDescriptor>;
```
dsharma-dc: Prefer using standard terminology 

```suggestion
    fn list_snapshot_by_origin(&self) -> Vec<VolumeSnapshotDescriptor>;
```

io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -44,6 +38,163 @@ use crate::{
...
+    /// Create snapshot clone.
+    async fn create_clone(
+        &self,
+        clone_param: CloneParams,
+    ) -> Result<Self::Lvol, Self::Error>;
+
+    /// Get clone list based on snapshot_uuid.
+    fn list_clones_by_snapshot_uuid(&self) -> Vec<Self::Lvol>;
+
+    // Get a Snapshot Iterator.
```
dsharma-dc, line 93:
```suggestion
    /// Get a Snapshot Iterator.
```

io-engine/src/replica_backend.rs
```diff
@@ -65,6 +77,51 @@ pub trait ReplicaOps: LogicalVolume {
         -> Result<(), crate::pool_backend::Error>;
 }
 
+#[async_trait::async_trait(?Send)]
+pub trait ReplicaSnapshotOps {
```
dsharma-dc: Do we really need to keep two separate traits `ReplicaSnapshotOps` and `SnapshotOps` when both are in generic backend to be implemented by replica of all backend types? It seems the idea is to separate based on `snapshot creation` vs `clone creation using snapshot`. I'd suggest to either clearly document, or perhaps merge the two traits?

io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -44,6 +38,163 @@ use crate::{
 
 use super::{BsError, Lvol, LvsError, LvsLvol};
 
+/// Result for low-level Lvol calls.
+pub type LvolResult = Result<*mut spdk_lvol, Errno>;
+
+///  Traits gives the common snapshot/clone interface for Local/Remote Lvol.
+#[async_trait::async_trait(?Send)]
+pub trait LvolSnapshotOps {
```
dsharma-dc, line 51: There will be a corresponding `LogicalVolumeSnapshotOps` for say lvm backend, when it supports snapshot?

dsharma-dc (approved):

io-engine/src/lvs/lvol_snapshot.rs
tiagolobocastro: I did not change this naming, already existing.

io-engine/src/lvs/lvol_snapshot.rs
tiagolobocastro: I don't think so; this is only kept here to avoid making further changes.
The other generic traits should suffice, I hope.

io-engine/src/replica_backend.rs
tiagolobocastro: `ReplicaSnapshotOps` could be merged into `ReplicaOps`, it's separated only since it's logically a different operation.

`SnapshotOps` should be separate as some of its operations don't make sense for a replica, they're snapshot specific. If we ever add cloning capabilities to replica directly, then we might need some changes there, perhaps.

tiagolobocastro: As discussed moved `ReplicaSnapshotOps` into `ReplicaOps`

tiagolobocastro:
bors try

tiagolobocastro:
bors cancel

tiagolobocastro:
bors try

tiagolobocastro:
weird, a comment caused a build error... also locally I see files modified which shouldn't be...
bors try

dsharma-dc:
> also locally I see files modified which shouldn't be

I also noticed this thing today in my local workspace, hmm.

dsharma-dc:
Is the modification to refactor `create_snapshot_remote` for clarity still pending?

tiagolobocastro:
> Is the modification to refactor `create_snapshot_remote` for clarity still pending?

No, but I had left the unused ones in the lvs specific trait. I've now removed those as well.

dsharma-dc (approved):

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

PR #1691, opened 2024-07-15
feat: add RDMA option to cli args and configs

dsharma-dc:


dsharma-dc:
bors try

dsharma-dc:
bors try

io-engine/src/subsys/config/opts.rs
```diff
@@ -97,6 +97,8 @@ pub struct NvmfTgtConfig {
     pub opts: NvmfTcpTransportOpts,
```
tiagolobocastro, line 97: Do we need NvmfRdmaTransportOpts?

tiagolobocastro (commented):
We should add rdma feature to the features proto api?

io-engine/src/subsys/config/opts.rs
dsharma-dc: > Do we need NvmfRdmaTransportOpts?

Yes, that is part of the upcoming changes.

dsharma-dc:
> We should add rdma feature to the features proto api?

Yes, have noted that in the doc. We might need change in Nexus share API too but not likely in first phase.

dsavitskiy (approved):

tiagolobocastro (approved):

dsharma-dc:
bors merge

PR #1692, opened 2024-07-16
fix: use auto-detected sector size for blockdev

dsharma-dc:
This fixes the behaviour where we pass 512 as sector size if the disk uri doesn't contain blk_size parameter. This causes pool creation failure if the underlying disk has a different sector size e.g. 4096. Instead of passing 512, we now pass 0 which lets spdk detect the device's sector size and use that value.

### DIFF
--- a/io-engine-tests/src/lib.rs
+++ b/io-engine-tests/src/lib.rs
@@ -188,6 +188,28 @@ pub fn truncate_file_bytes(path: &str, size: u64) {
     assert!(output.status.success());
 }
 
+/// Automatically assign a loopdev to path
+pub fn setup_loopdev_file(path: &str, sector_size: Option<u64>) -> String {
+    let log_sec = sector_size.unwrap_or(512);
+
+    let output = Command::new("losetup")
+        .args(["-f", "--show", "-b", &format!("{log_sec}"), path])
+        .output()
+        .expect("failed exec losetup");
+    assert!(output.status.success());
+    // return the assigned loop device
+    String::from_utf8(output.stdout).unwrap().trim().to_string()
+}
+
+/// Detach the provided loop device.
+pub fn detach_loopdev(dev: &str) {
+    let output = Command::new("losetup")
+        .args(["-d", dev])
+        .output()
+        .expect("failed exec losetup");
+    assert!(output.status.success());
+}
+
 pub fn fscheck(device: &str) {
     let output = Command::new("fsck")
         .args([device, "-n"])
--- a/io-engine/tests/lvs_pool.rs
+++ b/io-engine/tests/lvs_pool.rs
@@ -16,14 +16,33 @@ use std::pin::Pin;
 
 pub mod common;
 
-static DISKNAME1: &str = "/tmp/disk1.img";
-static DISKNAME2: &str = "/tmp/disk2.img";
+static TESTDIR: &str = "/tmp/io-engine-tests";
+static DISKNAME1: &str = "/tmp/io-engine-tests/disk1.img";
+static DISKNAME2: &str = "/tmp/io-engine-tests/disk2.img";
+static DISKNAME3: &str = "/tmp/io-engine-tests/disk3.img";
 
 #[tokio::test]
 async fn lvs_pool_test() {
-    common::delete_file(&[DISKNAME1.into(), DISKNAME2.into()]);
+    // Create directory for placing test disk files
+    // todo: Create this from some common place and use for all other tests too.
+    let _ = std::process::Command::new("mkdir")
+        .args(["-p"])
+        .args([TESTDIR])
+        .output()
+        .expect("failed to execute mkdir");
+
+    common::delete_file(&[
+        DISKNAME1.into(),
+        DISKNAME2.into(),
+        DISKNAME3.into(),
+    ]);
     common::truncate_file(DISKNAME1, 128 * 1024);
     common::truncate_file(DISKNAME2, 128 * 1024);
+    common::truncate_file(DISKNAME3, 128 * 1024);
+
+    //setup disk3 via loop device using a sector size of 4096.
+    let ldev = common::setup_loopdev_file(DISKNAME3, Some(4096));
+
     let args = MayastorCliArgs {
         reactor_mask: "0x3".into(),
         ..Default::default()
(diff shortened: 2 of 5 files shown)


dsharma-dc:
bors try

io-engine-tests/src/lib.rs
```diff
@@ -188,6 +188,28 @@ pub fn truncate_file_bytes(path: &str, size: u64) {
     assert!(output.status.success());
 }
 
+/// Automatically assign a loopdev to path
+pub fn setup_loopdev_file(path: &str, sector_size: Option<u64>) -> String {
```
tiagolobocastro, line 192: We don't need a loopdevice, we should keep uses of aio/uring to fallocated file as we want to potentially make it more robust and use it for prod replication for certain use cases

dsharma-dc: > We don't need a loopdevice, we should keep uses of aio/uring to fallocated file as we want to potentially make it more robust and use it for prod replication for certain use cases

It's a helper added for this case and potentially any future uses. With fallocated files we can't simulate larger sector size, but loopdev allows us to do that.

dsavitskiy: IIRC at some point I wanted to add something like this to test helpers, but ended up with some other solution.

dsavitskiy (approved):

io-engine-tests/src/lib.rs
tiagolobocastro: > > We don't need a loopdevice, we should keep uses of aio/uring to fallocated file as we want to potentially make it more robust and use it for prod replication for certain use cases
> 
> It's a helper added for this case and potentially any future uses. With fallocated files we can't simulate larger sector size, but loopdev allows us to do that.

Ah I see :)

io-engine/src/bdev/mod.rs
```diff
@@ -48,6 +50,30 @@ pub trait GetName {
...
 }
 
+/// Check if the path indicates a blockdev, or a regular file. It will be
+/// helpful for cases where we use a regular file e.g. a tmpfs file, as a pool
+/// backend. XXX: If we can't clearly determine if a path is blockdev(error
+/// conditions), we return false today so that caller side can decide what to
+/// do.
+pub fn path_is_blockdev(path: &str) -> bool {
+    // Create a mutable stat struct
+    let mut stat_result: stat = unsafe { std::mem::zeroed() };
```
tiagolobocastro: Why are you using unsafe here? Is it not possible to achieve the same using safe std api?

io-engine/tests/lvs_pool.rs
```diff
@@ -381,4 +445,6 @@ async fn lvs_pool_test() {
     .await;
 
     common::delete_file(&[DISKNAME2.into()]);
+    common::detach_loopdev(ldev.as_str());
```
tiagolobocastro, line 483: To avoid leaking loopdevices I'd suggest using RAII and also perhaps detaching leaked devices in the cargo cleanup script

io-engine/src/bdev/mod.rs
dsharma-dc: > Why are you using unsafe here? Is it not possible to achieve the same using safe std api?

I don't see any, since `stat` is an external libc type here. This function's `unsafe` usage is widespread in Mayastor code.

io-engine/tests/lvs_pool.rs
dsharma-dc: hmm, RAII here may be too much to do in test. w.r.t cleanup from cargo script, are you suggesting doing `losetup -D` from `clean-cargo-tests.sh` ? That might detach something we didn't attach as well. The problem with detaching loopdev is that if there is no user anymore, the detach succeeds, otherwise the device lingers with a `auto clear` flag and doesn't go away. like below:
```
NAME       SIZELIMIT OFFSET AUTOCLEAR RO BACK-FILE                    DIO LOG-SEC
/dev/loop0       0   0         1       0 /home/user/disk1              0     512
```

io-engine/src/bdev/mod.rs
tiagolobocastro: std::fs::metadata(path)?.is_block_device() ?

io-engine/tests/lvs_pool.rs
tiagolobocastro: Not -D, only the devices which we had created. Is it possible to add some kind of flag? Or perhaps we can place the backing files in specific directory and detach those?

> otherwise the device lingers with a auto clear flag and doesn't go away. like below:

You mean if the file gets deleted first, then loopdev gets stuck?

dsharma-dc: > Or perhaps we can place the backing files in specific directory and detach those?

I can keep keep the backing file in `/tmp/loop_disks/<filename>` instead of directly under `/tmp`. Let the cargo test do detachments as it is doing now. Additionally, from the cleanup script do device detach for each device by getting backing filename like `losetup -j /tmp/loop_disks/disk1.img -O NAME --noheadings`, for each backing file found under `/tmp/loop_disks/`. wdyt?

> You mean if the file gets deleted first, then loopdev gets stuck?

No, not if backing file is deleted. It is most likely an open descriptor or a mounted filesystem etc left uncleaned for the device. e.g. I had a loopdev used by lvm, and now that I deleted lvm setup the device isn't detaching.

io-engine/src/bdev/mod.rs
dsharma-dc: > std::fs::metadata(path)?.is_block_device() ?

Looks relevant, would be something like `std::fs::metadata(path)?.file_type().is_block_device()`. Let me try.

dsharma-dc: Updated to use `std::fs:metadata` now.

dsharma-dc:
bors try

io-engine/tests/lvs_pool.rs
tiagolobocastro: Yeah sounds good, maybe I'd suggest /tmp/io-engine/ to make it clear it's our stuff? Really we should do that for other stuff as well tbh

io-engine/src/bdev/mod.rs
tiagolobocastro: Nice :+1:

io-engine/tests/lvs_pool.rs
dsharma-dc: Done.

tiagolobocastro (approved):

dsharma-dc:
bors merge

scripts/clean-cargo-tests.sh
```diff
@@ -1,8 +1,27 @@
+#!/bin/env bash
```
tiagolobocastro, line 1:
```suggestion
#!/usr/bin/env bash
```

tiagolobocastro:
```suggestion
#!/usr/bin/env bash
```

dsharma-dc:
bors merge

PR #1701, opened 2024-07-29
feat: add RDMA listener to Mayastor Nvmf target

dsharma-dc:
This adds the capability to listen for rdma connections to the Mayastor Nvmf target if the rdma feature is enabled during installation. Any Nvmf subsystem facing the host i.e. the nexus nvmf subsystem will now be able to support tcp and rdma both.

### DIFF
--- a/io-engine/src/core/bdev.rs
+++ b/io-engine/src/core/bdev.rs
@@ -11,7 +11,7 @@ use snafu::ResultExt;
 use spdk_rs::libspdk::{spdk_bdev, spdk_get_ticks_hz};
 
 use crate::{
-    bdev::bdev_event_callback,
+    bdev::{bdev_event_callback, nexus::NEXUS_MODULE_NAME},
     bdev_api::bdev_uri_eq,
     core::{
         share::{NvmfShareProps, Protocol, Share, UpdateProps},
@@ -210,6 +210,7 @@ where
     ) -> Result<Self::Output, Self::Error> {
         let me = unsafe { self.get_unchecked_mut() };
         let props = NvmfShareProps::from(props);
+        let is_nexus_bdev = me.driver() == NEXUS_MODULE_NAME;
 
         let ptpl = props.ptpl().as_ref().map(|ptpl| ptpl.path());
 
@@ -232,7 +233,7 @@ where
             .await
             .context(ShareNvmf {})?;
 
-        subsystem.start().await.context(ShareNvmf {})
+        subsystem.start(is_nexus_bdev).await.context(ShareNvmf {})
     }
 
     fn create_ptpl(&self) -> Result<Option<PtplProps>, Self::Error> {
--- a/io-engine/src/core/env.rs
+++ b/io-engine/src/core/env.rs
@@ -834,6 +834,11 @@ impl MayastorEnvironment {
             .map(|s| s.clone())
     }
 
+    /// Check if RDMA needs to be enabled for Mayastor nvmf target.
+    pub fn rdma(&self) -> bool {
+        self.rdma
+    }
+
     /// Detects IP address for NVMF target by the interface specified in CLI
     /// arguments.
     fn detect_nvmf_tgt_iface_ip(iface: &str) -> Result<String, String> {
--- a/io-engine/src/subsys/config/opts.rs
+++ b/io-engine/src/subsys/config/opts.rs
@@ -35,6 +35,7 @@ use std::{
 };
 
 use crate::core::MayastorEnvironment;
+use strum_macros::{AsRefStr, EnumString, EnumVariantNames};
 
 pub trait GetOpts {
     fn get(&self) -> Self;
@@ -84,6 +85,30 @@ impl GetOpts for NexusOpts {
 /// Must be equal to the size of `spdk_nvmf_target_opts.crdt`.
 pub const TARGET_CRDT_LEN: usize = 3;
 
+#[derive(Clone, Default, EnumString, EnumVariantNames, AsRefStr)]
+#[strum(serialize_all = "lowercase")]
+pub enum NvmfTgtTransport {
+    Rdma,
+    #[default]
+    Tcp,
+}
+
+impl NvmfTransportOpts {
+    /// Tweak a few opts more suited for rdma.
+    fn for_rdma(mut self) -> Self {
+        self.in_capsule_data_size = try_from_env(
+            "NVMF_RDMA_IN_CAPSULE_DATA_SIZE",
+            self.in_capsule_data_size,
+        );
+        self.io_unit_size = try_from_env("NVMF_RDMA_IO_UNIT_SIZE", 8192); // SPDK_NVMF_RDMA_MIN_IO_BUFFER_SIZE
+        self.data_wr_pool_size =
+            try_from_env("NVMF_RDMA_DATA_WR_POOL_SIZE", 4095); // SPDK_NVMF_RDMA_DEFAULT_DATA_WR_POOL_SIZE
+        self.num_shared_buf =
+            try_from_env("NVMF_RDMA_NUM_SHARED_BUF", self.num_shared_buf);
+        self
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(default, deny_unknown_fields)]
 pub struct NvmfTgtConfig {
@@ -94,11 +119,13 @@ pub struct NvmfTgtConfig {
     /// NVMF target Command Retry Delay in x100 ms.
     pub crdt: [u16; TARGET_CRDT_LEN],
     /// TCP transport options
-    pub opts: NvmfTcpTransportOpts,
+    pub opts_tcp: NvmfTransportOpts,
     /// NVMF target interface (ip, mac, name or subnet).
     pub interface: Option<String>,
     /// Enable RDMA for NVMF target or not
     pub rdma: Option<bool>,
+    /// RDMA transport options.
+    pub opts_rdma: NvmfTransportOpts,
 }
 
 impl From<NvmfTgtConfig> for Box<spdk_nvmf_target_opts> {
@@ -126,9 +153,10 @@ impl Default for NvmfTgtConfig {
             name: "mayastor_target".to_string(),
             max_namespaces: 2048,
             crdt: args.nvmf_tgt_crdt,
-            opts: NvmfTcpTransportOpts::default(),
+            opts_tcp: NvmfTransportOpts::default(),
             interface: None,
             rdma: None,
+            opts_rdma: NvmfTransportOpts::default().for_rdma(),
         }
     }
 }
@@ -139,10 +167,10 @@ impl GetOpts for NvmfTgtConfig {
     }
 }
 
-/// Settings for the TCP transport
+/// Nvmf settings for the transports
 #[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
 #[serde(default, deny_unknown_fields)]
-pub struct NvmfTcpTransportOpts {
+pub struct NvmfTransportOpts {
     /// max queue depth
     max_queue_depth: u16,
     /// max qpairs per controller
@@ -264,7 +292,7 @@ where
     }
 }
 
-impl Default for NvmfTcpTransportOpts {
+impl Default for NvmfTransportOpts {
     fn default() -> Self {
         Self {
             max_queue_depth: try_from_env("NVMF_TCP_MAX_QUEUE_DEPTH", 32),
@@ -283,7 +311,7 @@ impl Default for NvmfTcpTransportOpts {
             acceptor_poll_rate: try_from_env("NVMF_ACCEPTOR_POLL_RATE", 10_000),
             zcopy: try_from_env("NVMF_ZCOPY", 1) == 1,
             ack_timeout: try_from_env("NVMF_ACK_TIMEOUT", 0),
-            data_wr_pool_size: try_from_env("NVMF_DATA_WR_POOL_SIZE", 0),
+            data_wr_pool_size: 0,
         }
     }
 }
@@ -291,8 +319,8 @@ impl Default for NvmfTcpTransportOpts {
 /// we cannot add derives for YAML to these structs directly, so we need to
 /// copy them. The upside though, is that if the FFI structures change, we will
 /// know about it during compile time.
-impl From<NvmfTcpTransportOpts> for spdk_nvmf_transport_opts {
-    fn from(o: NvmfTcpTransportOpts) -> Self {
+impl From<NvmfTransportOpts> for spdk_nvmf_transport_opts {
+    fn from(o: NvmfTransportOpts) -> Self {
         struct_size_init!(
             Self {
                 max_queue_depth: o.max_queue_depth,
--- a/io-engine/src/subsys/nvmf/subsystem.rs
+++ b/io-engine/src/subsys/nvmf/subsystem.rs
@@ -48,6 +48,7 @@ use spdk_rs::{
         spdk_nvmf_subsystem_state_change_done,
         spdk_nvmf_subsystem_stop,
         spdk_nvmf_tgt,
+        spdk_nvmf_tgt_get_transport,
         SPDK_NVME_SCT_GENERIC,
         SPDK_NVME_SC_CAPACITY_EXCEEDED,
         SPDK_NVME_SC_RESERVATION_CONFLICT,
@@ -68,8 +69,13 @@ use crate::{
     ffihelper::{cb_arg, done_cb, AsStr, FfiResult, IntoCString},
     lvs::Lvol,
     subsys::{
+        config::opts::NvmfTgtTransport,
         make_subsystem_serial,
-        nvmf::{transport::TransportId, Error, NVMF_TGT},
+        nvmf::{
+            transport::{TransportId, RDMA_TRANSPORT},
+            Error,
+            NVMF_TGT,
+        },
         Config,
     },
 };
@@ -798,7 +804,10 @@ impl NvmfSubsystem {
     }
 
     // we currently allow all listeners to the subsystem
-    async fn add_listener(&self) -> Result<(), Error> {
+    async fn add_listener(
+        &self,
+        transport: NvmfTgtTransport,
+    ) -> Result<(), Error> {
         extern "C" fn listen_cb(arg: *mut c_void, status: i32) {
             let s = unsafe { Box::from_raw(arg as *mut oneshot::Sender<i32>) };
             s.send(status).unwrap();
@@ -807,8 +816,8 @@ impl NvmfSubsystem {
         let cfg = Config::get();
 
         // dont yet enable both ports, IOW just add one transportID now
-
-        let trid_replica = TransportId::new(cfg.nexus_opts.nvmf_replica_port);
+        let trid_replica =
+            TransportId::new(cfg.nexus_opts.nvmf_replica_port, transport);
 
         let (s, r) = oneshot::channel::<i32>();
         unsafe {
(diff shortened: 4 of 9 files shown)


dsharma-dc:
bors try

dsharma-dc:
bors try

dsharma-dc:
bors try

io-engine/src/subsys/config/opts.rs
```diff
@@ -94,11 +181,13 @@ pub struct NvmfTgtConfig {
...
     pub crdt: [u16; TARGET_CRDT_LEN],
     /// TCP transport options
-    pub opts: NvmfTcpTransportOpts,
+    pub opts_tcp: NvmfTransportOpts,
     /// NVMF target interface (ip, mac, name or subnet).
     pub interface: Option<String>,
     /// Enable RDMA for NVMF target or not
     pub rdma: Option<bool>,
+    /// RDMA transport options
+    pub opts_rdma: NvmfTransportOpts,
```
dsharma-dc, line 128: @tiagolobocastro  I've kept `opts_tcp` and `opts_rdma` separate here in Config as that looks a clearer logical separation in the configuration which is loaded once. Modifying for rdma during load via a new `for_rdma()`.

io-engine/src/core/env.rs
```diff
@@ -834,6 +834,11 @@ impl MayastorEnvironment {
             .map(|s| s.clone())
     }
 
+    /// Check if RDMA needs to be enabled for Mayastor target.
```
```suggestion
    /// Check if RDMA needs to be enabled for Mayastor nexus target.
```

io-engine/src/subsys/config/opts.rs
```diff
@@ -84,6 +84,93 @@ impl GetOpts for NexusOpts {
...
+    zcopy: bool,
+    /// ACK timeout in milliseconds
+    ack_timeout: u32,
+    /// Size of RDMA data WR pool
+    data_wr_pool_size: u32,
+}
+
+impl NvmfTransportOpts {
+    fn for_rdma(mut self) -> Self {
+        self.in_capsule_data_size = 8192;
```
tiagolobocastro: Shouldn't this be 4k (in line with spdk)?

io-engine/src/subsys/config/opts.rs
```diff
@@ -84,6 +84,93 @@ impl GetOpts for NexusOpts {
...
+        self.io_unit_size = 8192;
```
tiagolobocastro:
```suggestion
        self.io_unit_size = SPDK_NVMF_RDMA_DEFAULT_MAX_IO_SIZE / SPDK_NVMF_MAX_SGL_ENTRIES;
```

io-engine/src/subsys/config/opts.rs
```diff
@@ -84,6 +84,93 @@ impl GetOpts for NexusOpts {
...
+        self.data_wr_pool_size = 4095;
+        self
+    }
+}
+
+impl Default for NvmfTransportOpts {
+    fn default() -> Self {
+        Self {
+            max_queue_depth: try_from_env("NVMF_TCP_MAX_QUEUE_DEPTH", 32),
```
tiagolobocastro: I think the downside of doing it this way is that now we can't control the max q depth just for tcp?

io-engine/src/subsys/config/opts.rs
```diff
@@ -84,6 +84,93 @@ impl GetOpts for NexusOpts {
...
+            ),
+            num_shared_buf: try_from_env("NVMF_TCP_NUM_SHARED_BUF", 2047),
+            buf_cache_size: try_from_env("NVMF_TCP_BUF_CACHE_SIZE", 64),
+            dif_insert_or_strip: false,
+            max_aq_depth: 32,
+            abort_timeout_sec: 1,
+            acceptor_poll_rate: try_from_env("NVMF_ACCEPTOR_POLL_RATE", 10_000),
+            zcopy: try_from_env("NVMF_ZCOPY", 1) == 1,
+            ack_timeout: try_from_env("NVMF_ACK_TIMEOUT", 0),
+            data_wr_pool_size: try_from_env("NVMF_DATA_WR_POOL_SIZE", 0),
```
tiagolobocastro: And here it's NVMF but actually we overwrite it on for_rdma anyway, so end up modifying it for tcp?

io-engine/src/subsys/config/opts.rs
tiagolobocastro: I suggest having a different way, example:
rdmaOpts:
  field: env_var("NVMF_RDMA_....", env_var("NVMF", rdma default ? common default
So, first check for transport specific override, then generic override, then specific default, and only if nothing sets it then common default

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -798,7 +804,7 @@ impl NvmfSubsystem {
     }
 
     // we currently allow all listeners to the subsystem
-    async fn add_listener(&self) -> Result<(), Error> {
+    async fn add_listener(&self, xprt: NvmfTgtTransport) -> Result<(), Error> {
```
tiagolobocastro:
```suggestion
    async fn add_listener(&self, transport: NvmfTgtTransport) -> Result<(), Error> {
```

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -958,9 +979,15 @@ impl NvmfSubsystem {
     }
 
     /// get ANA state
+    /// XXX: The SPDK NVME multipath is transport protocol independent. Should
+    /// we keep returning the ana state here by default using TCP transport
+    /// as today?
```
tiagolobocastro, line 987: does this mean, that we should instead return multiple states, one per transport type?

io-engine/src/subsys/nvmf/subsystem.rs
```diff
@@ -1118,7 +1163,10 @@ impl NvmfSubsystem {
     pub fn uri_endpoints(&self) -> Option<Vec<String>> {
         if let Some(v) = self.listeners_to_vec() {
             let nqn = self.get_nqn();
-            Some(v.iter().map(|t| format!("{t}/{nqn}")).collect::<Vec<_>>())
+            let retvec = Some(
```
tiagolobocastro: what's the change here?

io-engine/src/subsys/nvmf/target.rs
```diff
@@ -268,19 +278,92 @@ impl Target {
...
+                        The target will however keep running with \
+                        only tcp listener, with performance expectations of tcp", e
+                    );
+                });
+        }
         self.next_state();
         Ok(())
     }
 
+    /// Listen for incoming connections by default we only listen on the replica
```
tiagolobocastro: comment a bit confusing

io-engine/src/subsys/nvmf/transport.rs
```diff
@@ -57,8 +62,49 @@ pub async fn add_tcp_transport() -> Result<(), Error> {
...
+            // XXX: How to reset Target.rdma = false
+            // todo: add event mechanism for Target and Nvmfsubsystem
+            warn!(
+                "RDMA enablement failed due to rdma transport add failure.\
+                The target will however keep running with only tcp, \
+                with performance expectations of tcp."
+            ));
+
+        if ret.is_err() {
+            return Ok(());
```
tiagolobocastro, line 87: could probably add the warning here though

io-engine/src/subsys/nvmf/transport.rs
```diff
@@ -57,8 +62,49 @@ pub async fn add_tcp_transport() -> Result<(), Error> {
...
+
+    if add_rdma {
+        info!("Adding RDMA transport for Mayastor Nvmf target");
+        let mut opts = cfg.nvmf_tgt_conf.opts_rdma.into();
+        let transport = unsafe {
+            spdk_nvmf_transport_create(RDMA_TRANSPORT.as_ptr(), &mut opts)
+        };
+
+        let ret = transport
+            .to_result(|_| Error::Transport {
```
tiagolobocastro: not sure what's happening here, we're adding error but then we seem to just clear the error below?

io-engine/src/core/env.rs
dsharma-dc: Not `nexus` here, rather `Mayastor nvmf target`. The self is `MayastorEnvironment` and it's checked in io-engine start path.

io-engine/src/subsys/config/opts.rs
dsharma-dc: Rather use `SPDK_NVMF_RDMA_MIN_IO_BUFFER_SIZE` then? which is set as `SPDK_NVMF_RDMA_DEFAULT_MAX_IO_SIZE / SPDK_NVMF_MAX_SGL_ENTRIES` in spdk.

io-engine/src/subsys/nvmf/subsystem.rs
dsharma-dc: Yeah that's what I was thinking, as a vector. But it seems we might be ok not doing that and keep as-is.

io-engine/src/subsys/nvmf/subsystem.rs
dsharma-dc: You can ignore this for now. I'll revert from this change. The caller of this function is being modified in a separate PR for URI changes.

io-engine/src/subsys/nvmf/target.rs
dsharma-dc: It's an existing comment in `listen()` that got copied here too :) It is about that redundant nexus port not being used as we discussed. I'll modify the comment to clarify a bit.

io-engine/src/subsys/nvmf/transport.rs
dsharma-dc: If `to_result` ffi helper returns `Err(Error::Transport)` (which isn't used anyway except checking is_err()), we are generating a `warn` trace, but return without failing later.

io-engine/src/subsys/nvmf/transport.rs
dsharma-dc: Modifying to this:

```
        let ret = transport
            .to_result(|_| Error::Transport {
                source: Errno::UnknownErrno,
                msg: "failed to create RDMA transport".into(),
            });

        if let Err(e) = ret {
            // XXX: How to reset Target.rdma = false
            // todo: add event mechanism for Target and Nvmfsubsystem
            warn!(
                "RDMA enablement failed {e}.\
                The target will however keep running with only tcp, \
                with performance expectations of tcp."
            );
            return Ok(());
        }
```

io-engine/src/subsys/config/opts.rs
dsharma-dc: yeah, SPDK_NVMF_RDMA_DEFAULT_IN_CAPSULE_DATA_SIZE should be ok. I picked 8192 from somewhere in spdk documentation.

io-engine/src/subsys/config/opts.rs
dsharma-dc: > And here it's NVMF but actually we overwrite it on for_rdma anyway, so end up modifying it for tcp?

this opt is irrelevant for TCP anyway, and is not used underneath. Since it's part of common struct for us, the naming of env variable seems to be chosen so.

io-engine/src/subsys/nvmf/transport.rs
dsharma-dc: Simplified now.

io-engine/src/subsys/config/opts.rs
dsharma-dc: These constants are defined in .c file directly and not in public headers :(  , so using values.

io-engine/src/subsys/config/opts.rs
dsharma-dc: what if the env has both transport specific overrides?
On the other hand, most of these opts are ideally not supposed to be transport specific(except those which are) but since there was only TCP till now, we have the envs named specific to TCP.

tiagolobocastro: > On the other hand, most of these opts are ideally not supposed to be transport specific(except those which are)

Which ones are transport specific?

> but since there was only TCP till now, we have the envs named specific to TCP.

Yes which is why I'm saying if some of those are also applicable for RDMA then do we need to have different env var names if we want to control them independently?

dsharma-dc: > Which ones are transport specific?

e.g ack_timeout is valid only for TCP, and data_wr_size is valid only for RDMA.

> Yes which is why I'm saying if some of those are also applicable for RDMA then do we need to have different env var names if we want to control them independently?

I'd rather argue for renaming all current env variables generically without transport name, to lessen the circus required for setting up opts. If required to override a select few separately for RDMA, do it via RDMA specific env from inside `for_rdma()`. Let's get on a call tomorrow to sort this out.

dsharma-dc: addressed as discussed.

io-engine/src/subsys/config/opts.rs
dsharma-dc: addressed as discussed.

io-engine/src/subsys/nvmf/subsystem.rs
dsharma-dc: modified to `transport` :)

io-engine/src/core/bdev.rs
```diff
@@ -232,7 +233,7 @@ where
             .await
             .context(ShareNvmf {})?;
 
-        subsystem.start().await.context(ShareNvmf {})
+        subsystem.start(is_nexus_bdev).await.context(ShareNvmf {})
```
Abhinandan-Purkait, line 236: Is the indentation off here by chance?

io-engine/src/subsys/config/opts.rs
```diff
@@ -84,6 +84,100 @@ impl GetOpts for NexusOpts {
 /// Must be equal to the size of `spdk_nvmf_target_opts.crdt`.
 pub const TARGET_CRDT_LEN: usize = 3;
 
+#[derive(Clone, Default, Debug, PartialEq, Serialize, Deserialize)]
+pub enum NvmfTgtTransport {
+    Rdma,
+    #[default]
+    Tcp,
+}
```
Abhinandan-Purkait, line 94: Why not use strum macros?

Abhinandan-Purkait (approved):

io-engine/src/core/bdev.rs
Abhinandan-Purkait: Sorry, nevamind, I think it was github UI that showed it in a distorted way.

io-engine/src/subsys/config/opts.rs
dsharma-dc: Thanks, this is something new. Added `strum` annotations now.

dsharma-dc:
bors try

io-engine/src/subsys/config/opts.rs
```diff
@@ -94,11 +177,13 @@ pub struct NvmfTgtConfig {
     /// NVMF target Command Retry Delay in x100 ms.
     pub crdt: [u16; TARGET_CRDT_LEN],
     /// TCP transport options
-    pub opts: NvmfTcpTransportOpts,
+    pub opts_tcp: NvmfTransportOpts,
     /// NVMF target interface (ip, mac, name or subnet).
     pub interface: Option<String>,
     /// Enable RDMA for NVMF target or not
     pub rdma: Option<bool>,
+    /// RDMA transport options
```
tiagolobocastro:
```suggestion
    /// RDMA transport options.
```

io-engine/src/subsys/config/opts.rs
```diff
@@ -264,58 +316,32 @@ where
...
-                zcopy: o.zcopy,
-                reserved61: Default::default(),
-                ack_timeout: o.ack_timeout,
-                data_wr_pool_size: o.data_wr_pool_size,
-            },
-            opts_size
-        )
+impl From<NvmfTransportOpts> for spdk_nvmf_transport_opts {
+    fn from(o: NvmfTransportOpts) -> Self {
+        Self {
```
tiagolobocastro: Why is this not being initialized the same as before with struct_size_init! ?

io-engine/src/subsys/config/opts.rs
```diff
@@ -84,6 +85,88 @@ impl GetOpts for NexusOpts {
...
+pub enum NvmfTgtTransport {
+    Rdma,
+    #[default]
+    Tcp,
+}
+
+/// Nvmf settings for the transports
+#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct NvmfTransportOpts {
```
tiagolobocastro: any reason to move this?

io-engine/src/subsys/nvmf/target.rs
```diff
@@ -221,11 +225,14 @@ impl Target {
         });
     }
 
-    /// Listen for incoming connections by default we only listen on the replica
-    /// port
+    /// Listen for incoming connections, by default we only listen on the
+    /// replica port i.e. NVMF_PORT_REPLICA
```
tiagolobocastro:
```suggestion
    /// replica port i.e. NVMF_PORT_REPLICA.
```

io-engine/src/subsys/nvmf/target.rs
```diff
@@ -57,6 +57,8 @@ pub struct Target {
     poll_group_count: u16,
     /// The current state of the target
     next_state: TargetState,
+    /// Whether the target supports RDMA transport
```
tiagolobocastro:
```suggestion
    /// Whether the target supports RDMA transport.
```

io-engine/src/subsys/nvmf/target.rs
```diff
@@ -268,19 +278,92 @@ impl Target {
...
+
+        if self.rdma {
+            // listen RDMA also.
+            let _ = self.listen_rdma()
+                .map_err(|e| {
+                    warn!(
+                        "failed to listen rdma on address. err: {}:\
+                        The target will however keep running with \
+                        only tcp listener, with performance expectations of tcp", e
+                    );
```
tiagolobocastro:
```suggestion
                        "failed to listen rdma on address. err: {e}:\
                        The target will however keep running with \
                        only tcp listener, with performance expectations of tcp"
                    );
```

io-engine/src/subsys/nvmf/transport.rs
```diff
@@ -32,16 +34,19 @@ use crate::{
 static TCP_TRANSPORT: Lazy<CString> =
     Lazy::new(|| CString::new("TCP").unwrap());
 
-pub async fn add_tcp_transport() -> Result<(), Error> {
+pub static RDMA_TRANSPORT: Lazy<CString> =
+    Lazy::new(|| CString::new("RDMA").unwrap());
```
tiagolobocastro, line 38: btw can't these be lowercase? then the display would just work?

tiagolobocastro (approved):

io-engine/src/subsys/nvmf/transport.rs
dsharma-dc: It'll may be work for the case where we use this const to just generate `spdk_nvme_transport_id` ourselves and hand marshal the `trstring` there. But this const is also used in transport create path where spdk does some validations based on this, hence don't want to touch it.

io-engine/src/subsys/config/opts.rs
dsharma-dc: good catch, not sure if this was something prompted to do by build or a rookie mistake. I've set it back to using macro.

io-engine/src/subsys/config/opts.rs
dsharma-dc: It was likely moved up during rearranging duplicated opts structures. Put back.

dsharma-dc:
bors try-

dsharma-dc:
bors try

dsharma-dc:
bors merge

PR #1703, opened 2024-07-29
build: support podman for load/pushing container images

tiagolobocastro:
By default docker is tried, and then podman.
Support is simplistically verified with the exit code from --version.
First choice can be overridden with DOCKER env variable.


tiagolobocastro:
Making the same changes on all repos would be a pain, so instead I'm adding a common release script which all repos can make use of: https://github.com/openebs/mayastor-dependencies/pull/88

tiagolobocastro:
bors try

nix/pkgs/images/default.nix
```diff
@@ -110,14 +113,44 @@ in
...
+in
+{
+  # keep for back-compat, in case it's necessary
+  inherit mayastor-io-engine mayastor-io-engine-dev mayastor-casperf mayastor-fio-spdk;
+
+  release = {
+    mayastor = {
+      io-engine = mayastor-io-engine;
+      io-engine-client = mayastor-io-engine-client;
+      casperf = mayastor-casperf;
```
dsharma-dc, line 144: do we really need casperf for release build?

dsharma-dc (approved):

Abhinandan-Purkait (approved):

nix/pkgs/images/default.nix
tiagolobocastro: Yes, in fact you might ask the opposite, do we need it for debug builds :)
Sometimes it's useful to test it in debug because the debug build is quicker

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

Issue #1710, opened 2024-08-06
Reactor::block_on can lead to connection errors


tiagolobocastro:
**Describe the bug**
Sometimes connecting to a target can fail:
```
ERROR mayastor::spdk:nvme_fabric.c:596] Connect command failed, rc -125, trtype:TCP adrfam:IPv4 traddr:127.0.0.1 trsvcid:8420 subnqn:nqn.2019-05.io.openebs:replica-0-513   
ERROR mayastor::spdk:nvme_tcp.c:2017] Failed to poll NVMe-oF Fabric CONNECT command 
```

**To Reproduce**
A bit of luck and a lot of concurrent nexus creation and destruction and/or rebuilds.

**Expected behavior**
Either nexus creation fails or rebuild ioqpair fails to connect.

**Additional context**
Seems the issue is caused by usage of Reactor::block_on which can prevent processing of messages which have been taken out of thread message ring, and therefore are not processed during the block_on

As a WA increasing the fabrics connect timeout can help mitigate this.

tiagolobocastro:
Released as part of 2.7.1

PR #1711, opened 2024-08-06
[ BUGFIX ] Fabrics connect timeouts

tiagolobocastro:
Reactor block_on may prevent spdk thread messages from running and therefore this can lead to starvation of messages pulled from the thread ring, which are not polled during the block_on.
There are still a few uses remaining, most during init setup, so mostly harmless, though the Nexus Bdev destruction which runs on blocking code, does still contain a block_on.

---

    fix(nvmf/target): remove usage of block_on
    
    Split creating from starting the subsystem.
    This way we can start the subsystem in master reactor, and then move
    to the next spdk subsystem.
    
---

    fix(nexus-child/unplug): remove usage of block_on
    
    Initially this block_on was added because the remove callback was running in blocking
    fashion, but this has since changed and unplug is actually called from async context.
    As such, we don't need the block_on and simply call the async code directly.
    Also, simplify complete notification, as we can simply close the sender.
    
---

    fix(nvmx/qpair): return errno with absolute value
    
    Otherwise a returned negative value translates into an unknown Errno.
    
---

    feat: allow custom fabrics connect timeout
    
    Allows passing this via env NVMF_FABRICS_CONNECT_TIMEOUT.
    Also defaults it to 3s for now, rather than 500ms.
    
### DIFF
--- a/io-engine/src/bdev/nvmx/uri.rs
+++ b/io-engine/src/bdev/nvmx/uri.rs
@@ -227,6 +227,12 @@ impl<'probe> NvmeControllerContext<'probe> {
             )
             .with_transport_retry_count(
                 Config::get().nvme_bdev_opts.transport_retry_count as u8,
+            )
+            .with_fabrics_connect_timeout_us(
+                crate::subsys::config::opts::try_from_env(
+                    "NVMF_FABRICS_CONNECT_TIMEOUT",
+                    1_000_000,
+                ),
             );
 
         let hostnqn = template.hostnqn.clone().or_else(|| {
--- a/io-engine/src/subsys/nvmf/target.rs
+++ b/io-engine/src/subsys/nvmf/target.rs
@@ -27,7 +27,7 @@ use spdk_rs::libspdk::{
 
 use crate::{
     constants::NVME_CONTROLLER_MODEL_ID,
-    core::{Cores, Mthread, Reactor, Reactors},
+    core::{Cores, Mthread, Reactors},
     ffihelper::{AsStr, FfiResult},
     subsys::{
         nvmf::{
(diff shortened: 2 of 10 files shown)


io-engine/src/bdev/nvmx/uri.rs
```diff
@@ -227,6 +227,12 @@ impl<'probe> NvmeControllerContext<'probe> {
             )
             .with_transport_retry_count(
                 Config::get().nvme_bdev_opts.transport_retry_count as u8,
+            )
+            .with_fabrics_connect_timeout_us(
+                crate::subsys::config::opts::try_from_env(
+                    "NVMF_FABRICS_CONNECT_TIMEOUT",
+                    3_000_000,
```
tiagolobocastro: Probably now we can leave this as 500ms by default since we uncovered the real culprit?
(can still be overridden by env variable)

```suggestion
                    500_000,
```

tiagolobocastro:
bors try

tiagolobocastro:
bors try

io-engine/src/bdev/nvmx/uri.rs
tiagolobocastro: I've left it as 1s for now, just to give some more slack just in case.

io-engine/src/subsys/nvmf/target.rs
```diff
@@ -355,13 +350,20 @@ impl Target {
 
     /// Final state for the target during init.
     pub fn running(&mut self) {
-        self.enable_discovery();
-        info!(
-            "nvmf target accepting new connections and is ready to roll..{}",
-            '\u{1F483}'
-        );
+        let discovery = self.enable_discovery();
```
dsharma-dc: With this change, `enable_discovery` is not really starting the discovery subsystem so perhaps rename the function to `get_discovery_subsystem` or `create_discovery_subsystem`

dsharma-dc (approved):

dsharma-dc:
Do we know the reason why historically these paths were using blocking on futures to complete?

dsavitskiy (approved):

io-engine/src/subsys/nvmf/target.rs
tiagolobocastro: Yep, done

tiagolobocastro:
> Do we know the reason why historically these paths were using blocking on futures to complete?

the unplug code was originally not async, and suspect that's why block_on was used.

The others I don't know, perhaps was just used as a "shotcut"...

tiagolobocastro:
Resolves #1710 
bors merge

tiagolobocastro:
oops typo
bors cancel

tiagolobocastro:
bors merge

PR #1725, opened 2024-08-22
feat(lvs): adding support for pool grow and related tests

dsavitskiy:
* Pool grow API support added to LVS.
* Additional pool stats added.
* Added support for MD reservation hint.
* Related CLI subcommand added. 'pool list' subcommand extended to print additional pool info.
* Related tests added.
* For testing purposes, a non-safe support for resizing malloc bdev added.
* To simplify adding new pool and replica parameters, LVS inner create-related APIs refactored to accept parameters via PoolArgs and ReplicaArgs instead of function arguments.
* Added a example tool (lvs-eval) that creates an LVS with the given parameters and prints it internal configuration (allocated clusters, extent pages, etc.)

### DIFF
--- a/io-engine/src/grpc/v1/pool.rs
+++ b/io-engine/src/grpc/v1/pool.rs
@@ -16,6 +16,7 @@ use crate::{
     },
     lvs::{BsError, LvsError},
     pool_backend::{
+        self,
         FindPoolArgs,
         IPoolFactory,
         ListPoolArgs,
@@ -68,6 +69,11 @@ impl From<&destroy_snapshot_request::Pool> for FindPoolArgs {
 }
 impl From<ExportPoolRequest> for FindPoolArgs {
     fn from(value: ExportPoolRequest) -> Self {
+        Self::name_uuid(value.name, &value.uuid)
+    }
+}
+impl From<GrowPoolRequest> for FindPoolArgs {
+    fn from(value: GrowPoolRequest) -> Self {
         Self::name_uuid(value.name, &value.uuid)
     }
 }
@@ -183,8 +189,16 @@ impl TryFrom<CreatePoolRequest> for PoolArgs {
             disks: args.disks,
             uuid: args.uuid,
             cluster_size: args.cluster_size,
+            md_args: args.md_args.map(|md| md.into()),
             backend: backend.into(),
         })
+    }
+}
+impl From<PoolMetadataArgs> for pool_backend::PoolMetadataArgs {
+    fn from(params: PoolMetadataArgs) -> Self {
+        Self {
+            md_resv_ratio: params.md_resv_ratio,
+        }
     }
 }
 impl From<PoolType> for PoolBackend {
@@ -256,6 +270,7 @@ impl TryFrom<ImportPoolRequest> for PoolArgs {
             disks: args.disks,
             uuid: args.uuid,
             cluster_size: None,
+            md_args: None,
             backend: backend.into(),
         })
     }
@@ -299,6 +314,7 @@ impl PoolGrpc {
                 uuid: args.uuid,
                 thin: args.thin,
                 entity_id: args.entity_id,
+                use_extent_table: None,
             })
             .await
         {
@@ -337,6 +353,10 @@ impl PoolGrpc {
         self.pool.export().await?;
         Ok(())
     }
+    async fn grow(&self) -> Result<(), tonic::Status> {
+        self.pool.grow().await?;
+        Ok(())
+    }
     /// Access the `PoolOps` from this wrapper.
     pub(crate) fn as_ops(&self) -> &dyn PoolOps {
         self.pool.deref()
@@ -361,6 +381,18 @@ impl From<&dyn PoolOps> for Pool {
             committed: value.committed(),
             pooltype: PoolType::from(value.pool_type()) as i32,
             cluster_size: value.cluster_size(),
+            page_size: value.page_size(),
+            disk_capacity: value.disk_capacity(),
+            md_info: value.md_props().map(|md| md.into()),
+        }
+    }
+}
+impl From<pool_backend::PoolMetadataInfo> for PoolMetadataInfo {
+    fn from(value: pool_backend::PoolMetadataInfo) -> Self {
+        Self {
+            md_page_size: value.md_page_size,
+            md_pages: value.md_pages,
+            md_used_pages: value.md_used_pages,
         }
     }
 }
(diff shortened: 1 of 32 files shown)


dsavitskiy:
bors try

dsavitskiy:
bors try

dsavitskiy:
bors try

io-engine/src/grpc/v1/pool.rs
```diff
@@ -623,4 +655,32 @@ impl PoolRpc for PoolService {
...
+                    info!("{:?}", request.get_ref());
+
+                    let pool =
+                        GrpcPoolFactory::finder(request.into_inner()).await?;
+
+                    let previous_pool = Pool::from(pool.as_ops());
+                    pool.grow().await.map_err(Into::<Status>::into)?;
+                    let current_pool = Pool::from(pool.as_ops());
+
+                    Ok(GrowPoolResponse {
```
tiagolobocastro, line 692: Maybe add an info message here? Pool grown to x capacity?

io-engine/src/grpc/v1/pool.rs
```diff
@@ -623,4 +655,32 @@ impl PoolRpc for PoolService {
...
+            GrpcClientContext::new(&request, function_name!()),
+            async move {
+                crate::spdk_submit!(async move {
+                    info!("{:?}", request.get_ref());
+
+                    let pool =
+                        GrpcPoolFactory::finder(request.into_inner()).await?;
+
+                    let previous_pool = Pool::from(pool.as_ops());
+                    pool.grow().await.map_err(Into::<Status>::into)?;
```
tiagolobocastro, line 674: How does this fail if the device is same size?
Does it return error, or just return pool as is?

io-engine/src/lvm/vg_pool.rs
```diff
@@ -358,11 +358,18 @@ impl VolumeGroup {
...
         self.size
     }
+
     /// Get the volume group cluster size.
     pub(crate) fn cluster_size(&self) -> u64 {
         4 * 1024 * 1024
     }
 
+    /// Get the volume group page size.
+    pub(crate) fn page_size(&self) -> u64 {
```
tiagolobocastro: hmm maybe page_size should be optional?

io-engine/tests/lvs_grow.rs
```diff
@@ -0,0 +1,167 @@
...
+        assert!(approx_eq(
+            lvs.disk_capacity() as f64,
+            lvs.capacity() as f64,
+            10.0
+        ));
+
+        // Resize the malloc bdev.
+        unsafe {
+            let name = BDEV_NAME.to_owned();
+            resize_malloc_disk(name.into_cstring().as_ptr(), SIZE_AFTER_MB);
```
tiagolobocastro: Can we also add grow test for aio device with a file?

tiagolobocastro (approved):

io-engine/src/grpc/v1/pool.rs
dsavitskiy: It is idempotent: calling multiple times in a sequence yields the same result. If the device did not grow, and the pool stays the same, there's no error.
To detect if the pool has actually grown during the call, one has to compare the returned "previous_pool" vs "current_pool".
(SPDK internally just returns success if the pool size did not change.)

io-engine/src/lvm/vg_pool.rs
dsavitskiy: Done

io-engine/src/grpc/v1/pool.rs
dsavitskiy: Done

dsavitskiy:
bors try

io-engine/tests/lvs_grow.rs
dsavitskiy: Ok; I will add.

dsavitskiy:
bors try

dsavitskiy:
bors try

io-engine/src/lvs/lvs_store.rs
```diff
@@ -226,6 +234,26 @@ impl Lvs {
...
 
+    /// Returns blobstore page size.
+    pub fn page_size(&self) -> u64 {
+        unsafe { spdk_bs_get_page_size(self.blob_store()) }
+    }
+
+    /// Returns blobstore metadata page size.
+    pub fn md_page_size(&self) -> u64 {
+        unsafe { spdk_bs_get_page_size(self.blob_store()) }
+    }
```
dsharma-dc: `md_page_size()` looks redundant. There can be only a single page size for system and it's ok to use only `page_size()` for that.

io-engine/src/lvs/lvs_store.rs
```diff
@@ -409,41 +437,64 @@ impl Lvs {
...
-    /// Create a pool on base bdev
-    pub async fn create(
-        name: &str,
-        bdev: &str,
-        uuid: Option<String>,
-        cluster_size: Option<u32>,
+    /// Converts floating point metadata reservation ratio into SPDK's format.
+    fn mdp_ratio(args: &PoolArgs) -> Result<u32, LvsError> {
+        if let Some(h) = args.md_args.as_ref().and_then(|p| p.md_resv_ratio) {
+            if h > 0.0 {
```
dsharma-dc, line 438: Should we better ensure this is `>1.0` ? Otherwise it indicates shrinkage.

io-engine/src/lvs/lvs_store.rs
```diff
@@ -731,6 +783,35 @@ impl Lvs {
...
+    pub async fn grow(&self) -> Result<(), LvsError> {
+        info!("{self:?}: growing lvs...");
+
+        let (s, r) = pair::<i32>();
+
+        unsafe {
+            let lvs = self.as_inner_ptr();
+
+            // Update block count on spdk_bs_bdev.
+            spdk_bdev_update_bs_blockcnt((*lvs).bs_dev);
```
dsharma-dc, line 792: Is this update automatically reverted in case grow fails?

io-engine/src/lvs/lvs_store.rs
dsavitskiy: Ok, I will change.

io-engine/src/lvs/lvs_store.rs
dsavitskiy: Nope; it means reserve less space for metadata.
E.g., 0.5 means allocate 1 MD page per 2 segments.
0.25 means allocate 1 MD page per 4 segments.

io-engine/src/lvs/lvs_store.rs
dsavitskiy: No, but it is not needed.

In SPDK blobstore, there's another layer of abstraction/indirection on top of SPDK bdev: spdk_bs_bdev. It is needed for provide a uniform interface for possible Blob storage backends: normal bdev, a virtual zero-filled backend for thin provisioning, and some other stuff. Basically, it is like a simplified version of bdev struct, added for needs of blob store code.

For performance reasons (my guess), spdk_bs_bdev struct caches the underlying device size. This function spdk_bdev_update_bs_blockcnt(), it just updates this cached device size so it reflects the actual bdev size. It has nothing to do with pool grow itself.

(My opinion is that blob store should update that cache size automatically in response to bdev resize event, but SPDK does not implement that. Probably, they will fix that in future.)

dsavitskiy:
bors try

dsavitskiy:
bors try

dsharma-dc (approved):

dsavitskiy:
bors merge

PR #1726, opened 2024-08-26
feat: use transport specific nvmf uri scheme

dsharma-dc:


dsavitskiy (approved):

io-engine/src/target/nvmf.rs
```diff
@@ -35,9 +35,14 @@ pub async fn unshare(uuid: &str) -> Result<(), NvmfError> {
...
     if let Some(ss) = NvmfSubsystem::nqn_lookup(uuid) {
-        // for now we only pop the first but we can share a bdev
-        // over multiple nqn's
-        ss.uri_endpoints().unwrap().pop()
+        // for now we only pop the most relevant but we can share a bdev
+        // over multiple nqn's. An nvmf+rdma uri will be prioritised for
+        // returning from here. If not present, then nvmf+tcp will be returned.
+        let mut uris = ss.uri_endpoints().expect("no uri endpoints");
+        let rdma_uri =
+            uris.iter().find(|u| u.starts_with("nvmf+RDMA")).cloned();
```
tiagolobocastro: Should this be lowercase?
```suggestion
            uris.iter().find(|u| u.starts_with("nvmf+rdma")).cloned();
```

io-engine/src/bdev/dev.rs
```diff
@@ -66,7 +66,12 @@ pub(crate) mod uri {
             }
             "malloc" => Ok(Box::new(malloc::Malloc::try_from(&url)?)),
             "null" => Ok(Box::new(null_bdev::Null::try_from(&url)?)),
-            "nvmf" => Ok(Box::new(nvmx::NvmfDeviceTemplate::try_from(&url)?)),
+            // keeping nvmf scheme so existing tests(if any, setting this
+            // scheme) work. The replicas and nexus however should
+            // always be exposing nvmf+tcp or nvmf+rdma now.
+            "nvmf" | "nvmf+tcp" | "nvmf+rdma" => {
```
tiagolobocastro: So nvmf is now equivalent to nvmf+tcp?
What about when we expose both tcp and rdma?

io-engine/src/target/nvmf.rs
dsharma-dc: > Should this be lowercase?

The string constants in Mayastor and SPDK are capital case today so the `trstring` automatically comes in capital case. In some places, where required the case-insensitive comparison will be done.

io-engine/src/bdev/dev.rs
dsharma-dc: > So nvmf is now equivalent to nvmf+tcp?

Yeah, but not expecting `nvmf` to be generated anymore.

>  What about when we expose both tcp and rdma?

`nvmf` may well be removed later. Also, I will hold to merge this change until control plane changes are also ready to handle new uris.

tiagolobocastro: > nvmf may well be removed later. Also, I will hold to merge this change until control plane changes are also ready to handle new uris.

This would be a problem for back compatibility. Maybe if RDMA is not enabled we can keep using just nvmf?

Also I don't think I got the answer to, what if rdma is enabled, meaning we expose both tcp and rdma, which uri would we expose?
nvmf+rdma?
nvmf+tcp+rdma?
A list of URIs? nvmf+tcp:://....., nvmf+rdma:///....

dsharma-dc: When RDMA is enabled, the uri will be `nvmf+rdma` (tcp will be implied).

> Maybe if RDMA is not enabled we can keep using just nvmf?

This would need special handling in some places which may not be likeable. I'd say we can retain `nvmf` too and back-compat can be handled, e.g. new control plane talking to old dataplane will see `nvmf` uri and infer tcp behaviour?

tiagolobocastro: > When RDMA is enabled, the uri will be nvmf+rdma (tcp will be implied).

Seems a bit confusing to imply tcp when using rdma? 
Whilst today tcp is used as fallback, this may not always be the case? What if we also use fc at some point, how would it work then?

Would it not be clearer to use +rdma+tcp for example? Or expose a list of URIs

io-engine/src/target/nvmf.rs
tiagolobocastro: Could we pass `trstring` as lowercase ?
Otherwise we can modify the display in one of 2 ways:
1. generate the type string from the trtype using match
2. use tolower directly there

io-engine/src/bdev/dev.rs
dsharma-dc: > Seems a bit confusing to imply tcp when using rdma?

I think it's a fair internal assumption and keeps things simpler to handle.

> Whilst today tcp is used as fallback, this may not always be the case? 

I'd think this behaviour will remain so always. Even if it doesn't, we may just need to change caller side to not fallback. Even if(at all) fc comes in, same fallback behaviour will remain. fc and rdma interconnect can't be together anyway.
I earlier thought about uri list too, but that changes the API path where we just send a string (single uri).

Abhinandan-Purkait: Just a thought, since we are having this discussion, would it be more clear if this becomes a query param, like `?transport=tcp+rdma` that way the scheme remains same and we don't have to think about backward comp. ?

io-engine/src/target/nvmf.rs
dsharma-dc: I can do this in Formatter for `TransportId`. to_lowercase creates a new String, but should be ok here.

```
        write!(
            f,
            "nvmf+{}://{}:{}",
            self.0.trstring.as_str().to_lowercase(),
            self.0.traddr.as_str(),
            self.0.trsvcid.as_str()
        )
    }
```

io-engine/src/bdev/dev.rs
tiagolobocastro: As long as nvmf is kept when rdma is disabled, then I think backward compat is solved. Given that you'd need new control-plane to support rdma anyway.
Also to consider, if we want to support fc in the future then query parameter won't work because the traddr is different.
I think going forward, a list of URI's probably would be the saner way where you provide a list of potential connections. Also this would work in case we want to specify multiple networks for multipath for example.
But for now rdma+tcp probably is fine, and we can add list later.

gila: The nvme bootspec defines nvme URLs; might want to consult that.

dsharma-dc: Thanks @gila .
We seem to be aligned with the guidance in spec that says uri scheme is to be `nvme+<trtype>`. We just slightly diverge and intend to use `nvmf+<trtype>+<trtype>` (i.e. `nvmf+rdma+tcp` to explicitly indicate support of both protocols) while still remaining compliant to RFC 3986.

libnvme-rs/src/nvme_uri.rs
```diff
@@ -42,12 +42,14 @@ impl Drop for NvmeStringWrapper {
...
 enum NvmeTransportType {
     Tcp,
+    Rdma,
 }
 
 impl NvmeTransportType {
     fn to_str(&self) -> &str {
         match self {
             NvmeTransportType::Tcp => "tcp",
+            NvmeTransportType::Rdma => "rdma",
```
dsharma-dc, line 52: This is only used by tests so assuming `rdma` for connect upon seeing NvmeTransportType::Rdma , instead of fallback behaviour handling.

Abhinandan-Purkait (approved):

tiagolobocastro (approved):

dsharma-dc:
bors try

dsharma-dc:
> ## try
> Build failed:
> 
> * [continuous-integration/jenkins/branch](https://mayastor-ci.mayadata.io/job/Mayastor/job/trying/1329/display/redirect)

Adapt a regex in a test according to new uri scheme.

bors try

dsharma-dc:
bors try

dsharma-dc:
bors merge

PR #1727, opened 2024-08-27
test(wipe): allow wipes on snapshots

tiagolobocastro:
This is actually not the correct behaviour, but previously e2e was dependent on this.
Until the API is modified to handle this properly, revert to the previous behaviour to ensure tests are running.

### DIFF
--- a/io-engine/src/grpc/v1/test.rs
+++ b/io-engine/src/grpc/v1/test.rs
@@ -111,8 +111,10 @@ impl TestRpc for TestService {
                                 ),
                                 None => None,
                             };
+                            let args = FindReplicaArgs::new(&args.uuid);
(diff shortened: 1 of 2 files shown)


r1jt (approved):

dsharma-dc (approved):
ok for interim. We probably need a new resource type `snapshot` in the API

tiagolobocastro:
> ok for interim. We probably need a new resource type `snapshot` in the API

Yes indeed.
bors merge

tiagolobocastro:
"New" failure, assert on nvmf_tcp_req_complete with incorrect state, which doesn't get logged :(
Not related to this PR so moving on:
bors merge

tiagolobocastro:
bors merge

Issue #1739, opened 2024-09-17
Add support for virtblk type block devices as valid disks for pool creation


salilhpe:
**Describe the bug**
I am trying to setup mayastor on a ubuntu VM running in KVM, the disk that I want to add is /dev/vdb, however kubectl get mayastor block-devices <node>  does not list this device. Tough I can create the diskpool with this device.
If I do a call with the --all flag I get:
 /dev/vdb    disk       2TiB      no                       /devices/pci0000:00/0000:00:02.4/0000:05:00.0/virtio3/block/vdb            252     16      <none>      <none>                                <none>  <none>                                    "/dev/disk/by-path/pci-0000:05:00.0", "/dev/disk/by-path/virtio-pci-0000:05:00.0" 

**To Reproduce**
Setup mayastor on KVM VM

**Expected behavior**
block-devices should show the available disks.

**Screenshots**
block devices in  /proc/devices
Block devices:
 7 loop
 8 sd
 9 md
 11 sr
 65 sd
 66 sd
 67 sd
 68 sd
 69 sd
 70 sd
 71 sd
128 sd
129 sd
130 sd
131 sd
132 sd
133 sd
134 sd
135 sd
252 virtblk
253 device-mapper
254 mdp
259 blkext

And vdb is using 252.
$ ls -l /dev/block |grep vdb
lrwxrwxrwx 1 root root 6 Sep 12 10:32 252:16 -> ../vdb

** OS info (please complete the following information):**
 - Distro: Ubuntu
 - Kernel version
 - MayaStor revision or container image :2.5.1

**Additional context**

tiagolobocastro:
This is now available as of 2.7.1

PR #1740, opened 2024-09-17
fix(host-devices): extend allowed major list

tiagolobocastro:
Extends major list with the Local/Experimental 240-254 range and 255 reserved.


tiagolobocastro:
Resolves #1739

dsharma-dc (approved):

Abhinandan-Purkait (approved):

Abhinandan-Purkait:
@tiagolobocastro Are we sure pools can be created on those devices?

tiagolobocastro:
> @tiagolobocastro Are we sure pools can be created on those devices?

As long as it's a block device don't see why not. We're not stopping anyone today as we don't check this when creating the pool itself, it just changes the "available" reported on the plugin.

tiagolobocastro:
bors merge

PR #1746, opened 2024-10-03
feat: add mayastor nvmf rdma capability flag in features

dsharma-dc:


io-engine/src/subsys/nvmf/target.rs
```diff
@@ -104,7 +107,10 @@ impl Target {
             tgt: NonNull::dangling(),
             poll_group_count: 0,
             next_state: TargetState::Init,
-            rdma: MayastorEnvironment::global_or_default().rdma(),
+            // If mayastor args/env request this to be true then it'll
```
dsavitskiy: s/request/requests

dsavitskiy (approved):

io-engine/src/core/env.rs
```diff
@@ -292,11 +293,12 @@ impl MayastorFeatures {
         let lvm = env::var("ENABLE_LVM").as_deref() == Ok("true");
         let snapshot_rebuild =
             env::var("ENABLE_SNAPSHOT_REBUILD").as_deref() == Ok("true");
-
+        let rdma_capable_target = is_mayastor_nvmf_tgt_rdma_capable();
```
tiagolobocastro: hmm, we're conflating things a bit here, features exposes if it's capable, but here we're checking if it's capable and the target is listening?
I think we should separate those two things

dsharma-dc: The definition of "being rdma capable" for a running mayastor instance target here is whether it has rdma transport created and can take rdma connections, else it's a usual tcp mayastor target instance . What can we separate here?

dsharma-dc: or should we simply expose this as being capable in meaning that the target can take `--enable-rdma` flag for rdma support?

dsharma-dc: A nuance here is that `ENABLE_RDMA` env could be set but it doesn't guarantee that the target will eventually be rdma capable, until we can successfully setup the target for rdma in runtime.

tiagolobocastro: I think we can simply expose whether we're rdma capable for now, which would be the "--enable-rdma" feature gate.
If we want to know if we're listening for the transport RDMA then we can add more messages to expose runtime information.
Example, we could expose subsystem listeners.

dsharma-dc: That might make control plane's life difficult as it would see the io-engine is rdma capable upon registration, but rdma configuration could've actually failed.
I don't see any issue in exposing capability based on complete (one-time) runtime info, considering this is a little different due to "enablement" isn't guaranteeing "usability".

tiagolobocastro: Let's image we start binding to more than one network (which I'm sure we will at some point), how would you then expose your rdma feature if you succeeded listening on only 1 of them?
What if we want start listening on networks dynamically?
IMHO, conflating capability with usability is what'd make control plane's life difficult.

dsharma-dc: On the other hand, we'd be complicating API if we want to get real view of capability.

>What if we want start listening on networks dynamically?

What does this mean? I imagine it would certainly come with much bigger challenges compared to exposing capability.

tiagolobocastro: I'm not sure how adding another message containing the listener and the listen transport complicates it?

dsharma-dc: That new message would most likely be fetched with a new gRPC call to io-engine if control plane wants to do something? which basically means the capability info we get upon registration isn't reliable already.

tiagolobocastro: We don't need a new gRPC call, we just need to add a new message with this different kind of information.

dsharma-dc: so that's essentially doing the same thing, just splitting the info into two messages and stashing in same grpc(registration). (¬_¬)

tiagolobocastro: not the same thing exactly, it's ensuring we're not overloading capacity to do something versus any runtime failures in using something.
for example, you would be able to tell that a node supports RDMA and be able to tell RDMA is not enabled. If you mix them together, then you can only tell RDMA is not enabled, you don't know if it's supported or not.

dsharma-dc: let's go ahead with simple env based capability exposure.

tiagolobocastro: But then it won't be sufficient for how you intended to use it?
If you're really adverse to split it into two flags for some reason, then keep it as you had it (reading rdma listener) but please rename the proto to something that makes it very clear what's it doing so we don't make mistake of using it incorrectly in the future.
Ex: nvmfRdmaActive

dsharma-dc: > But then it won't be sufficient for how you intended to use it?

I don't see any immediate use-case in control-plane where we might have to do something based on concrete evidence of active rdma in io-engine. Control-plane logic as it stands today is driven mainly by uri scheme which inherently indicates active rdma. So I can rename it to `rdmaCapableIoEngine` or let it be `rdmaCapableTarget`

tiagolobocastro: If there's no need for active rdma port check, then rdmaCapableIoEngine or even just rdma sounds good.

dsharma-dc: changed to rdmaCapableIoEngine.

tiagolobocastro (approved):

dsharma-dc:
bors try

dsharma-dc:
bors merge

PR #1755, opened 2024-10-11
Reuse Rebuild IO handles

tiagolobocastro:
fix(rebuild): reuse rebuild IO handles
    
    Reuses the rebuild IO handles, rather than attempting to allocate
    them per rebuild task.
    The main issue with handle allocation on the fly is that the target
    may have not cleaned up a previous IO qpair connection, and so the
    connect may fail. We started seeing this more on CI because we forgot
    to cherry-pick a commit increasing the retry delay.
    However, after inspecting a bunch of user support bundles I see that
    we still have occasional connect errors. Rather than increasing the
    timeout, we attempt here to reuse the handles, thus avoid the
    problem almost entirely.
    
---

    refactor(rebuild): rebuild completion is not an error
    
    When the rebuild has been complete, if we wait for it this fails because
    the channels are not longer available.
    Instead, simply return the rebuild state, since this is what we want anyway.
    
### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "spdk-rs"]
 	path = spdk-rs
-	url = https://github.com/openebs/spdk-rs
+	url = ../spdk-rs.git
 	branch = develop
 [submodule "utils/dependencies"]
 	path = utils/dependencies
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1665,6 +1665,7 @@ dependencies = [
  "crossbeam",
  "derive_builder",
  "devinfo",
+ "either",
  "env_logger",
  "etcd-client",
  "event-publisher",
--- a/io-engine/Cargo.toml
+++ b/io-engine/Cargo.toml
@@ -74,7 +74,7 @@ libc = "0.2.149"
 log = "0.4.20"
 md5 = "0.7.0"
 merge = "0.1.0"
-nix = { version = "0.27.1", default-features = false, features = [ "hostname", "net", "socket", "ioctl" ] }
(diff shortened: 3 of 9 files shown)


dsharma-dc (approved):
lgtm.
As I understand we'll now make all 16 rebuild tasks use the same connection for reading/writing the rebuild segments. Is the rebuild descriptor destroyed today as soon as rebuild finishes and the nexus channels are reconfigured to add rebuilt bdev's handle? We wouldn't want a lingering handle in rebuild descriptor while the rebuilt child bdev handle has been added to main nexus channels too.

tiagolobocastro:
> lgtm. As I understand we'll now make all 16 rebuild tasks use the same connection for reading/writing the rebuild segments. 

Yes, each task has the immutable `RebuildDescriptor` and reuse the handles from there.

> Is the rebuild descriptor destroyed today as soon as rebuild finishes and the nexus channels are reconfigured to add rebuilt bdev's handle? We wouldn't want a lingering handle in rebuild descriptor while the rebuilt child bdev handle has been added to main nexus channels too.

Yes that's right, the rebuild backend "runs away" when the rebuild is started. When the rebuild completes the backend terminates, and on the `drop`, the handles are also dropped.
Eventually we can also improve things here by moving the handle to the tasks themselves, allowing us to drop the handles if the rebuild is paused for example. Also that could be the first step to allowing the rebuild to run on multi-cores too.

For now simply moving the handle to descriptor is good enough for solving the connection issues.

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #1759, opened 2024-10-23
test: add a test to enable and test rdma connect

dsharma-dc:
- Added a simple test to create and publish a nexus. Connect it over rdma.
- The connect API is modified to take transport parameter.
- The test automatically detects a suitable interface to create a soft-RDMA device, creates it and deletes it at the end of the test.

### DIFF
--- a/io-engine-tests/src/lib.rs
+++ b/io-engine-tests/src/lib.rs
@@ -8,6 +8,7 @@ use std::{io, io::Write, process::Command, time::Duration};
 
 use crossbeam::channel::{after, select, unbounded};
 use once_cell::sync::OnceCell;
+use regex::Regex;
 use run_script::{self, ScriptOptions};
 use url::{ParseError, Url};
 
@@ -35,6 +36,8 @@ pub mod replica;
 pub mod snapshot;
 pub mod test;
 pub mod test_task;
+
+use cli_tools::run_command_args;
 
 pub use compose::MayastorTest;
 
@@ -548,6 +551,86 @@ pub fn reactor_run_millis(milliseconds: u64) {
         s.send(())
     });
     reactor_poll!(r);
+}
+
+// We try to figure out the net interface that is being used to connect
+// to internet(public IP). Then use that interface name to setup rdma rxe
+// device.
+//
+// # host we want to "reach"
+// host=google.com
+//
+// # get the ip of that host (works with dns and /etc/hosts. In case we get
+// # multiple IP addresses, we just want one of them
+// host_ip=$(getent ahosts "$host" | awk '{print $1; exit}')
+//
+// # only list the interface used to reach a specific host/IP. We only want the
+// part # between dev and the remainder (use grep for that)
+// ip route get "$host_ip" | grep -Po '(?<=(dev ))(\S+)'
+pub fn setup_rdma_rxe_device() -> String {
+    let test_host = "google.com";
+    let ns_entries = run_command_args(
+        "getent",
+        vec!["ahosts", test_host],
+        Some("get ip of test host"),
+    )
+    .unwrap();
+
+    let test_host_ip = ns_entries
+        .1
+        .first()
+        .unwrap()
+        .to_string_lossy()
+        .split_whitespace()
+        .next()
+        .unwrap()
+        .to_owned();
+
+    let ent = &run_command_args(
+        "ip",
+        vec!["route", "get", test_host_ip.as_str()],
+        Some("get routing entry"),
+    )
+    .unwrap()
+    .1[0];
+
+    // match/grep the interface name
+    let pattern = r"dev (\S+)";
+    let re = Regex::new(pattern).unwrap();
+
+    let iface = re
+        .captures(ent.to_str().unwrap())
+        .unwrap()
+        .get(1)
+        .expect("interface not found")
+        .as_str();
+    println!("Using interface {} to create rdma rxe device", iface);
+
+    // now create the software rdma device.
+    let _ = run_command_args(
+        "rdma",
+        vec![
+            "link",
+            "add",
+            "io-engine-rxe0",
+            "type",
+            "rxe",
+            "netdev",
+            iface,
+        ],
+        Some("Create rxe device"),
+    );
+
+    iface.to_string()
+}
+
+pub fn delete_rdma_rxe_device() {
+    let _ = run_command_args(
+        "rdma",
+        vec!["link", "delete", "io-engine-rxe0"],
+        Some("Delete rxe device"),
+    )
+    .expect("rxe device delete");
 }
 
 pub fn composer_init() {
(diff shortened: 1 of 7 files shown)


dsharma-dc:
The changes for composer dependencies are here: https://github.com/openebs/mayastor-dependencies/pull/96

io-engine-tests/src/lib.rs
```diff
@@ -550,6 +553,84 @@ pub fn reactor_run_millis(milliseconds: u64) {
...
+            .get(1)
+            .expect("interface not found")
+            .as_str();
+        println!("Using interface {} to create rdma rxe device", iface);
+    
+        // now try to install rdma_rxe module and create rxe device.
+        // todo: More dynamic checks for module may be needed depending on platform.
+        // Assume linux-modules-extra-$(uname -r) is installed.
+        let _ = run_command_args(
+            "modprobe",
```
tiagolobocastro: I'd rather not do this and have this be a pre-requisites, just like nvme-*.

io-engine-tests/src/lib.rs
```diff
@@ -550,6 +553,84 @@ pub fn reactor_run_millis(milliseconds: u64) {
...
+//
+// # get the ip of that host (works with dns and /etc/hosts. In case we get
+// # multiple IP addresses, we just want one of them
+// host_ip=$(getent ahosts "$host" | awk '{print $1; exit}')
+//
+// # only list the interface used to reach a specific host/IP. We only want the part
+// # between dev and the remainder (use grep for that)
+// ip route get "$host_ip" | grep -Po '(?<=(dev ))(\S+)'
+pub fn setup_rdma_rxe_device() -> String {
+        let test_host = "google.com";
```
tiagolobocastro: formatting seems off

io-engine/tests/nvmf.rs
```diff
@@ -195,3 +249,84 @@ async fn nvmf_set_target_interface() {
...
+
+    let url =  url::Url::parse(device_uri.as_str()).unwrap();
+    assert!(url.scheme() == "nvmf+rdma+tcp");
+
+    let host = url.host_str().unwrap();
+    let nqn = format!("{NVME_NQN_PREFIX}:{}", nexus_name());
+    let conn_status = nvme_connect(host, &nqn, "rdma", true);
+    assert!(conn_status.success());
+
+    tokio::time::sleep(Duration::from_secs(2)).await;
```
tiagolobocastro: what's the sleep for here? Are there issues if we disconnect straight away?

io-engine-tests/src/lib.rs
```diff
@@ -550,6 +553,84 @@ pub fn reactor_run_millis(milliseconds: u64) {
...
+        let _ = run_command_args(
+            "rdma",
+            vec!["link", "add", "rxe0", "type", "rxe", "netdev", iface],
+            Some("Create rxe device"),
+        );
+    
+        iface.to_string()
+    }
+    
+    pub fn delete_rdma_rxe_device() {
```
tiagolobocastro: also should update cleanup script to clean these up in case of a crash

io-engine-tests/src/lib.rs
dsharma-dc: yeah I think this can be removed from here, I've added installation in bash script for cargo test anyway.

io-engine-tests/src/lib.rs
dsharma-dc: > formatting seems off

fixed

io-engine/tests/nvmf.rs
dsharma-dc: > what's the sleep for here? Are there issues if we disconnect straight away?

no issues, works ok. It was added while testing manually, removed now.

io-engine-tests/src/lib.rs
dsharma-dc: > also should update cleanup script to clean these up in case of a crash

deleting from cleanup script too now.

dsharma-dc: not uninstalling the module though.

Abhinandan-Purkait (approved):

scripts/cargo-test.sh
```diff
@@ -19,6 +19,12 @@ trap cleanup_handler INT QUIT TERM HUP EXIT
 export PATH=$PATH:${HOME}/.cargo/bin
 set -euxo pipefail
 
+# setup extra modules as we need rdma-rxe for rdma tests
+sudo apt -y install rdmacm-utils
```
tiagolobocastro: hmm you can't assume everyone and ci is using ubuntu.
I think we should leave this out, should be part of pre-requisites, just like hugepages and nvme-tcp

scripts/clean-cargo-tests.sh
```diff
@@ -21,6 +21,9 @@ done
 # Delete the directory too
 nix-sudo rmdir --ignore-fail-on-non-empty "/tmp/io-engine-tests" 2>/dev/null
 
+# If there was a soft rdma device created and left undeleted by nvmf rdma test,
+# delete that now. Not removing rdma-rxe kernel module.
+sudo rdma link delete rxe0 2>/dev/null
```
tiagolobocastro: Don't use sudo directly.
Btw maybe we can use name io-engine-rxe0? (or can we not dictate the name?)

```suggestion
nix-sudo rdma link delete rxe0 2>/dev/null
```

dsharma-dc: name is arbitrary, we can use `io-engine-rxe0`

scripts/cargo-test.sh
dsharma-dc: removed installation of package and added warning if kernel modules aren't installed.

scripts/clean-cargo-tests.sh
dsharma-dc: renamed to `io-engine-rxe0`

dsharma-dc:
bors try

tiagolobocastro (approved):

dsharma-dc:
bors try

dsharma-dc:
CI machines have been setup with required modules now.

bors try

dsharma-dc:
bors try

pchandra19:
bors try

pchandra19:
bors try

dsharma-dc:
I tried running the test manually and I notice the issue is that rxe rdma device is not getting correct IP address of the chosen interface assigned. This is the issue that we see when overlay network is present. As a result rdma listen is failing. I'll mark the test as `#ignore` for now and merge with that. 

IP assigned to rdma device on eth0 (compare only last byte of dot decimal notation)
`io-engine-rxe0	1	1	0000:0000:0000:0000:0000:ffff:87b5:b781	111.222.333.129 	v2	eth0`

Actual IP of eth0 from ifconfig

```
eth0: flags=xxxx<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 111.222.333.161  netmask 255.255.255.255  broadcast 0.0.0.0
```

dsharma-dc:
bors try

dsharma-dc:
bors merge

PR #1771, opened 2024-11-13
Introduce new CSAL ftl bdev

MaisenbacherD:
For this PR to be merged the PR https://github.com/openebs/spdk-rs/pull/67
for spdk-rs needs to be resolved and the submodule reference must be
updated accordingly. :)

The SPDK ftl bdev allows to create a layered device with a (fast) cache
device for buffering writes that get eventually flushed out sequentially
to a base device. SDPK ftl is also known as the Cloud Storage
Acceleration Layer (CSAL).

This kind of device potentially enables the use of emerging storage
interfaces like Zoned Namespace (ZNS) or Flexible Data Placement (FDP)
capable NVMe devices. Up to this point, those NVMe command sets are not
yet supported in SPDK upstream. However, the acceleration aspect of a
fast cache device already adds value.
With future support of new devices for SDPK ftl, Mayastor would already
be capable of utilizing those features simply by upgrading SDPK.

For now, the `ftl_mount_fs` test cases are hidden behind the `ignore`
attribute because PCIe devices are required for this test until
SPDK v24.09 is picked up, which introduces variable sector size emulation
for ftl devices.
To run these tests use the following:
```
RUST_LOG=TRACE cargo test -- --test-threads 1 --test ftl_mount_fs --nocapture --ignored
```

This patch introduces a new ftl device uri scheme:
```
ftl:///<ftl-device-name>?bbdev=<dev-uri-nested-encoding>&cbdev=<dev-uri-nested-encoding>
```

`<dev-uri-nested-encoding>` can be any already valid device uri where '?'
are replaced with '!' and '&' are replaced with '|'.
With SPDK v24.05 only PCIe devices will work where the cache device is
formatted to 4KiB+64B LBA format and the base device to 4KiB LBA format.

From SPDK v24.09 on, any device with a block size of 4KiB will work for
both cache and base device.

CSAL paper reference:
https://dl.acm.org/doi/10.1145/3627703.3629566

### DIFF
--- a/doc/test.md
+++ b/doc/test.md
@@ -29,8 +29,12 @@ Mayastor's unit tests, integration tests, and documentation tests via the conven
 Mayastor uses [spdk][spdk] which is quite senistive to threading. This means tests need to run one at a time:
 
 ```bash
-cargo test -- --test-threads 1
+cd io-engine
+RUST_LOG=TRACE cargo test -- --test-threads 1 --nocapture
 ```
+
+## Testing your own SPDK version
+To test your custom SPDK version please refere to the [spdk-rs documentation](https://github.com/openebs/spdk-rs/blob/develop/README.md#custom-spdk)
 
 ## Running the end-to-end test suite
 
@@ -54,6 +58,29 @@ Then, to run the tests:
 ```bash
 ./node_modules/mocha/bin/mocha test_csi.js
 ```
+
+## Using PCIe NVMe devices in cargo tests while developing
+
+When developing new features, testing those with real PCIe devices in the process might come in handy.
+In order to do so, the PCIe device first needs to be bound to the vfio driver:
+
+```bash
+sudo PCI_ALLOWED="<PCI-ADDRESS>" ./spdk-rs/spdk/scripts/setup.sh
+```
+
+The bdev name in the cargo test case can then follow the PCIe URI pattern:
+
+```rust
+static BDEVNAME1: &str = "pcie:///<PCI-ADDRESS>";
+```
+
+After testing the device may be rebound to the NVMe driver:
+
+```bash
+sudo PCI_ALLOWED="<PCI-ADDRESS>" ./spdk-rs/spdk/scripts/setup.sh reset
+```
+
+Please do not submit pull requests with active cargo test cases that require PCIe devices to be present.
 
 [spdk]: https://spdk.io/
 [doc-run]: ./run.md
--- a/io-engine/src/bdev/ftl.rs
+++ b/io-engine/src/bdev/ftl.rs
@@ -1,0 +1,325 @@
+//! The Ftl device is a representation of an SPDK (CSAL) ftl bdev, which allows
+//! to create a layered device with a persistent (fast) cache device for
+//! buffering writes that get eventually flushed out sequentially to a base
+//! device.
+//!
+//! # Uri
+//! ftl:///$ftl_device_name?bbdev=$bbdev_uri_percent_encoded&
+//! cbdev=$cbdev_uri_percent_encoded
+//!
+//! The bbdev_uri and cbdev_uri need to use percent encoding on '?' (= '%3F')
+//! and '&' (= '%26') segment dividers. This is needed so we can successfully
+//! parse the ftl uri. Optionally, the sub-uris may also be fully
+//! percent-encoded.
+//!
+//! # Parameters
+//! name: A name for the ftl device, example: "ftl-1".
+//! alias: Ftl device URI which can be used to open the bdev.
+//! uuid: A UUID that can be set to reference the resulting SPDK bdev.
+//! bbdev_uri: Nested device uri for the ftl base bdev, which is (partially)
+//!            percent encoded. This device currently requires to be a
+//!            non-volatile memory device which is has an LBA format of 4KiB
+//!            and is at least 20GB large.
+//! cbdev_uri: Nested device uri for the ftl base bdev, which is (partially)
+//!            percent encoded. This device currently requires to be a
+//!            non-volatile memory device which is has an LBA format of 4KiB
+//!            data size, 64B metadata size and is at least 20GB large.
+//!            From SPDK v24.09 on there is no requirement for the metadata
+//!            portion as Variable Sector Size (VSS) emulation is enabled.
+//!
+//! # Examples
+//! ftl:///ftl-1?bbdev=pcie:///0000:01:00.0&cbdev=pcie:///0000:02:00.0
+//!
+//! ftl:///ftl-1?bbdev=pcie%3A%2F%2F%2F0000%3A01%3A00.0&cbdev=pcie%3A%2F%2F%
+//! 2F0000%3A02%3A00.0
+//!
+//! ftl:///ftl-1?bbdev=pcie:///0000:01:00.0%
+//! 3Finvalidoption=test&cbdev=pcie:///0000:02:00.0
+//!
+//! ftl:///ftl-2?bbdev=aio:///tmp/basedev.img%3Fblk_size=4096&cbdev=aio:///tmp/
+//! cachedev.img%3Fblk_size=4096
+use std::{
+    collections::HashMap,
+    convert::TryFrom,
+    fmt::{Debug, Formatter},
+};
+
+use core::ffi::c_void;
+
+use async_trait::async_trait;
+use futures::channel::oneshot;
+use log::info;
+use nix::errno::Errno;
+use percent_encoding::percent_decode_str;
+use snafu::{OptionExt, ResultExt};
+use std::mem;
+use url::Url;
+
+use spdk_rs::{
+    ffihelper::errno_result_from_i32,
+    libspdk::{
+        bdev_ftl_create_bdev,
+        bdev_ftl_delete_bdev,
+        ftl_bdev_info,
+        spdk_ftl_conf,
+        spdk_ftl_get_default_conf,
+        spdk_ftl_mode,
+    },
+    UntypedBdev,
+};
+
+use crate::{
+    bdev::{dev::reject_unknown_parameters, util::uri, CreateDestroy, GetName},
+    bdev_api::{self, bdev_create, bdev_destroy, BdevError},
+    core::VerboseError,
+    ffihelper::{cb_arg, done_errno_cb, ErrnoResult},
+};
+
+/// An ftl bdev specified via URI.
+pub struct Ftl {
+    /// The name of the ftl-bdev we created.
+    name: String,
+    /// Alias which can be used to open the bdev.
+    alias: String,
+    /// Uuid of the spdk bdev.
+    uuid: Option<uuid::Uuid>,
+    /// Ftl's base bdev URI.
+    bbdev_uri: String,
+    /// Ftl's cache bdev URI.
+    cbdev_uri: String,
+}
+
+impl Debug for Ftl {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "Ftl '{}' (bbdev_uri {} cbdev_uri {})",
+            self.name, self.bbdev_uri, self.cbdev_uri
+        )
+    }
+}
+
+impl TryFrom<&Url> for Ftl {
+    type Error = BdevError;
+
+    fn try_from(uri: &Url) -> Result<Self, Self::Error> {
+        let segments = uri::segments(uri);
+        if segments.is_empty() {
+            return Err(BdevError::InvalidUri {
+                uri: uri.to_string(),
+                message: "empty path".to_string(),
+            });
+        }
+
+        let mut parameters: HashMap<String, String> =
+            uri.query_pairs().into_owned().collect();
+
+        let uuid = uri::uuid(parameters.remove("uuid")).context(
+            bdev_api::UuidParamParseFailed {
+                uri: uri.to_string(),
+            },
+        )?;
+
+        let encoded_bbdev_uri =
+            parameters.remove("bbdev").context(bdev_api::InvalidUri {
+                uri: uri.to_string(),
+                message: String::from("No bbdev parameter found"),
+            })?;
+
+        let bbdev_uri = percent_decode_str(&encoded_bbdev_uri)
+            .decode_utf8()
+            .map_err(|e| BdevError::InvalidUri {
+                uri: uri.to_string(),
+                message: format!(
+                    "Could not percent decode bbdev_uri sub-uri - {}",
+                    e
+                ),
+            })?
+            .to_string();
+
+        let encoded_cbdev_uri =
+            parameters.remove("cbdev").context(bdev_api::InvalidUri {
+                uri: uri.to_string(),
+                message: String::from("No cbdev parameter found"),
+            })?;
+
+        let cbdev_uri = percent_decode_str(&encoded_cbdev_uri)
+            .decode_utf8()
+            .map_err(|e| BdevError::InvalidUri {
+                uri: uri.to_string(),
+                message: format!(
+                    "Could not percent decode cbdev_uri sub-uri - {}",
+                    e
+                ),
+            })?
+            .to_string();
+
+        // Device parameter checking for bbdev and cbdev are done in
+        // bdev_ftl_create_bdev
+        reject_unknown_parameters(uri, parameters)?;
+
+        Ok(Self {
+            name: uri.path()[1 ..].into(),
+            alias: uri.to_string(),
+            uuid,
+            bbdev_uri,
+            cbdev_uri,
+        })
+    }
+}
+
+impl GetName for Ftl {
+    fn get_name(&self) -> String {
+        self.name.clone()
+    }
+}
+
+pub extern "C" fn ftl_bdev_init_fn_cb(
+    _ptr: *const ftl_bdev_info,
+    sender_ptr: *mut c_void,
+    errno: i32,
+) {
+    info!("{:?}: ftl_bdev_init_fn_cb", errno);
+    let sender = unsafe {
+        Box::from_raw(sender_ptr as *mut oneshot::Sender<ErrnoResult<()>>)
+    };
+    sender
+        .send(errno_result_from_i32((), errno))
+        .expect("done callback receiver side disappeared");
+}
+
+#[async_trait(?Send)]
+impl CreateDestroy for Ftl {
+    type Error = BdevError;
+
+    async fn create(&self) -> Result<String, Self::Error> {
+        if UntypedBdev::lookup_by_name(&self.name).is_some() {
+            return Err(BdevError::BdevExists {
+                name: self.name.clone(),
+            });
+        }
+
+        let ftl_dev_name = CString::new(self.name.clone()).unwrap();
+
+        debug!(
+            "{:?}: Creating ftl-bdev '{:?}' with bbdev {:?} and cbdev {:?}.",
+            self, ftl_dev_name, self.bbdev_uri, self.cbdev_uri
+        );
+
+        use std::ffi::CString;
+
+        let base_dev_name = bdev_create(&self.bbdev_uri).await?;
+        let cache_dev_name = match bdev_create(&self.cbdev_uri).await {
+            Ok(dev_name) => dev_name,
+            Err(err) => {
+                if let Err(e) = bdev_destroy(&self.bbdev_uri).await {
+                    error!("{:?} bbdev cleanup error: {}", self, e);
+                }
+                return Err(err);
+            }
+        };
+
+        let (s, r) = oneshot::channel::<ErrnoResult<()>>();
+
+        let spdk_ftl_conf_size = mem::size_of::<spdk_ftl_conf>() as u64;
+        let mut ftl_conf = spdk_ftl_conf {
+            ..unsafe { mem::zeroed() }
+        };
+        unsafe { spdk_ftl_get_default_conf(&mut ftl_conf, spdk_ftl_conf_size) };
+        ftl_conf.name = ftl_dev_name.as_ptr() as *mut i8;
+        ftl_conf.base_bdev = base_dev_name.as_ptr() as *mut i8;
+        ftl_conf.cache_bdev = cache_dev_name.as_ptr() as *mut i8;
+        ftl_conf.fast_shutdown = true;
+        ftl_conf.verbose_mode = true;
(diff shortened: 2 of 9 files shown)


MaisenbacherD:
Discussion question: The `spdk_ftl_conf` allows us to set configurations like `core_mask` and `l2p_dram_limit`. Are those configurations worth exposing through the ftl uri, or should we go with the default configuration?

doc/test.md
```diff
@@ -55,6 +59,29 @@ Then, to run the tests:
...
+static BDEVNAME1: &str = "pcie:///<PCI-ADDRESS>";
+```
+
+After testing the device may be rebound to the NVMe driver:
+
+```bash
+sudo PCI_ALLOWED="<PCI-ADDRESS>" ./spdk-rs/spdk/scripts/setup.sh reset
+```
+
+Please do not submit pull requests with active cargo test cases that require PCIe devices to be present.
```
tiagolobocastro, line 83: Maybe in the future we can have a feature flag for PCIe tests which can be run on a dedicated test system :)
Btw, can we add virtual PCIe devices with qemu? That could be a way of running those tests without having spare pcie nvme devices

io-engine/src/bdev/ftl.rs
```diff
@@ -0,0 +1,258 @@
...
+use async_trait::async_trait;
+use futures::channel::oneshot;
+use log::info;
+use nix::errno::Errno;
+use snafu::{OptionExt, ResultExt};
+use std::mem;
+use url::Url;
+
+use spdk_rs::ffihelper::errno_result_from_i32;
+use spdk_rs::{
```
tiagolobocastro: did you run cargo fmt?

io-engine/src/bdev/ftl.rs
```diff
@@ -0,0 +1,258 @@
...
+};
+
+use crate::{
+    bdev::{dev::reject_unknown_parameters, util::uri, CreateDestroy, GetName},
+    bdev_api::{self, bdev_create, bdev_destroy, BdevError},
+    core::VerboseError,
+    ffihelper::{cb_arg, done_errno_cb, ErrnoResult},
+};
+
+pub struct Ftl {
```
tiagolobocastro, line 79: Add simple doc-comment explaining what `Ftl` is?

io-engine/src/bdev/ftl.rs
```diff
@@ -0,0 +1,258 @@
...
+    /// the name of the ftl-bdev we created
```
tiagolobocastro: Try to start with a `C`apital and end with a `.` - makes the auto-generated rust docs look better IIRC

```suggestion
    /// The name of the ftl-bdev we created.
```

io-engine/src/bdev/ftl.rs
```diff
@@ -0,0 +1,258 @@
...
+    name: String,
+    /// alias which can be used to open the bdev
+    alias: String,
+    /// uuid of the spdk bdev
+    uuid: Option<uuid::Uuid>,
+    // ftl's base bdev
+    bbdev_uri: String,
+    // ftl's cache bdev (must support VSS DIX mode)
```
tiagolobocastro: What is VSS ?
DIX is the data integrity stuff? So the cache bdev has to support blocksize + 8B, for example?
Any other requirements for this cache device? Does it need to be non-volatile so ex it can't be a ram device?

io-engine/src/bdev/ftl.rs
```diff
@@ -0,0 +1,258 @@
+//!
+//! The ftl bdev creates a CSAL-FTL disk.
```
tiagolobocastro: Can you document how the uri looks like? for example like for https://github.com/openebs/mayastor/blob/6374141862abf68b848e54a5dea2072d6c064819/io-engine/src/bdev/nx.rs#L1

io-engine/src/bdev/ftl.rs
```diff
@@ -0,0 +1,258 @@
...
+
+        let cbdev_uri = parameters
+            .remove("cbdev")
+            .context(bdev_api::InvalidUri {
+                uri: uri.to_string(),
+                message: String::from("No cbdev parameter found"),
+            })?
+            .to_string()
+            .replace("!", "?")
+            .replace("|", "&");
```
tiagolobocastro: This is because we can't have those in the "sub-uris" ? Would be good to document that at the top of the file as well.
The nx and lvs probably have the same issue, maybe I never tried using query parameters there
Can around this as well by using percent encoding?

io-engine/src/bdev/ftl.rs
```diff
@@ -0,0 +1,258 @@
...
+        debug!(
+            "{:?}: Creating ftl-bdev '{:?}' with bbdev {:?} and cbdev {:?}.",
+            self,
+            CString::new(self.name.clone()).unwrap(),
+            self.bbdev_uri,
+            self.cbdev_uri
+        );
+
+        use std::ffi::CString;
+        extern crate libc;
```
tiagolobocastro: Is this needed?

io-engine/src/bdev/ftl.rs
```diff
@@ -0,0 +1,258 @@
...
+
+        let base_dev_name = bdev_create(&self.bbdev_uri).await?;
+        let cache_dev_name = bdev_create(&self.cbdev_uri).await?;
+
+        let l_name_ptr =
+            CString::new(self.name.clone()).unwrap().into_raw() as *mut i8;
```
tiagolobocastro: This is leaking memory?
Does it have to be mut or can we use a reference? example:
```suggestion
        let l_name =
            CString::new(self.name.clone()).unwrap();
```
```
 ftl_conf.name = l_name.as_ptr();
```

io-engine/src/bdev/ftl.rs
```diff
@@ -0,0 +1,258 @@
...
+            ..unsafe { mem::zeroed() }
+        };
+        unsafe { spdk_ftl_get_default_conf(&mut ftl_conf, spdk_ftl_conf_size) };
+        ftl_conf.name = l_name_ptr;
+        ftl_conf.base_bdev = l_bname_ptr;
+        ftl_conf.cache_bdev = l_cname_ptr;
+        ftl_conf.fast_shutdown = true;
+        ftl_conf.verbose_mode = true;
+        // SPDK_FTL_MODE_CREATE = (1 << 0)
+        ftl_conf.mode = 1 << 0;
```
tiagolobocastro: We don't have access to `SPDK_FTL_MODE_CREATE` ?

io-engine/src/bdev/ftl.rs
```diff
@@ -0,0 +1,258 @@
...
+            return Ok(self.get_name());
+        }
+
+        Err(BdevError::BdevNotFound {
+            name: self.get_name(),
+        })
+    }
+
+    async fn destroy(self: Box<Self>) -> Result<(), Self::Error> {
+        debug!("{:?}: deleting", self);
```
tiagolobocastro, line 288: nit:
```suggestion
        debug!("{self:?}: deleting");
```

io-engine/src/bdev/ftl.rs
```diff
@@ -0,0 +1,258 @@
...
+
+        let mut result =
+            if let Some(mut bdev) = UntypedBdev::lookup_by_name(&self.name) {
```
tiagolobocastro: nit: you can transmute Option to Result or use let else to reduce indentation on the happy path

io-engine/src/bdev/ftl.rs
```diff
@@ -0,0 +1,258 @@
...
+
+                r.await
+                    .context(bdev_api::BdevCommandCanceled {
+                        name: self.name.clone(),
+                    })?
+                    .context(bdev_api::DestroyBdevFailed { name: self.name })
+            } else {
+                Err(BdevError::BdevNotFound { name: self.name })
+            };
+        if let Err(e) = bdev_destroy(&self.bbdev_uri).await {
```
tiagolobocastro: Shouldn't we reverse the order here?
Destroy the backing bdevs first, and then the self bdev?

io-engine/src/core/env.rs
```diff
@@ -1088,6 +1125,11 @@ impl MayastorEnvironment {
         // ensure we are within the context of a spdk thread from here
         Mthread::primary().set_current();
 
+        // To enable SPDK tracing set self.num_entries (eg. to 32768).
+        if self.num_entries > 0 {
+            self.init_spdk_tracing();
```
tiagolobocastro, line 1135: Pardon my ignorance, this num_entries existing already?
What is this tracing actually doing here?

io-engine/tests/ftl_mount_fs.rs
```diff
@@ -0,0 +1,141 @@
+//TODO: with spdk v24.09 ftl devices support variable sector size emulation. At that point we can
+//enable testing with aio devices, which do not require a metadata section in the LBA format.
+//For now we are hiding this test case behind the `ignore` attribute.
```
tiagolobocastro:
```suggestion
//! TODO: with spdk v24.09 ftl devices support variable sector size emulation. At that point we can
//! enable testing with aio devices, which do not require a metadata section in the LBA format.
//! For now we are hiding this test case behind the `ignore` attribute.
```

tiagolobocastro: Could you please point me to the commit which enables this emulation, just for my curiosity? :)

doc/test.md
MaisenbacherD: Do you mean a rust feature flag, such that PCIe support can be selected at compile time?
It makes sense to me! I briefly thought about adding a feature flag because of the failing test case but went with the `ignore` attribute for simplicity.  
I can rework this to be a rust feature flag within this PR, however, if I attempt this (https://doc.rust-lang.org/cargo/reference/features.html) I get the following error:
```
error: failed to parse manifest at `/home/dennis/src/mayastor/Cargo.toml`

Caused by:
  this virtual manifest specifies a `features` section, which is not allowed
```
Do you know what needs to be done here?

A QEMU NVMe emulated device via PCIe would work for testing. One just needs to make sure that the cache device is configured with `ms=8` https://qemu-project.gitlab.io/qemu/system/devices/nvme.html#metadata.

io-engine/src/bdev/ftl.rs
MaisenbacherD: Yes I did, but it did not merge the spdk_rs prefixes for some reason 🤔
I fixed it manually.

io-engine/src/bdev/ftl.rs
MaisenbacherD: Done :)
Also improved the docs comment on top of the file.

io-engine/src/bdev/ftl.rs
MaisenbacherD: I improved the docs comment on top of the file.

VSS = Variable Sector Size
DIX = Data Integrity Extension (aka Protection Information (PI))

The cache bdev needs to have an LBA format of 4KiB data size and 64B metadata size.
It also needs to be non-volatile memory if you don't want to lose data :)

io-engine/src/bdev/ftl.rs
MaisenbacherD: Yes, it is because we can't have the `sub-uris`. I have to try the percent encoding. Lets leave this conversation as unresolved for now :)

io-engine/src/bdev/ftl.rs
MaisenbacherD: Removed. This was a relict during development.

io-engine/src/bdev/ftl.rs
MaisenbacherD: Good catch, yes it is leaking memory! I reworked this section.

io-engine/src/bdev/ftl.rs
MaisenbacherD: Thanks for addressing this!
I tried importing `SPDK_FTL_MODE_CREATE` via `enum spdk_ftl_mode` which is located in `spdk/include/spdk/ftl.h` alongside `struct spdk_ftl_conf` which I am also using here.

However, when doing

```
use spdk_rs::{
    ffihelper::errno_result_from_i32,
    libspdk::{
        bdev_ftl_create_bdev, bdev_ftl_delete_bdev, ftl_bdev_info,
        spdk_ftl_conf, spdk_ftl_get_default_conf, spdk_ftl_mode,
    },
    UntypedBdev,
};
```

I am getting the following error:

```
error[E0432]: unresolved import `spdk_rs::libspdk::spdk_ftl_mode`
  --> io-engine/src/bdev/ftl.rs:44:51
   |
44 |         spdk_ftl_conf, spdk_ftl_get_default_conf, spdk_ftl_mode,
   |                                                   ^^^^^^^^^^^^^
   |                                                   |
   |                                                   no `spdk_ftl_mode` in `libspdk`
   |                                                   help: a similar name exists in the module: `spdk_ftl_conf`
```

What am I doing wrong here? I can't figure out what to change in spdk-rs' build.rs to include this enum.

io-engine/src/bdev/ftl.rs
MaisenbacherD: Can't do that:
```
error[E0382]: borrow of moved value: `self`
   --> io-engine/src/bdev/ftl.rs:244:17
    |
243 |     async fn destroy(self: Box<Self>) -> Result<(), Self::Error> {
    |                      ---- value moved here
244 |         debug!("{self:?}: deleting");
    |                 ^^^^^^^^ value borrowed here after move

```

io-engine/src/bdev/ftl.rs
MaisenbacherD: `bdev_ftl_delete_bdev` requires the backing bdev's to be present.

io-engine/src/core/env.rs
MaisenbacherD: `num_entries` already exists in `MayastorEnvironment` which loosely resembles `spdk_app_opts` and indicates the number of trace entries per core.
If this `num_entries > 0` we initialize the SPDK tracing framework.

io-engine/tests/ftl_mount_fs.rs
MaisenbacherD: The commit which enables this:
https://review.spdk.io/gerrit/c/spdk/spdk/+/19628/82

You can also look at the `Submitted together` commits that did the actual work :)

io-engine/src/bdev/ftl.rs
MaisenbacherD: The percent-encoding should be possible. However, the user specified URIs will get very convoluted:
See the very simple example (without additional options in the nested URIs) with my proposed encoding vs percent-encoding:
```
ftl:///ftl-1?bbdev=pcie:///0000:01:00.0&cbdev=pcie:///0000:02:00.0
```
vs.
```
ftl:///ftl-1?bbdev=pcie%3A%2F%2F%2F0000%3A01%3A00.0&cbdev=pcie%3A%2F%2F%2F0000%3A02%3A00.0
```

What do you think how we should proceed?

doc/test.md
tiagolobocastro: > Do you know what needs to be done here?

The feature flag would need to be added to the io-engine crate, see some examples: https://github.com/openebs/mayastor/blob/2825e5eb8042999718c6017229385b731f9b9b5a/io-engine/Cargo.toml#L12

I'll leave it up to you, fine to keep ignore for now as well.
<br>

> A QEMU NVMe emulated device via PCIe would work for testing. One just needs to make sure that the cache device is configured with ms=8 https://qemu-project.gitlab.io/qemu/system/devices/nvme.html#metadata.

Maybe something we may be able to do here then. We run the python BDD tests in a vm, maybe we can configure it with an emulated PCIe device and be able to run these tests then?
Alternatively we can setup VM's on demand from the tests themselves? With nix we can setup test vm's easily, example: https://github.com/openebs/mayastor-extensions/blob/develop/tests/helm/test.nix 
But I don't think there's emulated nvme support there...

io-engine/src/bdev/ftl.rs
tiagolobocastro: Can we use a mix, example, only use percent decoding for the ? and & ?

io-engine/src/bdev/ftl.rs
tiagolobocastro: Do you need to add it to `allowlist_type` on the spdk-rs build.rs?

io-engine/src/bdev/ftl.rs
tiagolobocastro: oh, interesting...

io-engine/src/bdev/ftl.rs
tiagolobocastro: nvm I must have been day dreaming, makes perfect sense to destroy the top-down...

tiagolobocastro (approved):

io-engine/src/bdev/ftl.rs
MaisenbacherD: Thanks that works!
I was so close :D While testing I had already added the `rustified_enum` but forgot the `allowlist_type`.
I opened another PR on spdk-rs for this: https://github.com/openebs/spdk-rs/pull/69

io-engine/src/bdev/ftl.rs
MaisenbacherD: Yes this works. I rewrote this parsing section and now we can use partial percent-encoding for `?` and `&` . Optionally you can also use percent-encoding for the entire sub-uri.
Thanks for this suggestion :)

doc/test.md
MaisenbacherD: I now introduced a new feature flag :)


Yes we should definitely look into running the test in an emulated QEMU environment or something similar.

That on demand VM setup you linked looks interesting. It would take a bit time to play with it and figure out the capabilities.
Should we address the CI setup within this PR?

tiagolobocastro: We can do in another PR, let's get this one merged :)

io-engine/src/bdev/ftl.rs
tiagolobocastro: Awesome :)

io-engine/src/bdev/ftl.rs
tiagolobocastro: Great, approved that one too!

tiagolobocastro:
bors try

MaisenbacherD:
Addressed the linter errors :)

doc/test.md
MaisenbacherD: I will talk with you outside of this PR to discuss this :)

dsharma-dc:
bors try

dsharma-dc:
Linter failures still. Could you please run below script in your workspace:
`./scripts/rust-style.sh`

Please refer https://gist.github.com/dsharma-dc/a14488638a7815561571882759e7921e

MaisenbacherD:
> Linter failures still. Could you please run below script in your workspace:

Thanks for the hint :) I just ran the script.

io-engine/src/bdev/ftl.rs
```diff
@@ -0,0 +1,310 @@
...
+//!
+//! The bbdev_uri and cbdev_uri need to use percent encoding on '?' (= '%3F')
+//! and '&' (= '%26') segment dividers. This is needed so we can successfully
+//! parse the ftl uri. Optionally, the sub-uris may also be fully
+//! percent-encoded.
+//!
+//! # Parameters
+//! name: A name for the ftl device, example: "ftl-1".
+//! alias: Ftl device URI which can be used to open the bdev.
+//! uuid: A UUID that can be seet to reference the resutling SPDK bdev.
```
dsharma-dc:
```suggestion
//! uuid: A UUID that can be set to reference the resulting SPDK bdev.
```

tiagolobocastro:
bors try

MaisenbacherD:
I updated the spdk-rs submodule reference to the merged commit from https://github.com/openebs/spdk-rs/pull/69 :)

MaisenbacherD:
bors try

tiagolobocastro:
You should be able to use bors with this:
bors d+

MaisenbacherD:
Resolved some more linter issues :)

MaisenbacherD:
bors d+

MaisenbacherD:
bors try

io-engine/src/bdev/ftl.rs
```diff
@@ -0,0 +1,310 @@
...
+
+        if errno != 0 {
+            let err = BdevError::CreateBdevFailed {
+                source: Errno::from_raw(errno.abs()),
+                name: self.name.clone(),
+            };
+
+            error!("{:?} error: {}", self, err.verbose());
+
+            return Err(err);
```
dsharma-dc, line 259: Shouldn't we destroy the base bdev and cache bdev created earlier here, in case of ftl bdev creation failed?

MaisenbacherD: I agree :) Will address this tomorrow.

MaisenbacherD:
Updated spdk-rs submodule reference https://github.com/openebs/spdk-rs/pull/70#pullrequestreview-2465784790

io-engine/src/bdev/ftl.rs
MaisenbacherD: @dsharma-dc and @tiagolobocastro
Would it make sense to implement the drop trait for `Ftl` here where we always `bdev_destroy` `bbdev` and `cbdev`?
Further up in the code the creation of `cbdev` might fail and then we would need to destroy `bbdev` as well.

tiagolobocastro: Possibly though may be difficult as `Drop` is blocking. We could potentially just schedule the deletion to happen asynchronously, though this may become an awkward api.
Another question here is, what if the b and c devices are exist?  Presumably this fails on `bdev_create`?

MaisenbacherD: > Possibly though may be difficult as Drop is blocking. We could potentially just schedule the deletion to happen asynchronously, though this may become an awkward api.

Okay, then I will handle those cases individually :)

> Another question here is, what if the b and c devices are exist? Presumably this fails on bdev_create?

Yes it is going to fail, and I would propose to keep it like this because we don't want to accidentally reformat a drive that is already in use :)

MaisenbacherD:
@dsharma-dc Thanks for your comments :)
I addressed them.

MaisenbacherD:
bors try

dsharma-dc (approved):

MaisenbacherD:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
doh, had missed a couple
bors merge

PR #1780, opened 2024-12-09
fix(nexus): don't persist if child faults during nexus create

dsharma-dc:
When nexus is being created with a single child, and that child goes into retire path before nexus is open, then the child gets persisted as unhealthy. This will cause volume to never be able to attach later on.

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_persistence.rs
+++ b/io-engine/src/bdev/nexus/nexus_persistence.rs
@@ -103,6 +103,15 @@ impl<'n> Nexus<'n> {
                     };
                     nexus_info.children.push(child_info);
                 });
+                // We started with this child because it was healthy in etcd, or isn't there at all.
+                // Being unhealthy here means it is undergoing a fault/retire before nexus is open.
+                if nexus_info.children.len() == 1 && !nexus_info.children[0].healthy {
+                    warn!("{self:?} Not persisting: the only child went unhealthy during nexus creation");
+                    return Err(Error::NexusCreate {
+                        name: self.name.clone(),
+                        reason: "only child is unhealthy".to_string(),
+                    });
+                }
             }
             PersistOp::AddChild { child_uri, healthy } => {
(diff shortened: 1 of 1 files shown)


io-engine/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -103,6 +103,15 @@ impl<'n> Nexus<'n> {
                     };
                     nexus_info.children.push(child_info);
                 });
+                // We started with this child because it was healthy in etcd, or isn't there at all.
+                // Being unhealthy here means it is undergoing a fault/retire before nexus is open.
+                if nexus_info.children.len() == 1 && !nexus_info.children[0].healthy {
```
tiagolobocastro, line 108: How do we get here?
We create the nexus, and open the children (or child in this case) and when we're saving the nexus info the child gets faulted at the same time?

dsharma-dc: yeah, the child is undergoing retire path while nexus creation isn't complete. We also see the persist call from retire path emits this warning since the create path persisted replica as unhealthy.
>[2024-11-13T18:06:02.134805509+00:00  WARN io_engine::bdev::nexus::nexus_bdev_children:nexus_bdev_children.rs:1055] Nexus 'ba4de70e-0ecb-4488-bb2a-40abb924a634' [open]: no healthy replicas persent in persistent store when retiring replica 'nvmf://x.x.x.x:8420/nqn.2019-05.io.openebs:ef00f725-b47d-4720-aa69-184cf3ab0114?uuid=ef00f725-b47d-4720-aa69-184cf3ab0114':
                            not persisting the replica state

dsharma-dc:
bors try

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

dsharma-dc:
bors merge

PR #1782, opened 2024-12-11
Cherry pick PR #1780

dsharma-dc:
When nexus is being created with a single child, and that child goes into retire path before nexus is open, then the child gets persisted as unhealthy. This will cause volume to never be able to attach later on.


dsharma-dc:
bors try

sinhaashish (approved):

Abhinandan-Purkait (approved):

dsharma-dc:
bors merge

PR #1786, opened 2024-12-18
chore: increase ps_retries and silence log once logged

Abhinandan-Purkait:
- Increase the ps_retries to 300.
- Silently retry.

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_persistence.rs
+++ b/io-engine/src/bdev/nexus/nexus_persistence.rs
@@ -104,9 +104,12 @@ impl<'n> Nexus<'n> {
                     };
                     nexus_info.children.push(child_info);
                 });
-                // We started with this child because it was healthy in etcd, or isn't there at all.
-                // Being unhealthy here means it is undergoing a fault/retire before nexus is open.
-                if nexus_info.children.len() == 1 && !nexus_info.children[0].healthy {
+                // We started with this child because it was healthy in etcd, or
+                // isn't there at all. Being unhealthy here
+                // means it is undergoing a fault/retire before nexus is open.
+                if nexus_info.children.len() == 1
+                    && !nexus_info.children[0].healthy
+                {
                     warn!("{self:?} Not persisting: the only child went unhealthy during nexus creation");
                     return Err(Error::NexusCreate {
                         name: self.name.clone(),
@@ -224,6 +227,7 @@ impl<'n> Nexus<'n> {
         };
 
         let mut retry = PersistentStore::retries();
+        let mut logged = false;
         loop {
             let Err(err) = PersistentStore::put(&key, &info.inner).await else {
                 trace!(?key, "{self:?}: the state was saved successfully");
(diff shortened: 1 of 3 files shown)


io-engine/src/core/env.rs
```diff
@@ -183,7 +183,7 @@ pub struct MayastorCliArgs {
     )]
     /// Persistent store timeout.
     pub ps_timeout: Duration,
-    #[clap(long = "ps-retries", default_value = "30")]
+    #[clap(long = "ps-retries", default_value = "255")]
     /// Persistent store operation retries.
     pub ps_retries: u8,
```
Abhinandan-Purkait: Should this be u16 or u32? isn't u8 is kind of low?

Abhinandan-Purkait: We can go as high as 255 in this case.

dsharma-dc: u16 should be ok to use here.

io-engine/src/core/env.rs
```diff
@@ -183,7 +183,7 @@ pub struct MayastorCliArgs {
     )]
     /// Persistent store timeout.
     pub ps_timeout: Duration,
-    #[clap(long = "ps-retries", default_value = "30")]
+    #[clap(long = "ps-retries", default_value = "255")]
```
tiagolobocastro: Modifying this might affect runtime of existing test.
Can leave this as is and modify  through the helm chart instead?

io-engine/src/core/env.rs
tiagolobocastro: Yeah maybe use u16

io-engine/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -238,10 +242,13 @@ impl<'n> Nexus<'n> {
...
             }
 
-            error!(
-                "{self:?}: failed to persist nexus information, \
-                will retry ({retry} left): {err}"
-            );
+            if !logged {
+                error!(
+                    "{self:?}: failed to persist nexus information, \
+                    will retry ({retry} left): {err}"
```
tiagolobocastro: :)

```suggestion
                    will silently retry ({retry} left): {err}"
```

io-engine/src/core/env.rs
Abhinandan-Purkait: Need to add a new value in that case. Can do that.

io-engine/src/core/env.rs
Abhinandan-Purkait: Done.

Abhinandan-Purkait: https://github.com/openebs/mayastor-extensions/pull/592

io-engine/src/core/env.rs
Abhinandan-Purkait: https://github.com/openebs/mayastor-extensions/pull/592

dsharma-dc (approved):

tiagolobocastro (approved):

Abhinandan-Purkait:
bors merge

PR #1788, opened 2024-12-18
fix(self-shutdown): abort frozen ios when unsharing shutdown nexus

tiagolobocastro:
These frozen IOs prevent the nexus from shutting down. We don't have any hooks today to do this whilsts the target is stopping so we add a simple loop which tries a number of times. Ideally we should get some sort of callback to trigger this.

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev.rs
@@ -1114,6 +1114,18 @@ impl<'n> Nexus<'n> {
         Ok(())
     }
 
+    /// Aborts all frozen IOs of a shutdown Nexus.
+    /// # Warning
+    /// These aborts may translate into I/O errors for the initiator.
+    pub async fn abort_shutdown_frozen_ios(&self) {
+        if self.status() == NexusStatus::Shutdown {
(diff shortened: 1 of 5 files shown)


dsharma-dc:
The frozen IOs are held up in our NexusChannel and not yet **(re)submitted into the child device io channel**, because in self shutdown path we haven't brought back io mode to normal. So wondering why they hold up nexus nvmf subsystem unshare/stop? Perhaps since the IOs have entered the nexus subsystem and not ack'ed?

dsharma-dc (approved):

tiagolobocastro:
> The frozen IOs are held up in our NexusChannel and not yet **(re)submitted into the child device io channel**, because in self shutdown path we haven't brought back io mode to normal. So wondering why they hold up nexus nvmf subsystem unshare/stop? Perhaps since the IOs have entered the nexus subsystem and not ack'ed?

Haven't looked into the subsystem for a while but IIRC it wont stop because there's outstanding IO. When the last IO completes the subsystem is stopped.

abhilashshetty04 (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #1791, opened 2024-12-20
feat(io-engine): add persistent store transaction API

dsharma-dc:
This change adds API to do Compare-and-Swap operations to the persistent store via io-engine.


dsharma-dc:
bors try

io-engine/src/store/etcd.rs
```diff
@@ -49,6 +49,25 @@ impl Store for Etcd {
...
+        key: &K,
+        cmps: Vec<Compare>,
+        ops_success: Vec<TxnOp>,
+        ops_failure: Option<Vec<TxnOp>>,
+    ) -> Result<TxnResponse, StoreError> {
+        let fops = ops_failure.map_or(vec![], |v| v);
+        self.0
+            .txn(Txn::new().when(cmps).and_then(ops_success).or_else(fops))
+            .await
+            .context(TxnErr {
```
tiagolobocastro, line 66: Do we get an error on the logs which is clear as to what happened?

io-engine/src/store/etcd.rs
```diff
@@ -49,6 +49,25 @@ impl Store for Etcd {
...
+    async fn txn_kv<K: StoreKey>(
+        &mut self,
+        key: &K,
+        cmps: Vec<Compare>,
+        ops_success: Vec<TxnOp>,
+        ops_failure: Option<Vec<TxnOp>>,
+    ) -> Result<TxnResponse, StoreError> {
+        let fops = ops_failure.map_or(vec![], |v| v);
+        self.0
+            .txn(Txn::new().when(cmps).and_then(ops_success).or_else(fops))
```
tiagolobocastro, line 64: what is this `or_else` doing?

io-engine/src/store/etcd.rs
```diff
@@ -49,6 +49,25 @@ impl Store for Etcd {
         Ok(())
     }
 
+    /// Executes a transaction for the given key. If the compares succeed, then
+    /// ops_success will be executed atomically, otherwise ops_failure will be
+    /// executed atomically.
+    async fn txn_kv<K: StoreKey>(
+        &mut self,
+        key: &K,
+        cmps: Vec<Compare>,
```
tiagolobocastro, line 58: Why do we need an array here? Can't we just do .when([cmp]) ?

dsharma-dc: > Why do we need an array here? Can't we just do .when([cmp]) ?

If we need to check multiple compare conditions, then caller need to build a list/vector of Compares and pass here.

io-engine/src/store/etcd.rs
dsharma-dc: ops in the `or_else` part get executed atomically if the compare conditions fail.

io-engine/src/store/etcd.rs
dsharma-dc: > Do we get an error on the logs which is clear as to what happened?

It's difficult to get complete info as to what failed as part of transaction. The `succeeded()` validation  in transaction response is deterministic info of whether transaction compares succeed or fail.

tiagolobocastro: So what does it look like on the logs with this change?

io-engine/src/store/etcd.rs
tiagolobocastro: Perhaps I'd suggest start simpler by allowing a single compare first

io-engine/src/store/etcd.rs
dsharma-dc: You mean from io-engine logging? or api usage. For io-engine it'll be something like below, this is not final yet:

```
[2025-01-08T12:05:37.051454598+00:00  INFO io_engine::persistent_store:persistent_store.rs:208] Executing transaction for key /openebs.io/mayastor/apis/v0/clusters/761ab476f4f4489b900ded45a464a79d/namespaces/default/volume/1c4e66fd-3b33-4439-b504-d49aba53da26/nexus/2f2455c8-e237-4586-848d-762f26ebb044/info.
[2025-01-08T12:05:37.052938551+00:00 ERROR io_engine::bdev::nexus::nexus_persistence:nexus_persistence.rs:216] Nexus '1c4e66fd-3b33-4439-b504-d49aba53da26' [open]: failed to update persistent store, will shutdown the nexus: failed to save nexus state 1c4e66fd-3b33-4439-b504-d49aba53da26
```

io-engine/src/store/etcd.rs
dsharma-dc: > Perhaps I'd suggest start simpler by allowing a single compare first

That shouldn't matter here. The underlying client api anyway takes a vector. Our caller will use a single Compare for now.

io-engine/src/store/etcd.rs
tiagolobocastro: Would be good to have some more context here on the failure.
Also I hope we distinguish compare errors from general etcd write failures which should be retried?

dsharma-dc: > Would be good to have some more context here on the failure. Also I hope we distinguish compare errors from general etcd write failures which should be retried?

Yes the general error if received as `Err` from txn API has to be dealt with retries by caller. However if a Put TxnOp failure happens after a successful compare, then the etcd handles it harshly on server side:
```
	_, err := executeTxn(ctx, lg, txnWrite, rt, txnPath, txnResp)
	if err != nil {
		if isWrite {
			// CAUTION: When a txn performing write operations starts, we always expect it to be successful.
			// If a write failure is seen we SHOULD NOT try to recover the server, but crash with a panic to make the failure explicit.
			// Trying to silently recover (e.g by ignoring the failed txn or calling txn.End() early) poses serious risks:
			// - violation of transaction atomicity if some write operations have been partially executed
			// - data inconsistency across different etcd members if they applied the txn asymmetrically
			lg.Panic("unexpected error during txn with writes", zap.Error(err))
		} else {
			lg.Error("unexpected error during readonly txn", zap.Error(err))
		}
	}

Comment (dsharma-dc) [reply io-engine/src/store/etcd.rs:66]: > Would be good to have some more context here on the failure.

Have extended caller side error handling(not in this PR) by printing `source` in `StateSaveFailed` error to add more context in error, as below.
```
[2025-01-09T11:29:00.330005232+00:00  INFO io_engine::persistent_store:persistent_store.rs:208] Executing transaction for key /openebs.io/mayastor/apis/v0/clusters/761ab476f4f4489b900ded45a464a79d/namespaces/default/volume/1c4e66fd-3b33-4439-b504-d49aba53da26/nexus/486a4605-1004-409b-b24c-9786c8ae279d/info.
[2025-01-09T11:29:00.331331260+00:00 ERROR io_engine::bdev::nexus::nexus_persistence:nexus_persistence.rs:216] Nexus '1c4e66fd-3b33-4439-b504-d49aba53da26' [open]: failed to update persistent store, will shutdown the nexus: failed to save nexus state 1c4e66fd-3b33-4439-b504-d49aba53da26, Failed to do 'transaction' for key /openebs.io/mayastor/apis/v0/clusters/761ab476f4f4489b900ded45a464a79d/namespaces/default/volume/1c4e66fd-3b33-4439-b504-d49aba53da26/nexus/486a4605-1004-409b-b24c-9786c8ae279d/info. Error io error: Txn CompareOp failed
```

Abhinandan-Purkait (approved):

io-engine/local-randrw-0-verify.state
Abhinandan-Purkait, line 1: Removed this file?

dsharma-dc: > Remove this file?

This file seems to have been checked-in with some earlier PR mistakenly. It's a fio verify-async state file and is not needed. So just deleting as part of this PR.

Abhinandan-Purkait: Okay.

tiagolobocastro (approved):

dsharma-dc:
bors merge

PR #1793, opened 2024-12-26
ci: switch from jenkins to github actions

tiagolobocastro:
Jenkins has become troublesome to maintain and makes it harder for users
    to see and test output without adding new users.
    We could probably spend some time improving that, but the current Jenkins
    is also setup on systems managed by the sponsoring company so it makes
    sense to move away from that at least.
    In order to run the CI tests we need a relatively powerful CI system so
    we are making use of the ubuntu-latest-16-cores runners.

    The tests seem to run slower than on Jenkins. Part of it may be explained
    because the dependencies are not pre-cache and also the pre-requisites
    which take time to install.

### DIFF
--- a/.github/workflows/bdd.yml
+++ b/.github/workflows/bdd.yml
@@ -1,0 +1,62 @@
+name: BDD CI
+on:
+  workflow_call:
+
+env:
+  CARGO_TERM_COLOR: always
+  CARGO_INCREMENTAL: 0
+  CI: 1
+
+jobs:
+  bdd-tests:
+    runs-on: ubuntu-latest-16-cores
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          submodules: 'recursive'
+      - uses: DeterminateSystems/nix-installer-action@v14
+      - uses: DeterminateSystems/magic-nix-cache-action@v8
+      - name: Pre-populate nix-shell
+        run: |
+          export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' nix/sources.json -r)
+          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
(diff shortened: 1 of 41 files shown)


tiagolobocastro:
bors try

.github/workflows/bdd.yml
```diff
@@ -0,0 +1,62 @@
+name: BDD CI
+on:
+  workflow_call:
```
abhilashshetty04, line 3: Why we use on.`workflow_call`? and not `pull_request` or `push`.. on other files i see we trigger some on `push`

tiagolobocastro: Because this is part of the merge check, see pr-ci.yml

test/python/tests/nexus/docker-compose.yml
```diff
@@ -8,15 +8,15 @@ services:
     container_name: "ms0"
     image: rust:latest
     environment:
-        - MY_POD_IP=10.0.0.2
+        - MY_POD_IP=10.1.0.2
```
dsharma-dc, line 11: Why are subnet and IP changes needed in these files?

abhilashshetty04 (approved):

dsharma-dc (approved):

Abhinandan-Purkait (approved):

test/python/tests/nexus/docker-compose.yml
tiagolobocastro: Because it's already used in github actions runner :(
Though this is not great having all these ips hardcoded to be honest, this is bound to be an issue again!

tiagolobocastro:
Change of the guards :)
bors merge

tiagolobocastro:
ah, forgot one file...
bors cancel

tiagolobocastro:
bors merge

PR #1804, opened 2025-01-17
ci: ignore udisksd coredumps

tiagolobocastro:
Otherwise we detect this as a failure.
todo: instead, we should only detect crashes on processes started by the test.


dsharma-dc (approved):

niladrih (approved):

tiagolobocastro:
bors merge

PR #1811, opened 2025-01-31
Rename action and add ci-report

tiagolobocastro:
ci(gha): upload test report
    
    Add test script which generates a report tar with useful
    information such as journalctl logs and nvme devices.
    Upload this tar on github actions.
    
---

    ci: rename push job
    
    The previous name was image-build-test which clashed with the image
    build test :)

### DIFF
--- a/.github/workflows/bdd.yml
+++ b/.github/workflows/bdd.yml
@@ -41,7 +41,7 @@ jobs:
         run: |
           echo "TEST_START_DATE=$(date +"%Y-%m-%d %H:%M:%S")" >> $GITHUB_ENV
           nix-shell --run "./scripts/pytest-tests.sh"
-      - name: Test Report
+      - name: Pytest BDD Report
         if: always()
         uses: pmeier/pytest-results-action@main
         with:
@@ -50,6 +50,18 @@ jobs:
           display-options: a
           fail-on-empty: true
           title: Test results
+      - run: |
+          if [ -d test/python/reports ]; then
+            mkdir -p ci-report
+            mv test/python/reports ci-report/python
+          fi
+          nix-shell --run "./scripts/ci-report.sh"
+        if: failure()
+      - uses: actions/upload-artifact@v4
+        if: failure()
+        with:
+          name: ci-report-bdd
+          path: ./ci-report/ci-report.tar.gz
       - name: Cleanup
         if: always()
         run: nix-shell --run "./scripts/pytest-tests.sh --clean-all-exit"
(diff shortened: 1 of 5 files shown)


tiagolobocastro:
bors try

tiagolobocastro:
bors try

Abhinandan-Purkait (approved):

niladrih (commented):
Sequential test runs would overwrite previous reports? 
Maybe use mktemp to generate a unique directory/file name?

tiagolobocastro:
> Sequential test runs would overwrite previous reports? Maybe use mktemp to generate a unique directory/file name?

What do you mean by sequential test runs?

niladrih:
Out of CI, local runs back to back.

tiagolobocastro:
> Out of CI, local runs back to back.

If if the run fails, then the next run won't "run". At least the only way I've run it back to back.
I don't think we have a use case for this atm.

pchandra19:
> Sequential test runs would overwrite previous reports? Maybe use mktemp to generate a unique directory/file name?

The report will be stored as an artifact of the GA run. Each run should produce a separate artifact IMO , thus avoiding overrides.
@tiagolobocastro Since these artifacts do not seem to be stored for a very long time , I would suggest putting in a custom retention for these if possible (retention-days: X)

tiagolobocastro:
> > Sequential test runs would overwrite previous reports? Maybe use mktemp to generate a unique directory/file name?
> 
> The report will be stored as an artifact of the GA run. Each run should produce a separate artifact IMO , thus avoiding overrides. @tiagolobocastro Since these artifacts do not seem to be stored for a very long time , I would suggest putting in a custom retention for these if possible (retention-days: X)

The default is 90 days IIRC
Do you suggest a different value?

pchandra19:
> > > Sequential test runs would overwrite previous reports? Maybe use mktemp to generate a unique directory/file name?
> > 
> > 
> > The report will be stored as an artifact of the GA run. Each run should produce a separate artifact IMO , thus avoiding overrides. @tiagolobocastro Since these artifacts do not seem to be stored for a very long time , I would suggest putting in a custom retention for these if possible (retention-days: X)
> 
> The default is 90 days IIRC Do you suggest a different value?

I was thinking more like a week or two at max. for these unless you see a use case where we would need it for longer. if its the latter , I would suggest to keep the default and not change here.

tiagolobocastro:
> I was thinking more like a week or two at max. for these unless you see a use case where we would need it for longer. if its the latter , I would suggest to keep the default and not change here.

Provided it doesn't hinder the actions in any way I'd say 30 days, as sometimes we may not get a chance to go through a failure until later.
Alright, I'll leave it as is then.

niladrih (approved):

tiagolobocastro:
bors merge

PR #1824, opened 2025-02-25
feat(encryption): add crypto vbdev

dsharma-dc:
This adds a crypto vbdev support in Mayastor. A crypto vbdev is always created with a base bdev underneath e.g an aio bdev. A crypto vbdev needs an associated key and crypto opts, that are attached to the vbdev for its lifetime.

### DIFF
--- a/io-engine/src/bdev/crypto.rs
+++ b/io-engine/src/bdev/crypto.rs
@@ -1,0 +1,266 @@
+//!
+//! This modules deals with managing crypto vbdev. The crypto vbdev is
+//! a virtual bdev that needs a base bdev for it to be created. The crypto vbdev
+//! concept can be applied to encrypt the complete lvolstore or a particular lvol.
+//! For encrypting lvolstore(Mayastor diskpool), the crypto vbdev is created on top
+//! of base bdev of the disk pool e.g. aio bdev. For lvol level encryption, the
+//! crypto vbdev is to be created on top of the lvol vbdev.
+//!
+//! A general workflow example for creating a spdk crypto vbdev for encrypted lvolstore is as below:
+//!
+//! sudo $SPDK_SCRIPT_PATH/rpc.py accel_crypto_key_create -c AES_CBC -k 01234567891234560123456789deadc0 -n dskey_aes128_cbc_1
+//! sudo $SPDK_SCRIPT_PATH/rpc.py bdev_aio_create /dev/loop0 MyBaseBdev 512
+//! sudo $SPDK_SCRIPT_PATH/rpc.py bdev_crypto_create MyBaseBdev MyCryptoVbdev -n key_aes128_cbc
+//!
+//! The vbdev MyCryptoVbdev can now be used to create lvolstore on it.
+//!
+
+use futures::channel::oneshot;
+use io_engine_api::v1 as v1rpc;
+use nix::errno::Errno;
+use spdk_rs::{
+    ffihelper::IntoCString,
+    libspdk::{
+        accel_dpdk_cryptodev_enable, accel_dpdk_cryptodev_set_driver, create_crypto_disk,
+        create_crypto_opts_by_name, delete_crypto_disk, spdk_accel_assign_opc,
+        spdk_accel_crypto_key, spdk_accel_crypto_key_create, spdk_accel_crypto_key_create_param,
+        spdk_accel_crypto_key_get,
+    },
+};
+use std::{
+    fmt::{Debug, Display, Formatter},
+    os::raw::c_char,
+};
+
+use crate::{
+    bdev_api::BdevError,
+    core::CoreError,
+    ffihelper::{cb_arg, pair, FfiResult},
+};
+
+/// Representation of a crypto vbdev to be created in SPDK.
+/// TODO: Currently we don't fit crypto vbdev into the uri based bdev management
+/// as the information required to completely setup the crypto vbdev isn't part of the uri,
+/// and also the crypto uri isn't provided by user rather we'll have to do some masking
+/// using crypto:// scheme.
+#[allow(dead_code)]
+#[derive(Default)]
+pub(crate) struct Crypto {
+    // Name of the crypto vbdev.
+    pub name: String,
+    // Name of the base bdev.
+    pub base_bdev_name: String,
+    // Encryption key params.
+    pub key: EncryptionKey,
+}
+
+#[derive(Copy, Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
+/// Cipher mode for the encryption enablement. The default is kept as AES_XTS here
+/// because operations are default assigned to software module, which only supports
+/// AES_XTS. We don't expect to use default though.
+pub enum Cipher {
+    AesCbc,
+    #[default]
+    AesXts,
+}
+
+#[derive(Clone, Debug, Default, PartialEq)]
+/// A struct that represents the parameters required to create a key.
+pub struct EncryptionKey {
+    pub cipher: Cipher,
+    pub key_name: String,
+    pub key: String,
+    pub key_len: u32,
+    pub key2: Option<String>,
+    pub key2_len: Option<u32>,
+}
+
+impl Debug for Crypto {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "crypto vbdev '{}', 'base_bdev: {}', keyname '{}', cipher: '{:?}'",
+            self.name, self.base_bdev_name, self.key.key_name, self.key.cipher
+        )
+    }
+}
+
+impl Display for Cipher {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let c = match self {
+            Self::AesCbc => "AES_CBC",
+            Self::AesXts => "AES_XTS",
+        };
+        write!(f, "{c}")
+    }
+}
+
+impl From<v1rpc::common::Cipher> for Cipher {
+    fn from(value: v1rpc::common::Cipher) -> Self {
+        match value {
+            v1rpc::common::Cipher::AesCbc => Self::AesCbc,
+            v1rpc::common::Cipher::AesXts => Self::AesXts,
+        }
+    }
+}
+
+/// Driver types for dpdk_cryptodev module. We won't be using any other
+/// than aesni-mb.
+#[derive(Debug, Default)]
+enum AccelDpdkCryptodevDriverType {
+    #[default]
+    AccelDpdkCryptodevAesniMb,
+    _AccelDpdkCryptodevQat,
+    _AccelDpdkCryptodevMlx5Pci,
+    _AccelDpdkCryptodevUadk,
+    _AccelDpdkCryptodevLast,
+}
+
+impl Display for AccelDpdkCryptodevDriverType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let drv = match self {
+            Self::AccelDpdkCryptodevAesniMb => "crypto_aesni_mb",
+            Self::_AccelDpdkCryptodevQat => "crypto_qat",
+            Self::_AccelDpdkCryptodevMlx5Pci => "mlx5_pci",
+            Self::_AccelDpdkCryptodevUadk => "crypto_uadk",
+            Self::_AccelDpdkCryptodevLast => "crypto_driver_invalid",
+        };
+        write!(f, "{drv}")
+    }
+}
+
+// This function enables the dpdk_cryptodev module with aesni_mb driver in spdk,
+// and assigns the encrypt and decrypt operation to it, which by default are assigned to
+// software module initially.
+#[allow(dead_code)]
+pub fn enable_dpdk_cryptodev_accel_module() -> Result<(), CoreError> {
+    info!("Enabling accel dpdk_cryptodev module");
+    unsafe {
+        accel_dpdk_cryptodev_enable();
+        let setdrv_ret = accel_dpdk_cryptodev_set_driver(
+            AccelDpdkCryptodevDriverType::AccelDpdkCryptodevAesniMb
+                .to_string()
+                .into_cstring()
+                .as_ptr(),
+        );
+        if setdrv_ret != 0 {
+            return Err(CoreError::InitCryptoModule {
+                reason: format!("Unsupported driver: {setdrv_ret:?}"),
+            });
+        }
+        let err_enc = spdk_accel_assign_opc(8, "dpdk_cryptodev".into_cstring().as_ptr());
+        let err_dec = spdk_accel_assign_opc(9, "dpdk_cryptodev".into_cstring().as_ptr());
+        trace!("assign opc err_enc: {err_enc}, err_dec: {err_dec}");
+        if err_enc != 0 || err_dec != 0 {
+            return Err(CoreError::InitCryptoModule {
+                reason: format!(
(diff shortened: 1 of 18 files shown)


dsharma-dc:
bors try

io-engine/src/bdev/crypto.rs
```diff
@@ -0,0 +1,281 @@
...
+    // Name of the base bdev.
+    pub base_bdev_name: String,
+    // Encryption key params.
+    pub key: EncryptionKey,
+}
+
+#[derive(Copy, Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
+/// Cipher mode for the encryption enablement. The default is kept as AES_XTS here
+/// because operations are default assigned to software module, which only supports
+/// AES_XTS. We don't expect to use default though without inputs from upstream.
```
tiagolobocastro: What does this mean, without inputs from upstream?

io-engine/src/bdev/crypto.rs
```diff
@@ -0,0 +1,281 @@
...
+pub enum Cipher {
+    AesCbc,
+    #[default]
+    AesXts,
+    Unsupported,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
+pub struct EncryptionKey {
```
tiagolobocastro, line 69: missing doc comments

io-engine/src/bdev/crypto.rs
```diff
@@ -0,0 +1,281 @@
...
+};
+
+/// Representation of a crypto vbdev to be created in SPDK.
+/// TODO: Currently we don't fit crypto vbdev into the uri based bdev management
+/// as the information required to completely setup the crypto vbdev isn't part of the uri,
+/// and also the crypto uri isn't provided by user rather we'll have to do some masking
+/// using crypto:// scheme.
+#[allow(dead_code)]
+#[derive(Default)]
+pub struct Crypto {
```
tiagolobocastro: Do these need to be pub or is pub(crate) enough?

io-engine/src/bdev/crypto.rs
```diff
@@ -0,0 +1,281 @@
...
+    pub key2: Option<String>,
+    pub key2_len: Option<u32>,
+}
+
+impl Debug for Crypto {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "crypto vbdev '{}', 'base_bdev: {}', keyname '{}', key '{}', key2 '{:?}'",
+            self.name, self.base_bdev_name, self.key.key_name, self.key.key, self.key.key2
```
tiagolobocastro: should we really do this, seems unsafe to output the raw key?

io-engine/src/bdev/crypto.rs
```diff
@@ -0,0 +1,281 @@
...
+        )
+    }
+}
+
+impl Display for Cipher {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{}",
+            match self {
```
tiagolobocastro: nit: create var for the str would reduce the loc

io-engine/src/bdev/crypto.rs
```diff
@@ -0,0 +1,281 @@
...
+                Self::AesCbc => "AES_CBC",
+                Self::AesXts => "AES_XTS",
+                Self::Unsupported => "UNSUPPORTED",
+            }
+        )
+    }
+}
+
+impl From<i32> for Cipher {
```
tiagolobocastro: Why is this needed?

io-engine/src/bdev/crypto.rs
```diff
@@ -0,0 +1,281 @@
...
+        k2.into_cstring().into_raw()
+    } else {
+        std::ptr::null_mut()
+    };
+    let key_name = key_params.key_name.as_str().into_cstring().into_raw();
+
+    unsafe {
+        debug!(
+            "Creating {} key {} | {k2} of size {}",
+            CStr::from_ptr(cipher).to_str().unwrap(),
```
tiagolobocastro: Why are we converting it from ptr when we have it already?

io-engine/src/bdev/crypto.rs
```diff
@@ -0,0 +1,281 @@
...
+            CStr::from_ptr(key).to_str().unwrap(),
```
tiagolobocastro: again, this doesn't seem safe

io-engine/src/bdev/crypto.rs
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    r.await
+        .expect("callback gone while deleting crypto disk")
+        .to_result(|e| BdevError::DestroyBdevFailed {
+            source: Errno::from_raw(e),
+            name: vbdev_name.to_string(),
+        })?;
+
+    info!("crypto vbdev destroyed successfully");
```
```suggestion
    info!("crypto vbdev {vbdev_name} destroyed successfully");
```

io-engine/src/bdev/crypto.rs
```diff
@@ -0,0 +1,281 @@
...
+            });
+        }
+        Ok(())
+    }
+}
+
+/// A simple wrapper around `spdk_accel_crypto_key_create`. The keys' strings here are
+/// in hexlified format.
+fn create_crypto_key(key_params: &EncryptionKey) -> Result<*mut spdk_accel_crypto_key, BdevError> {
+    let cipher = key_params.cipher.to_string().into_cstring().into_raw();
```
tiagolobocastro: Do you need to transfer ownership of the string to spdk?
If not then probably don't use into_raw. Otherwise you need to clean it up, it's leaking memory here.

io-engine/src/bdev/crypto.rs
```diff
@@ -0,0 +1,281 @@
...
+    base_bdev_name: &str,
+    key_params: &EncryptionKey,
+) -> Result<(), BdevError> {
+    // Create the key.
+    let key = create_crypto_key(key_params)?;
+
+    // Setup the crypto opts using the key now.
+    let crypto_opts_ptr = unsafe {
+        create_crypto_opts_by_name(
+            crypto_vbdev_name.into_cstring().into_raw(),
```
tiagolobocastro: same comment here, it's leaking memory, probably as_ptr works here as IIRC this helper functions copies the strings

io-engine/src/core/mod.rs
```diff
@@ -273,6 +273,10 @@ pub enum CoreError {
     WipeFailed {
         source: wiper::Error,
     },
+    #[snafu(display("Failed to init crypto module: {}", reason))]
```
tiagolobocastro:
```suggestion
    #[snafu(display("Failed to init crypto module: {reason}"))]
```

dsharma-dc: Does it work inline formatted like `{reason}` with snafu display? I see all other error codes here formatting outside?

io-engine/src/bdev/crypto.rs
dsharma-dc: Can't do this directly as vbdev_name is a reference to name from bdev which will print garbage after destruction. Can take a `String` input, would that be ok?

io-engine/src/bdev/crypto.rs
dsharma-dc: yeah, shouldn't print that. Remained from debugging.

io-engine/src/bdev/crypto.rs
dsharma-dc: Basically we don't decide a default cipher if not incoming from grpc. That won't be the case anyway so I'll remove this statement.

io-engine/src/bdev/crypto.rs
tiagolobocastro: Oh isn't `vbdev_name` the input from this function?

io-engine/src/core/mod.rs
tiagolobocastro: Yeah it works, the existing code predates this rust feature

io-engine/src/bdev/crypto.rs
dsharma-dc: > Oh isn't `vbdev_name` the input from this function?

No, It's from caller like `destroy_crypto_vbdev(bdev.name())` where bdev is of type `Bdev>()>`

tiagolobocastro: Ah I see, we'll have to clone it then.
But also, then I think destroy take the Bdev by value rather than the `&str` name? Otherwise this is kinda breaking the lifetime of &'a str.

io-engine/src/bdev/crypto.rs
dsharma-dc: to translate from grpc enum.

tiagolobocastro: should use the type from grpc, and not the i32 raw type

io-engine/src/bdev/crypto.rs
dsharma-dc: > Do these need to be pub or is pub(crate) enough?

pub(crate) seems enough.

io-engine/src/bdev/crypto.rs
dsharma-dc: Made the parameter `String` so we don't interfere with lifecycle.

io-engine/src/bdev/crypto.rs
dsharma-dc: modified to use as_ptr()

io-engine/src/bdev/crypto.rs
dsharma-dc: > Do you need to transfer ownership of the string to spdk? If not then probably don't use into_raw. Otherwise you need to clean it up, it's leaking memory here.

Modified

dsharma-dc:
bors try

dsharma-dc:
bors try-

dsharma-dc:
bors try-

dsharma-dc:
bors try

dsharma-dc:
bors try-

dsharma-dc:
bors try

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
```diff
@@ -211,6 +253,26 @@ async fn create(mut ctx: Context, matches: &ArgMatches) -> crate::Result<()> {
         .to_owned();
 
     let uuid = matches.get_one::<String>("uuid");
+    let cipher = matches
+        .get_one::<String>("cipher")
+        .unwrap_or(&String::from("invalid_cipher"))
```
tiagolobocastro: perhaps can use Option rather than this string, then just check against option; like so:

match (cipher, key, key2)

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
```diff
@@ -54,6 +54,27 @@ pub fn subcommands() -> Command {
                 .required(false)
                 .value_parser(PoolType::types().to_vec())
                 .default_value(PoolType::Lvs.as_ref()),
+        )
+        .arg(
+            Arg::new("cipher")
+                .short('c')
+                .long("cipher")
+                .help("The cipher to use for encryption")
+                .required(false),
```
tiagolobocastro: try adding require kind of options to see if it simplifies things

io-engine/src/grpc/v1/pool.rs
```diff
@@ -226,13 +277,24 @@ impl TryFrom<ImportPoolRequest> for PoolArgs {
...
+            name: args.name.clone(),
+            disks: args.disks.clone(),
+            uuid: args.uuid.clone(),
             cluster_size: None,
             md_args: None,
             backend: backend.into(),
+            enc_key,
+            crypto_vbdev_name: args
+                .encryption
+                .clone()
```
tiagolobocastro:
```suggestion
                .as_ref()
```

io-engine/src/subsys/config/pool.rs
```diff
@@ -159,9 +159,11 @@ struct Pool {
     #[serde(skip_serializing)]
     replicas: Option<Vec<Replica>>,
     backend: PoolBackend,
+    /// Is the pool enrypted
```
tiagolobocastro:
```suggestion
    /// Is the pool enrypted.
```

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
```diff
@@ -54,6 +54,29 @@ pub fn subcommands() -> Command {
...
                 .value_parser(PoolType::types().to_vec())
                 .default_value(PoolType::Lvs.as_ref()),
+        )
+        .arg(
+            Arg::new("cipher")
+                .short('c')
+                .long("cipher")
+                .help("The cipher to use for encryption")
+                .required(false)
+                .requires("encryption-key"),
```
dsharma-dc, line 64: @tiagolobocastro  Improved clap cli args usage added.

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
```diff
@@ -312,6 +402,20 @@ async fn import(mut ctx: Context, matches: &ArgMatches) -> crate::Result<()> {
         })?
         .to_owned();
     let uuid = matches.get_one::<String>("uuid");
+    let cipher = matches
+        .get_one::<String>("cipher")
+        .unwrap_or(&String::from("invalid_cipher"))
```
tiagolobocastro: do the same changes here as above

io-engine/src/subsys/config/pool.rs
```diff
@@ -186,6 +194,8 @@ impl From<LvsBdev> for Pool {
                 .unwrap_or_else(|| base.name().to_string())],
             replicas: None,
             backend: PoolBackend::Lvs,
+            // XXX: Check how we use this and set correctly.
+            encrypted: false,
```
tiagolobocastro, line 198: hmm we can use this to tackle the potential issue we discussed earlier, of issuing create call with enc key to a previous version of io-engine. the control-plane could then check the pool state and if pool is reported as not encrypted then should fail creation, and perhaps even trigger pool deletion.

CC @Abhinandan-Purkait

nix/pkgs/io-engine/cargo-package.nix
```diff
@@ -91,6 +92,7 @@ let
       systemdMinimal.dev
       utillinux.dev
       rdma-core
+      keyutils
```
tiagolobocastro, line 95: note related to this PR:
somehow I wonder if the spdk-rs refactoring isn't complete yet, as we could be pulling dependencies from there rather than duplicating

tiagolobocastro (approved):

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
dsharma-dc: > perhaps can use Option rather than this string, then just check against option; like so:
> 
> match (cipher, key, key2)

It's indeed an Option in create path. I missed to change in import path. Changing now. Looks like you commented on old diff?

io-engine/src/subsys/config/pool.rs
dsharma-dc: yeah, this might be one way. We must indeed trigger deletion if such situation arises.

Abhinandan-Purkait (approved):

dsharma-dc:
bors merge

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
dsharma-dc: > try adding require kind of options to see if it simplifies things

Sorry I didn't understand this, could you please elaborate a bit?

PR #1827, opened 2025-03-06
request encrypted pool crypto vbdev

dsharma-dc:
Given encryption related args available from create pool request, this
creates a crypto vbdev based on the incoming parameters, and then
requests the lvolstore creation on that crypto vbdev.

### DIFF
--- a/io-engine/src/bin/io-engine-client/v1/pool_cli.rs
+++ b/io-engine/src/bin/io-engine-client/v1/pool_cli.rs
@@ -130,7 +130,8 @@ pub fn subcommands() -> Command {
                 .short('e')
                 .long("xts-key")
                 .help("encryption key2 required for AES_XTS")
-                .required_if_eq("cipher", "AES_XTS"),
+                .required_if_eq("cipher", "AES_XTS")
+                .required(false),
         );
 
     let destroy = Command::new("destroy")
--- a/io-engine/src/lvs/lvs_store.rs
+++ b/io-engine/src/lvs/lvs_store.rs
@@ -20,7 +20,10 @@ use url::Url;
 use super::{BsError, ImportErrorReason, Lvol, LvsError, LvsIter, PropName, PropValue};
 
 use crate::{
-    bdev::{uri, PtplFileOps},
+    bdev::{
+        crypto::{create_crypto_vbdev_on_base_bdev, destroy_crypto_vbdev},
+        uri, PtplFileOps,
+    },
     bdev_api::{bdev_destroy, BdevError},
     core::{
         logical_volume::LogicalVolume, snapshot::LvolSnapshotOps, Bdev, IoType, NvmfShareProps,
@@ -316,11 +319,19 @@ impl Lvs {
             name: args.name.clone(),
         })?;
 
+        // If we are requesting for an encrypted pool, then we should match existing pool(if any)
+        // by pool's bdev name as crypto bdev name.
+        let pool_bdev_name = if let Some(c) = args.crypto_vbdev_name.as_ref() {
+            c
+        } else {
+            &parsed.get_name()
+        };
+
         // At any point two pools with the same name should
         // not exists so returning error
         if let Some(pool) = Self::lookup(&args.name) {
             let pool_name = pool.base_bdev().name().to_string();
-            return if pool_name.as_str() == parsed.get_name() {
+            return if pool_name.as_str() == pool_bdev_name {
                 Err(LvsError::Import {
                     source: BsError::VolAlreadyExists {},
                     name: args.name.clone(),
@@ -353,7 +364,35 @@ impl Lvs {
             Ok(name) => Ok(name),
         }?;
 
-        let pool = Self::import(&args.name, &bdev).await?;
+        if let Some(ref cname) = args.crypto_vbdev_name {
+            // Only if bdev doesn't exist. e.g. direct import path.
+            if UntypedBdev::lookup_by_name(cname).is_none() {
+                if let Some(e) = args.enc_key {
+                    match create_crypto_vbdev_on_base_bdev(cname, &bdev, &e) {
+                        Ok(_) => {}
+                        Err(berr) => {
+                            let _ = parsed.destroy().await.map_err(|e| {
+                                    error!(
+                                        "failed to delete base_bdev {bdev} after failed crypto vbdev creation. {e:?}"
+                                    );
+                                });
+                            return Err(LvsError::PoolCreate {
+                                source: BsError::LvsCryptoVbdev {
+                                    source: match berr {
+                                        BdevError::CreateBdevFailed { source, .. } => source,
+                                        _ => Errno::EINVAL,
+                                    },
+                                },
+                                name: args.name.clone(),
+                            });
+                        }
+                    }
+                }
+            }
+        }
+
+        let bsdev_name = args.crypto_vbdev_name.as_ref().unwrap_or(&bdev);
+        let pool = Self::import(&args.name, bsdev_name).await?;
         // Try to destroy the pending snapshots without catching
         // the error.
         Lvol::destroy_pending_discarded_snapshot().await;
@@ -400,7 +439,11 @@ impl Lvs {
         let bdev = args.disks[0].clone();
 
         let pool_name = args.name.clone().into_cstring();
-        let bdev_name = bdev.into_cstring();
+        let bdev_name = if let Some(ref c) = args.crypto_vbdev_name {
+            c.clone().into_cstring()
+        } else {
+            bdev.into_cstring()
+        };
 
         let cluster_size = if let Some(cluster_size) = args.cluster_size {
             if cluster_size % ROUND_TO_MB == 0 {
@@ -496,8 +539,14 @@ impl Lvs {
         let disk = Self::parse_disk(args.disks.clone())?;
 
         info!(
-            "Creating or importing lvs '{}' from '{}'...",
-            args.name, disk
+            "Creating or importing {enc} lvs '{}' from '{}'...",
+            args.name,
+            disk,
+            enc = if args.crypto_vbdev_name.is_some() {
+                "encrypted"
+            } else {
+                "non-encrypted"
+            }
         );
 
         let bdev_ops = uri::parse(&disk).map_err(|e| LvsError::InvalidBdev {
@@ -505,8 +554,16 @@ impl Lvs {
             name: args.name.clone(),
         })?;
 
+        // If we are requesting for an encrypted pool, then we should lookup existing pool(if any)
+        // by pool's bdev name as crypto bdev name.
+        let pool_bdev_name = if let Some(c) = args.crypto_vbdev_name.as_ref() {
+            c
+        } else {
+            &bdev_ops.get_name()
+        };
+
         if let Some(pool) = Self::lookup(&args.name) {
-            return if pool.base_bdev().name() == bdev_ops.get_name() {
+            return if pool.base_bdev().name() == pool_bdev_name {
                 Err(LvsError::PoolCreate {
                     source: BsError::VolAlreadyExists {},
                     name: args.name.clone(),
@@ -518,7 +575,6 @@ impl Lvs {
                 })
             };
         }
-
         // Create the underlying ndev.
         let bdev_name = match bdev_ops.create().await {
             Err(e) => match e {
@@ -538,6 +594,31 @@ impl Lvs {
             Ok(name) => Ok(name),
         }?;
 
+        // Create crypto bdev now if required.
+        if let Some(ref cname) = args.crypto_vbdev_name {
+            if let Some(ref e) = args.enc_key {
+                match create_crypto_vbdev_on_base_bdev(cname, &bdev_name, e) {
+                    Ok(_) => {}
+                    Err(berr) => {
+                        let _ = bdev_ops.destroy().await.map_err(|e| {
+                                error!(
+                                    "failed to delete base_bdev {bdev_name} after failed crypto vbdev creation. {e:?}"
+                                );
+                            });
+                        return Err(LvsError::PoolCreate {
+                            source: BsError::LvsCryptoVbdev {
+                                source: match berr {
+                                    BdevError::CreateBdevFailed { source, .. } => source,
+                                    _ => Errno::EINVAL,
+                                },
+                            },
+                            name: args.name.clone(),
+                        });
+                    }
+                }
+            }
+        }
+
         match Self::import_from_args(args.clone()).await {
             Ok(pool) => Ok(pool),
             // try to create the pool
@@ -545,6 +626,7 @@ impl Lvs {
                 source: BsError::CannotImportLvs {},
                 ..
             }) => {
+                let cbdev_name: Option<String> = args.crypto_vbdev_name.clone();
                 match Self::create_from_args_inner(PoolArgs {
                     disks: vec![bdev_name.clone()],
                     ..args
@@ -552,6 +634,15 @@ impl Lvs {
                 .await
                 {
                     Err(create) => {
+                        // destroy crypto vbdev first.
+                        if let Some(c) = cbdev_name.as_ref() {
+                            let _ = destroy_crypto_vbdev(c.clone()).await.map_err(|e| {
+                                error!(
+                                    "failed to delete crypto vbdev {c} after failed pool creation. {e}"
+                                );
+                            });
+                        }
+
                         let _ = bdev_ops.destroy().await.map_err(|_e| {
                             // we failed to delete the base_bdev be loud about it
                             // there is not much we can do about it here, likely
@@ -664,7 +755,7 @@ impl Lvs {
         // when destroying a pool unshare all volumes
         self.unshare_all().await;
 
-        let base_bdev = self.base_bdev();
+        let mut base_bdev = self.base_bdev();
 
         let evt = self.event(EventAction::Delete);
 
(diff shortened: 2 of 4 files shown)


dsharma-dc:
This PR is built on previous open PR, hence the commits from previous one are included here. As part of this review only 36fb22ca6e3e5c1e72c9c09b5b052f7837c1bb45 needs to be reviewed.

io-engine/src/bdev/crypto.rs
```diff
@@ -0,0 +1,269 @@
...
+/// Representation of a crypto vbdev to be created in SPDK.
+/// TODO: Currently we don't fit crypto vbdev into the uri based bdev management
+/// as the information required to completely setup the crypto vbdev isn't part of the uri,
+/// and also the crypto uri isn't provided by user rather we'll have to do some masking
+/// using crypto:// scheme.
+#[allow(dead_code)]
+#[derive(Default)]
+pub(crate) struct Crypto {
+    // Name of the crypto vbdev.
+    pub name: String,
```
tiagolobocastro: should all be pub(crate)

io-engine/src/bdev/crypto.rs
```diff
@@ -0,0 +1,269 @@
...
+/// Cipher mode for the encryption enablement. The default is kept as AES_XTS here
+/// because operations are default assigned to software module, which only supports
+/// AES_XTS. We don't expect to use default though.
+pub enum Cipher {
+    AesCbc,
+    #[default]
+    AesXts,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
```
tiagolobocastro: Do we serialize this?

io-engine/src/bdev/crypto.rs
```diff
@@ -0,0 +1,269 @@
...
+    _AccelDpdkCryptodevUadk,
+    _AccelDpdkCryptodevLast,
+}
+
+impl Display for AccelDpdkCryptodevDriverType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{}",
+            match self {
```
tiagolobocastro: nit: can put on variable to reduce line count

io-engine/src/bdev/crypto.rs
```diff
@@ -0,0 +1,269 @@
...
+                Self::_AccelDpdkCryptodevLast => "crypto_driver_invalid",
+            }
+        )
+    }
+}
+
+// This function enables the dpdk_cryptodev module with aesni_mb driver in spdk,
+// and assigns the encrypt and decrypt operation to it, which by default are assigned to
+// software module initially.
+#[allow(dead_code)]
```
tiagolobocastro: is this not called?

io-engine/src/bdev/crypto.rs
```diff
@@ -0,0 +1,269 @@
...
+        accel_dpdk_cryptodev_enable();
+        let setdrv_ret = accel_dpdk_cryptodev_set_driver(
+            AccelDpdkCryptodevDriverType::AccelDpdkCryptodevAesniMb
+                .to_string()
+                .into_cstring()
+                .as_ptr(),
+        );
+        if setdrv_ret != 0 {
+            return Err(CoreError::InitCryptoModule {
+                reason: format!("Unsupported driver: {:?}", setdrv_ret),
```
```suggestion
                reason: format!("Unsupported driver: {setdrv_ret:?}"),
```

io-engine/src/bdev/crypto.rs
```diff
@@ -0,0 +1,269 @@
...
+            });
+        }
+        let err_enc = spdk_accel_assign_opc(8, "dpdk_cryptodev".into_cstring().as_ptr());
+        let err_dec = spdk_accel_assign_opc(9, "dpdk_cryptodev".into_cstring().as_ptr());
+        info!("assign opc err_enc: {err_enc}, err_dec: {err_dec}");
```
tiagolobocastro: do we need this as info?

io-engine/src/bdev/crypto.rs
```diff
@@ -0,0 +1,269 @@
...
+    unsafe {
+        let ret = spdk_accel_crypto_key_create(&create_key_params);
+        if ret != 0 {
+            Err(BdevError::CreateBdevFailed {
+                source: Errno::from_raw(ret.abs()),
+                name: format!("Key create failed: {:?}", create_key_params.key_name),
+            })
+        } else {
+            let key_ptr = spdk_accel_crypto_key_get(create_key_params.key_name);
+            trace!("Key '{:?}' created.", key_params.key_name);
```
```suggestion
            trace!("Key '{:?}' created", key_params.key_name);
```

io-engine/src/bdev/lvs.rs
```diff
@@ -209,6 +214,9 @@ impl Lvs {
             cluster_size: None,
             md_args: None,
             backend: PoolBackend::Lvs,
+            enc_key: self.key.clone(),
+            // XXX: Is this path ever exercised apart from test, or casperf perhaps?
```
tiagolobocastro: you mean no one reads this name?

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
```diff
@@ -54,6 +54,27 @@ pub fn subcommands() -> Command {
...
+        .arg(
+            Arg::new("encryption-key")
+                .short('k')
+                .long("encryption-key")
+                .help("The encryption key of the pool in hexlified format")
+                .required(false),
+        )
+        .arg(
+            Arg::new("xts-key")
+                .short('e')
```
tiagolobocastro: what does e stand for?

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
```diff
@@ -251,6 +313,23 @@ async fn create(mut ctx: Context, matches: &ArgMatches) -> crate::Result<()> {
         None => None,
     };
 
+    let enc_key_msg = enc_key.map(|k| v1rpc::common::EncryptionKey {
+        key_name: "key_".to_owned() + k.as_str(),
```
tiagolobocastro: why are we adding the prefix?

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
```diff
@@ -251,6 +313,23 @@ async fn create(mut ctx: Context, matches: &ArgMatches) -> crate::Result<()> {
...
+        key: k.clone().into(),
+        key_length: (k.len() * 4) as u32,
+        key2: xts_key.as_ref().map(|k2| k2.clone().into()),
+        key2_length: xts_key.map(|x| { x.len() * 4 } as u32),
+    });
+
+    let enc_msg = enc_key_msg.map(|e| {
+        v1rpc::common::create_pool_request::Encryption::Data(v1rpc::common::EncryptionData {
+            cipher: v1rpc::common::Cipher::from_str_name(&cipher)
+                .unwrap()
```
tiagolobocastro:
```suggestion
                ?
```

io-engine/src/core/mod.rs
```diff
@@ -318,6 +322,7 @@ impl ToErrno for CoreError {
             Self::Ptpl { .. } => Errno::EIO,
             Self::SnapshotCreate { source, .. } => source,
             Self::WipeFailed { .. } => Errno::EIO,
+            Self::InitCryptoModule { .. } => Errno::EINVAL,
```
tiagolobocastro: Does einval always make sense here? Is it always bad input?

tiagolobocastro (commented):
Haven't finished

io-engine/src/bdev/crypto.rs
dsharma-dc: > should all be pub(crate)

For now I'm not using this. Will modify when required.

io-engine/src/bdev/crypto.rs
dsharma-dc: > is this not called?

It'll be once that ipsec_mb issue is resolved and we start enabling dpdk_cryptodev drivers.

io-engine/src/bdev/crypto.rs
dsharma-dc: > do we need this as info?

converted to trace

io-engine/src/core/mod.rs
dsharma-dc: > Does einval always make sense here? Is it always bad input?

Yeah, the cases where we might get this as per code today is e.g. trying to assign encrypt/decrypt ops to dpdk module without enabling it, or say enabling a wrong dpdk driver.

io-engine/src/bdev/crypto.rs
dsharma-dc: > Do we serialize this?

No, removed these derives. 
btw, please see this. Might be clearer to review changes in respective PRs.
https://github.com/openebs/mayastor/pull/1827#issuecomment-2702928221

io-engine/src/bdev/lvs.rs
dsharma-dc: > you mean no one reads this name?

I mean this `create` path. I see only `CreateDestroy` of Lvol is directly calling this, but I don't think that's the usual prod path?

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
dsharma-dc: > what does e stand for?

Taken from spdk's shorthands :) extra key perhaps.

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
dsharma-dc: > why are we adding the prefix?

Just to give the key some name as the key create call needs a name.

Abhinandan-Purkait (approved):

io-engine/src/lvs/lvs_store.rs
```diff
@@ -347,7 +358,30 @@ impl Lvs {
             Ok(name) => Ok(name),
         }?;
 
-        let pool = Self::import(&args.name, &bdev).await?;
+        if let Some(ref cname) = args.crypto_vbdev_name {
+            // Only if bdev doesn't exist. e.g. direct import path.
+            if UntypedBdev::lookup_by_name(cname).is_none() {
```
tiagolobocastro, line 369: Should we check if the cname refers to our crypto bdev?

io-engine/src/lvs/lvs_store.rs
```diff
@@ -347,7 +358,30 @@ impl Lvs {
...
+                if let Some(e) = args.enc_key {
+                    match create_crypto_vbdev_on_base_bdev(cname, &bdev, &e) {
+                        Ok(_) => {}
+                        Err(_) => {
+                            let _ = parsed.destroy().await.map_err(|_e| {
+                                    error!(
+                                        "failed to delete base_bdev {bdev} after failed crypto vbdev creation"
```
tiagolobocastro: can we place the error in the error string here?

io-engine/src/lvs/lvs_store.rs
```diff
@@ -347,7 +358,30 @@ impl Lvs {
...
+                                    );
+                                });
+                            return Err(LvsError::PoolCreate {
+                                source: BsError::LvsCryptoVbdev {},
```
tiagolobocastro: Include error from `create_crypto_vbdev_on_base_bdev`

io-engine/src/lvs/lvs_store.rs
```diff
@@ -347,7 +358,30 @@ impl Lvs {
...
+                                name: args.name.clone(),
+                            });
+                        }
+                    }
+                }
+            }
+        }
+
+        let bsdev_name = args.crypto_vbdev_name.as_ref().unwrap_or(&bdev);
+        let pool = Self::import(&args.name, bsdev_name).await?;
```
tiagolobocastro, line 395: Do we need to delete up the crypto bdev here if import fails?
Also wondering, do we also need to delete the disk bdev?

io-engine/src/lvs/lvs_store.rs
```diff
@@ -532,20 +583,50 @@ impl Lvs {
             Ok(name) => Ok(name),
         }?;
 
+        // Create crypto bdev now if required.
+        if let Some(ref cname) = args.crypto_vbdev_name {
+            if let Some(ref e) = args.enc_key {
+                match create_crypto_vbdev_on_base_bdev(cname, &bdev_name, e) {
+                    Ok(_) => {}
+                    Err(_) => {
```
tiagolobocastro: same comments here regarding keeping the errors?

io-engine/src/lvs/lvs_store.rs
```diff
@@ -671,16 +752,42 @@ impl Lvs {
...
+        // If the base_bdev is a crypto vbdev then we need to destroy both - the crypto vbdev and it's base.
+        if base_bdev.driver() == "crypto" {
+            let cbdev = base_bdev.crypto_base_bdev();
+
+            let _ = destroy_crypto_vbdev(base_bdev.name().to_string())
+                .await
+                .map_err(|e| LvsError::Destroy {
+                    source: e,
+                    name: self_str.clone(),
+                });
```
tiagolobocastro, line 789: we're mapping the error here, but then not doing anything with it?

dsharma-dc: Yeah, we wouldn't want to bail out without attempting base bdev deletion so kept it like this. Perhaps we can spit an error msg like other existing paths where we just emit an error saying we couldn't delete bdev?

io-engine/src/lvs/lvs_store.rs
dsharma-dc: > Do we need to delete up the crypto bdev here if import fails? Also wondering, do we also need to delete the disk bdev?

We can let it be I suggest. Firstly, the check otherwise would be error specific because we treat `CannotImportLvs` error which is usual as a green signal to go ahead and create pool.
Secondly, if it is some other error we bubble it up and a possible retry should work without needing to recreate bdevs.

io-engine/src/lvs/lvs_store.rs
dsharma-dc: done

io-engine/src/lvs/lvs_store.rs
dsharma-dc: > Should we check if the cname refers to our crypto bdev?

Wdym by `our crypto bdev` here? Before we're here we have already done the pool name existence.

io-engine/src/lvs/lvs_store.rs
dsharma-dc: done

io-engine/src/lvs/lvs_store.rs
dsharma-dc: > can we place the error in the error string here?

done

dsharma-dc:
bors try

dsharma-dc:
bors try-

io-engine/src/lvs/lvs_store.rs
tiagolobocastro: I mean, we're just checking if a bdev named as `cname` exists, should we check if it's created with our parameters?

io-engine/src/lvs/lvs_store.rs
tiagolobocastro: This may cause some issues if we ever to do things such as move storage across nodes, and leave these bdevs open here (ex: for cloud use cases).
But since we already have this issue for non-encrypted pools, maybe we can create separate ticket to handle this.

io-engine/src/lvs/lvs_store.rs
tiagolobocastro: But if we don't want to do anything, then also there's no point in doing the mapping, it's a no-op?
Also, if we can't delete the crypto bdev, then we won't be able to delete the disk bdev because it's still in-use?

io-engine/src/lvs/lvs_store.rs
dsharma-dc: > I mean, we're just checking if a bdev named as `cname` exists, should we check if it's created with our parameters?

I don't think we need to do that. SPDK guarantees unique bdev names. We derive crypto vbdev name from pool name like `crypto_<mypoolname>`. So if the pool name `mypoolname` doesn't exist it's unlikely that such a crypto vbdev name will exist(unless some cleanup left behind).

tiagolobocastro: Seems a bit risky, especially considering we don't always clean up the bdevs as seen on another comment thread.
Let's say pool creation fails or whatever, and crypto bdev is left.
Then user tries to create pool with the same name but a different encryption key. What happens if we reuse the old crypto bdev?

io-engine/src/lvs/lvs_store.rs
dsharma-dc: > But if we don't want to do anything, then also there's no point in doing the mapping, it's a no-op? 

Yes, we can skip mapping altogether.

> Also, if we can't delete the crypto bdev, then we won't be able to delete the disk bdev because it's still in-use?

In a normal case, if a base bdev is directly deleted then the associated crypto vbdev layered on it also gets deleted. So if for some reason deleting crypto vbdev first failed, attempting base bdev deletion is another chance to succeed deleting crypto bdev.

io-engine/src/lvs/lvs_store.rs
dsharma-dc: I can think of few cases based on comment:

1. **Pool creation fails, crypto bdev left behind** -  Create pool with same name and different key will fail because crypto bdev name exists. There is no blobstore also anyway because pool creation failed.
2. **crypto bdev left behind during pool destroy** - If we have attempted crypto bdev deletion it means blobstore is already removed/deinitialized, so no risk of corrupting blobstore.
3. **blobstore deinit itself failed** - In this case other validations should kick in as they hold today.(barring the case that we discussed offline about directly using disk)

tiagolobocastro: So the blobstore doesn't exist, but a creation attempt of the same pool with a different secret would end up reusing the previous crypto bdev which would result in an unimportable pool?

dsharma-dc: As discussed offline, I've created a separate task to look into practical details of this scenario because when we validate against key params we'll have to interact with spdk to get currently attached key object details.

tiagolobocastro (approved):

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
dsharma-dc: It doesn't allow to use `?` here
```
cannot use the `?` operator in a closure that returns `io_engine_api::v1::common::create_pool_request::Encryption`
```

dsharma-dc:
bors try

dsharma-dc:
bors merge

dsharma-dc:
bors cancel

dsharma-dc:
bors merge

dsharma-dc:
bors merge

PR #1833, opened 2025-03-12
feat(encryption): fetch secret params from secret source

dsharma-dc:
When no encryption parameters present create pool request => non-encrypted pool created. no parsing required.

When encryption parameters present source in create/import pool request:
1. if source file/Secret not found - create/import fails.
2. if file/Secret found but data has parsing issues -> create/import fails.
3. if file/Secret parsed successfully -> create/import succeeds.

Basically if a secret source is provided(either a file or a K8s secret), we should always fail if there is any trouble getting params.

### DIFF
--- a/io-engine/Cargo.toml
+++ b/io-engine/Cargo.toml
@@ -114,6 +114,7 @@ sysfs = { path = "../sysfs" }
 version-info = { path = "../utils/dependencies/version-info", features = ["deps-logs-head"] }
 events-api = { path = "../utils/dependencies/apis/events" }
 event-publisher = { path = "../utils/dependencies/event-publisher" }
+secret-provider = { path = "../utils/dependencies/secret-provider" }
 tracing-filter = { path = "../utils/dependencies/tracing-filter" }
 
 [dependencies.serde]
--- a/io-engine/src/grpc/v1/pool.rs
+++ b/io-engine/src/grpc/v1/pool.rs
@@ -17,6 +17,7 @@ use io_engine_api::v1::{
     replica::destroy_replica_request,
     snapshot::destroy_snapshot_request,
 };
+use secret_provider::secret_data;
 use std::{
     convert::{TryFrom, TryInto},
     fmt::Debug,
@@ -24,6 +25,15 @@ use std::{
     panic::AssertUnwindSafe,
 };
 use tonic::{Request, Status};
+
+pub type PoolCreateEncryptionParams = create_pool_request::Encryption;
+pub type PoolImportEncryptionParams = import_pool_request::Encryption;
+
+enum PoolEncryptionParams {
+    Create(PoolCreateEncryptionParams),
+    Import(PoolImportEncryptionParams),
+    NoEncryptionParams,
+}
 
 impl From<DestroyPoolRequest> for FindPoolArgs {
     fn from(value: DestroyPoolRequest) -> Self {
@@ -55,6 +65,48 @@ impl From<GrowPoolRequest> for FindPoolArgs {
     fn from(value: GrowPoolRequest) -> Self {
         Self::name_uuid(value.name, &value.uuid)
     }
+}
+
+/// Helper routine to extract Encryption params from the Create or Import pool request.
+async fn util_fetch_secret_params(
+    params: &PoolEncryptionParams,
+) -> Result<Option<PoolEncKey>, Status> {
+    let enc_key = match params {
+        PoolEncryptionParams::Create(enc_arg) => {
+            match enc_arg.clone() {
+                // Would have been nice to deduplicate the code if we could call secret_data using a
+                // trait object returned by library. But it doesn't seem to work here because the
+                // library call having generic type in definition isn't allowing that.
+                create_pool_request::Encryption::Data(ckd) => Some(
+                    ckd.try_into()
+                        .map_err(|e: LvsError| Status::invalid_argument(e.to_string()))?,
+                ),
+                create_pool_request::Encryption::Secret(cks) => {
+                    let secret_params = secret_data(cks.secret.as_str())
+                        .await
+                        .map_err(|e| Status::invalid_argument(e.to_string()))?;
+                    trace!("[create pool] Received encryption params: {secret_params:?}");
+                    Some(secret_params)
+                }
+            }
+        }
+        PoolEncryptionParams::Import(enc_arg) => match enc_arg.clone() {
+            import_pool_request::Encryption::Data(ikd) => Some(
+                ikd.try_into()
+                    .map_err(|e: LvsError| Status::invalid_argument(e.to_string()))?,
+            ),
+            import_pool_request::Encryption::Secret(iks) => {
+                let secret_params = secret_data(iks.secret.as_str())
+                    .await
+                    .map_err(|e| Status::invalid_argument(e.to_string()))?;
+                trace!("[import pool] Received encryption params: {secret_params:?}");
+                Some(secret_params)
+            }
+        },
+        PoolEncryptionParams::NoEncryptionParams => None,
+    };
+
+    Ok(enc_key)
 }
 
 /// RPC service for mayastor pool operations
@@ -189,12 +241,6 @@ impl TryFrom<CreatePoolRequest> for PoolArgs {
             }
         }
 
-        let enc_key = match args.encryption.clone() {
-            Some(create_pool_request::Encryption::Data(kd)) => kd.try_into().ok(),
-            Some(create_pool_request::Encryption::Secret(_ks)) => None,
-            _ => None,
-        };
-
         Ok(Self {
             name: args.name.clone(),
             disks: args.disks.clone(),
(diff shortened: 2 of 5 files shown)


dsharma-dc:
bors try

io-engine/src/grpc/v1/pool.rs
```diff
@@ -531,11 +661,16 @@ impl PoolRpc for PoolService {
                 crate::spdk_submit!(async move {
                     info!("{:?}", request.get_ref());
 
-                    let factory =
-                        GrpcPoolFactory::new(PoolBackend::try_from(request.get_ref().pooltype)?)?;
-                    factory
-                        .create(PoolArgs::try_from(request.into_inner())?)
-                        .await
+                    let req = request.into_inner();
```
tiagolobocastro: Could we get the encryption parameters completely outside of the spdk reactors, before calling self.locked? Then that would simplify the functions above? Wdyt?

dsharma-dc: > Could we get the encryption parameters completely outside of the spdk reactors, before calling self.locked? Then that would simplify the functions above? Wdyt?

Possibly. I'll have to try that. In that case we'll still have to do at least a async tokio task spawn? or are we already running in a tokio runtime before doing self.locked?

tiagolobocastro: No, we're already running in tokio, so would simplify things a lot by removing the trampolines

dsharma-dc: I've done this change in a separate commit now. ptal.
Since we're just fetching params, I wouldn't think we should be concerned about any race here outside lock?

io-engine/src/grpc/v1/pool.rs
```diff
@@ -57,6 +68,85 @@ impl From<GrowPoolRequest> for FindPoolArgs {
     }
 }
 
+/// Parse and return encryption parameters from a Kubernetes secret or a file.
+async fn _params_get_from_secret_source(secret_source_name: &str) -> Result<PoolEncKey, Status> {
+    let secret_params = match current_platform_type() {
+        PlatformType::K8s => {
+            let client = kube::Client::try_default()
+                .await
+                .expect("Should be able to create kube client");
```
tiagolobocastro: Should return an error

io-engine/src/grpc/v1/pool.rs
```diff
@@ -57,6 +68,85 @@ impl From<GrowPoolRequest> for FindPoolArgs {
     }
 }
 
+/// Parse and return encryption parameters from a Kubernetes secret or a file.
+async fn _params_get_from_secret_source(secret_source_name: &str) -> Result<PoolEncKey, Status> {
```
tiagolobocastro: This should also be placed on a dependencies crate

io-engine/src/grpc/v1/pool.rs
```diff
@@ -57,6 +68,85 @@ impl From<GrowPoolRequest> for FindPoolArgs {
...
+}
+
+/// Helper routine to extract Encryption params from the Create or Import pool request.
+async fn util_fetch_secret_params(
+    params: &PoolEncryptionParams,
+) -> Result<Option<PoolEncKey>, Status> {
+    let enc_key = match params {
+        PoolEncryptionParams::Create(enc_arg) => {
+            match enc_arg.clone() {
+                // Would have been nice to deduplicate the code if we could call secret_data using a
```
tiagolobocastro, line 77: Isn't the problem here that we defined the create and import types separately, rather than reusing? (I mean in the protobuf files)

```protobuf
oneof encryption {
    EncryptionData data = 5;
    EncryptionSecret secret = 6;
  }
```

If we put the oneof on a spearate definition, then we can reuse it?

io-engine/src/grpc/v1/pool.rs
tiagolobocastro: hmm yeah I think should not be a concern this the grpc should not be modifying the secrets themselves.

io-engine/src/grpc/v1/pool.rs
dsharma-dc: > If we put the oneof on a spearate definition, then we can reuse it?

We had noticed this and tried to put this into common proto. But then, that requires a type to be defined for use as oneof. So that was resulting into again some nested types and complex handling.

io-engine/src/grpc/v1/pool.rs
dsharma-dc: > This should also be placed on a dependencies crate

I can do that in a separate PR as it will need to be made generic based, currently it's tied to PoolEncKey type.

io-engine/src/grpc/v1/pool.rs
dsharma-dc: > Should return an error

done

dsharma-dc:
bors try

io-engine/src/grpc/v1/pool.rs
tiagolobocastro: I guess it depends where we want to pay the price. This way we're paying it here by duplicating since we now have 2 different types for the same thing essentially.

io-engine/src/grpc/v1/pool.rs
tiagolobocastro: I would rather to that first, and not have this repo know anything about "PlatformDeployer" or even "kube" crate

Abhinandan-Purkait: I think that's a fair point, let's extend the secret provider to handle the platform bits as well, unless we want more granular control on various platform.

dsharma-dc: The function is now moved to the secret-provider crate in dependencies, and usage is updated in this PR.
https://github.com/openebs/mayastor-dependencies/pull/119

Abhinandan-Purkait (approved):

io-engine/Cargo.toml
```diff
@@ -114,6 +114,8 @@ sysfs = { path = "../sysfs" }
 version-info = { path = "../utils/dependencies/version-info", features = ["deps-logs-head"] }
 events-api = { path = "../utils/dependencies/apis/events" }
 event-publisher = { path = "../utils/dependencies/event-publisher" }
+kube = { version = "0.94.2", features = ["derive", "runtime"] }
```
dsharma-dc: ah, this must be moved to secret-provider's Cargo.toml

dsharma-dc:
bors try

io-engine/src/grpc/v1/pool.rs
```diff
@@ -57,6 +67,42 @@ impl From<GrowPoolRequest> for FindPoolArgs {
...
+                    let secret_params = secret_data(cks.secret.as_str())
+                        .await
+                        .map_err(|e| Status::invalid_argument(e.to_string()))?;
+                    trace!("[create pool] Received encryption params: {secret_params:?}");
+                    Some(secret_params)
+                }
+            }
+        }
+        PoolEncryptionParams::Import(enc_arg) => match enc_arg.clone() {
+            import_pool_request::Encryption::Data(ikd) => ikd.try_into().ok(),
```
tiagolobocastro: hmm should these try_into be failed rather than ok'ed?
I might be misunderstanding this, but if we ok() does it mean we end up creating the pool without encryption rather than fail with invalid args?

dsharma-dc: > hmm should these try_into be failed rather than ok'ed? I might be misunderstanding this, but if we ok() does it mean we end up creating the pool without encryption rather than fail with invalid args?

Good catch indeed. I made to sure never return a `None` in case of Encryption::Secret, but this was also meant to be changed. It would've impacted io-engine-client only though.
Fixed now.

io-engine/src/grpc/v1/pool.rs
```diff
@@ -57,6 +67,48 @@ impl From<GrowPoolRequest> for FindPoolArgs {
...
+) -> Result<Option<PoolEncKey>, Status> {
+    let enc_key = match params {
+        PoolEncryptionParams::Create(enc_arg) => {
+            match enc_arg.clone() {
+                // Would have been nice to deduplicate the code if we could call secret_data using a
+                // trait object returned by library. But it doesn't seem to work here because the
+                // library call having generic type in definition isn't allowing that.
+                create_pool_request::Encryption::Data(ckd) => {
+                    Some(ckd.try_into().map_err(|_| {
+                        Status::invalid_argument("Couldn't parse raw secret params".to_string())
```
tiagolobocastro: Is the returned error from try_into useful, if so we could either use it fully or make it mart of the message

```suggestion
                    Some(ckd.try_into().map_err(|error| {
                        Status::invalid_argument(error.to_string())
```

io-engine/src/grpc/v1/pool.rs
```diff
@@ -525,17 +568,28 @@ impl GrpcPoolFactory {
...
                     info!("{:?}", request.get_ref());
-
                     let factory =
                         GrpcPoolFactory::new(PoolBackend::try_from(request.get_ref().pooltype)?)?;
-                    factory
-                        .create(PoolArgs::try_from(request.into_inner())?)
-                        .await
+                    let mut pool_args = PoolArgs::try_from(request.into_inner())?;
+
+                    pool_args.crypto_vbdev_name = enc_key
```
tiagolobocastro: Rather than making args mut here, and duplicating the crypto_ format we can add an with_encryption method to PoolArgs, something like this:

factory.create(PoolArgs::try_from(request.into_inner())?.with_encryption(enc_key))

tiagolobocastro (approved):

io-engine/src/grpc/v1/pool.rs
dsharma-dc: done.

dsharma-dc:
bors merge

PR #1837, opened 2025-03-17
chore: expose diskpool encryption feature for Mayastor

dsharma-dc:


Abhinandan-Purkait (approved):

io-engine/src/bin/io-engine.rs
```diff
@@ -97,6 +97,13 @@ fn start_tokio_runtime(args: &MayastorCliArgs) {
         warn!("RDMA is requested to be enabled for Mayastor NVMEoF target");
     }
 
+    // Use as-is basis for diskpool encryption since the spdk build is enabled
+    // --with-crypto.
+    // TODO: Once dpdk crypto modules are supported, we'll provide a switch
+    // later on to control the behaviour of choosing to use accel_sw based
```
tiagolobocastro, line 103: hmm interesting, should we expose the encryption algorithms as well?

tiagolobocastro (approved):

io-engine/src/bin/io-engine.rs
dsharma-dc: > hmm interesting, should we expose the encryption algorithms as well?

Not many choices there to expose AFAIK. accel_sw only supports AES_XTS-128. With dpdk modules, only AES_CBC-128 and AES_CBC-256.

dsharma-dc:
bors merge

PR #1840, opened 2025-03-21
Fix IPv6 support

tiagolobocastro:
fix(tgt-iface): reject scope ids on ipv6
    
    Start to add support for scope ids on ipv6.
    This is not yet supported on the nvmf target, I suspect because there's
    a bug on the socket interface which doesn't return the scope id as part
    of the traddr, though this is mostly a guess so far.
    As such, for the moment we detect but reject scope ids.
    
---

    fix(tgt-iface): support multiple ips per adapter
    
    In case adapter has multiple ipv6 ips, then the libc function
    getifaddrs returns multiple elements with the same name and so
    putting them on a map was actually erasing some ips!
    
    However, this now means we need to handle trying to pick one
    of the ips!
    For now we prioritize ipv4 vs 6 IF the grpc is also ipv4.
    Also we prefer ipv6 unique local, link local and !multicast.
    
---

    fix(tgt-iface): support ipv6 on tgt-iface
    
    The tgt-iface allows us a different way of specifying which interface,
    subnet or ip to bind the nvmf target to.
    This was also hardcoded to work with ipv4 mostly.
    
---

    fix(ipv6): support ipv6 for grpc and nvmf target
    
    In a few places ipv4 was still pretty much expected, leading into a
    few different crashes.
    Now we try to parse always as std::net::IpAddr which is an enum variant
    that can be either ipv4 or ipv6.

### DIFF
--- a/io-engine/src/bin/io-engine-client/context.rs
+++ b/io-engine/src/bin/io-engine-client/context.rs
@@ -135,7 +135,10 @@ impl Context {
             .unwrap_or('b');
         // Ensure the provided host is defaulted & normalized to what we expect.
         let host = if let Some(host) = matches.get_one::<String>("bind") {
-            let uri = host.parse::<Uri>().context(InvalidUri)?;
+            let uri = match host.parse::<Uri>().context(InvalidUri) {
+                Ok(uri) => Ok(uri),
+                Err(error) => format!("[{host}]").parse::<Uri>().map_err(|_| error),
+            }?;
             let mut parts = uri.into_parts();
             if parts.scheme.is_none() {
                 parts.scheme = Scheme::from_str("http").ok();
--- a/io-engine/src/core/env.rs
+++ b/io-engine/src/core/env.rs
@@ -817,38 +817,52 @@ impl MayastorEnvironment {
             None => ("name", iface),
         };
 
-        let pred: Box<dyn Fn(&nic::Interface) -> bool> = match cls {
-            "name" => Box::new(|n| n.name == name),
+        let map_ok = |filter: bool, n: nic::Interface| -> Option<nic::Interface> {
+            if filter {
+                Some(n)
+            } else {
+                None
+            }
+        };
+
+        let pred: Box<dyn FnMut(nic::Interface) -> Option<nic::Interface>> = match cls {
+            "name" => Box::new(|n| map_ok(n.name == name, n)),
             "mac" => {
                 let mac = Some(name.parse::<nic::MacAddr>()?);
-                Box::new(move |n| n.mac == mac)
+                Box::new(move |n| map_ok(n.mac == mac, n))
             }
             "ip" => {
                 let addr = nic::parse_ip(name)?;
-                Box::new(move |n| n.ip_match(addr))
+                Box::new(move |n| n.matched_ip_kind(addr))
             }
             "subnet" => {
                 let (subnet, mask) = nic::parse_ip_subnet(name)?;
-                Box::new(move |n| n.ip_subnet_eq(subnet, mask))
+                Box::new(move |n| n.matched_subnet_kind(subnet, mask))
             }
             _ => {
                 return Err(format!("Invalid NVMF target interface: '{iface}'"));
             }
         };
 
-        let mut nics: Vec<_> = nic::find_all_nics().into_iter().filter(pred).collect();
+        let mut nics: Vec<_> = nic::find_all_nics().into_iter().filter_map(pred).collect();
 
-        let res = match nics.pop() {
+        // In case of matching by name, mac or subnet, the adapter may have multiple ips, and
+        // in this case how to prioritize them?
+        // For now, we can at least match ipv4 or ipv6 to match the grpc.
+        let env = MayastorEnvironment::global_or_default();
+        let grpc = env.grpc_endpoint.expect("Should always be set");
+        nics.sort_by(|a, b| a.sort(b, grpc.is_ipv4()));
+
+        let res: nic::Interface = match nics.pop() {
             None => Err(format!("Network interface matching '{iface}' not found")),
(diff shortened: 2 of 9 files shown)


Abhinandan-Purkait (approved):

dhess:
Thanks for this! Will the Helm chart need any corresponding fixes?

tiagolobocastro:
> Thanks for this! Will the Helm chart need any corresponding fixes?

No, should just work I think.
Here's this PR running on an ipv6 kind cluster:

```
❯ kubectl-mayastor get nodes
 ID           GRPC ENDPOINT                  STATUS  VERSION
 kind-worker  [fc00:f853:ccd:e793::3]:10124  Online  v1.0.0-1072-g4af708e36958
```

tiagolobocastro:
There's another issue when publishing volume, an extra 0 is being added:
` deviceUri: nvmf://[fc00:f853:ccd:e793::3]:0:8420/nqn.2019-05.io.openebs:880bc436-cca4-41d6-8c77-d481091389d9?hostnqn=nqn.2019-05.io.openebs:node-name:kind-worker`

`[fc00:f853:ccd:e793::3]:0:8420` should be `[fc00:f853:ccd:e793::3]:8420`

tiagolobocastro:
Ok, scope id's are not quite working on the nvmf target, I suspect the socket is simply not returning the correctly formatted string with the %.
Anyhow for now we have to reject scope ids.

Other than that things mostly work at this layer. 
~~Then we fail on the control-plane which doesn't seem to handle ipv6 on the nvmf connection either, so that's needs fixing but on a different repo.~~
~~EDIT: Nevermind, one last bug here, need to format ipv6 socket with the brackets...~~
EDIT2: There's indeed a small bug on the nvmf library we use, I'll raise PRs for that too

bors try

tiagolobocastro:
With another change on control-plane, this now works fully:

```
nvme-subsys3 - NQN=nqn.2019-05.io.openebs:880bc436-cca4-41d6-8c77-d481091389d9
               hostnqn=nqn.2019-05.io.openebs:node-name:kind-worker
               iopolicy=numa
\
 +- nvme3 tcp traddr=fc00:f853:ccd:e793::3,trsvcid=8420,src_addr=fc00:f853:ccd:e793::3 live
```

io-engine/src/bin/io-engine-client/context.rs
```diff
@@ -135,7 +135,10 @@ impl Context {
             .unwrap_or('b');
         // Ensure the provided host is defaulted & normalized to what we expect.
         let host = if let Some(host) = matches.get_one::<String>("bind") {
-            let uri = host.parse::<Uri>().context(InvalidUri)?;
+            let uri = match host.parse::<Uri>().context(InvalidUri) {
+                Ok(uri) => Ok(uri),
+                Err(error) => format!("[{host}]").parse::<Uri>().map_err(|_| error),
```
dsharma-dc, line 140: Should this formatting change here since you removed `[]` from host returned by Url crate?

io-engine/src/core/nic.rs
```diff
@@ -128,33 +186,135 @@ impl Interface {
...
+            if let Some(socket) = self.inet6.addr {
+                return Some(SIpAddr::from(socket));
+            }
+        } else {
+            if let Some(socket) = self.inet6.addr {
+                return Some(SIpAddr::from(socket));
+            }
+            if let Some(socket) = self.inet.addr {
+                return Some(SIpAddr::from(socket));
+            }
```
dsharma-dc: Can remove code duplication here perhaps, something like:

```
    let (first, second) = if prefer_ipv4 {
        (&self.inet.addr, &self.inet6.addr)
    } else {
        (&self.inet6.addr, &self.inet.addr)
    };

    first.or_else(|| second).map(SIpAddr::from)
```

io-engine/src/core/env.rs
```diff
@@ -809,86 +834,90 @@ impl MayastorEnvironment {
...
-            ));
+            )),
         }
+    }
 
-        Ok(res.inet.addr.unwrap().to_string())
+    fn prefer_ipv4(grpc: std::net::SocketAddr) -> bool {
+        match grpc {
+            std::net::SocketAddr::V4(_) => true,
+            std::net::SocketAddr::V6(socket) if socket.ip().is_unspecified() => true,
```
dsharma-dc, line 899: Isn't this case of ipv6 similar to ipv4 with 0.0.0.0? What's the reason we want to prefer v4 here?

tiagolobocastro: This is to keep existing behavior, otherwise we might start using ipv6 addresses for nvmf whereas today we're using ipv4

io-engine/src/core/nic.rs
tiagolobocastro: indeed well spotted, done

tiagolobocastro: It failed to build actually, because the types are different, so had to do this:
```rust
let (first, second) = if prefer_ipv4 {
            (
                self.inet.addr.map(Into::into),
                self.inet6.addr.map(Into::into),
            )
        } else {
            (
                self.inet6.addr.map(Into::into),
                self.inet.addr.map(Into::into),
            )
        };

        first.or(second)
```

io-engine/src/bin/io-engine-client/context.rs
tiagolobocastro: That's because the Uri type requires you to use the [ ] when parsing even if no port is present.

dsharma-dc (approved):

tiagolobocastro:
bors merge

dhess:
Thanks! Will this fix be included in the next release?

tiagolobocastro:
> Thanks! Will this fix be included in the next release?

Yes will be included in the next release, perhaps in a patch v2.8.1

PR #1843, opened 2025-03-25
Cherry-pick 1840

tiagolobocastro:
fix(tgt-iface): reject scope ids on ipv6
    
    Start to add support for scope ids on ipv6.
    This is not yet supported on the nvmf target, I suspect because there's
    a bug on the socket interface which doesn't return the scope id as part
    of the traddr, though this is mostly a guess so far.
    As such, for the moment we detect but reject scope ids.
    
---

    fix(tgt-iface): support multiple ips per adapter
    
    In case adapter has multiple ipv6 ips, then the libc function
    getifaddrs returns multiple elements with the same name and so
    putting them on a map was actually erasing some ips!
    
    However, this now means we need to handle trying to pick one
    of the ips!
    For now we prioritize ipv4 vs 6 IF the grpc is also ipv4.
    Also we prefer ipv6 unique local, link local and !multicast.
    
---

    fix(tgt-iface): support ipv6 on tgt-iface
    
    The tgt-iface allows us a different way of specifying which interface,
    subnet or ip to bind the nvmf target to.
    This was also hardcoded to work with ipv4 mostly.
    
---

    fix(ipv6): support ipv6 for grpc and nvmf target
    
    In a few places ipv4 was still pretty much expected, leading into a
    few different crashes.
    Now we try to parse always as std::net::IpAddr which is an enum variant
    that can be either ipv4 or ipv6.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -641,9 +641,9 @@ checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
 
 [[package]]
 name = "cc"
-version = "1.2.1"
+version = "1.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fd9de9f2205d5ef3fd67e685b0df337994ddd4495e2a28d185500d0e1edfea47"
+checksum = "1fcb57c740ae1daf453ae85f16e37396f672b039e00d9d866e07ddb24e328e3a"
 dependencies = [
  "jobserver",
  "libc",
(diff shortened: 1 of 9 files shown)


dsharma-dc (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
ah, dependencies still needs aproving...
bors cancel

tiagolobocastro:
bors merge

PR #1846, opened 2025-04-02
fix(lvs/export): destroy bdevs correctly considering crypto pool

dsharma-dc:
Copies the bdev destroy logic as we do in pool destroy path, to the pool export path.

### DIFF
--- a/io-engine/src/lvs/lvs_store.rs
+++ b/io-engine/src/lvs/lvs_store.rs
@@ -672,7 +672,7 @@ impl Lvs {
         info!("{}: exporting lvs...", self_str);
 
         let pool = self.name().to_string();
-        let base_bdev = self.base_bdev();
+        let mut base_bdev = self.base_bdev();
         let (s, r) = pair::<i32>();
 
         self.unshare_all().await;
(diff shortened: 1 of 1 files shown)


io-engine/src/lvs/lvs_store.rs
```diff
@@ -688,12 +688,37 @@ impl Lvs {
...
+            let cbdev = base_bdev.crypto_base_bdev();
+
+            let _ = destroy_crypto_vbdev(base_bdev.name().to_string())
+                .await
+                .map_err(|e| {
+                    error!(
+                        "failed to delete crypto vbdev {:?} during lvs export. {e}",
+                        base_bdev.name()
+                    );
+                });
```
tiagolobocastro: If this fails is there any point trying to destroy the base bdev?

io-engine/src/lvs/lvs_store.rs
```diff
@@ -688,12 +688,37 @@ impl Lvs {
...
-        bdev_destroy(&base_bdev.bdev_uri_original_str().unwrap_or_default())
-            .await
-            .map_err(|e| LvsError::Destroy {
+        // If the base_bdev is a crypto vbdev then we need to destroy both - the crypto vbdev and it's base.
+        if base_bdev.driver() == "crypto" {
+            let cbdev = base_bdev.crypto_base_bdev();
+
+            let _ = destroy_crypto_vbdev(base_bdev.name().to_string())
+                .await
+                .map_err(|e| {
```
tiagolobocastro: if let Err

tiagolobocastro (approved):

io-engine/src/lvs/lvs_store.rs
dsharma-dc: > If this fails is there any point trying to destroy the base bdev?

Yes, we get another chance. The destroy of base bdev destroys the crypto vbdev sitting on it as well.

io-engine/src/lvs/lvs_store.rs
dsharma-dc: done

dsharma-dc:
bors try

Abhinandan-Purkait (approved):

dsharma-dc:
bors merge

PR #1848, opened 2025-04-04
feat(encryption): allow pools on same node to share keys

dsharma-dc:
The pools in same node(io-engine) can now share encryption key. Earlier, we used to set a crypto vbdev as key owner which means that during crypto vbdev destroy, the key was also destroyed. Now we don't set a crypto vbdev as owner and hence need to manage the key lifecycle using a hashmap that keeps track of which key is being used by which all crypto vbdevs.

### DIFF
--- a/io-engine/src/bdev/crypto.rs
+++ b/io-engine/src/bdev/crypto.rs
@@ -24,7 +24,7 @@ use spdk_rs::{
         accel_dpdk_cryptodev_enable, accel_dpdk_cryptodev_set_driver, create_crypto_disk,
         create_crypto_opts_by_name, delete_crypto_disk, spdk_accel_assign_opc,
         spdk_accel_crypto_key, spdk_accel_crypto_key_create, spdk_accel_crypto_key_create_param,
-        spdk_accel_crypto_key_get,
+        spdk_accel_crypto_key_destroy, spdk_accel_crypto_key_get,
     },
 };
 use std::{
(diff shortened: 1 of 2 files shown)


dsharma-dc:
Tested by creating 15 pools distributed among 3 keys. Shutdown the io-engine, recreate/import of pools worked ok. Destroyed pools, and keys were destroyed by the last user of a particular key.

dsharma-dc:
Opening for review. If necessary we can take this for MVP, otherwise for the next drop.

tiagolobocastro (commented):
What if we simply create the key using the pool name itself? Then it's lifecycle can match the pool and we don't need this extra complexity

dsharma-dc:
> What if we simply create the key using the pool name itself? Then it's lifecycle can match the pool and we don't need this extra complexity

This change rather intends to detach the lifecycle of key from the lifecycle of any particular pool/cryptobdev.

abhilashshetty04:
> The pools in same node(io-engine) can now share encryption key. Earlier, we used to set a crypto vbdev as key owner which means that during crypto vbdev destroy, the key was also destroyed. Now we don't set a crypto vbdev as owner and hence need to manage the key lifecycle using a hashmap that keeps track of which key is being used by which all crypto vbdevs.

Does this mean in an Io-engine, When the first crypto bdev it creates the encryption key and the same key is used for all subsequent crypto bdev that gets created?

dsharma-dc:
@abhilashshetty04 
> Does this mean in an Io-engine, When the first crypto bdev it creates the encryption key and the same key is used for all subsequent crypto bdev that gets created?

Yes. with this change the `K8s_Secret:Pools:SPDK_keys` are `1:N:1` relation, true sharing.  With the proposal discussed, the relation is `1:N:N`, a higher cost on memory, but can be easily done. Today what we have is `N:N:N`.
@tiagolobocastro

io-engine/src/bdev/crypto.rs
```diff
@@ -38,6 +38,82 @@ use crate::{
...
+            unsafe {
+                let key = spdk_accel_crypto_key_get(
+                    key_name.to_string().into_cstring().as_ptr() as *mut c_char
+                );
+                if key.is_null() {
+                    warn!("Non-existent key in SPDK.");
+                } else {
+                    let err = spdk_accel_crypto_key_destroy(key);
+                    if err != 0 {
+                        error!("Failed to destroy spdk accel crypto key");
```
```suggestion
                        error!("Failed to destroy spdk accel crypto key: {err}");
```

io-engine/src/bdev/crypto.rs
```diff
@@ -38,6 +38,82 @@ use crate::{
...
+                    }
+                }
+            }
+            map.remove(key_name);
+        }
+    }
+}
+
+/// Gets all crypto vbdev names using a given key_name.
+#[allow(dead_code)]
```
tiagolobocastro: Is this needed?

io-engine/src/bdev/crypto.rs
```diff
@@ -200,6 +276,12 @@ fn create_crypto_key(key_params: &EncryptionKey) -> Result<*mut spdk_accel_crypt
     };
 
     unsafe {
+        // Check if the key with this name exists. If it does, then we reuse that.
+        let existing_key = spdk_accel_crypto_key_get(create_key_params.key_name);
+        if !existing_key.is_null() {
+            info!("Key {:?} exists already, reusing it.", key_params.key_name);
```
tiagolobocastro, line 291: Maybe add the pool name as well to make it clear

io-engine/src/bdev/crypto.rs
```diff
@@ -271,6 +369,13 @@ pub fn create_crypto_vbdev_on_base_bdev(
         })
     } else {
         info!("crypto vbdev {crypto_vbdev_name} created on base bdev {base_bdev_name}");
+        add_key_user(key_params.key_name.clone(), crypto_vbdev_name.to_string());
```
tiagolobocastro: Shouldn't the key be added on the create_crypto_key?
If the pool create fails, don't we need to delete it?

dsharma-dc: > Shouldn't the key be added on the create_crypto_key?

Seems more logical to add once the `user` i.e. the crypto vbdev is also created, which happens  after `create_crypto_key`.

> If the pool create fails, don't we need to delete it?

When the pool create fails, we usually destroy bdevs that are already created. This means we also destroy crypto vbdev via `destroy_crypto_vbdev`, where the remove call is placed.

tiagolobocastro: Ah I see, so the cleanup would take care of it.

Though is there a race condition here when:
pool1 with secret A exists
pool2 create with secret A, secret A exists, but name not added yet
pool1 destroy, remove it's bdev name, and now list is empty so it destroys the the secret key?

dsharma-dc: This window is possible. This can be plugged by making a  recheck-and-create step for key inside the `add_key_user`. OR 
`create_crypto_key` can be always called from inside `add_key_user`, which will make map entry before crypto bdev is created. If bdev creation fails later, cleanup path will attempt entry removal which we'll have to do special handling for ENOENT(because bdev doesn't exist).

tiagolobocastro: The second approach seems more solid if we handle the create crypto from the locked path

io-engine/src/bdev/crypto.rs
dsharma-dc: getting pool name here is difficult as this is behind API and only bdevs are the thing here.

dsharma-dc:
bors try

Abhinandan-Purkait (approved):

dsharma-dc:
bors try

abhilashshetty04 (approved):

dsharma-dc:
bors merge

dsharma-dc:
bors merge

PR #1851, opened 2025-04-15
fix(ipv6): set correct adrfam and ip for nvmx

tiagolobocastro:
Removes the [ ] from the ip address for nvmx.
Sets the correct address family.

todo: probably we should refactor the builders to use proper types rather than just strings...


dsharma-dc (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #1862, opened 2025-05-12
chore(ci): inject github token post nix installation

pchandra19:
Tested here : https://github.com/openebs/mayastor/actions/runs/14969447595/job/42046846759

This PR address the security advisory where the github token was being exposed.
As a solution:
We are first installing nix (without gh token) and then injecting the token in a seperate step.
Also, as a safety measure we are making if a token is present.

### DIFF
--- a/.github/workflows/bdd.yml
+++ b/.github/workflows/bdd.yml
@@ -10,22 +10,44 @@ env:
 jobs:
   bdd-tests:
     runs-on: ubuntu-latest-16-cores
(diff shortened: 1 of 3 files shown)


Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

pchandra19:
bors merge

avishnu (approved):

PR #1879, opened 2025-06-20
feat(persistence): update replica health condition as atomic operation

dsharma-dc:
Fail the transaction operation and shutdown the nexus to avoid succeeding IO via remaining replicas. This ensures data integrity in case the replica being marked here has been picked up as source of truth elsewhere for the volume.

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev_error.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev_error.rs
@@ -176,7 +176,7 @@ pub enum Error {
     InvalidReservation { reservation: u8 },
     #[snafu(display("failed to update share properties {}", name))]
     UpdateShareProperties { source: CoreError, name: String },
-    #[snafu(display("failed to save nexus state {}", name))]
+    #[snafu(display("failed to save nexus state {name}, {source}"))]
     SaveStateFailed { source: StoreError, name: String },
 }
 
--- a/io-engine/src/bdev/nexus/nexus_persistence.rs
+++ b/io-engine/src/bdev/nexus/nexus_persistence.rs
@@ -1,5 +1,10 @@
 use super::{IoMode, Nexus, NexusChild};
-use crate::{persistent_store::PersistentStore, sleep::mayastor_sleep};
+use crate::{
+    persistent_store::{to_json_byte_vec, PersistentStore},
+    sleep::mayastor_sleep,
+    store::store_defs::StoreError,
+};
+use etcd_client::Error as EtcdErr;
 use serde::{Deserialize, Serialize};
 use std::time::Duration;
 
@@ -8,10 +13,10 @@ use super::Error;
 /// Information associated with the persisted NexusInfo structure.
 pub struct PersistentNexusInfo {
     /// Structure that is written to the persistent store.
-    inner: NexusInfo,
+    pub inner: NexusInfo,
     /// Key to use to persist the NexusInfo structure.
     /// If `Some` the key has been supplied by the control plane.
-    key: Option<String>,
+    pub key: Option<String>,
 }
 
 impl PersistentNexusInfo {
@@ -31,12 +36,19 @@ impl PersistentNexusInfo {
 
 /// Definition of the nexus information that gets saved in the persistent
 /// store.
-#[derive(Serialize, Deserialize, Debug, Default)]
+#[derive(Clone, Serialize, Deserialize, Debug, Default)]
 pub struct NexusInfo {
     /// Nexus destroyed successfully.
     pub clean_shutdown: bool,
+    /// Nexus needs to be shutdown.
+    pub do_self_shutdown: bool,
     /// Information about children.
     pub children: Vec<ChildInfo>,
+}
+pub struct NexusInfoTxn<'a> {
+    key_info: &'a mut PersistentNexusInfo,
+    // Expected value for the key.
+    expected: NexusInfo,
 }
 
 /// Definition of the child information that gets saved in the persistent
@@ -96,6 +108,7 @@ impl<'n> Nexus<'n> {
                 // expect the NexusInfo structure to contain default values.
                 assert!(nexus_info.children.is_empty());
                 assert!(!nexus_info.clean_shutdown);
+                assert!(!nexus_info.do_self_shutdown);
                 self.children_iter().for_each(|c| {
                     let child_info = ChildInfo {
                         uuid: NexusChild::uuid(c.uri()).expect("Failed to get child UUID."),
@@ -164,11 +177,40 @@ impl<'n> Nexus<'n> {
 
                 let uuid = NexusChild::uuid(child_uri).expect("Failed to get child UUID.");
 
+                let expected_value = nexus_info.clone();
                 nexus_info.children.iter_mut().for_each(|c| {
                     if c.uuid == uuid {
                         c.healthy = *healthy;
                     }
                 });
+
+                let mut txn = NexusInfoTxn {
+                    key_info: &mut persistent_nexus_info,
+                    expected: expected_value,
+                };
+
+                // Try executing the transaction. If the nexus info key's value isn't what we
+                // expect, shutdown this nexus. This can only happen if do_self_shutdown is
+                // found true in etcd for this key, which can only be set by core-agent's
+                // republish path. In this situation, the newly published nexus could've
+                // picked up the replica that we are trying to mark unhealthy here. Shutting
+                // down this nexus here ensures we don't let this IO succeed to other replicas
+                // after marking this one as unhealthy.
+                match self.save_txn(&mut txn).await {
+                    Ok(_) => {
+                        self.set_nexus_io_mode(IoMode::Normal).await;
+                        return Ok(());
+                    }
+                    Err(e) => {
+                        error!(
+                            "{self:?}: failed to update persistent store txn, \
+                            will shutdown the nexus: {e}"
+                        );
+                        self.try_self_shutdown();
+
+                        return Err(e);
+                    }
+                }
             }
             PersistOp::Shutdown => {
                 // Only update the clean shutdown variable. Do not update the
@@ -241,6 +283,84 @@ impl<'n> Nexus<'n> {
             if mayastor_sleep(Duration::from_secs(1)).await.is_err() {
                 error!("{self:?}: failed to wait for sleep");
             }
+        }
+    }
+
+    async fn save_txn(&self, info: &mut NexusInfoTxn<'_>) -> Result<(), Error> {
+        // If a key has been provided, use it to store the NexusInfo; use the
+        // nexus uuid as the key otherwise.
+        let key = match &info.key_info.key {
+            Some(k) => k.clone(),
+            None => self.uuid().to_string(),
+        };
+
+        let mut retry = PersistentStore::retries();
+
+        let new_value = to_json_byte_vec(&info.key_info.inner);
+        let expected_value = to_json_byte_vec(&info.expected);
+        let mut logged = false;
+
+        loop {
+            match PersistentStore::txn_create_execute(&key, &new_value, &expected_value).await {
+                Ok(txn_resp) => {
+                    if let Some(current_value) = txn_resp {
+                        let val = serde_json::from_slice::<NexusInfo>(&current_value).unwrap();
+
+                        // The server had likely received the transaction and executed it, but
+                        // client here saw a timeout. So if the current value is same as what we intended
+                        // to set, then consider success. Don't trust any other value and shutdown.
+                        if current_value == new_value {
+                            info!("value for key {key} already updated: {val:?}");
+                            return Ok(());
+                        }
+
+                        warn!("current state found: key - {key}, value - {val:?}");
+
+                        // This nexus won't be used again but let's still update this field with what's
+                        // found in etcd.
+                        info.key_info.inner_mut().do_self_shutdown = val.do_self_shutdown;
+
+                        return Err(Error::SaveStateFailed {
+                            source: StoreError::Txn {
+                                key: key.clone(),
+                                source: EtcdErr::IoError(std::io::Error::new(
+                                    std::io::ErrorKind::Other,
+                                    "Txn CompareOp failed",
+                                )),
+                            },
+                            name: self.name.clone(),
+                        });
+                    } else {
+                        // Don't need to check individual op responses.
+                        debug!(?key, "{self:?}: the state was saved successfully via txn");
+                        return Ok(());
+                    }
+                }
+
+                Err(err) => {
+                    retry -= 1;
+                    if retry == 0 {
+                        return Err(Error::SaveStateFailed {
+                            source: err,
+                            name: self.name.clone(),
+                        });
+                    }
+
+                    if !logged {
+                        error!(
+                            "{self:?}: failed to persist nexus info transaction: {err}\
+                        will silently retry ({retry} left): {err}"
+                        );
+                        logged = true;
+                    }
+
+                    // Allow some time for the connection to the persistent
+                    // store to be re-established before retrying the operation.
+                    if mayastor_sleep(Duration::from_secs(1)).await.is_err() {
+                        error!("{self:?}: failed to wait for sleep");
+                    }
+                }
+            };
         }
     }
 }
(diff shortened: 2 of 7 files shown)


dsharma-dc:
bors try

io-engine/src/bdev/nexus/nexus_bdev_error.rs
```diff
@@ -176,7 +176,7 @@ pub enum Error {
     InvalidReservation { reservation: u8 },
     #[snafu(display("failed to update share properties {}", name))]
     UpdateShareProperties { source: CoreError, name: String },
-    #[snafu(display("failed to save nexus state {}", name))]
+    #[snafu(display("failed to save nexus state {}, {}", name, source))]
```
tiagolobocastro:
```suggestion
    #[snafu(display("failed to save nexus state {name}, {source}"))]
```

io-engine/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -31,14 +36,40 @@ impl PersistentNexusInfo {
...
     pub clean_shutdown: bool,
+    /// Nexus needs to be shutdown.
+    pub do_self_shutdown: bool,
     /// Information about children.
     pub children: Vec<ChildInfo>,
 }
 
+impl Display for NexusInfo {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{{children: [")?;
```
tiagolobocastro: Could you just call the debug impl here?

io-engine/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -1,7 +1,12 @@
 use super::{IoMode, Nexus, NexusChild};
-use crate::{persistent_store::PersistentStore, sleep::mayastor_sleep};
+use crate::{
+    persistent_store::{to_json_byte_vec, PersistentStore},
+    sleep::mayastor_sleep,
+    store::store_defs::StoreError,
+};
+use etcd_client::{Compare, CompareOp, Error as EtcdErr, TxnOp, TxnOpResponse};
```
tiagolobocastro: this is breaking the `PersistentStore` API, which will complicate replacing it with a proxy (there's a ticket for that somewhere).
We should abstract this behavior inside the `PersistentStore`

io-engine/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -164,11 +196,40 @@ impl<'n> Nexus<'n> {
...
 
+                let expected_value = nexus_info.clone();
                 nexus_info.children.iter_mut().for_each(|c| {
                     if c.uuid == uuid {
                         c.healthy = *healthy;
                     }
                 });
+
+                let mut txn = NexusInfoTxn {
+                    key_info: &mut persistent_nexus_info,
```
tiagolobocastro, line 188: do we need this cross call mutation?

io-engine/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -243,4 +304,88 @@ impl<'n> Nexus<'n> {
...
+        let new_value = to_json_byte_vec(&info.key_info.inner);
+        let expected_value = to_json_byte_vec(&info.expected);
+        info!(
+            "Executing transaction for key {}, value {}, expected value {}.",
+            key,
+            String::from_utf8_lossy(&new_value),
+            String::from_utf8_lossy(&expected_value)
+        );
+
+        let cmp = Compare::value(key.to_string(), CompareOp::Equal, expected_value);
```
tiagolobocastro: we should hide this inside the `PersistentStore` itself

io-engine/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -243,4 +304,88 @@ impl<'n> Nexus<'n> {
...
+        let put_op = TxnOp::put(key.to_string(), new_value, None);
+        let or_else_op = TxnOp::get(key.to_string(), None);
```
tiagolobocastro: Do we need this last get?
Could we just compare and write, and fail should the compare not match for any reason

dsharma-dc:
bors try

io-engine/src/bdev/nexus/nexus_persistence.rs
dsharma-dc: > Do we need this last get? Could we just compare and write, and fail should the compare not match for any reason

This is required. A scenario I uncovered is below. The test modified in this patch hits this sometimes.

`In case of etcd issue like network partition, the transaction reached and got executed by etcd, but client saw operation timeout. The retry in this case has transaction response succeeded=false because the value in keystore is not what we expect. But if that value is same as what we are trying to set, then we don’t need to do anything further. In contrast, if the value found in keystore is not the same as what we are trying to set, we consider CompareOp failed and fail the state save operation and nexus goes into shutdown.`

io-engine/src/bdev/nexus/nexus_persistence.rs
dsharma-dc: > Could you just call the debug impl here?

Sure, will remove this, was using during initial testing.

io-engine/src/bdev/nexus/nexus_persistence.rs
dsharma-dc: > this is breaking the `PersistentStore` API, which will complicate replacing it with a proxy (there's a ticket for that somewhere). We should abstract this behavior inside the `PersistentStore`

Refactored the API. Didn't make it specific to Nexus info, rather a little more generic that it can be used for transaction that `do put op only, and do get upon failure`.

dsharma-dc:
bors try

dsharma-dc:
bors try

io-engine/src/bdev/nexus/nexus_persistence.rs
dsharma-dc: mutating this in the failure case to set the new flag's value in the in-core object `key_info` here, just in case required to debug.

io-engine/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -31,14 +36,40 @@ impl PersistentNexusInfo {
...
+//            }
+//            write!(f, "uuid: {}, healthy: {}", cinfo.uuid, cinfo.healthy)?;
+//        }
+//        write!(
+//            f,
+//            "], clean_shutdown: {}, do_self_shutdown: {}}}",
+//            self.clean_shutdown, self.do_self_shutdown
+//        )
+//    }
+//}
```
Abhinandan-Purkait: Do we need this?

Abhinandan-Purkait (approved):

io-engine/src/store/etcd.rs
```diff
@@ -55,17 +55,41 @@ impl Store for Etcd {
...
+                    .or_else([or_else_op]),
+            )
             .await
             .context(TxnErr {
                 key: key.to_string(),
-            })
+            })?;
+
+        if !txn_resp.succeeded() {
+            if let TxnOpResponse::Get(g) = &txn_resp.op_responses()[0] {
```
tiagolobocastro: Need to check if len > 1?

io-engine/src/store/etcd.rs
```diff
@@ -55,17 +55,41 @@ impl Store for Etcd {
...
+                if let Some(kv) = g.kvs().first() {
```
tiagolobocastro, line 80: better to take here rather copying the value on return

io-engine/src/store/etcd.rs
```diff
@@ -55,17 +55,41 @@ impl Store for Etcd {
...
+                    let current_value = kv.value();
+                    return Ok(Some(current_value.to_owned()));
+                }
+            } else {
+                // This is needed because absence of above response data shouldn't get
+                // implicitly assumed as success. With this the top level caller can
+                // compare this returned blank value with some expected value.
+                return Ok(Some(vec![]));
```
tiagolobocastro: Shouldn't we return a StoreError here as this seems like an etcd bug if we hit this?

io-engine/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -243,4 +304,74 @@ impl<'n> Nexus<'n> {
...
+        let new_value = to_json_byte_vec(&info.key_info.inner);
+        let expected_value = to_json_byte_vec(&info.expected);
+
+        loop {
+            match PersistentStore::txn_create_execute(&key, &new_value, &expected_value).await {
+                Ok(txn_resp) => {
+                    if let Some(current_value) = txn_resp {
+                        let val = serde_json::from_slice::<NexusInfo>(&current_value).unwrap();
+
+                        warn!("current state found: key - {key}, value - {val:?}");
```
tiagolobocastro: move this to after the if statement?

io-engine/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -243,4 +304,74 @@ impl<'n> Nexus<'n> {
...
+
+                Err(err) => {
+                    retry -= 1;
+                    if retry == 0 {
+                        return Err(Error::SaveStateFailed {
+                            source: err,
+                            name: self.name.clone(),
+                        });
+                    }
+                    error!("{self:?}: failed to persist nexus info transaction: {err}");
```
tiagolobocastro: Use the same log once strategy we have in `save`

io-engine/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -243,4 +304,74 @@ impl<'n> Nexus<'n> {
...
+                                source: EtcdErr::IoError(std::io::Error::new(
+                                    std::io::ErrorKind::Other,
+                                    "Txn CompareOp failed",
+                                )),
+                            },
+                            name: self.name.clone(),
+                        });
+                    } else {
+                        // Don't need to check individual op responses.
+                        debug!(?key, "{self:?}: state save transaction successful");
```
tiagolobocastro: same as `save`

```suggestion
                        trace!(?key, "{self:?}: the state was saved successfully");
```

io-engine/src/store/store_defs.rs
```diff
@@ -1,7 +1,7 @@
 //! Definition of a trait for a key-value store together with its error codes.
 
 use async_trait::async_trait;
-use etcd_client::{Compare, Error, TxnOp, TxnResponse};
+use etcd_client::Error;
```
tiagolobocastro, line 4: Not needed now I hope?

io-engine/src/persistent_store.rs
```diff
@@ -196,18 +195,27 @@ impl PersistentStore {
...
-    pub async fn txn(
+    /// Executes a transaction with Put op for the given key.
+    /// On failure, Get the key value.
+    pub async fn txn_create_execute(
         key: &impl StoreKey,
-        cmps: Vec<Compare>,
-        ops_success: Vec<TxnOp>,
-        ops_failure: Option<Vec<TxnOp>>,
-    ) -> Result<TxnResponse, StoreError> {
+        new_value: &[u8],
```
tiagolobocastro, line 202: These should be `&impl StoreValue` ?

io-engine/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -31,14 +36,40 @@ impl PersistentNexusInfo {
...
+//            self.clean_shutdown, self.do_self_shutdown
+//        )
+//    }
+//}
+
+pub struct NexusInfoTxn<'a> {
+    pub key_info: &'a mut PersistentNexusInfo,
+    // Expected value for the key.
+    pub expected: NexusInfo,
+}
```
tiagolobocastro: Do any of these need to be pub?

io-engine/src/store/etcd.rs
dsharma-dc: > Shouldn't we return a StoreError here as this seems like an etcd bug if we hit this?

We can return StoreError. In that case, we'll retry the transaction and it's going to fail again. The only difference is if in the retry failure we now have this `Get` response and it is what we intended to set, then we take success path. With current approach we are more eagerly shutting down.

io-engine/src/store/etcd.rs
dsharma-dc: > better to take here rather copying the value on return

do you mean this?
`if let Some(kv) = g.kvs().first().take()` . This also gives &KeyValue and then kv.value() is anyway &[u8]

io-engine/src/store/etcd.rs
dsharma-dc: > Need to check if len > 1?

It's ok not to check here. Either there will be one or nothing(handled in else path).

io-engine/src/store/store_defs.rs
dsharma-dc: > Not needed now I hope?

This has been there since long time, embedded into our error codes. ¯\_(ツ)_/¯

io-engine/src/store/etcd.rs
tiagolobocastro: hmm, if len is 0 it will crash?

io-engine/src/store/etcd.rs
tiagolobocastro: but won't vec![] lead into a program crash?

`let val = serde_json::from_slice::<NexusInfo>(&current_value).unwrap();`

io-engine/src/store/store_defs.rs
tiagolobocastro: oh it's used in the error variants, got it

io-engine/src/store/etcd.rs
tiagolobocastro: ah because it's a vector, you'd have to use something like `if let Some(kv) = g.take_kvs().remove(0)`
never mind then, more trouble than it's worth

io-engine/src/store/etcd.rs
dsharma-dc: > hmm, if len is 0 it will crash?

Is this leading to out-of-bound way of accessing? Then yeah, it can crash. Should be using first() or check len here.

io-engine/src/store/etcd.rs
dsharma-dc: Here I've changed to return StoreError already. Haven't pushed yet.

io-engine/src/store/etcd.rs
tiagolobocastro: If the len is 0 and you try to access [0] then yes it will

io-engine/src/persistent_store.rs
dsharma-dc: > These should be `&impl StoreValue` ?

We are using the serialized value here in this file for comparison below, and the StoreValue trait for u8 doesn't seem to work out of the box.

dsharma-dc:
bors try

io-engine/src/store/store_defs.rs
```diff
@@ -108,10 +108,9 @@ pub trait Store: Sync + Send + Clone {
     async fn txn_kv<K: StoreKey>(
         &mut self,
         key: &K,
-        cmps: Vec<Compare>,
-        ops_success: Vec<TxnOp>,
-        ops_failure: Option<Vec<TxnOp>>,
-    ) -> Result<TxnResponse, StoreError>;
+        new_value: Vec<u8>,
```
tiagolobocastro, line 112: I'd suggest renaming the function now, and also amend the description (I think it's on the impl rather than here, which we should also move to here instead)

tiagolobocastro (approved):

io-engine/src/store/store_defs.rs
dsharma-dc: > I'd suggest renaming the function now, and also amend the description (I think it's on the impl rather than here, which we should also move to here instead)

Naming seems ok to me. What do you suggest? Amended the description.

tiagolobocastro: was thinking something like `compare_and_write`

dsharma-dc: > was thinking something like `compare_and_write`

Let me change it to `put_kv_cas()`, similar to core-agent `put_obj_cas()`. So there is a respective `put_kv()` and `put_obj` for them.

dsharma-dc:
bors merge

PR #1901, opened 2025-08-20
fix: exit with error if gRPC port is busy

tiagolobocastro:
By default, the panic behaviour is to unwind, meaning we don't exit. In case we fail to even start gRPC we should exit, as there's nothing to do.

In this commit, we now internally raise a signal SIGUSR1 which then triggers the shutdown of the reactor et all.

todo:
We should also consider catching any panic and do the same, or is there any case where we'd want to unwind and remain running?


abhilashshetty04 (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #1897, opened 2025-08-18
do bdev rescan before calling lvs grow function

abhilashshetty04:
This PR does `bdev_aio_rescan` before calling grow function.

Also some changes in error handling which will help in control plane logic.

PR #1903, opened 2025-08-29
proto changes for pool expansion

abhilashshetty04:


io-engine/src/bin/io-engine-client/v1/pool_cli.rs
```diff
@@ -184,8 +184,8 @@ pub fn subcommands() -> Command {
                 .default_value(PoolType::Lvs.as_ref()),
         );
 
-    let grow = Command::new("grow")
-        .about("Grow a storage pool to fill the entire underlying device")
+    let expand = Command::new("expand")
+        .about("Expand a storage pool to fill the entire underlying device")
```
dsharma-dc:
```suggestion
        .about("Expand a storage pool to span the entire underlying device")
```

io-engine/src/lvs/lvs_error.rs
```diff
@@ -297,6 +305,8 @@ impl ToErrno for LvsError {
             Self::CloneConfigFailed { .. } => Errno::EINVAL,
             Self::WipeFailed { .. } => Errno::EINVAL,
             Self::ResourceLockFailed { .. } => Errno::EBUSY,
+            Self::MaxExpansionFactorParse { .. } => Errno::EINVAL,
+            Self::MaxExpansionFactorFormat { .. } => Errno::EINVAL,
```
dsharma-dc: Let's not introduce two new errors just for parsing the input. You might want to use LvsError::PoolCreate itself with some differing internal codes.

abhilashshetty04: This is just to give detailed error on each type of error.  PoolCreate requires BsError(seems to be blobstore specific) . This is just validations that we do on user inputs.
 
I have added 1 more if conversion fails (GiB, TiB > Bytes). Shall we just have 1 error type for factors and 1 for Capacity based input? If not, I can have 1 common for MaxExpansion input.

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
```diff
@@ -546,24 +549,48 @@ async fn grow(mut ctx: Context, matches: &ArgMatches) -> crate::Result<()> {
...
+            name: Some(name.clone()),
+            pooltype: pooltype.map(|pooltype| v1rpc::pool::PoolTypeValue {
+                value: v1rpc::pool::PoolType::from(pooltype) as i32,
+            }),
+            uuid: None,
+        })
+        .await
+        .context(GrpcStatus)?;
+
+    let pool = &list_response.get_ref().pools[0];
```
tiagolobocastro, line 579: this may crash if there's no pool?

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
```diff
@@ -546,24 +549,48 @@ async fn grow(mut ctx: Context, matches: &ArgMatches) -> crate::Result<()> {
...
 
-    if old_cap == new_cap {
-        println!("Pool capacity did not change: {new_cap} bytes");
-    } else {
-        println!("Pool capacity was {old_cap}, now {new_cap} bytes");
-    }
+    match ctx.output {
+        OutputFormat::Json => {}
+        OutputFormat::Default => {
+            println!("pool expanded from {previous_capacity} to {current_capacity}");
```
tiagolobocastro, line 598: still missing the case for "capacity did not change" ?

io-engine/src/grpc/v1/pool.rs
```diff
@@ -709,39 +710,23 @@ impl PoolRpc for PoolService {
         .await
     }
 
+    async fn grow_pool(&self, _request: Request<GrowPoolRequest>) -> GrpcResult<GrowPoolResponse> {
+        unimplemented!()
```
tiagolobocastro: Let's return an error here

io-engine/src/lvs/lvs_store.rs
```diff
@@ -415,20 +422,34 @@ impl Lvs {
...
+    fn md_resv_ratio(args: &PoolArgs, capacity: u64) -> Result<u32, LvsError> {
+        let param = match args.md_args.as_ref().and_then(|p| p.max_expansion.clone()) {
+            Some(p) => p,
+            None => return Ok(100),
+        };
+        let factor = if let Some(stripped) = param.strip_suffix('x') {
+            stripped
+                .parse::<f64>()
+                .map_err(|_| LvsError::MaxExpansionParse {
+                    msg: format!("Failed to parse factor {stripped} into float"),
```
```suggestion
                .map_err(|error| LvsError::MaxExpansionParse {
                    msg: format!("Failed to parse max_expansion {stripped} as float: {error}"),
```

io-engine/src/lvs/lvs_store.rs
```diff
@@ -415,20 +422,34 @@ impl Lvs {
...
+                })?
+        } else if param.ends_with('B') {
+            let expand_bytes = Byte::from_str(&param).map_err(|_| LvsError::MaxExpansionParse {
+                msg: format!("Max expansion size {param} can't be parsed into bytes"),
```
```suggestion
            let expand_bytes = Byte::from_str(&param).map_err(|error| LvsError::MaxExpansionParse {
                msg: format!("Failed to parse max_expansion {param} as bytes: {error}"),
```

io-engine/src/lvs/lvs_store.rs
```diff
@@ -415,20 +422,34 @@ impl Lvs {
...
-        }
+            return Err(LvsError::MaxExpansionParse {
+                msg: format!("Max expansion factor {param} does not end with x or B"),
+            });
+        };
+        // SPDK ensures that pool has enough pages in used_cluster_mask to track the current device size.
+        // So, If maxExpansion results in < 1x it still shouldn't matter. Its same as having
+        // default reservation. However, it does impact how many md pages per cluster is reserved.
+        // For ex. If factor turns out to be 0.5 then blobstore reserves 1 page per 2 clusters.
+        // So lets ensures that we pass atleast default reservation.
```
```suggestion
        // The blobstore ensures that we have enough pages in used_cluster_mask to track the current device size.
        // So, If maxExpansion results in < 1x it still shouldn't matter. Its same as having
        // default reservation. However, it does impact how many md pages per cluster are reserved.
        // For ex, if the factor turns out to be 0.5 then the blobstore reserves 1 page per 2 clusters.
        // So lets ensures that we pass at least default reservation.
```

io-engine/src/pool_backend.rs
```diff
@@ -234,6 +234,7 @@ pub trait IPoolProps {
     fn committed(&self) -> u64;
     fn md_props(&self) -> Option<PoolMetadataInfo>;
     fn encrypted(&self) -> bool;
+    fn max_expandable_size(&self) -> u64;
```
tiagolobocastro: And make lvm return None

```suggestion
    fn max_expandable_size(&self) -> Option<u64>;
```

io-engine/src/lvs/lvs_error.rs
dsharma-dc: ok. yeah if we can have only one error message that distinguishes different parsing errors, perhaps by a `message` field, that should be ok.

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
abhilashshetty04: That's there in other PR. We return error

https://github.com/openebs/mayastor/pull/1897/files#diff-679e16b5e7851f2cb7d5ad905905e498b5d5bb0383020273d269eaab1829d468R261

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
```diff
@@ -546,24 +549,55 @@ async fn grow(mut ctx: Context, matches: &ArgMatches) -> crate::Result<()> {
...
+            pooltype: pooltype.map(|pooltype| v1rpc::pool::PoolTypeValue {
+                value: v1rpc::pool::PoolType::from(pooltype) as i32,
+            }),
+            uuid: None,
+        })
+        .await
+        .context(GrpcStatus)?;
+
+    if list_response.get_ref().pools.is_empty() {
+        return Err(ClientError::ResourceNotFound {
```
dsharma-dc: You can just use tonic::Status code NotFound here. Since it's pool resource grpc, don't need to explicitly define a resource type.

io-engine/src/lvs/lvs_store.rs
```diff
@@ -466,9 +490,10 @@ impl Lvs {
                 msg: format!("{cluster_size}, larger than max limit {MAX_CLUSTER_SIZE}"),
             });
         }
-
-        let mdp_ratio = Self::mdp_ratio(&args)?;
-
+        let bdev =
+            UntypedBdev::lookup_by_name(&bdev).ok_or(LvsError::BdevNotFound { name: bdev })?;
+        let bdev_capacity = bdev.num_blocks() * bdev.block_len() as u64;
+        let mdp_ratio = Self::md_resv_ratio(&args, bdev_capacity)?;
```
dsharma-dc: As discussed, we can do all this on caller side and stuff the info in PoolArgs::PoolMetadataArgs, then use it here.

dsharma-dc: on second thought, keep the stuff here and just use the standard way of returning BdevNotFound, without new error code, as done in other places in create pool path. Because moving to caller side will incur additional checks for crypto vs normal bdev.

abhilashshetty04: The PoolMetadataArgs are args mapped from grpc message `PoolMetadataArgs`.

I have used following error and doing bdev lookup in the same place before calling  `md_resv_ratio` fn as discussed.

```
LvsError::InvalidBdev {
      source: BdevError::BdevNotFound {
          name: bdev.to_string(),
      },
      name: bdev.to_string(),
  }
```

abhilashshetty04: Done

io-engine/src/lvs/lvs_store.rs
```diff
@@ -208,6 +209,12 @@ impl Lvs {
         unsafe { spdk_bs_get_used_md(self.blob_store()) }
     }
 
+    /// Size upto which blobstore can be expanded.
+    pub fn max_expandable_size(&self) -> Option<u64> {
+        // TODO: Use spdk function when changes gets merged.
+        Some(0)
```
dsharma-dc, line 215: May be just use Some(MAX_INT) here until spdk changes are in so that the testing doesn't possibly run into erratic behaviour due to 0

io-engine/src/lvs/lvs_store.rs
```diff
@@ -415,20 +422,37 @@ impl Lvs {
...
+            return Err(LvsError::MaxExpansionParse {
+                msg: format!("Max expansion factor {param} does not end with x or B"),
+            });
+        };
+        // The Blobstore ensures that we have enough pages in used_cluster_mask to track the current device size.
+        // So, If maxExpansion results in < 1x it still shouldn't matter. Its same as having
+        // default reservation. However, it does impact how many md pages per cluster are reserved.
+        // For ex. if the factor turns out to be 0.5 then blobstore reserves 1 page per 2 clusters.
+        // So lets ensures that we pass at least default reservation.
+        Ok((factor.max(1.0) * 100.0) as u32)
```
dsharma-dc, line 455: If we are changing usual default to 200, then this also should be perhaps made 200 here.

dsharma-dc (approved):

io-engine/src/lvs/lvs_store.rs
abhilashshetty04: I guess that should be fine. Even if we pass MAX_INT test cant rely on that. Test for the time being do some smaller grow and ensure framework is set up. Once we have SPDK changes they can refer max_expandable_size.

io-engine/src/lvs/lvs_store.rs
abhilashshetty04: We eventually decided not have have different default then SPDK. Let user specify themselves if they want to have expansion capability for new Pools. I have mentioned default as 1x in oep as well.

niladrih (approved):

Abhinandan-Purkait (approved):

io-engine/src/lvs/lvs_store.rs
dsharma-dc: > We eventually decided not have have different default then SPDK. Let user specify themselves if they want to have expansion capability for new Pools. I have mentioned default as 1x in oep as well.

So what if someone forgets to specify for some reason but later on intends to expand, and realise they don't have headroom?

abhilashshetty04: I thought If expansion is critical, users should know and decide what factor they want. Making it default “2x” hides an important configuration knob. They will anyway see `max_expandable_size` and know they have limited expansion capability.

abhilashshetty04:
bors merge

abhilashshetty04:
bors try

abhilashshetty04:
bors merge

abhilashshetty04:
bors cancel

abhilashshetty04:
bors merge

PR #1906, opened 2025-09-15
ci: filter git scissor line in commit lint hook

urso:
Updates pre-commit hook to exclude Git's scissor line when running commitlint, preventing the hook from processing contents intended to be excluded from the final commit message.

I have git configured to add a diff of all changes to my editor when writing the commit message. But unfortunately long lines and rust code tend to fail the commitlint. The filter will remove the diff before sending the commit message to commitlint.

For example for this PR git did append the following to the message:

```
#
# ------------------------ >8 ------------------------
# Do not modify or remove the line above.
# Everything below it will be ignored.
diff --git c/.pre-commit-config.yaml i/.pre-commit-config.yaml
index a81f5daa..32342d32 100644
--- c/.pre-commit-config.yaml
+++ i/.pre-commit-config.yaml
@@ -35,7 +35,7 @@ repos:
         name: Commit Lint
         description: Runs commitlint against the commit message.
         language: system
-        entry: bash -c "npm install @commitlint/config-conventional @commitlint/cli; cat $1 | grep -v '^#' | npx commitlint"
+        entry: bash -c "npm install @commitlint/config-conventional @commitlint/cli; cat $1 | awk '/^# -+.*>8.*-+/{exit} 1' | grep -v '^#' | npx commitlint"
         args: [$1]
         stages: [commit-msg]
     -   id: python-check
```

We look for the `# ------------------------ >8 ------------------------` line and remove everything until the end.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,7 +35,7 @@ repos:
         name: Commit Lint
         description: Runs commitlint against the commit message.
         language: system
-        entry: bash -c "npm install @commitlint/config-conventional @commitlint/cli; cat $1 | grep -v '^#' | npx commitlint"
+        entry: bash -c "npm install @commitlint/config-conventional @commitlint/cli; cat $1 | awk '/^# -+.*>8.*-+/{exit} 1' | grep -v '^#' | npx commitlint"
         args: [$1]
         stages: [commit-msg]
     -   id: python-check
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,7 +35,7 @@ repos:
         name: Commit Lint
         description: Runs commitlint against the commit message.
         language: system
-        entry: bash -c "npm install @commitlint/config-conventional @commitlint/cli; cat $1 | awk '/^# -+.*>8.*-+/{exit} 1' | grep -v '^#' | npx commitlint"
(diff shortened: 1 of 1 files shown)


tiagolobocastro (approved):
Maybe can you also change this to use commitlint directly rather than installing via npm?
https://github.com/openebs/mayastor-control-plane/blob/a678742bd944a50a2a844d1caabe2ae37f3dd9f5/.pre-commit-config.yaml#L39C73-L39C83

Otherwise can be done as different PR, up to you

urso:
> Maybe can you also change this to use commitlint directly rather than installing via npm?

Good idea. Let me give that a try

urso:
Done, updated PR to use commitlint via nix shell environment

shell.nix
```diff
@@ -25,6 +25,7 @@ let
 
     cfg = {
       buildInputs = with pkgs; [
+        commitlint
```
tiagolobocastro: It should be there already, coming from the default buildInputs from spdk-rs?

urso: Oh, I haven't checked spdk-rs/nix/shell. Indeed it already has commitlint.

dsharma-dc (approved):

Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

tiagolobocastro:
Thanks @urso !
bors merge

PR #1908, opened 2025-09-15
feat(raid): add RaidBdev

urso:
Add RaidBDev to core module. The RaidBdevs provides us with a more convenient API over having to use low-level SPDK raid_bdev functions throughout the code base.

Note: I have a working Raid0 implementation (MVP) available in my feature-raid0 branch. For context how the changes in here fit in the full implementation see: https://github.com/openebs/mayastor/compare/develop...urso:mayastor:feature-raid0?expand=1

### DIFF
--- a/io-engine/src/core/raid.rs
+++ b/io-engine/src/core/raid.rs
@@ -1,0 +1,502 @@
+use std::{
+    fmt::{Debug, Formatter},
+    ptr::{self, NonNull},
+    time::Duration,
+};
+
+use futures::channel::oneshot;
+use nix::errno::Errno;
+
+use crate::{
+    bdev_api::BdevError,
+    core::UntypedBdev,
+    ffihelper::{cb_arg, done_errno_cb, ErrnoResult, IntoCString},
+};
+
+use spdk_rs::{
+    libspdk,
+    libspdk::{
+        raid_bdev, raid_bdev_add_base_bdev, raid_bdev_create, raid_bdev_delete,
+        raid_bdev_find_by_name, spdk_uuid, RAID_BDEV_STATE_CONFIGURING, RAID_BDEV_STATE_OFFLINE,
+        RAID_BDEV_STATE_ONLINE,
+    },
+    PollerBuilder, Uuid as SpdkUuid,
+};
+
+pub struct RaidBdev {
+    inner: NonNull<raid_bdev>,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RaidState {
+    /// RAID is waiting for more devices to be added
+    WaitingForDevices,
+    /// RAID has sufficient devices and is transitioning to online (starting up)
+    StartingUp,
+    /// RAID is fully online and operational
+    Online,
+    /// RAID is offline/failed
+    Offline,
+    /// RAID state is unknown or invalid
+    Unknown,
+}
+
+impl AsRef<str> for RaidState {
+    fn as_ref(&self) -> &str {
+        match self {
+            RaidState::WaitingForDevices => "waiting for devices",
+            RaidState::StartingUp => "starting up",
+            RaidState::Online => "online",
+            RaidState::Offline => "offline",
+            RaidState::Unknown => "unknown",
+        }
+    }
+}
+
+impl RaidState {
+    pub fn as_str(&self) -> &str {
+        self.as_ref()
+    }
+}
+
+impl std::fmt::Display for RaidState {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.as_ref())
+    }
+}
+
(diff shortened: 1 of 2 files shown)


io-engine/src/core/raid.rs
```diff
@@ -0,0 +1,502 @@
...
+    fn as_ref(&self) -> &str {
+        match self {
+            RaidState::WaitingForDevices => "waiting for devices",
+            RaidState::StartingUp => "starting up",
+            RaidState::Online => "online",
+            RaidState::Offline => "offline",
+            RaidState::Unknown => "unknown",
+        }
+    }
+}
```
Abhinandan-Purkait, line 54: nit: You can sure strum macros

io-engine/src/core/raid.rs
```diff
@@ -0,0 +1,502 @@
...
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct RaidDeviceCounts {
+    /// number of base bdevs discovered and configured
+    pub discovered: u8,
+    /// number of base bdevs we expect to be operational (can be less for degraded arrays - depends on raid level)
+    pub operational: u8,
+    /// total number of base bdevs comprising the RAID array
+    pub total: u8,
```
Abhinandan-Purkait, line 75: Doc comments, should start with Capital and end with `.`. It would be good if we can follow this, even if at some places it was not followed prior to your changes. Thanks!

io-engine/src/core/raid.rs
```diff
@@ -0,0 +1,502 @@
...
+        let (s, r) = oneshot::channel::<Result<(), RaidWaitError>>();
+
+        let start_time = Instant::now();
+        let raid_bdev = RaidBdev::from_ptr(self.inner.as_ptr());
+        let mut sender = Some(s);
+        let _poller = PollerBuilder::<()>::new()
+            .with_name(&format!("raid_wait_{}", self.name()))
+            .with_interval(Duration::from_millis(100))
+            .with_poll_fn(move |_| {
+                // Reconstruct RaidBdev from the raw pointer
```
abhilashshetty04, line 342: This comment goes before line 336?

io-engine/src/core/raid.rs
```diff
@@ -0,0 +1,502 @@
...
+
+        match raid_bdev.state {
+            RAID_BDEV_STATE_ONLINE => RaidState::Online,
+            RAID_BDEV_STATE_OFFLINE => RaidState::Offline,
+            RAID_BDEV_STATE_CONFIGURING => {
+                // Check if we have enough devices to start transitioning to online
+                if device_counts.discovered >= device_counts.operational {
+                    RaidState::StartingUp
+                } else {
+                    RaidState::WaitingForDevices
```
abhilashshetty04, line 294: Shouldn't it be `WaitingForDevice` when `device_counts.total < device_counts.discovered`? Because raid bdev might have enough bdevs, some could be just not operational.  Also, do we know when num_base_bdevs_operational is decremented?

What happens when `num_base_bdevs_operational < min_base_bdevs_operational`?

io-engine/src/core/raid.rs
```diff
@@ -0,0 +1,502 @@
...
+        strip_size_kb: u32,
+        num_devices: u8,
+        level: i32,
+    ) -> Result<Self, BdevError> {
+        let cname = name.to_string().into_cstring();
+
+        let uuid_ptr = uuid.as_ref().map_or(ptr::null(), |uuid| {
+            let spdk_uuid = SpdkUuid::from(*uuid).into_raw();
+            &spdk_uuid as *const spdk_uuid
+        });
```
susobhandey, line 164: let spdk_uuid = uuid.map(|u| SpdkUuid::from(u).into_raw());
let uuid_ptr = spdk_uuid
    .as_ref()
    .map_or(ptr::null(), |u| u as *const spdk_uuid);   we need to do like this right now it's lifetime is end of create() what do you think?

io-engine/src/core/raid.rs
```diff
@@ -0,0 +1,502 @@
...
+
+        let mut raid_bdev_ptr: *mut raid_bdev = ptr::null_mut();
+        let errno = unsafe {
+            raid_bdev_create(
+                cname.as_ptr(),
+                strip_size,
+                num_devices,
+                level,
+                false,
+                uuid_ptr,
```
susobhandey, line 181: Invalid dead pointer

io-engine/src/core/raid.rs
```diff
@@ -0,0 +1,502 @@
...
+            return None;
+        }
+
+        Some(RaidBdev::from_ptr(raid_bdev_ptr))
+    }
+}
+
+impl RaidBdev {
+    /// Create a RaidBdev wrapper from a raw pointer
+    pub fn from_ptr(ptr: *mut raid_bdev) -> Self {
```
susobhandey, line 146: better to be unsafe

PR #1911, opened 2025-09-30
set PR_SET_IO_FLUSHER for io-engine 

daclash:
This pull request introduces a new functionality to set the PR_SET_IO_FLUSHER
flag in io-engine for improved I/O process handling.  The changes add a new
module that implements a helper function to invoke the prctl system call with
the PR_SET_IO_FLUSHER flag and integrates this call into the io-engine startup
routine. This ensures that processes involved in I/O operations are configured
correctly.

Follow-on changes:

- Update runner scripts to include CAP_SYS_RESOURCE capability
- Update systemd service examples in documentation
- Required for PR_SET_IO_FLUSHER system call in direct execution contexts

### DIFF
--- a/io-engine/src/bin/io-engine.rs
+++ b/io-engine/src/bin/io-engine.rs
@@ -21,6 +21,7 @@ use io_engine::{
     eventing::Event,
     grpc, logger,
     persistent_store::PersistentStoreBuilder,
+    prctl::Prctl,
     subsys::Registration,
 };
 use version_info::fmt_package_info;
@@ -272,6 +273,12 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     }
 
     info!("{}", fmt_package_info!());
+
+    if let Err(err) = Prctl::set_io_flusher() {
+        error!("Failed to set PR_SET_IO_FLUSHER, CAP_SYS_RESOURCE is required error: {err}");
+    } else {
+        info!("PR_SET_IO_FLUSHER is configured");
+    }
 
     // Handle diagnostics-related commands before initializing the agent.
     // Once diagnostics command is executed (regardless of status), exit the
--- a/io-engine/src/prctl.rs
+++ b/io-engine/src/prctl.rs
@@ -1,0 +1,22 @@
+use libc::c_int;
+use std::io::{Error, Result};
+
+const PR_SET_IO_FLUSHER: c_int = 57;
+
+pub struct Prctl;
+
+impl Prctl {
+    /// From man page of prctl: If a user process is involved in the block layer
+    /// or filesystem I/O path, and can allocate memory while processing I/O requests
+    /// it must set arg2 to 1. This will put the process in the IO_FLUSHER  state,
+    /// which allows it special treatment to make progress when allocating memory.
+    /// If arg2 is 0, the process will clear the IO_FLUSHER state, and the default
+    /// behavior will be used.
+    pub fn set_io_flusher() -> Result<()> {
+        let ret = unsafe { libc::prctl(PR_SET_IO_FLUSHER, 1, 0, 0, 0) };
+        if ret != 0 {
+            return Err(Error::last_os_error());
+        }
+        Ok(())
+    }
+}
--- a/doc/run.md
+++ b/doc/run.md
@@ -230,7 +230,7 @@ systemd.services.mayastor = {
     Type = "forking";
     User = "nobody";
     ExecStart = "${pkgs.mayastor}/bin/mayastor";
-    AmbientCapabilities = "CAP_SETPCAP CAP_SYS_ADMIN CAP_IPC_LOCK CAP_SYS_NICE";
+    AmbientCapabilities = "CAP_SETPCAP CAP_SYS_ADMIN CAP_IPC_LOCK CAP_SYS_NICE CAP_SYS_RESOURCE";
   };
 };
 ```
@@ -245,7 +245,7 @@ After=network.target
 Description=A cloud native declarative data plane.
 
 [Service]
-AmbientCapabilities=CAP_SETPCAP CAP_SYS_ADMIN CAP_IPC_LOCK CAP_SYS_NICE
+AmbientCapabilities=CAP_SETPCAP CAP_SYS_ADMIN CAP_IPC_LOCK CAP_SYS_NICE CAP_SYS_RESOURCE
 ExecStart=/usr/bin/mayastor
 Type=simple
 User=nobody
--- a/io-engine-bench/.cargo/runner.sh
+++ b/io-engine-bench/.cargo/runner.sh
@@ -20,10 +20,10 @@ fi
 #
 # * Set `cap_setpcap` to be able to set [ambient capabilities](https://lwn.net/Articles/636533/) which can be inherited
 # by children.
-# * Set `cap_sys_admin,cap_ipc_lock,cap_sys_nice` as they are required by the io-engine.
+# * Set `cap_sys_admin,cap_ipc_lock,cap_sys_nice,cap_sys_resource` as they are required by the io-engine.
 ${MAYBE_SUDO} capsh \
-  --caps="cap_setpcap+iep cap_sys_admin,cap_ipc_lock,cap_sys_nice+iep" \
-  --addamb=cap_sys_admin --addamb=cap_ipc_lock --addamb=cap_sys_nice \
+  --caps="cap_setpcap+iep cap_sys_admin,cap_ipc_lock,cap_sys_nice,cap_sys_resource+iep" \
+  --addamb=cap_sys_admin --addamb=cap_ipc_lock --addamb=cap_sys_nice --addamb=cap_sys_resource \
   -- -c "${ARGS}"
 
 if [ -d "$TARGET_CRITERION" ]; then
--- a/io-engine/.cargo/runner.sh
+++ b/io-engine/.cargo/runner.sh
@@ -13,8 +13,8 @@ fi
 #
 # * Set `cap_setpcap` to be able to set [ambient capabilities](https://lwn.net/Articles/636533/) which can be inherited
 # by children.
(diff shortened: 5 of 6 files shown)


io-engine/src/bin/io-engine.rs
```diff
@@ -273,6 +274,15 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     info!("{}", fmt_package_info!());
 
+    if let Err(err) = Prctl::set_io_flusher() {
+        error!(
+            "Failed to set PR_SET_IO_FLUSHER, CAP_SYS_RESOURCE is required error: {}",
+            err
```
tiagolobocastro: nit:
```suggestion
            "Failed to set PR_SET_IO_FLUSHER, CAP_SYS_RESOURCE is required error: {err}"
```

io-engine/src/prctl.rs
```diff
@@ -0,0 +1,23 @@
...
+pub struct Prctl;
+
+impl Prctl {
+    // From man page of prctl: If a user process is involved in the block layer
+    // or filesystem I/O path, and can allocate memory while processing I/O requests
+    // it must set arg2 to 1. This will put the process in the IO_FLUSHER  state,
+    // which allows it special treatment to make progress when allocating memory.
+    // If arg2 is 0, the process will clear the IO_FLUSHER state, and the default
+    // behavior will be used.
+
```
tiagolobocastro: Using 3 /// for doc comments

```suggestion
    /// From man page of prctl: If a user process is involved in the block layer
    /// or filesystem I/O path, and can allocate memory while processing I/O requests
    /// it must set arg2 to 1. This will put the process in the IO_FLUSHER  state,
    /// which allows it special treatment to make progress when allocating memory.
    /// If arg2 is 0, the process will clear the IO_FLUSHER state, and the default
    /// behavior will be used.
```

io-engine/src/bin/io-engine.rs
tiagolobocastro: Was wondering if we should abort here, though this may fail on pre 5.9 kernels right?
From recollection we had a small user portion on 5.4, so if were to abort in failure it might impact them, so perhaps better to just error as you have here

tiagolobocastro (commented):
bors try

io-engine/src/bin/io-engine.rs
daclash: > From recollection we had a small user portion on 5.4, so if were to abort in failure it might impact them, so perhaps better to just error as you have here.

Correct.  v5.4 kernels will fail this call so I chose to log an error.
However, I am happy to make a make a change here if you want.

The oldest LTS kernel is v5.4.299 and that version does not support `PR_SET_IO_FLUSHER`.
The second oldest LTS kernel is v5.10.244 and that version does support `PR_SET_IO_FLUSHER`.

```text
host /workspace/linux $ git tag -n -l | grep -Ee 'v5[.]4[.][0-9][0-9][0-9]+' | tail -1
v5.4.299        This is the 5.4.299 stable release

host /workspace/linux $ git checkout v5.4.299
...

host /workspace/linux $ grep . -nHIREe 'PR_SET_IO_FLUSHER'
host /workspace/linux $

host /workspace/linux $ git checkout v5.10.244
...

host /workspace/linux $ grep . -nHIREe 'PR_SET_IO_FLUSHER'
./include/uapi/linux/prctl.h:247:#define PR_SET_IO_FLUSHER              57
./kernel/sys.c:2529:    case PR_SET_IO_FLUSHER:
./tools/include/uapi/linux/prctl.h:247:#define PR_SET_IO_FLUSHER                57
```

daclash: I applied the suggested change locally, changed to pass pre-commit, and force pushed to keep the commit history simple.  The changes are visible using <https://github.com/daclash/mayastor/compare/44217aa0761e31660448e34c4ec515f0d3b005d9..e9c8fa4558d356eaa30168b0f3114f2d3d87da83>.

io-engine/src/prctl.rs
daclash: I applied the suggested change locally, changed to pass pre-commit, and force pushed to keep the commit history simple.  The changes are visible using <https://github.com/daclash/mayastor/compare/44217aa0761e31660448e34c4ec515f0d3b005d9..e9c8fa4558d356eaa30168b0f3114f2d3d87da83>.

daclash:
`cargo test` with both <https://github.com/openebs/mayastor/pull/1911> and <https://github.com/openebs/mayastor-dependencies/pull/135> shows the following output:

```text
host /workspace/acsa-mayastor $ nix-shell --run 'cd io-engine && RUST_LOG=TRACE cargo test --features=io-engine-testing -- --test-threads 1 --nocapture --test compose_up_down' 2>&1 | grep -i flusher
ms2:[2025-09-29T20:03:42.751741798+00:00  INFO io_engine:io-engine.rs:280] PR_SET_IO_FLUSHER is configured
ms1:[2025-09-29T20:03:42.510464406+00:00  INFO io_engine:io-engine.rs:280] PR_SET_IO_FLUSHER is configured

host /workspace/acsa-mayastor $ git log --oneline @{u}..
26a8f5c6 (HEAD -> user/daclash/pr-set-io-flusher, origin/user/daclash/pr-set-io-flusher) fix: add CAP_SYS_RESOURCE capability for PR_SET_IO_FLUSHER support

host /workspace/acsa-mayastor $ cd ./utils/dependencies/
host /workspace/acsa-mayastor/utils/dependencies $ git log -1 --oneline
c2694e6 (HEAD) fix: add CAP_SYS_RESOURCE capability to Docker containers
```

io-engine/src/bin/io-engine.rs
tiagolobocastro: Alright, let's keep it as is then, thank you

tiagolobocastro (approved):
thanks @daclash!

tiagolobocastro:
bors delegate+

daclash:
bors try

dsharma-dc:
Do we any measured evidence of Mayastor improvements by setting the IO_FLUSHER state? perhaps under say memory stress conditions?

daclash:
@dsharma-dc we had a non-Mayastor usermode filesystem daemon that required this change. The observable symptom was a hung filesystem IO, typically occurring in CI jobs. The specific test case did not stress memory but did perform synchronous IO. The deadlock was **very** difficult to triage because the hung kernel threads did not appear in the `dmesg` log. We applied the fix to our Mayastor forks because the similar deadlock paths exist. The detailed description below describes the two possible paths. We have not noticed any negative impact, performance or otherwise, after applying the change.


These are the potential deadlock scenarios for Mayastor based on the two known kernel deadlock scenarios for the non-Mayastor product. Deadlock seems possible and the downside to setting `PR_SET_IO_FLUSHER` is minimal. The scenarios mention `pwritev()` system calls for simplicity. `bdev_uring_writev()` uses `io_uring_prep_writev*()` but the io_uring write eventually reaches the same kernel path. [file-backed-disks] are used in scenario B but something similar is likely for "physical" disks.

[file-backed-disks]: https://openebs.io/docs/user-guides/replicated-storage-user-guide/replicated-pv-mayastor/additional-information/tips#using-file-backed-disks

Mayastor potential deadlock scenario A (deadlock on io-engine resource):

1. The workload pod mounts the Nexus bdev.
2. The workload pod executes multiple `pwritev()` system calls to a single large file that remain in the kernel page cache.
3. The workload pod executes `fdatasync()` which causes the kernel to start sending writes to the Nexus bdev.
4. The Nexus write code path encounters a page fault.
5. The page fault results in a page reclaim for the single large file.
6. The Nexus begins to execute the write associated with the page reclaim.
7. The Nexus thread executing the write in 4 is still blocked and may own resources.
8. The write in 6 may be blocked waiting for the thread in 4.
   a. The wait is on either the blocked thread itself or a resource owned by the blocked thread.

Mayastor potential deadlock scenario B (deadlock on kernel inode lock):

1. The workload pod mounts the Nexus bdev
   a. One replica must be local.
2. The workload pod executes multiple `pwritev()` system calls to a single large file that remain in the kernel page cache.
3. The workload pod executes `fdatasync()` which causes the kernel to start sending writes to the Nexus bdev.
4. The Nexus sends a write to the local replica bdev.
5. The local replica bdev makes a `pwritev()` system call to the file backed disk.
6. The `pwritev()` system call acquires the kernel inode lock.
7. The `pwritev()` system call triggers a page reclaim associated with the single large file in 2.
8. The kernel sends a write to the Nexus to satisfy the page reclaim.
9. The Nexus sends the write to the local replica bdev.
10. The local replica bdev makes a `pwritev()` system call to the file backed disk.
11. The `pwritev()` system call is blocked because the kernel inode lock is held.



The deadlocks were difficult to triage because the hung thread did not appear in the hung kernel task list.  The thread that is holding the resource has a state of `TASK_WAKEKILL` and `TASK_UNINTERRUPTIBLE`. The hung kernel task list in `dmesg` does not include task with a state that includes `TASK_WAKEKILL`.

Setting `PR_SET_IO_FLUSHER` on the io-engine process prevents the page reclaim in A5 and B7 from occurring, which avoids the deadlock.

dsharma-dc (approved):

tiagolobocastro:
@dsharma-dc also pls review https://github.com/openebs/mayastor-dependencies/pull/135

tiagolobocastro:
> @dsharma-dc also pls review [openebs/mayastor-dependencies#135](https://github.com/openebs/mayastor-dependencies/pull/135)

It's merged @daclash, you may update the module here and then issue the merge here whenever you're ready

daclash:
bors try

daclash:
bors try

daclash:
bors try-

daclash:
bors status

daclash:
bors try

tiagolobocastro:
Thank you @daclash :rocket:

daclash:
@tiagolobocastro thank you for the review and explaining the process!

PR #1914, opened 2025-10-06
chore: add staging ci for mayastor

pchandra19:


.github/workflows/image.yml
```diff
@@ -32,5 +37,16 @@ jobs:
...
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
       - name: Build and push the release images
-        run: ./scripts/release.sh
+        run: |
+          if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            nix-shell --run "./scripts/release.sh --tag ${{ inputs.tag }} --registry ghcr.io/openebs/mayastor/dev"
```
tiagolobocastro: We don't need the nix-shell here? Or if we do let's not use the default one which brings in rust compiler spdk et all which is not needed here

scripts/staging/shell.nix
```diff
@@ -0,0 +1,20 @@
...
+    oras
+    crane
+    yq-go
+    jq
+  ] ++ pkgs.lib.optional (builtins.getEnv "IN_NIX_SHELL" == "pure" && pkgs.system != "aarch64-darwin") [
+    docker
+    git
+    curl
+    nix
+    kubernetes-helm-wrapped
```
tiagolobocastro, line 17: do we need this here?

scripts/staging/validate.sh
```diff
@@ -0,0 +1,145 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+SCRIPT_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]:-"$0"}")")"
+ROOT_DIR="${SCRIPT_DIR}/../.."
+: "${PARENT_ROOT_DIR:=$ROOT_DIR}"
+
+source "$ROOT_DIR/mayastor/scripts/utils/log.sh"
```
tiagolobocastro: have you tested this? There's no such thing in this repo...
also there's no charts for release here, that's on extensions

scripts/staging/mirror-images.sh
```diff
@@ -0,0 +1,61 @@
...
+
+set -euo pipefail
+
+SCRIPT_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]:-"$0"}")")"
+ROOT_DIR="${SCRIPT_DIR}/../.."
+
+# if PARENT_ROOT_DIR is not defined use the one below
+: "${PARENT_ROOT_DIR:=$ROOT_DIR}"
+
+source "$ROOT_DIR/mayastor/scripts/utils/log.sh"
```
tiagolobocastro: also does not exist

scripts/staging/log.sh
tiagolobocastro, line 1: With new dependencies module : https://github.com/openebs/mayastor/pull/1918
We can take this from utils/dependencies instead

scripts/staging/mirror-images.sh
tiagolobocastro, line 1: also we could move this to dependencies?

scripts/staging/validate.sh
tiagolobocastro, line 1: Wonder if the entire staging actually can move to dependencies?
CC @Abhinandan-Purkait

pchandra19: I thought of doing so ..but the validate script will differ slightly in extensions as it also has to validate charts. Here and in CP there are just images (no charts)

tiagolobocastro: we can make the script accept parameters for example; if not specified then it'd do images only

PR #1915, opened 2025-10-06
Don't let 1 bad nexus lockup the entire nexus subsystem

tiagolobocastro:
refactor(grpc/nexus): don't take subsystem lock on create/destroy
    
    A nexus gets sometimes locked up during unshare, and we're not yet sure of
    how this happens (ie IO stuck)
    This change loosens the sequential creation and destruction of nexuses,
    allowing us to interact with other nexuses when another is locked up.
    
    We could also enforce sequential during the happy path, for the subsystem
    itself?
    
    We should also start reporting the stuck nexuses in some way?
    
---

    chore: update commitlint config file
    
    This should have been modified since I lasted updated the packages, but
    somehow I must have missed this!

### DIFF
--- a/io-engine/src/grpc/v1/nexus.rs
+++ b/io-engine/src/grpc/v1/nexus.rs
@@ -75,7 +75,8 @@ impl NexusService {
                     Some(g) => Some(g),
                     None => {
                         return Err(Status::deadline_exceeded(
-                            "Failed to acquire access to object within given timeout".to_string(),
+                            "Failed to acquire access to subsystem within given timeout"
+                                .to_string(),
                         ))
                     }
                 }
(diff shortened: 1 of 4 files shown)


dsharma-dc (approved):

io-engine/src/grpc/v1/nexus.rs
```diff
@@ -373,7 +374,7 @@ impl NexusRpc for NexusService {
         let ctx = GrpcClientContext::new(&request, function_name!());
         let args = request.into_inner();
 
-        self.serialized(ctx, args.uuid.clone(), true, async move {
+        self.serialized(ctx, args.uuid.clone(), false, async move {
```
Abhinandan-Purkait, line 377: What does the true/false change here?

Abhinandan-Purkait (approved):

io-engine/src/grpc/v1/nexus.rs
tiagolobocastro: whether the entire service is locked, in addition to the nexus uuid hash

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1922, opened 2025-10-19
fix: encrypted pool expansion workflow

abhilashshetty04:
We were rescanning base bdev of the Pool. Base bdev for the encrypted pool is crypto bdev which cant be resized. Now, we get the Aio bdev backing the crypto bdev.

crypto bdev gets notified via event when aio is resized after which it adjusts the blockcnt.

### DIFF
--- a/io-engine/src/lvs/lvs_store.rs
+++ b/io-engine/src/lvs/lvs_store.rs
@@ -9,14 +9,14 @@ use nix::errno::Errno;
 use pin_utils::core_reexport::fmt::Formatter;
 
 use spdk_rs::libspdk::{
-    bdev_aio_rescan, spdk_bdev_update_bs_blockcnt, spdk_blob_store, spdk_bs_free_cluster_count,
-    spdk_bs_get_cluster_size, spdk_bs_get_max_growable_size, spdk_bs_get_md_len,
-    spdk_bs_get_page_size, spdk_bs_get_used_md, spdk_bs_total_data_cluster_count, spdk_lvol,
-    spdk_lvol_opts, spdk_lvol_opts_init, spdk_lvol_store, spdk_lvs_grow_live,
-    vbdev_get_lvol_store_by_name, vbdev_get_lvol_store_by_uuid, vbdev_get_lvs_bdev_by_lvs,
-    vbdev_lvol_create_with_opts, vbdev_lvs_create, vbdev_lvs_create_with_uuid, vbdev_lvs_destruct,
-    vbdev_lvs_import, vbdev_lvs_unload, LVOL_CLEAR_WITH_NONE, LVOL_CLEAR_WITH_UNMAP,
-    LVS_CLEAR_WITH_NONE,
+    bdev_aio_rescan, bdev_uring_rescan, spdk_bdev_update_bs_blockcnt, spdk_blob_store,
+    spdk_bs_free_cluster_count, spdk_bs_get_cluster_size, spdk_bs_get_max_growable_size,
+    spdk_bs_get_md_len, spdk_bs_get_page_size, spdk_bs_get_used_md,
+    spdk_bs_total_data_cluster_count, spdk_lvol, spdk_lvol_opts, spdk_lvol_opts_init,
+    spdk_lvol_store, spdk_lvs_grow_live, vbdev_get_lvol_store_by_name,
+    vbdev_get_lvol_store_by_uuid, vbdev_get_lvs_bdev_by_lvs, vbdev_lvol_create_with_opts,
+    vbdev_lvs_create, vbdev_lvs_create_with_uuid, vbdev_lvs_destruct, vbdev_lvs_import,
+    vbdev_lvs_unload, LVOL_CLEAR_WITH_NONE, LVOL_CLEAR_WITH_UNMAP, LVS_CLEAR_WITH_NONE,
 };
 use url::Url;
 
@@ -443,7 +443,7 @@ impl Lvs {
                 Byte::from_str(&param).map_err(|error| LvsError::MaxExpansionParse {
                     msg: format!("Failed to parse max_expansion {param} as bytes: {error}"),
                 })?;
-            expand_bytes.as_u64() as f64 / capacity as f64
+            (expand_bytes.as_u64() as f64 / capacity as f64).ceil()
         } else {
             return Err(LvsError::MaxExpansionParse {
                 msg: format!("Max expansion factor {param} does not end with x or B"),
@@ -880,6 +880,7 @@ impl Lvs {
     #[tracing::instrument(level = "debug", err)]
     pub async fn grow(&self) -> Result<(), LvsError> {
         info!("{self:?}: growing lvs...");
+        let lvs_name = self.name();
 
         let disk_bdev = self
             .base_bdev()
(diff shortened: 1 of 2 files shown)


io-engine/src/lvs/lvs_store.rs
```diff
@@ -880,31 +881,48 @@ impl Lvs {
...
                     name: self.base_bdev().name().to_string(),
                 });
             }
+
+            if self.encrypted() {
+                // When the underlying AIO bdev is resized, crypto bdev receives SPDK_BDEV_EVENT_RESIZE
+                // event. The crypto bdev then adjusts its block count accordingly. To ensure that this resize
+                // operation completes before proceeding, we wait briefly for the crypto bdev to update.
+                // This delay gives the crypto bdev time to process the resize event asynchronously.
+                let rx = mayastor_sleep(std::time::Duration::from_secs(1));
```
tiagolobocastro: Can't we wait for that event to occur rather than just sleep?

io-engine/src/lvs/lvs_store.rs
```diff
@@ -880,31 +881,48 @@ impl Lvs {
...
             name: self.name().to_string(),
         })?;
+
         if url.scheme() != "malloc" {
+            let bdev = disk_bdev.name().into_cstring();
             info!(
-                "Attempting to rescan bdev {:?} part of lvs {:?}, uri {:?}",
-                self.base_bdev().name(),
+                "Attempting to rescan bdev {:?} part of lvs {:?}",
+                bdev,
```
```suggestion
                "Attempting to rescan bdev {bdev:?} part of lvs {:?}",
```

Also I see we removed the uri, can we not use `uri_str` here?

io-engine/src/lvs/lvs_store.rs
abhilashshetty04: I saw it reflecting new size after 1 second in my testing. I suppose we can wait for 2 or 3 secs with 500ms interval and see if size changes. Ill try that.

io-engine/src/lvs/lvs_store.rs
abhilashshetty04: Is it required? We are resizing the bdev right? This message has lvs and bdev name.

io-engine/src/lvs/lvs_store.rs
abhilashshetty04: If crypto and aio bdev size matches it confirms event is handled by crypto bdev and blockcnt is adjusted.

io-engine/src/lvs/lvs_store.rs
```diff
@@ -939,6 +956,31 @@ impl Lvs {
         Ok(())
     }
 
+    // When the underlying AIO bdev is resized, crypto bdev receives SPDK_BDEV_EVENT_RESIZE
+    // event. The crypto bdev then adjusts its block count accordingly. To ensure that this resize
+    // operation completes before proceeding, we wait briefly for the crypto bdev to update.
+    // This delay gives the crypto bdev time to process the resize event asynchronously.
+    async fn crypto_resized(&self) -> bool {
```
dsharma-dc:
```suggestion
    async fn crypto_vbdev_resized(&self) -> bool {
```

dsharma-dc (approved):

io-engine/src/lvs/lvs_store.rs
dsharma-dc: > Can't we wait for that event to occur rather than just sleep?

For that a way could be to perhaps wait on a oneshot receiver which is written by a callback from spdk, but that's going to complicate the implementation too much I suppose. If it was a bdev that we hold descriptor for, then it could've been easier to do via `bdev_event_callback`

io-engine/src/lvs/lvs_store.rs
tiagolobocastro: Only mentioned as you had it before, could be useful

io-engine/src/lvs/lvs_store.rs
```diff
@@ -880,31 +881,47 @@ impl Lvs {
...
-                "Attempting to rescan bdev {:?} part of lvs {:?}, uri {:?}",
-                self.base_bdev().name(),
+                "Attempting to rescan bdev {:?} part of lvs {:?}",
+                bdev,
                 self.name(),
-                self.base_bdev().bdev_uri_str().unwrap_or_default()
             );
 
-            let errno = unsafe { bdev_aio_rescan(cname.as_ptr() as *mut std::os::raw::c_char) };
+            let errno = unsafe { bdev_aio_rescan(bdev.as_ptr() as *mut std::os::raw::c_char) };
```
tiagolobocastro: oh just noticed something here, what if the device is a pcie or if it's an uring? Then `bdev_aio_rescan` is not correct?

io-engine/src/lvs/lvs_store.rs
```diff
@@ -939,6 +956,31 @@ impl Lvs {
...
+            let disk_bdev = base_bdev
+                .crypto_base_bdev()
+                .map(Bdev::new)
+                .unwrap_or_else(|| self.base_bdev());
+            let disk_bdev_size = disk_bdev.num_blocks() * disk_bdev.block_len() as u64;
+            let crypto_bdev_size = base_bdev.num_blocks() * base_bdev.block_len() as u64;
+            if crypto_bdev_size == disk_bdev_size {
+                return true;
+            } else {
+                let rx = mayastor_sleep(std::time::Duration::from_millis(500));
```
tiagolobocastro: How long does it typically take? If less then it might be worth sleeping less time with more max retries to avoid blocking `gRPC`

```suggestion
                let rx = mayastor_sleep(std::time::Duration::from_millis(100));
```

io-engine/src/lvs/lvs_store.rs
dsharma-dc: good catch. Only aio and uring support rescan AFAIK, so need to handle accordingly.

io-engine/src/lvs/lvs_store.rs
abhilashshetty04: It completes at 2nd iteration. So takes more then 500ms.

io-engine/src/lvs/lvs_store.rs
abhilashshetty04: Added it back.

io-engine/src/lvs/lvs_store.rs
abhilashshetty04: Handled uring now. There is no pcie rescan function in SPDK. We might have to recreate bdev. Is pcie expand possible anyway?

io-engine/src/lvs/lvs_store.rs
tiagolobocastro: I guess you don't know if it completes at 2nd iteration? You slept 500ms but it could have completed anytime?

io-engine/src/lvs/lvs_store.rs
```diff
@@ -879,32 +880,51 @@ impl Lvs {
...
+            let bdev = disk_bdev.name().into_cstring();
+            info!("Attempting to rescan bdev: {uri_str} part of lvs {lvs_name}");
 
-            let errno = unsafe { bdev_aio_rescan(cname.as_ptr() as *mut std::os::raw::c_char) };
+            let errno = if url.scheme() == "uring" {
+                unsafe { bdev_uring_rescan(bdev.as_ptr() as *mut std::os::raw::c_char) }
+            } else if url.scheme() == "aio" {
+                unsafe { bdev_aio_rescan(bdev.as_ptr() as *mut std::os::raw::c_char) }
+            } else {
+                0
```
tiagolobocastro: then above rather than checking for malloc we should check for uring/aio only?

io-engine/src/lvs/lvs_store.rs
tiagolobocastro: I think it's possible yes, nvme namespaces are configurable.
Easiest way to test is probably by creating a pool on top of an nvme-tcp since you can simply use mayastor and resize

io-engine/src/lvs/lvs_store.rs
abhilashshetty04: No i had added loggers before just to see both bdev size. I'm okay with sleeping less time with more retries.

tiagolobocastro: Let's use less sleep time with more retries then

tiagolobocastro (approved):

abhilashshetty04:
bors merge

abhilashshetty04:
bors cancel

abhilashshetty04:
bors merge

PR #1932, opened 2025-12-02
fix(nvme): create nvme bdev with spdk_nvme_ctrlr_opts

tiagolobocastro:
Setup and pass spdk_nvme_ctrlr_opts as a parameter when creating the bdev. This is required because the nvme create bdev function no longer handles a nullptr parameter.


abhilashshetty04 (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #1933, opened 2025-12-03
fix(nvmf/hosts): handle existing allowed host einval

tiagolobocastro:
refactor(nvmf/hosts): reuse existing connected_hosts
    
    Rather than re-loop the hosts, simply reuse the connected hosts,
    and disconnect any which is not on the new allowed list.
    
---

    fix(nvmf/hosts): handle existing allowed host einval
    
    On the latest spdk, we cannot add the same host back as it's now rejecting
    it with einval.
    Instead, we check the existing hosts and add only the new ones.


abhilashshetty04 (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #1935, opened 2025-12-09
fix(bdev/destroy): unload pool when disk is hotremoved

tiagolobocastro:
When destroying a bdev as part of a hot-removal, the uri used to destroy the bdev may not match exactly the alias which was added on the creation.
In this case we'll leak a bdev alias which will cause issues if we attempt to re-create the bdev as it'll have a stale alias pointer.

We've modified spdk to automatically delete all aliases on unregister, and so we don't need to manually remove them, thus fixing this scenario.


Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1937, opened 2025-12-16
test(bdd/rebuild): don't wait too long for rebuilds

tiagolobocastro:
I'm not entirely sure what the test is checking for, but it has a hardcoded sleep which doesn't work if rebuild is too fast. Reduced it for now, though probably we might want a different approach.

Also modified the wait for completion to use a retrying approach, which is already better than the hardocded sleep there.

Also noticed the rebuilds don't always succeed! I suspect this is because we're adding the same replica too quickly before the child was removed cleanly. This should probably be fixed.


abhilashshetty04 (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #1942, opened 2026-01-29
Expose DiskPool Errors and Alerts

tiagolobocastro:
feat: implement diskpool errors info and alerts

    It's only implemented for Lvs pools since we are using the
    functionality in the SPDK base bdev.

---

    feat: show alert status and errors on tabled output

---

    feat: add clear error subcommand to the cli

---

    refactor: pull spdk-rs error stats

    Add new reset_stats_ext to retrieve stats with errors.

---

    chore: pull api submodule changes

    Pulls the api changes for the pool errors.
    Adds blanket implementation.

### DIFF
--- a/io-engine/src/bin/io-engine-client/v1/pool_cli.rs
+++ b/io-engine/src/bin/io-engine-client/v1/pool_cli.rs
@@ -729,9 +729,10 @@ async fn list(mut ctx: Context, matches: &ArgMatches) -> crate::Result<()> {
 }
 
 fn pool_state_to_str(idx: i32) -> &'static str {
-    match v1rpc::pool::PoolState::try_from(idx).unwrap() {
+    match v1rpc::pool::PoolState::try_from(idx).unwrap_or_default() {
         v1rpc::pool::PoolState::PoolUnknown => "unknown",
         v1rpc::pool::PoolState::PoolOnline => "online",
+        v1rpc::pool::PoolState::PoolSuspected => "suspected",
         v1rpc::pool::PoolState::PoolDegraded => "degraded",
         v1rpc::pool::PoolState::PoolFaulted => "faulted",
     }
--- a/io-engine/src/bin/io-engine-client/v1/pool_cli.rs
+++ b/io-engine/src/bin/io-engine-client/v1/pool_cli.rs
@@ -5,7 +5,7 @@ use crate::{
 use byte_unit::Byte;
 use clap::{Arg, ArgMatches, Command};
 use colored_json::ToColoredJson;
-use io_engine_api::v1 as v1rpc;
+use io_engine_api::{v1 as v1rpc, v1::pool::Pool};
 use snafu::ResultExt;
 use std::{convert::TryFrom, str::FromStr};
 use strum::VariantNames;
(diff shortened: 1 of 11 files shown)


tiagolobocastro:
bors try

tiagolobocastro:
bors try

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
```diff
@@ -728,11 +760,55 @@ async fn list(mut ctx: Context, matches: &ArgMatches) -> crate::Result<()> {
...
+            println!("pool stats error cleared, current: {:#?}", pool.errors);
+        }
+    };
+
+    Ok(())
+}
+
 fn pool_state_to_str(idx: i32) -> &'static str {
-    match v1rpc::pool::PoolState::try_from(idx).unwrap() {
+    match v1rpc::pool::PoolState::try_from(idx).unwrap_or_default() {
```
abhilashshetty04, line 794: Should we panic here as we know it should not fail unwrap()?

tiagolobocastro: If it's a new state, then show as "PoolUnknown" which is the default, there's no need for panic

io-engine/src/core/env.rs
```diff
@@ -280,6 +280,36 @@ pub struct MayastorCliArgs {
...
+pub struct PoolCliArgs {
+    /// I/O error count threshold.
+    /// After this many errors a pool alert is raised as Warning.
+    #[clap(long, default_value = "64")]
+    pub io_error_threshold: u64,
+
+    /// I/O stall deadline.
+    /// If an I/O is stuck longer than this period, then the pool is considered stalled and a
+    /// Critical alert is raised.
+    /// The pool disk will also be reset and the stall will be cleared once complete and
```
abhilashshetty04, line 300:
```suggestion
    /// The pool disk will also be reset and the stall will be cleared once IO completes and
```

io-engine/src/core/env.rs
```diff
@@ -280,6 +280,36 @@ pub struct MayastorCliArgs {
...
+    /// I/O flows again.
+    #[clap(long, default_value = "222s", value_parser = humantime::parse_duration)]
```
abhilashshetty04: Shall we make io_stall_deadline a little lower? 222 seconds is very long for IO to be timing out.

io-engine/src/core/env.rs
```diff
@@ -280,6 +280,36 @@ pub struct MayastorCliArgs {
...
+    pub io_stall_deadline: Duration,
+
+    /// I/O stall transitions threshold.
+    /// After this many transitions within the window, a pool alert is raised as Warning.
```
abhilashshetty04, line 306: Also these, 3 transitions in 6h seems may not seem flaky. What if Pool goes through first 3 transitions very quickly then is completely fine for the rest of the window.  We are keeping a recovered Pool in Warning state.

Can we do for example: 12 transitions in 1 hr ?

io-engine/src/core/env.rs
```diff
@@ -280,6 +280,36 @@ pub struct MayastorCliArgs {
...
+    #[clap(long, default_value = "3")]
+    pub io_stall_transition_threshold: u64,
+
+    /// I/O stall transitions threshold.
```
abhilashshetty04: /// I/O stall transition window.
Time window (in seconds) during which stall ↔ resume state transitions are tracked for flakiness detection.

io-engine/src/core/env.rs
tiagolobocastro: the intention here is waiting for the reset itself to complete as a first step

io-engine/src/core/env.rs
tiagolobocastro: We're not detecting timeouts here, we're trying to detect stalls
That being said, we can have a lower value here by default, and specify production values on the helm chart

abhilashshetty04 (approved):

io-engine/src/core/env.rs
```diff
@@ -280,6 +280,38 @@ pub struct MayastorCliArgs {
...
+    #[clap(long, default_value = "64")]
+    pub io_error_threshold: u64,
+
+    /// I/O stall deadline.
+    /// If an I/O is stuck longer than this period, then the pool is considered stalled and a
+    /// Critical alert is raised.
+    /// The pool disk will also be reset and the stall will be cleared once complete and
+    /// I/O flows again.
+    #[clap(long, default_value = "30s", value_parser = humantime::parse_duration)]
+    pub io_stall_deadline: Duration,
```
niladrih: This can be a humantime::Duration directly, i.e.

```suggestion
    #[clap(long, default_value = "30s")]
    pub io_stall_deadline: humantime::Duration,
```

io-engine/src/core/env.rs
```diff
@@ -280,6 +280,38 @@ pub struct MayastorCliArgs {
...
+    /// I/O stall transitions threshold.
+    /// After this many transitions within the window, a pool alert is raised as Warning.
+    #[clap(long, default_value = "3")]
+    pub io_stall_transition_threshold: u64,
+
+    /// I/O stall transitions window.
+    /// Time window during which stall ↔ resume state transitions are tracked
+    /// for flakiness detection.
+    #[clap(long, default_value = "3h", value_parser = humantime::parse_duration)]
+    pub io_stall_transition_window: Duration,
```
niladrih: Same as above, don't have to register the value_parser, humantime::Duration implements `FromStr`, so it works out of the box.

io-engine/src/core/env.rs
```diff
@@ -280,6 +280,38 @@ pub struct MayastorCliArgs {
...
+    #[clap(flatten)]
+    pub pool: PoolCliArgs,
+}
+
+/// DiskPool related arguments.
+#[derive(Debug, clap::Parser, Clone)]
+pub struct PoolCliArgs {
+    /// I/O error count threshold.
+    /// After this many errors a pool alert is raised as Warning.
+    #[clap(long, default_value = "64")]
```
niladrih: nit: can just use default_value_t here.

niladrih (approved):

io-engine/src/core/env.rs
tiagolobocastro: Indeed, done

io-engine/src/core/env.rs
tiagolobocastro: I think I used parser thinking about adding min/max validation, but didn't end up implementing it
I'll leave as default for now, and maybe @abhilashshetty04 might add validations since my stall changes are just for scaffolding

tiagolobocastro:
bors merge

PR #1945, opened 2026-02-04
Nvmf - don't panic when subsystem pause races subsystem stop

tiagolobocastro:
test(nvmf): race stop with resume and pause

    There should be no crashes now when racing those.
    However, resume after stop will leave the subsystem active even
    though no share uri will be reported.
    This is a minor issue, though we should still address this.

---

    fix(nvmf): don't panic when pause races stop

    Pausing and/or stopping a subsystem may take some time if I/Os
    are stuck or haven't timed out.
    This may widen a race between between stop and pause where pause
    completes with an error if the subsystem is stopped but not yet
    deactivated.

    We now handle this by simply considering pause complete in this
    case since the subsystem is essentially paused.
    However this opens up a race where a resume may start a subsystem
    which was previously stopped but not yet deactivated!!
    I'm proposing that we handle this within SPDK and reject resumes
    when the subsystem is inactive/stopped.

---

    refactor(nvmf): use implicitly named arguments

    Also slightly simplify the pause handling with early break

---

    feat(tracing): enable dtracing and spdk tracing

    Pulls spdk-rs which allows us to configure spdk with dtracing and also
    parse cli args to enable spdk tracing.

    This can be very useful to debug spdk, ie subsystem state change events

---

    feat(io-engine): add coredump enablement

    Enables coredump by default in debug builds.
    Using spdk value of 5GiB limit if specific limit is not specified.

### DIFF
--- a/io-engine-tests/src/compose/mod.rs
+++ b/io-engine-tests/src/compose/mod.rs
@@ -74,6 +74,7 @@ impl<'a> MayastorTest<'a> {
     }
 
     pub fn new_ex(args: MayastorCliArgs, log_level: Option<&str>) -> MayastorTest<'static> {
+        io_engine::coredump::enable(io_engine::coredump::DEFAULT_CORE_LIMIT).ok();
         let (tx, rx) = bounded(1);
         mayastor_test_init_ex(args.log_format.unwrap_or_default(), log_level);
         let thdl = std::thread::Builder::new()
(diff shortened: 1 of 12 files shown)


tiagolobocastro:
bors try

abhilashshetty04 (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
@mergify backport release/2.9

PR #1946, opened 2026-02-05
CherryPick PR1945

tiagolobocastro:
fix(nvmf): don't panic when pause races stop

    Pausing and/or stopping a subsystem may take some time if I/Os
    are stuck or haven't timed out.
    This may widen a race between between stop and pause where pause
    completes with an error if the subsystem is stopped but not yet
    deactivated.

    We now handle this by simply considering pause complete in this
    case since the subsystem is essentially paused.
    However this opens up a race where a resume may start a subsystem
    which was previously stopped but not yet deactivated!!
    I'm proposing that we handle this within SPDK and reject resumes
    when the subsystem is inactive/stopped.

---

    refactor(nvmf): use implicitly named arguments

    Also slightly simplify the pause handling with early break

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev_children.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev_children.rs
@@ -1145,7 +1145,7 @@ impl<'n> Nexus<'n> {
 
                 // Step 5: Mark nexus as shutdown.
                 // Note: we don't persist nexus's state in ETCd as nexus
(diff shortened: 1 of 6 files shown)


Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

PR #1947, opened 2026-02-06
fix(nvmf/stuck): disconnect when target is in bad state

tiagolobocastro:
It was seen that sometimes the connection to the replica is in an odd state where adminq keeps failing to send the keep alive, and the nexus subsystem pause is hung.

The theory is that there are stuck IOs somehow. Prior to 2.11 the timing out of replica IOs was not always correct so this could somehow be playing a part here.

To attempt to unstuck the hang, we now wait up to the adminq keep alive timeout and then in case the nexus is pausing we now force disconnect the bad replica.


Abhinandan-Purkait (approved):

niladrih (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

tiagolobocastro:
@mergify backport develop

PR #1948, opened 2026-02-11
adding in memory structure to store pool information

abhilashshetty04:


io-engine/src/pool_information.rs
```diff
@@ -0,0 +1,41 @@
...
+/// in memory structure containing pool runtime information.
+pub static POOL_INFO: Lazy<RwLock<HashMap<String, RwLock<PoolInfo>>>> =
+    Lazy::new(|| RwLock::new(HashMap::new()));
+
+#[derive(Default, Debug, Clone, PartialEq, Eq)]
+pub struct PoolInfo {
+    /// Set when Pools io is under stall.
+    pub io_stalled: bool,
+    /// Contains list of timestamps when pool went into stall state.
+    pub transition_timestamps: Vec<Instant>,
```
dsharma-dc: Would be more useful perhaps as a pair of `Instant`(when it went to stall) and `Duration`(of stall). or the duration of stall is prefixed to `stall_transition_window` ?

abhilashshetty04: What do we do with the duration of stall? Will we act differently on longer stall duration. `stall_deadline` is 30s now by default. So if IO has not completed for 30s then we mark `io_stalled: true`.

https://github.com/openebs/mayastor/pull/1942/changes#diff-33eba726b9ecc1920d886fa6276f0d4287c6bfd1075236fa9f27fa6beea6028eR302

Currently, the list will have timestamps when we transition from `!io_stall -> io_stall` state.  During list_pool we will use `update_transition_timestamp` to exclude transitions which do not fit in the `stall_transition_window`. The idea was to eventually get the Pool in healthy state when IOs are no longer stalling.

Here is a bdd for this

  ( stall_transition_threshold and stall_transition_threshold is going to be helm parameter
   on IO-engine. This will allow administrator to set how many stall transitions are tolerable.
   If stall_transition_window is 30 mins and stall_transition_threshold is 10. Then at any point
   in time if there are 10 state transitions in 30 mins we will mark the Pool as Warning. Pool will
   get into Suspected state, Pool will be less preferred for scheduling. More then 1 transition in
   stall_transition_window will start raising Attention alert on the Pool.)

  Scenario Outline: DiskPool PoolAlertStatus set to Critical as stall_transition_threshold exceeds
    Given a Pool created with stall_transition_window set to 300s
    And stall_transition_threshold set to 5
    And stall_deadline set to 10s
    When a replica is scheduled on the Pool which has its target on the same node
    And FIO is scheduled on that volume
    When the DiskPool backend device I/O is stuck at time t
    # Consider control plane pool interval here also as before. Can be set very low as this is time
    # sensitive test.
    Then DiskPool status is set to Suspected eventually
    And the DiskPool PoolAlertStatus is set to Critical
    And PoolAlert is set to IoStall
    And the DiskPool backend device I/O resumes
    Then DiskPool status is Online eventually
    And the DiskPool PoolAlertStatus is set to Healthy
    Then the DiskPool backend device I/O is stuck at time t + 150s
    And IO resumes after <stall_deadline>
    And the DiskPool backend device IO gets stuck and resumes 2 more times before t + 300s
    And the DiskPool PoolAlertStatus is set to Attention
    And PoolAlert is set to IoStallIntermittent
    When the DiskPool backend device I/O is stuck at time t + 320s
    # Even though we transitioned 5 times, The first state change is not valid as its out of the
    # stall_transition_window period.
    Then DiskPool status is set to Suspected eventually
    And the DiskPool PoolAlertStatus is set to Critical
    And PoolAlert is set to IoStall
    Then the DiskPool backend device I/O resumes
    And DiskPool status is Online eventually
    And the DiskPool PoolAlertStatus is set to Attention
    And PoolAlert is set to IoStallIntermittent
    # This transition will trigger flaky check. Pool PoolAlertStatus will be Critical while IO is
    # stalled. When it recovers it will be set to Warning.
    When the DiskPool backend device I/O is stuck at time t + 350s
    Then DiskPool status is set to Suspected eventually
    And the DiskPool PoolAlertStatus is set to Critical
    And PoolAlert is set to IoStall
    And the DiskPool backend device I/O resumes
    Then DiskPool status is Online eventually
    And the DiskPool PoolAlertStatus is set to Warning
    And PoolAlert is set to IoStallIntermittentExc
    When t + 480s elapse
    # Oldest state transition record expires here
    Then the DiskPool PoolAlertStatus is set to Attention
    And PoolAlert is set to IoStallIntermittent
    When t + 650s elapse
    # All records expire here
    Then the DiskPool PoolAlertStatus is set to Healthy eventually

io-engine/src/lvs/lvs_store.rs
```diff
@@ -314,6 +316,10 @@ impl Lvs {
         } else {
             lvs.share_all().await;
             info!("{:?}: existing lvs imported successfully", lvs);
+            {
```
tiagolobocastro: you don't need a scope here

io-engine/src/lvs/lvs_store.rs
```diff
@@ -556,6 +562,10 @@ impl Lvs {
         match Self::lookup(&args.name) {
             Some(pool) => {
                 info!("{:?}: new lvs created successfully", pool);
+                {
```
tiagolobocastro: same here

io-engine/src/lvs/lvs_store.rs
```diff
@@ -314,6 +316,10 @@ impl Lvs {
         } else {
             lvs.share_all().await;
             info!("{:?}: existing lvs imported successfully", lvs);
+            {
+                let mut cache = get_write_lock();
+                cache.insert(name.to_string(), RwLock::new(PoolInfo::default()));
```
tiagolobocastro: Could you add a method to Lvs and put this there? Then just call it rather than repeat this code below

io-engine/src/lvs/lvs_store.rs
```diff
@@ -726,6 +736,14 @@ impl Lvs {
             base_bdev.name()
         );
 
+        {
+            let mut cache = get_write_lock();
+            cache.remove(self.name());
```
tiagolobocastro: hmm you can't use self.name here, this is after pool destruction

io-engine/src/lvs/lvs_store.rs
```diff
@@ -726,6 +736,14 @@ impl Lvs {
...
+            if cache.get(self.name()).is_none() {
```
tiagolobocastro: why is this needed?

io-engine/src/lvs/lvs_store.rs
```diff
@@ -834,6 +852,14 @@ impl Lvs {
             self_str
         );
 
+        {
+            let mut cache = get_write_lock();
+            cache.remove(self.name());
```
tiagolobocastro: Same here, please in function but specify the name as you can't use self here

io-engine/src/pool_information.rs
```diff
@@ -0,0 +1,41 @@
+use once_cell::sync::Lazy;
+use parking_lot::{RwLock, RwLockReadGuard, RwLockWriteGuard};
+use std::{
+    collections::HashMap,
+    time::{Duration, Instant},
+};
+
+/// in memory structure containing pool runtime information.
```
tiagolobocastro:
```suggestion
/// In-memory structure containing pool runtime information.
```

io-engine/src/pool_information.rs
```diff
@@ -0,0 +1,41 @@
...
+pub static POOL_INFO: Lazy<RwLock<HashMap<String, RwLock<PoolInfo>>>> =
+    Lazy::new(|| RwLock::new(HashMap::new()));
+
+#[derive(Default, Debug, Clone, PartialEq, Eq)]
+pub struct PoolInfo {
```
tiagolobocastro, line 14: Missing doc comment

io-engine/src/pool_information.rs
```diff
@@ -0,0 +1,41 @@
...
+use std::{
+    collections::HashMap,
+    time::{Duration, Instant},
+};
+
+/// in memory structure containing pool runtime information.
+pub static POOL_INFO: Lazy<RwLock<HashMap<String, RwLock<PoolInfo>>>> =
+    Lazy::new(|| RwLock::new(HashMap::new()));
+
+#[derive(Default, Debug, Clone, PartialEq, Eq)]
```
tiagolobocastro: Why is Eq needed?

io-engine/src/pool_information.rs
```diff
@@ -0,0 +1,41 @@
...
+pub struct PoolInfo {
+    /// Set when Pools io is under stall.
+    pub io_stalled: bool,
+    /// Contains list of timestamps when pool went into stall state.
+    pub transition_timestamps: Vec<Instant>,
+}
+
+impl PoolInfo {
+    /// Retains transition timestamps which havent elapsed stall_transition_window.
```
tiagolobocastro:
```suggestion
    /// Retains only transition timestamps that are within the stall_transition_window.
```

io-engine/src/pool_information.rs
```diff
@@ -0,0 +1,41 @@
...
+    pub fn update_transition_timestamp(&mut self, stall_transition_window: Duration) {
+        self.transition_timestamps
+            .retain(|ts| ts.elapsed() < stall_transition_window);
+    }
+}
+
+/// Returns write lock of the hashmap.
+pub fn get_write_lock() -> RwLockWriteGuard<'static, HashMap<String, RwLock<PoolInfo>>> {
```
tiagolobocastro: Maybe drop the `get_` since it's not typically used in rust.
Also maybe name it as what is is, rather than the type, ex:

```suggestion
pub fn pools_list() -> RwLockWriteGuard<'static, HashMap<String, RwLock<PoolInfo>>> {
```

io-engine/src/pool_information.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+/// Returns read lock of the hashmap.
+pub fn get_read_lock() -> RwLockReadGuard<'static, HashMap<String, RwLock<PoolInfo>>> {
+    POOL_INFO.read()
+}
+
+/// Initiatalizes in memory structure.
+pub fn init_pool_health_cache() {
+    Lazy::force(&POOL_INFO);
```
tiagolobocastro: Why is this needed, can't we just let it lazy initialise?

io-engine/src/pool_information.rs
tiagolobocastro: Could this grow too much if the flaky is too often vs the time window?

io-engine/src/lvs/lvs_store.rs
abhilashshetty04: It worked though. Not sure how. I will use a different variable.

io-engine/src/lvs/lvs_store.rs
abhilashshetty04: I may have to use Lvs::remove_cache something. Is that okay?

io-engine/src/lvs/lvs_store.rs
tiagolobocastro: If nothing modified the pointed data, it might work, that's why it's undefined behavior

io-engine/src/lvs/lvs_store.rs
tiagolobocastro: or `Lvs::remove_info` ?

io-engine/src/pool_information.rs
abhilashshetty04: Yes, it may. Do we want to limit the vec size to stall_transition_threshold. For example 3x of stall_transition_threshold?

abhilashshetty04: Now we create Vec for size 10x of stall_transition_threshold

io-engine/src/grpc/v1/pool.rs
```diff
@@ -461,13 +462,25 @@ impl Deref for PoolErrorsNt {
...
 }
 impl PoolErrorsNt {
-    fn new(environ: &MayastorEnvironment, stats: &BdevErrorStats) -> Self {
+    fn new(pool: &str, environ: &MayastorEnvironment, stats: &BdevErrorStats) -> Self {
+        let mut io_stalled = false;
+        let mut io_stall_transition_count: u64 = 0;
+
+        let cache = pool_info_read();
+
+        if let Some(pool_lock) = cache.get(pool) {
```
tiagolobocastro, line 471: Should we expose a `get_or_default()`, which would insert an entry if not found?

io-engine/src/lvs/lvs_store.rs
```diff
@@ -726,6 +742,8 @@ impl Lvs {
             base_bdev.name()
         );
 
+        Lvs::remove_info(&pool);
```
tiagolobocastro: This may leak entries as when device is removed with error, unload completes with an error code, but pool is unloaded

io-engine/src/pool_information.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// Set when Pools io is under stall.
+    pub io_stalled: bool,
+    /// Contains list of timestamps when pool went into stall state.
+    pub transition_timestamps: Vec<Instant>,
+}
+
+impl Default for PoolInfo {
+    fn default() -> Self {
+        let cli_args = MayastorEnvironment::global();
+        let capacity = cli_args.pool_args.io_stall_transition_threshold * 10;
```
tiagolobocastro: This may not be doing what you think, this is just setting initial capacity, it's not limiting the max pushes.
You should just use default, and then on the insertions you have to ensure oldest is removed before you push new entry

abhilashshetty04: Right, I will use VecDeque. It allows better way to remove from the front.

io-engine/src/pool_information.rs
```diff
@@ -0,0 +1,48 @@
...
+/*impl Default for PoolInfo {
+    fn default() -> Self {
+        let cli_args = MayastorEnvironment::global();
+        let capacity = cli_args.pool_args.io_stall_transition_threshold * 10;
+        Self {
+            io_stalled: false,
+            transition_timestamps: VecDeque::with_capacity(capacity as usize),
+        }
+    }
+}*/
```
tiagolobocastro:
```suggestion
```

tiagolobocastro (approved):

io-engine/src/grpc/v1/pool.rs
abhilashshetty04: It does not seem correct to modify from the list call.

abhilashshetty04:
bors try

Abhinandan-Purkait (approved):

abhilashshetty04:
bors merge

PR #1949, opened 2026-02-12
SPDK DiskError and Hot-removal patches

tiagolobocastro:
test: add lvs hot-removal
    
    The bdev aio and uring modules now hot-remove the bdev when they
    detect the backing device is hot-removed.
    This happens when they complete an io with -ENODEV, -EIO or positive error.
    
    Add a test which leverages ublk to achieve this by removing the ublk device.
    
---

    feat(ublk): add ublksrv nix package

---

    test: add disk errors test
    
    Makes uses of an LVM lvol allowing us to set it up to yield errors
    without detaching the device.
    
    We "import" the gRPC pool types, allowing us to test we're returning
    the correct states, alerts and errors.

---

    fix(lvm): cleanup stable dm entries when creating a vg
    
---

    test: don't panic twice on environment drop
    
    Also improve some testing asserts to output more information.
    
---

    build: pulls latest spdk-rs to include hotremoval fixes
    
---

    feat(lvm): add device-mapper utilities
    
    Adds methods for suspending, resuming, retrieving the table and
    updating the table.
    This can be useful for testing for example.
    
---

    feat(lvm): support commands with inputs
    
---

    fix(lvm): close fd above stderr
    
    On invocation, lvm requires that only the standard file descriptors stdin,
    stdout and stderr are available. If others are found, they get closed
    and messages are issued warning about the leak.
    
    Previously we were silencing the errors only, now we close the range from
    3 to 1024.
    todo: pick higher range?

---

    refactor: implement default for cli args
    
    Makes it less awkward to use.
    Also we should consider modifying the MayastorEnvironment.
    
---

    refactor: expose lvm vg create lvol
    
    This will allow us to use concrete types for testing.
    
    As a bonus we also avoid looking up the VG twice.
    Though something seems slightly off, there's some indirection
    back and forth from vg to lv.
    
---

    fix(lvm): handle lvol already exists error
    
    If already exists, then validates the parameters.
    
---

    feat: allow running the lvm code outside of spdk context
    
    Adds a no_spdk variable to the PoolArgs, which is only usable
    for the LVM backend currently, though could extend to ZFS ex.
    
    As a simplistic impl for now we won't attach the mayastor tag
    in this case. Probably not ideal, but for now should be ok.
    
    Also the lvm commands were being trampolined to the spdk reactor.
    This is easily removed by simply checking if we're in and spdk
    thread, and if not then no need to trampoline anyway so this
    was an easy win.

### DIFF
--- a/io-engine/src/lvm/dm_setup.rs
+++ b/io-engine/src/lvm/dm_setup.rs
@@ -1,0 +1,127 @@
+/// Possible device states are SUSPENDED, ACTIVE, and READ-ONLY.
+/// The dmsetup suspend command sets a device state to SUSPENDED.
+/// When a device is suspended, all I/O operations to that device stop.
+/// The dmsetup resume command restores a device state to ACTIVE.
+#[derive(Eq, PartialEq)]
+pub enum DmState {
+    Suspended,
+    Active,
+    ReadOnly,
+    Unknown(String),
+}
+impl From<String> for DmState {
+    fn from(value: String) -> Self {
+        let state = value.to_uppercase();
+        match state.as_str() {
+            "SUSPENDED" => Self::Suspended,
+            "ACTIVE" => Self::Active,
+            "READ-ONLY" => Self::ReadOnly,
+            _ => Self::Unknown(state),
+        }
+    }
+}
+impl std::fmt::Display for DmState {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let state = match self {
+            DmState::Suspended => "Suspended",
+            DmState::Active => "Active",
+            DmState::ReadOnly => "Read-Only",
+            DmState::Unknown(state) => state.as_str(),
+        };
+        write!(f, "{state}")
+    }
+}
+
+/// A device-mapper setup table.
+#[derive(Debug)]
+pub struct DmTable(pub String);
+
+impl std::fmt::Display for DmTable {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+/// Helper methods for running dmsetup commands.
+pub struct DmSetup;
+
+impl DmSetup {
+    /// Suspends the device.
+    ///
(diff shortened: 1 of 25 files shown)


tiagolobocastro:
bors try

tiagolobocastro:
bors try

io-engine/src/lvm/dm_setup.rs
```diff
@@ -0,0 +1,148 @@
...
+impl std::fmt::Display for DmTable {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+/// Helper methods for running dmsetup commands.
+pub struct DmSetup;
+
+impl DmSetup {
```
abhilashshetty04, line 48: There is no create method yet?

tiagolobocastro: No,  these are associated functions, which are called via the `lvm::LogicalVolume`

abhilashshetty04 (approved):

io-engine/src/lvm/dm_setup.rs
abhilashshetty04: We will have to create via lvm module for using dmsetup in lvstore specific case also? I have seen that we can do dmsetup on block device directly.

SECS=$(blockdev --getsz /dev/sdd)
 
dmsetup create newdisk --table "0 ${SECS} linear /dev/sdd 0"

tiagolobocastro: Yeah, this was not intended for general use case since on testing we have no guarantees that we have a certain block device available, like `/dev/sdd`.
So instead we can easily make one using the LVM functionality, since we already have that code for pools anyhow.
Although I guess we could also use loop devices, that would require some more code to ensure it's cleaned up cleanly in case of failures

abhilashshetty04: Yes, i think we already have loop device based tests.

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #1952, opened 2026-03-03
Improve the io-engine cli and logging ux

tiagolobocastro:
fix: set correct error code for max nvmf subsystem
    
    In this case we now return out_of_range status code with
    an errno of EMLINK.

---

    fix: expose errno in the grpc status metadata
    
    The grpc status codes are very limited, and makes it hard to determine
    what has actually happened.
    In the control-plane we've take a different approach where we return
    the status as a proto message.
    To avoid changing the entire api here, we're now exposing the errno
    as a metadata header.
    
    Also, create a better mapping from EIO/EILSEQ to data_loss code.
    
---

    refactor: change Errno trait to use reference
    
    This makes using it more flexible and allows us to consume the
    value in other paths whilst taking the errno a priori.
    
---

    style: tidy up error logging with display
    
    Some of the errors were still using the debug, and relying on
    the verbose error trait.
    This would in times generate very long error messages with repeated
    errors due to a mix of Display and Debug
    I tried to move everything to Display and not rely on the verbose
    trait at all.
    Also removed some unused errors fields and variants.

---

    style(pool/lvs): don't trace the pool args twice during create
    
    The pool create top method already sets the args in the span, so
    if we add it again we get a huge line with both which are the same!
    
---

    fix(pool/lvs): include generic errno error
    
    When an lvs pool operation fails with a gereric errno, that was
    not being displayed which is not helpful.
    
---

    build: use existing config from nix-shell and setup vscode
    
    Sometimes we override the nix-shell args, which leads to a different
    set of libs etc
    When we open an editor from inside this shell or when we run another
    nix-shell command, then we have to pick up the initial args, otherwise
    we may get incorrect or unexpected results.
    Also, setup the vscode env variables so when we're using an ssh connection
    for example, it'd pick up the correct variables as well.

---

    style: prefix cli flattened structs with prefix
    
    Adds prefix to spdk-tracing and pool error thresholds.
    This should help clarify what the values as for.
    
---

    style: improve cli help as well as default logging
    
    Improves the help menu to make it easier to use.
    Also de-clutters the default span logging when running in debug.
    Fixes the hostname being set in json logging.

### DIFF
--- a/io-engine/src/core/env.rs
+++ b/io-engine/src/core/env.rs
@@ -303,10 +303,18 @@ pub struct MayastorCliArgs {
 #[derive(Debug, clap::Parser, Clone)]
 pub struct SpdkTracingArgs {
     /// Number of trace entries per lcore.
-    #[clap(long, env = "SPDK_TRACING_ENTRIES", default_value_t = 0)]
+    #[clap(
+        long = "spdk-tracing-entries",
+        env = "SPDK_TRACING_ENTRIES",
+        default_value_t = 0
+    )]
     pub entries: u64,
     /// Number of user created threads.
-    #[clap(long, env = "SPDK_TRACING_THREADS", default_value_t = 1)]
+    #[clap(
+        long = "spdk-tracing-threads",
+        env = "SPDK_TRACING_THREADS",
+        default_value_t = 1
+    )]
     pub threads: u32,
 }
 impl Default for SpdkTracingArgs {
@@ -320,7 +328,11 @@ impl Default for SpdkTracingArgs {
 pub struct PoolCliArgs {
     /// I/O error count threshold. {n}
     /// After this many errors a pool alert is raised as Warning.
-    #[clap(long, env, default_value_t = 8)]
+    #[clap(
+        long = "pool-io-error-threshold",
+        env = "POOL_IO_ERROR_THRESHOLD",
+        default_value_t = 8
+    )]
     pub io_error_threshold: u64,
 
     /// I/O stall deadline. {n}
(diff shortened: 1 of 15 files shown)


io-engine/src/core/env.rs
```diff
@@ -314,28 +325,44 @@ impl Default for SpdkTracingArgs {
 /// DiskPool related arguments.
 #[derive(Debug, clap::Parser, Clone)]
 pub struct PoolCliArgs {
-    /// I/O error count threshold.
+    /// I/O error count threshold. {n}
     /// After this many errors a pool alert is raised as Warning.
-    #[clap(long, env, default_value_t = 8)]
+    #[clap(
+        long = "pool-io-error-threshold",
+        env = "POOL_IO_ERROR_THRESHOLD",
```
tiagolobocastro: @abhilashshetty04 this means the helm chart would slightly change here

tiagolobocastro, line 333: @abhilashshetty04 this means the helm chart would slightly change here

Abhinandan-Purkait (approved):

tiagolobocastro:
bors try

niladrih (approved):

io-engine/src/core/env.rs
```diff
@@ -314,28 +326,44 @@ impl Default for SpdkTracingArgs {
...
+        default_value_t = 3
+    )]
     pub io_stall_transition_threshold: u64,
 
-    /// I/O stall transitions window.
+    /// I/O stall transitions window. {n}
     /// Time window during which stall ↔ resume state transitions are tracked
     /// for flakiness detection.
-    #[clap(long, env, default_value = "3h")]
+    #[clap(
```
niladrih, line 362: You could arg instead of clap, because that's the new one, clap is still usable.

niladrih (approved):

io-engine/src/core/env.rs
tiagolobocastro: hmm indeed, I guess I'll change all these to arg then

tiagolobocastro:
bors try

io-engine/src/core/env.rs
tiagolobocastro: There's multiple binaries where this needs changing, I'll leave for another PR as this is already rather large

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

Issue #1803, opened 2025-01-17
RDMA capability between nexus and replicas


dsharma-dc:
**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**
Mayastor capability to internally use RDMA connection for IO dispatch between nexus  and replicas.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

yugchaudhari:
Hey @dsharma-dc, any update on this one? Asking because we're evaluating vLLM with GDS-enabled GPUs on Mayastor, the app→nexus RDMA path (shipped in v2.8.0) already covers the read flow when the nexus is co-located with a replica, but the internal nexus→replica path is the remaining gap for cases where the nexus ends up without a local replica, and for write-heavy paths (model checkpointing, rebuilds).

Is this on a roadmap or planned for any near-term iteration? Interested in helping move it forward if there's an existing direction or OEP I can build on.

yugchaudhari:
> Hey [@dsharma-dc](https://github.com/dsharma-dc), any update on this one? Asking because we're evaluating vLLM with GDS-enabled GPUs on Mayastor, the app→nexus RDMA path (shipped in v2.8.0) already covers the read flow when the nexus is co-located with a replica, but the internal nexus→replica path is the remaining gap for cases where the nexus ends up without a local replica, and for write-heavy paths (model checkpointing, rebuilds).
> 
> Is this on a roadmap or planned for any near-term iteration? Interested in helping move it forward if there's an existing direction or OEP I can build on.

cc @tiagolobocastro

PR #1953, opened 2026-03-05
feat(io-engine): expose SPDK transport_tos for RDMA QoS/DSCP marking

jleeh:
## Summary

  - Add `transport_tos` field to `NvmeBdevOpts` struct, replacing the hardcoded `0` value
  - Configurable via `NVME_TRANSPORT_TOS` environment variable (defaults to `0` for backwards compatibility)
  - Wire through both `From` impls (`spdk_bdev_nvme_opts ↔ NvmeBdevOpts`)

## Motivation

When using NVMe-oF over RDMA (RoCEv2), lossless transport requires end-to-end DSCP marking so that switches and NICs can classify storage traffic into a priority queue with Priority Flow Control (PFC) enabled.

SPDK already supports `transport_tos` on `spdk_bdev_nvme_opts`, but Mayastor hardcodes it to `0`. This means all RDMA packets are sent with default best-effort QoS, making it impossible to achieve lossless RoCE without kernel-level workarounds that only cover the target (responder) side.

With this change, operators can set `NVME_TRANSPORT_TOS=104` (DSCP 26 / AF31) on the io-engine container to mark initiator-side RDMA traffic, completing the end-to-end lossless path:

  - **Switch side**: DCB config trusts DSCP and maps AF31 → priority 3 with PFC
  - **NIC side**: `dcb app add dscp-prio 26:3` + `dcb pfc set prio-pfc 3:on`
  - **Target side**: `rdma_cm.cma_roce_tos=104` kernel parameter
  - **Initiator side**: `NVME_TRANSPORT_TOS=104` ← **this PR**

### DIFF
--- a/io-engine/src/subsys/config/opts.rs
+++ b/io-engine/src/subsys/config/opts.rs
@@ -382,6 +382,10 @@ pub struct NvmeBdevOpts {
     /// These strings are based on device serial number and namespace ID and
     ///  will always be the same for that device.
     pub generate_uuids: bool,
+    /// Type of Service value for RDMA transport connections.
+    /// Used to set DSCP for QoS classification (e.g., 104 = DSCP 26/AF31
+    /// for lossless RoCE traffic with Priority Flow Control).
+    pub transport_tos: u8,
 }
 
 impl GetOpts for NvmeBdevOpts {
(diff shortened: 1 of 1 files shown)


tiagolobocastro (approved):

tiagolobocastro:
This seems fine but btw we don't use RDMA on the initiator side, we've only implemented the target side.
Here's a ticket for adding initiator: https://github.com/openebs/mayastor/issues/1803

abhilashshetty04 (approved):

tiagolobocastro:
@jleeh probably you'd need to also expose this via the helm chart, do you want to raise a PR there as well?

bors merge

PR #1958, opened 2026-03-17
fix: speed up list-replicas gRPC

tiagolobocastro:
fix: lookup nvmf subsystem using rb-tree find methods
    
    This avoids iterating over the entire list of subsystems which can
    greatly reduce lookup when the nr of subsystems is in the thousands.
    
---

    refactor: lookup nqn for events using subsystem api
    
    Don't convert the nqn to a uuid since we can get the bdev
    from the subsystem itself.
    
---

    fix: use nqn as cstr instead of copying
    
    Avoid copying the nqn into a new string everything it's inspected.
    
---

    refactor: add nvme max-namespaces to the cli and env args
    
    This makes it easier to modify tests to override the values.
    todo: we should change all other random bits of env parsing args
    to use this.
    
---

    test: ensure list replicas don't take "too long"
    
    No sure what the expected duration should be, but 300ms seems reasonably
    long to list all replicas.
    
    Note that most of the time was measure to come from the subsystem get/list.
    
---

    refactor: add wipe super to ReplicaArgs
    
    This will allow us to create replicas without wiping the "super block",
    which would be useful for testing with a very large number of replicas,
    without having to wipe and allocate the initial clusters.

### DIFF
--- a/io-engine/examples/lvs-eval/main.rs
+++ b/io-engine/examples/lvs-eval/main.rs
@@ -176,6 +176,7 @@ async fn create_lvol(
         thin,
         entity_id: None,
         use_extent_table: Some(et),
+        wipe_super: true,
     };
 
     lvs.create_lvol_with_opts(opts).await.map_err(|err| {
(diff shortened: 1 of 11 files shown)


tiagolobocastro:
bors try

niladrih (approved):
Don't have a lot of context, but the code does look like it does what the commit messages say it does.

abhilashshetty04 (approved):

abhilashshetty04 (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #1960, opened 2026-03-22
fix: lvs and nexus error code mapping

tiagolobocastro:
refactor(nexus): convert nexus error into granular status
    
    Finer control of the nexus error into tonic status.
    Also includes the errno in the status header.
    
---

    fix(nexus/share): return oor and emlink when reached max subsystems
    
    Similar to lvol/share, return out of range with emlink when we've reached
    the maximum number of namespaces/subsystems.
    
---

    fix(pool/lvs): use invalid argument for invalid cluster size
    
    Return correct gRPC status code when creating a pool with an invalid
    cluster size.
    
    Also, don't default error core to gRPC status, as this makes it very
    easy to add a new error and rely on the default of internal error!

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev_error.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev_error.rs
@@ -6,7 +6,7 @@ use super::{ChildError, NbdError, NexusPauseState};
 
 use crate::{
     bdev_api::BdevError,
-    core::{CoreError, VerboseError},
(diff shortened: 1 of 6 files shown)


niladrih (approved):

abhilashshetty04 (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #1961, opened 2026-04-01
fix(lvs): handle I/O errors during lvs lifecycle operations

tiagolobocastro:
fix(lvs): destroy with failing disk
    
    Pull in the spdk fix for blobstore deletion with -EIO.
    The destroy can now destroy the backing bdev when it fails
    with -EIO.
    Also we update the test to include the destroy workflow.

---

    test(lvs): add pool operations with broken backend
    
    The destroy is not working due an spdk bug.
    
---

    fix(lvs): cleanup pool backend devices
    
    When destroy/export fails we should clean up the backend devices.
    Currently destroy cleanup is getting stuck due an spdk bug, so
    for now we leave it in case of -EIO.
    
---

    refactor(lvs): reduce duplicated code on destroy/export
    
    Adds common cleanup function for both destroy/export.
    
---

    refactor(lvs): reduce duplicate code and fix undo on failed creates
    
    With the new blobstore fixes, failed creates now take a different path
    depending on the error code.
    
    This means we now also have to cleanup in another "match arm", though is
    a fix as otherwise we would try to create the pool in case of any error
    and not just EILSEQ!!!!
    
    A lot of code was duplicated between the create and import, so I also
    refactored that.
    This also fixes a missing check for crypto bdev existence in the create
    path!
    
---

    build: pull latest spdk-rs with blobstore fixes

### DIFF
--- a/io-engine/examples/lvs-eval/display.rs
+++ b/io-engine/examples/lvs-eval/display.rs
@@ -13,7 +13,7 @@ use spdk_rs::libspdk::{
 pub async fn print_lvs(lvs: &Lvs) {
     print_separator("LVS", 0);
 
-    print_bdev(lvs.base_bdev());
+    print_bdev(lvs.base_bdev_());
     print_lvs_data(lvs);
     print_replicas(lvs);
 }
--- a/io-engine/src/bdev/lvs.rs
+++ b/io-engine/src/bdev/lvs.rs
@@ -245,11 +245,10 @@ impl Lvs {
     }
 
     async fn wipe_super(args: PoolArgs) -> Result<(), BdevError> {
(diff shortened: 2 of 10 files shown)


Abhinandan-Purkait (approved):

niladrih (approved):

abhilashshetty04 (approved):

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

PR #1963, opened 2026-04-07
Improve snapshot listing times

tiagolobocastro:
test: add multi-replica and clone check to lvol_snap_list
    
    Ensures the num_clones is set correctly
    Spreads snapshots across a number of replicas
    Creates a clone for each snapshot.
    
    todo: investigate warn message:
    Using passed in clear method 0x2 instead of stored value of 0x0
    
---

    fix(lvs): don't query snapshot parent for listing
    
    I'm not sure why this was being done as we can simply return the uuid
    itself, rather than query only to return the same value...
    It's possible that we wanted to return the uuid only if the parent
    still existings, but it's not clear from the api and would be a strange
    behaviour...
    
    So we now drop this, and simply return the parent uuid.

---

    refactor(lvs): reorder xattr to list parent id as the first
    
    This makes snapshot searching quicker, as we don't need to load
    the other attributes when it's a mismatch.
    
---

    refactor(lvs): clarify is_snapshot_clone api
    
    Renames is_snapshot_clone into clone_source, where the Lvol snapshot
    is returned if found.
    Added new api method is_clone which checks if the current Lvol is a
    cloned lvol.
    
    This both clarifies the previous method and also makes it more efficient
    as we don't need to lookup the source snapshot all the times.
    
---

    refactor(lvs): replace list clones logic with spdk immediate clone
    
    Rather than looping all bdevs, we get the list of clones directly
    from the snapshot using spdk_lvol_iter_immediate_clones.
    Further we trim this down to "real_clones" by filtering out snapshots
    from the returned list.
    
    todo: can we optimize this process, we don't need 2 lists?
    
---

    refactor(lvs): compare snapshot uuid_str with potential clone parent
    
    When listing all potential bdevs to find matching clones, don't bother
    looking up the snapshot from each potential clone.
    Instead, simply compare the uuid with the snapshot.
    
---

    refactor(lvs): add lvol uuid lookup function
    
    Lookup lvol directly, rather than iterating all bdevs.

---

    refactor(lvs): count snapshot clones using spdk_blob_get_clones
    
    This api is much faster than iterating all bdevs, map them to lvols and then
    iterating again all bdevs to find the clones...
    
    This reduces the lvol_snap_list all the way to 50ms!

---

    refactor(lvs): simplify bdev to Lvol mappings
    
    Simplify all the iterations which try to load a bdev as an lvol.
    
    Also when comparing lvol's when listing clones, don't use the uuid
    as the comparisson method as it's an expensive operation requiring
    allocation of memory and string copies.
    Rather, we can simply compare the spdk_lvol pointer.
    
    This further reduces the slvol_snap_list test by another ~.5s
    
---

    refactor(lvs/xattr): change blob attr api to use compile constants
    
    Using a type which can yield &CStr avoids having to convert strings
    to CString at runtime.
    This reduces the snapshot list test by ~0.5s.
    
---

    refactor: compare bdev uuid bytes rather than string
    
    It's faster to compare the 16bytes rather than the stringified version.
    
---

    refactor: use new api for blob xattr reading
    
    We don't always need an owned string for the attribute values, as sometimes
    we only want to check/compare the value.
    I refactored the apis to returned a string with the lifetime of the lvol or
    the specified blob.
    The internal function is marked as unsafe and should not be directly used
    as the returned value is a pointer from the lvol, and must not be used
    beyong the blob's lifetime.
    
---

    test: add lvs snapshot list timing test
    
    Creates 3000 snapshots from a single replica and time out long it takes
    to list all snapshots.

### DIFF
--- a/io-engine/src/lvs/lvol_snapshot.rs
+++ b/io-engine/src/lvs/lvol_snapshot.rs
@@ -729,7 +729,7 @@ impl LvolSnapshotOps for Lvol {
         };
         bdev.into_iter()
             .filter_map(Lvol::ok_from)
-            .filter(|b| b.is_snapshot_clone().is_some())
+            .filter(|b| b.is_clone())
             .collect::<Vec<Lvol>>()
     }
 
@@ -778,14 +778,14 @@ impl LvolSnapshotOps for Lvol {
             let parent = self
                 .lvs()
                 .lookup_lvol_by_uuid_str(self.blob_xattr(SnapshotXattrs::ParentId)?)?;
-            let parent_snap_lvol = parent.is_snapshot_clone()?;
+            let parent_snap_lvol = parent.clone_source()?;
 
             let usage = parent_snap_lvol.usage();
             let usage = total_ancestor_snap_size
                 - (usage.allocated_bytes_snapshots + usage.allocated_bytes);
             Some(usage)
         // if self is clone.
-        } else if self.is_snapshot_clone().is_some() {
+        } else if self.is_clone() {
             let sum = Lvol::list_all_lvol_snapshots(Some(self))
                 .iter()
                 .map(|v| v.snapshot_lvol().allocated())
--- a/io-engine/src/lvs/lvol_snapshot.rs
+++ b/io-engine/src/lvs/lvol_snapshot.rs
@@ -698,6 +698,7 @@ impl LvolSnapshotOps for Lvol {
 
     /// List clones based on snapshot_uuid.
     fn list_clones_by_snapshot_uuid(&self) -> Vec<Lvol> {
+        let su = self.as_inner_ref().uuid_str.as_ptr();
         let bdev = match UntypedBdev::bdev_first() {
             Some(b) => b,
             None => return Vec::new(), /* No devices available, no clones */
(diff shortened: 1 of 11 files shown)


tiagolobocastro:
bors try

niladrih (approved):

io-engine/src/lvs/lvol_snapshot.rs
```diff
@@ -698,22 +692,27 @@ impl LvolSnapshotOps for Lvol {
...
-            .filter(|b| b.driver() == "lvol")
-            .map(|b| Lvol::try_from(b).unwrap())
-            .filter_map(|b| {
-                let snap_lvol = b.is_snapshot_clone();
-                if snap_lvol.is_some() && snap_lvol.unwrap().uuid() == self.uuid() {
-                    Some(b)
-                } else {
-                    None
+        // todo: optimize this process, avoid multiple calls to spdk_blob_get_clones.
+        let clone_count = 100; /* self.blob_clone_count() as usize */
```
Abhinandan-Purkait, line 696: What is this hardcoded limit ensuring? Can't there be more than this?

Abhinandan-Purkait (approved):

io-engine/src/lvs/lvol_snapshot.rs
tiagolobocastro: It's not a limit, it's just the initial allocation

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

PR #1965, opened 2026-04-10
adding implementation for pool io stall

abhilashshetty04:
This PR:

Sets up timeout detection on blobstore desc on pool create and import. Also, adds callback handling incase of stall notification. We attempt to reset lvstore (works only for aio, not implemented for uring). When reset completes, we try to read superblock, when read completes we mark pool as IO resumed.

Adds unit test for the pool stall feature.

### DIFF
--- a/io-engine-tests/src/compose/mod.rs
+++ b/io-engine-tests/src/compose/mod.rs
@@ -42,6 +42,14 @@ impl<'a> MayastorTest<'a> {
         rx.await.unwrap()
     }
 
+    /// spawn a future on this mayastor instance without waiting for completion.
+    pub fn spawn_detached<F>(&self, future: F)
+    where
+        F: Future<Output = ()> + 'static,
+    {
+        self.reactor.send_future(future);
+    }
+
     pub fn send<F>(&self, future: F)
     where
         F: Future<Output = ()> + 'static,
--- a/io-engine/src/lvs/lvs_store.rs
+++ b/io-engine/src/lvs/lvs_store.rs
@@ -313,59 +313,56 @@ impl Lvs {
 
     /// Enable stall detection on a given Pool.
     fn enable_stall_detection(&self) {
-        let lvs = self.as_inner_ptr();
+        let lvs = self.name();
         let environ = MayastorEnvironment::global();
-        let stall_deadline = environ.pool_args.io_stall_deadline;
-        let std_d: std::time::Duration = stall_deadline.into();
-        let secs: u64 = std_d.as_secs();
-        let rc = unsafe { vbdev_lvs_set_timeout(lvs, secs, Some(Self::lvstore_timeout_cb)) };
+        let secs: u64 = environ.pool_args.io_stall_deadline.as_secs();
+        let rc = unsafe {
+            vbdev_lvs_set_timeout(self.as_inner_ptr(), secs, Some(Self::lvstore_timeout_cb))
+        };
         if rc != 0 {
             error!(
-                "rc :{rc} while enabling stall deadline on pool {}",
-                self.name()
+                "error while enabling stall deadline on pool {}: {:?}",
+                self.name(),
+                Errno::from_raw(rc)
             );
+        } else {
+            info!("enabled stall detection of {secs}s on {lvs}");
         }
-        info!("enabled stall detection on {}", self.name());
     }
 
-    /// Disable stall detection on a given Pool.
-    /// We do this once we receive first stall notification.
+    /// Disable stall detection on a given Pool if reset is submitted successfullly.
     /// This is to stop receiving flood of callbacks as SPDK sends notification
     /// for all IOs which are stuck on the Pool in loop.
-    fn disable_stall_detection(&self) {
-        let lvs = self.as_inner_ptr();
-        let rc = unsafe { vbdev_lvs_set_timeout(lvs, 0, None) };
-        if rc != 0 {
-            error!(
-                "rc :{rc} while disabling stall deadline on pool {}",
-                self.name()
-            );
-        }
-        info!("disabled stall detection on {}", self.name());
-    }
-
-    /// Marks the pool as stalled in the cache and disables stall detection to avoid callback floods.
-    /// It also triggers a reset of the underlying bdev.
-    fn io_stalled(&self) {
-        let cache = pool_info_read();
-        let lvs = self.name();
-        if let Some(pool_lock) = cache.get(lvs) {
-            let mut pool_mut = pool_lock.write();
-            if !pool_mut.io_stalled {
-                info!("Pool {lvs} has entered IO stall");
+    fn disable_stall_detection(&self, rc: i32) {
+        if rc == 0 {
+            let lvs = self.name();
+            if let Some(pool_lock) = pool_info_read().get(lvs) {
+                let mut pool_mut = pool_lock.write();
                 pool_mut.io_stalled = true;
-                self.disable_stall_detection();
-                let p = std::ptr::null_mut();
-                unsafe { vbdev_lvs_bs_bdev_reset(self.as_inner_ptr(), Some(Self::lvstore_reset_cb), p) };
+            }
+            let rc = unsafe { vbdev_lvs_set_timeout(self.as_inner_ptr(), 0, None) };
+            if rc != 0 {
+                error!("rc {rc} while disabling stall deadline on pool {lvs}");
+            } else {
+                info!("disabled stall detection on {lvs}");
             }
         }
     }
 
+    /// Attempts to reset the pool facing IO stall and disables stall detection
+    /// to avoid flood of callbacks if reset is submitted successfully.
+    fn mark_io_stalled(&self) {
+        let p = std::ptr::null_mut();
+        let rc = unsafe {
+            vbdev_lvs_bs_bdev_reset(self.as_inner_ptr(), Some(Self::lvstore_reset_cb), p)
+        };
+        self.disable_stall_detection(rc);
+    }
+
     /// Marks the pool as recovered in the cache, update transition timestamp and re-enables stall detection.
-    fn io_resume(&self) {
-        let cache = pool_info_read();
+    fn mark_io_resume(&self) {
         let lvs = self.name();
-        if let Some(pool_lock) = cache.get(lvs) {
+        if let Some(pool_lock) = pool_info_read().get(lvs) {
             let mut pool_mut = pool_lock.write();
             pool_mut.io_stalled = false;
             info!("Pool {lvs} recovered from IO stall");
@@ -657,31 +654,33 @@ impl Lvs {
     extern "C" fn lvstore_timeout_cb(ctx: *mut c_void, _bdev_io: *mut spdk_bdev_io) {
         let lv_store = ctx as *mut spdk_lvol_store;
         let lvs: Lvs = Lvs::from_inner_ptr(lv_store);
-        lvs.io_stalled();
+        lvs.mark_io_stalled();
     }
 
     /// Callback function called by SPDK when reset completes.
-    extern "C" fn lvstore_reset_cb(lvs: *mut spdk_lvol_store, _success: bool, _ctx: *mut c_void) {
-        let lvs: Lvs = Lvs::from_inner_ptr(lvs);
-        let bs = lvs.blob_store();
-        let ctx = Box::new(BsTimeoutCtx {
-            pool_name: lvs.name().to_string(),
-        });
-        let ctx_ptr = Box::into_raw(ctx) as *mut c_void;
+    extern "C" fn lvstore_reset_cb(lvs: *mut spdk_lvol_store, success: bool, _ctx: *mut c_void) {
+        let lvs_wrapper: Lvs = Lvs::from_inner_ptr(lvs);
+        let lvs_name = lvs_wrapper.name().to_string();
+        if let Ok(bdev) = lvs_wrapper.base_bdev() {
+            let driver = bdev.driver();
+            info!("reset completed with status: {success} on {lvs_name} of bdev type: {driver}");
+        }
         Reactors::master().send_future(async move {
-            unsafe { spdk_bs_read_super(bs, Some(Self::bs_super_read_cb), ctx_ptr) }
+            if let Some(lvs) = Lvs::lookup(&lvs_name) {
+                let bs = lvs.blob_store();
+                let lvs_ptr = lvs.as_inner_ptr();
+                unsafe {
+                    spdk_bs_read_super(bs, Some(Self::bs_super_read_cb), lvs_ptr as *mut c_void)
+                }
+            }
         });
     }
 
     /// Callback function called by SPDK when superblock read completes.
-    extern "C" fn bs_super_read_cb(ctx: *mut c_void, errno: i32) {
-        if errno == 0 {
-            let ctx_str = unsafe { &*(ctx as *mut BsTimeoutCtx) };
-            let pool = &ctx_str.pool_name;
-            let lvs = Lvs::lookup(pool).unwrap();
-            lvs.io_resume();
-        }
-        // Shall we attempt to read again if it fails?
+    extern "C" fn bs_super_read_cb(ctx: *mut c_void, _errno: i32) {
+        let lvs_raw = ctx as *mut spdk_lvol_store;
+        let lvs: Lvs = Lvs::from_inner_ptr(lvs_raw);
+        lvs.mark_io_resume();
     }
 
     /// Imports the pool if it exists, otherwise tries to create a new pool.
@@ -879,7 +878,7 @@ impl Lvs {
                 base_bdev = c;
             }
         }
-        info!(
+        debug!(
             "Deleting bdev {}, uri {:?}",
             base_bdev.name(),
             base_bdev.bdev_uri_original_str()
@@ -1352,8 +1351,4 @@ impl PtplFileOps for LvsPtpl {
     fn subpath(&self) -> std::path::PathBuf {
         std::path::PathBuf::from("pool/").join(self.uuid())
     }
-}
-
-struct BsTimeoutCtx {
-    pool_name: String,
 }
--- a/io-engine/src/lvs/lvs_store.rs
+++ b/io-engine/src/lvs/lvs_store.rs
@@ -5,18 +5,22 @@ use futures::channel::oneshot;
 use nix::errno::Errno;
 use parking_lot::RwLock;
 use pin_utils::core_reexport::fmt::Formatter;
-use std::{convert::TryFrom, fmt::Debug, os::raw::c_void, pin::Pin, ptr::NonNull, str::FromStr};
+use std::{
+    convert::TryFrom, fmt::Debug, os::raw::c_void, pin::Pin, ptr::NonNull, str::FromStr,
+    time::Instant,
+};
 use url::Url;
 
 use spdk_rs::libspdk::{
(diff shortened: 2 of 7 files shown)


abhilashshetty04:
bors try

abhilashshetty04:
bors try

io-engine/src/lvm/vg_pool.rs
```diff
@@ -277,7 +277,6 @@ impl VolumeGroup {
     /// > Note: The Vg is first exported and then destroyed.
     pub async fn destroy_(mut self, purge_always: bool) -> Result<(), Error> {
         self.export().await?;
-
```
tiagolobocastro, line 280: if there's no other modification here, please leave as is to avoid touching

io-engine/src/lvs/lvs_store.rs
```diff
@@ -293,6 +297,76 @@ impl Lvs {
...
     }
 
+    /// Enable stall detection on a given Pool.
+    fn enable_stall_detection(&self) {
+        let lvs = self.as_inner_ptr();
+        let environ = MayastorEnvironment::global();
+        let stall_deadline = environ.pool_args.io_stall_deadline;
+        let std_d: std::time::Duration = stall_deadline.into();
+        let secs: u64 = std_d.as_secs();
+        let rc = unsafe { vbdev_lvs_set_timeout(lvs, secs, Some(Self::lvstore_timeout_cb)) };
```
tiagolobocastro: Conversion is not required IIRC

```suggestion
        let secs: u64 = stall_deadline.as_secs();
```

io-engine/src/lvs/lvs_store.rs
```diff
@@ -293,6 +297,76 @@ impl Lvs {
...
+        if rc != 0 {
+            error!(
+                "rc :{rc} while enabling stall deadline on pool {}",
+                self.name()
+            );
+        }
+        info!("enabled stall detection on {}", self.name());
```
tiagolobocastro: This needs to be on else, otherwise it's very confusing...

```suggestion
        } else {
            info!("enabled stall detection on {}", self.name());
        }
```

io-engine/src/lvs/lvs_store.rs
```diff
@@ -293,6 +297,76 @@ impl Lvs {
...
+    fn disable_stall_detection(&self) {
+        let lvs = self.as_inner_ptr();
+        let rc = unsafe { vbdev_lvs_set_timeout(lvs, 0, None) };
+        if rc != 0 {
+            error!(
+                "rc :{rc} while disabling stall deadline on pool {}",
+                self.name()
+            );
+        }
+        info!("disabled stall detection on {}", self.name());
```
tiagolobocastro: same here...

io-engine/src/lvs/lvs_store.rs
```diff
@@ -293,6 +297,76 @@ impl Lvs {
...
+    fn enable_stall_detection(&self) {
+        let lvs = self.as_inner_ptr();
+        let environ = MayastorEnvironment::global();
+        let stall_deadline = environ.pool_args.io_stall_deadline;
+        let std_d: std::time::Duration = stall_deadline.into();
+        let secs: u64 = std_d.as_secs();
+        let rc = unsafe { vbdev_lvs_set_timeout(lvs, secs, Some(Self::lvstore_timeout_cb)) };
+        if rc != 0 {
+            error!(
+                "rc :{rc} while enabling stall deadline on pool {}",
```
tiagolobocastro: You can do something like so:

```suggestion
                "error while enabling stall deadline on pool {}: {}", self.name(), nix::Error::from(rc)
```

io-engine/src/lvs/lvs_store.rs
```diff
@@ -293,6 +297,76 @@ impl Lvs {
...
+                "rc :{rc} while disabling stall deadline on pool {}",
+                self.name()
+            );
+        }
+        info!("disabled stall detection on {}", self.name());
+    }
+
+    /// Marks the pool as stalled in the cache and disables stall detection to avoid callback floods.
+    /// It also triggers a reset of the underlying bdev.
+    fn io_stalled(&self) {
```
```suggestion
    fn mark_io_stalled(&self) {
```

io-engine/src/lvs/lvs_store.rs
```diff
@@ -561,6 +637,37 @@ impl Lvs {
...
+            unsafe { spdk_bs_read_super(bs, Some(Self::bs_super_read_cb), ctx_ptr) }
+        });
+    }
+
+    /// Callback function called by SPDK when superblock read completes.
+    extern "C" fn bs_super_read_cb(ctx: *mut c_void, errno: i32) {
+        if errno == 0 {
+            let ctx_str = unsafe { &*(ctx as *mut BsTimeoutCtx) };
+            let pool = &ctx_str.pool_name;
+            let lvs = Lvs::lookup(pool).unwrap();
```
tiagolobocastro: let's not unwrap...

io-engine/src/lvs/lvs_store.rs
```diff
@@ -561,6 +637,37 @@ impl Lvs {
...
+            lvs.io_resume();
+        }
+        // Shall we attempt to read again if it fails?
```
tiagolobocastro: Actually, why do we even need to check the errno?
As long as it's completed the I/O, we can consider the stall to be over - whether I/O is failing or not has no relation to stall and should be handled by the error logic

io-engine/src/lvs/lvs_store.rs
```diff
@@ -756,7 +863,7 @@ impl Lvs {
                 base_bdev = c;
             }
         }
-        trace!(
+        info!(
```
```suggestion
        debug!(
```

io-engine/tests/lvs_pool.rs
```diff
@@ -695,6 +704,168 @@ async fn lvs_errors() {
...
+    for i in 1..(IO_STALL_TRANSITION_THRESHOLD + 1) {
+        let state = lvol.dm_suspend().await.unwrap();
+
+        assert_eq!(state, DmState::Suspended);
+        ms.spawn_detached(async move {
+            let lvs = Lvs::lookup("tpool").unwrap();
+            let _ = lvs.grow().await;
+        });
+
+        sleep(std::time::Duration::from_secs(2)).await;
```
tiagolobocastro: what is this achieving?
If this is waiting for the grow, can't we wait for the spawn above?

io-engine/tests/lvs_pool.rs
```diff
@@ -695,6 +704,168 @@ async fn lvs_errors() {
...
+            alerts.warning,
+            vec![PoolAlert::IoStallIntermittentExc as i32]
+        );
+        assert_eq!(alerts.status(), PoolAlertStatus::Warning);
+    })
+    .await;
+
+    let ht = humantime::Duration::from_str(IO_STALL_TRANSITION_WINDOW).unwrap();
+    let std_d: std::time::Duration = ht.into();
+    sleep(std_d).await;
```
tiagolobocastro:
```suggestion
    sleep(humantime::Duration::from_str(IO_STALL_TRANSITION_WINDOW).unwrap().into()).await;
```

io-engine/src/lvs/lvs_store.rs
```diff
@@ -293,6 +297,76 @@ impl Lvs {
...
+                let p = std::ptr::null_mut();
+                unsafe {
+                    vbdev_lvs_bs_bdev_reset(self.as_inner_ptr(), Some(Self::lvstore_reset_cb), p)
+                };
+            }
+        }
+    }
+
+    /// Marks the pool as recovered in the cache, update transition timestamp and re-enables stall detection.
+    fn io_resume(&self) {
```
```suggestion
    fn mark_io_resumed(&self) {
```

io-engine-tests/src/compose/mod.rs
```diff
@@ -42,6 +42,13 @@ impl<'a> MayastorTest<'a> {
         rx.await.unwrap()
     }
 
+    pub fn spawn_detached<F>(&self, future: F)
```
tiagolobocastro, line 46: missing doc comment

io-engine/src/lvs/lvs_store.rs
```diff
@@ -579,6 +633,37 @@ impl Lvs {
...
+        let lv_store = ctx as *mut spdk_lvol_store;
+        let lvs: Lvs = Lvs::from_inner_ptr(lv_store);
+        lvs.io_stalled();
+    }
+
+    /// Callback function called by SPDK when reset completes.
+    extern "C" fn lvstore_reset_cb(lvs: *mut spdk_lvol_store, _success: bool, _ctx: *mut c_void) {
+        let lvs: Lvs = Lvs::from_inner_ptr(lvs);
+        let bs = lvs.blob_store();
+        let ctx = Box::new(BsTimeoutCtx {
```
tiagolobocastro: This needs to be freed, or use lvs pointer

io-engine/src/lvs/lvs_store.rs
```diff
@@ -579,6 +633,37 @@ impl Lvs {
...
+            pool_name: lvs.name().to_string(),
+        });
+        let ctx_ptr = Box::into_raw(ctx) as *mut c_void;
+        Reactors::master().send_future(async move {
+            unsafe { spdk_bs_read_super(bs, Some(Self::bs_super_read_cb), ctx_ptr) }
```
tiagolobocastro: I wonder if there's a potential issue here, you're passing bs via a future, but the blobstore could be gone by the time it executes, this may explain the issues seen on the shutdown path
You should probably pass the name to the future and do a lookup to get a new lvs pointer

io-engine/src/lvs/lvs_store.rs
```diff
@@ -307,6 +297,72 @@ impl Lvs {
...
+        let lvs = self.name();
+        if let Some(pool_lock) = pool_info_read().get(lvs) {
+            let mut pool_mut = pool_lock.write();
+            if !pool_mut.io_stalled {
+                info!("Pool {lvs} has entered IO stall");
+                pool_mut.io_stalled = true;
+                self.disable_stall_detection();
+                let p = std::ptr::null_mut();
+                unsafe {
+                    vbdev_lvs_bs_bdev_reset(self.as_inner_ptr(), Some(Self::lvstore_reset_cb), p)
```
tiagolobocastro: here missing error handling, what if rc !=0
suggest to not disable stall, unless rc == 0, so we can retry this next time

io-engine/tests/lvs_pool.rs
abhilashshetty04: We expect the Pool to be in Suspected state after 2 secs wait as stall_deadline is 1 sec. Grow will not complete by waiting for 2 secs anyway.

abhilashshetty04:
bors try

io-engine/src/lvs/lvs_store.rs
abhilashshetty04: I will try to reproduce this.

abhilashshetty04:
bors try

io-engine/src/lvs/lvs_store.rs
```diff
@@ -307,6 +311,83 @@ impl Lvs {
...
+            vbdev_lvs_set_timeout(self.as_inner_ptr(), secs, Some(Self::lvstore_timeout_cb))
+        };
+        if rc != 0 {
+            error!(
+                "error while enabling stall deadline on pool {}: {:?}",
+                self.name(),
+                Errno::from_raw(rc)
+            );
+        } else {
+            info!("enabled stall detection of {}s on {lvs}", secs);
```
```suggestion
            info!("enabled stall detection of {secs}s on {lvs}");
```

io-engine/src/lvs/lvs_store.rs
```diff
@@ -307,6 +311,83 @@ impl Lvs {
...
+    /// Disable stall detection on a given Pool.
+    /// We do this once we receive first stall notification.
+    /// This is to stop receiving flood of callbacks as SPDK sends notification
+    /// for all IOs which are stuck on the Pool in loop.
+    fn disable_stall_detection(&self, rc: i32) {
+        if rc == 0 {
+            let lvs = self.name();
+            let rc = unsafe { vbdev_lvs_set_timeout(self.as_inner_ptr(), 0, None) };
+            if rc != 0 {
+                error!("rc :{rc} while disabling stall deadline on pool {lvs}");
```
```suggestion
                error!("rc {rc} while disabling stall deadline on pool {lvs}");
```

io-engine/src/lvs/lvs_store.rs
```diff
@@ -307,6 +311,83 @@ impl Lvs {
...
+
+    /// Marks the pool as stalled in the cache and disables stall detection to avoid callback floods.
+    /// It also triggers a reset of the underlying bdev.
+    fn mark_io_stalled(&self) {
+        let lvs = self.name();
+        if let Some(pool_lock) = pool_info_read().get(lvs) {
+            let mut pool_mut = pool_lock.write();
+            if !pool_mut.io_stalled {
+                info!("Pool {lvs} has entered IO stall");
+                pool_mut.io_stalled = true;
```
tiagolobocastro: shouldn't this be set only if the reset call returned rc = 0 ?

io-engine/src/lvs/lvs_store.rs
```diff
@@ -579,6 +662,38 @@ impl Lvs {
...
+        lvs.mark_io_stalled();
+    }
+
+    /// Callback function called by SPDK when reset completes.
+    extern "C" fn lvstore_reset_cb(lvs: *mut spdk_lvol_store, success: bool, _ctx: *mut c_void) {
+        let lvs_wrapper: Lvs = Lvs::from_inner_ptr(lvs);
+        info!(
+            "reset completed with status: {} on bdev type: {}",
+            success,
+            lvs_wrapper.base_bdev().unwrap().driver()
```
```suggestion
            "reset completed with status: {success} on bdev type: {:?}",
            lvs_wrapper.base_bdev().map(|b| b.driver())
```

io-engine/src/lvs/lvs_store.rs
```diff
@@ -579,6 +662,38 @@ impl Lvs {
...
+        );
+        Reactors::master().send_future(async move {
+            if let Some(lvs_wrapper) = Lvs::lookup(lvs_wrapper.name()) {
```
tiagolobocastro: you can't reuse the `lvs_wrapper` here, you have to lookup the name directly

io-engine/src/lvs/lvs_store.rs
```diff
@@ -579,6 +662,38 @@ impl Lvs {
...
+        });
+    }
+
+    /// Callback function called by SPDK when superblock read completes.
+    extern "C" fn bs_super_read_cb(ctx: *mut c_void, errno: i32) {
+        if errno == 0 {
+            let lvs_raw = ctx as *mut spdk_lvol_store;
+            let lvs: Lvs = Lvs::from_inner_ptr(lvs_raw);
+            lvs.mark_io_resume();
+        }
```
tiagolobocastro: I think I mentioned it in the previous review, we don't and arguably shouldn't check errno here, I/O errors != stall

```suggestion
        let lvs_raw = ctx as *mut spdk_lvol_store;
        let lvs: Lvs = Lvs::from_inner_ptr(lvs_raw);
        lvs.mark_io_resume();
```

io-engine/tests/lvs_pool.rs
tiagolobocastro: Maybe don't hardcode it then, use MayastorEnvironment::read().pool.io_deadline * 2....

io-engine/src/lvs/lvs_store.rs
abhilashshetty04: That  way we would not be reporting pool as io_stalled. If we can't reset, we still set state and not disable stall detection. We will try again to reset as we will get callback.

io-engine/src/lvs/lvs_store.rs
abhilashshetty04: made the change.

io-engine/src/lvs/lvs_store.rs
tiagolobocastro: We shouldn't report stalled, if we can't check the reset
Image what happens if we enter here, set stalled as true, try to reset and fail... then pool I/O resumes
But now stalled remains true!

abhilashshetty04:
bors try

io-engine/src/lvs/lvs_store.rs
abhilashshetty04: Makes sense.

abhilashshetty04:
bors try

abhilashshetty04:
bors try

io-engine/src/lvs/lvs_store.rs
```diff
@@ -579,6 +650,35 @@ impl Lvs {
...
+        let lvs: Lvs = Lvs::from_inner_ptr(lv_store);
+        lvs.mark_io_stalled();
+    }
+
+    /// Callback function called by SPDK when reset completes.
+    extern "C" fn lvstore_reset_cb(lvs: *mut spdk_lvol_store, success: bool, _ctx: *mut c_void) {
+        let lvs_wrapper: Lvs = Lvs::from_inner_ptr(lvs);
+        if let Ok(bdev) = lvs_wrapper.base_bdev() {
+            let driver = bdev.driver();
+            info!("reset completed with status: {success} on bdev type: {driver}",);
```
tiagolobocastro, line 665:
```suggestion
            info!("reset completed with status: {success} on bdev type: {driver}");
```

io-engine/src/lvs/lvs_store.rs
tiagolobocastro: hmm this is still the same, you're just sending the raw pointer now

abhilashshetty04:
>> hmm this is still the same, you're just sending the raw pointer now

This will do the lookup of lvs by name. If it cant find we dont call the read function.

abhilashshetty04:
> > > hmm this is still the same, you're just sending the raw pointer now
> 
> This will do the lookup of lvs by name. If it cant find we dont call the read function.

Done

abhilashshetty04:
bors try

io-engine/src/lvs/lvs_store.rs
```diff
@@ -307,6 +311,71 @@ impl Lvs {
...
+
+    /// Marks the pool as recovered in the cache, update transition timestamp and re-enables stall detection.
+    fn mark_io_resume(&self) {
+        let lvs = self.name();
+        if let Some(pool_lock) = pool_info_read().get(lvs) {
+            let mut pool_mut = pool_lock.write();
+            pool_mut.io_stalled = false;
+            info!("Pool {lvs} recovered from IO stall");
+            let cli_args = MayastorEnvironment::global();
+            let max_entries = cli_args.pool_args.io_stall_transition_threshold * 10;
```
tiagolobocastro, line 370: Shouldn't we use io_stall_transition_window here?

tiagolobocastro (approved):

io-engine/src/lvs/lvs_store.rs
abhilashshetty04: `io_stall_transition_window` is a time interval. Here we pop entries if array has 10* `io_stall_transition_threshold`.

Abhinandan-Purkait (approved):

abhilashshetty04:
bors merge

io-engine/src/lvs/lvs_store.rs
tiagolobocastro: I cannot believe it, this still got merged incorrectly

tiagolobocastro: oh wrong comment, meant for the disabling the the timeout which must be scheduled on the app thread...
testing using the grow() is not really helpful since the I/O in this case is done in the app thread, so it's not catching the problem

PR #1966, opened 2026-04-12
feat(reactor): add SPDK interrupt mode support

### LINKED ISSUE #1745: Use SPDK dynamic scheduler
**Is your feature request related to a problem? Please describe.**
When running Mayastor on clusters which have variable workloads CPU usage and power consumption increases significantly. 

**Describe the solution you'd like**
SPDK has a dynamic scheduler which, I believe, should reduce the power consumption without significantly impacting performance. Crucially, when threads are idle they are switched into interrupt mode and stop polling.

**Additional context**
I've tried deploying Mayastor at home but the increased power consumption was enough for me to abandon it. I don't have much understanding of SPDK but thought I'd put this up as from my reading it seems like this might help. I'd welcome being further educated though!

jr42:
## Summary

Opt-in SPDK interrupt mode for io-engine: reactors sleep in
`fd_group_wait()` instead of busy-polling, reducing CPU from ~1000m
per core to <300m when idle.

- Enable with `ENABLE_INTERRUPT_MODE=true` (default: off, backward compatible)
- Follows Longhorn v2 hybrid pattern (LEP 2025-07-21): epoll for NVMe-oF
  TCP targets, timerfd polling for NVMe initiators
- Includes pytest compose wiring for interrupt mode testing

### Implementation

Reactor changes (`reactor.rs`, +271 lines):
- New `ReactorState::Interrupt` with `fd_group_wait`-based event loop
- Reactor-level `FdGroup` nests all thread fd_groups for hierarchical mux
- Wakeup eventfd (`FD_TYPE_EVENTFD`, auto-drained) for Rust future delivery
- Cross-core wake on thread schedule to prevent multi-core init deadlock
- Late fd_group nesting in `add_incoming()` for dynamic thread assignment
- Clean shutdown path restoring poll mode

### Why interrupt mode instead of SPDK's dynamic scheduler (#1745)

Mayastor implements its own reactor loop (`reactor.rs`), bypassing
SPDK's stock reactor entirely. SPDK's dynamic scheduler monitors thread
busyness *within SPDK's reactor* -- since mayastor's reactor replaces
it, the scheduler has nothing to observe or control.

Instead, we implement interrupt mode directly in the custom reactor,
following the same pattern validated by Longhorn v2 (LEP 2025-07-21).
This is simpler, more predictable, and doesn't require restructuring the
reactor to use SPDK's scheduler infrastructure. The dynamic scheduler
remains a future option if the reactor is ever migrated closer to SPDK's
stock implementation.

Depends-On: openebs/spdk-rs#105
Depends-On: openebs/spdk#70
Closes: #1745

## Test plan

- [x] 73/86 pytest tests pass in single-core interrupt mode
      (13 failures are env-related, identical in poll mode)
- [x] Multi-core validated (2-core smoke test)
- [x] Production: 3-node cluster, 16 volumes, CPU ~3000m to ~463m (85% reduction)
- [x] Rolling restart validated: volumes auto-recover, nexuses redistribute
- [ ] CI pipeline passes

### DIFF
--- a/io-engine/src/core/reactor.rs
+++ b/io-engine/src/core/reactor.rs
@@ -159,12 +159,9 @@ impl Reactors {
             // and the rest in poll mode is worse than either pure
             // configuration (see PR #1966 review).
             if interrupt_mode {
-                let rc = spdk_rs::Thread::interrupt_mode_enable();
-                assert_eq!(
-                    rc, 0,
-                    "Failed to enable SPDK interrupt mode (rc={}); \
+                spdk_rs::Thread::interrupt_mode_enable().expect(
+                    "Failed to enable SPDK interrupt mode; \
                      aborting rather than running in mixed mode",
-                    rc,
                 );
                 info!("SPDK interrupt mode enabled globally");
             }
@@ -328,11 +325,8 @@ impl Reactor {
         // (see PR #1966 review). ENOMEM or init-order violations at
         // boot are unrecoverable anyway.
         let (fgrp, wakeup_fd) = if interrupt_enabled {
-            let fg = spdk_rs::FdGroup::create().unwrap_or_else(|rc| {
-                panic!(
-                    "Reactor core {}: failed to create fd_group (rc={})",
-                    core, rc
-                )
+            let fg = spdk_rs::FdGroup::create().unwrap_or_else(|err| {
+                panic!("Reactor core {}: failed to create fd_group ({})", core, err)
             });
             let efd = unsafe { libc::eventfd(0, libc::EFD_NONBLOCK | libc::EFD_CLOEXEC) };
             assert!(
@@ -350,11 +344,11 @@ impl Reactor {
                 std::ptr::null_mut(),
                 spdk_rs::FD_TYPE_EVENTFD,
             )
-            .unwrap_or_else(|rc| {
+            .unwrap_or_else(|err| {
                 unsafe { libc::close(efd) };
                 panic!(
-                    "Reactor core {}: failed to add wakeup eventfd to fd_group (rc={})",
-                    core, rc
+                    "Reactor core {}: failed to add wakeup eventfd to fd_group ({})",
+                    core, err
                 )
             });
             info!("Reactor core {core}: interrupt mode enabled");
@@ -663,13 +657,13 @@ impl Reactor {
             if thread_fgrp.is_null() {
                 continue;
             }
-            if let Err(rc) = fgrp.nest(thread_fgrp) {
+            if let Err(err) = fgrp.nest(thread_fgrp) {
                 error!(
-                    "Core {}: nest failed for thread '{}' (rc={}) — \
+                    "Core {}: nest failed for thread '{}' ({}) — \
                      staying in poll mode",
                     self.lcore,
                     t.name(),
-                    rc,
+                    err,
                 );
                 for n in nested {
                     let _ = fgrp.unnest(n);
@@ -778,7 +772,7 @@ impl Reactor {
             .then_some(self.fgrp.as_ref())
             .flatten();
 
-        let mut fallback_reason: Option<(String, i32)> = None;
+        let mut fallback_reason: Option<(String, Errno)> = None;
         while let Some(i) = self.incoming.pop() {
             if fallback_reason.is_none() {
                 if let Some(fgrp) = nest_into_fgrp {
@@ -792,7 +786,7 @@ impl Reactor {
                                     self.lcore,
                                 );
                             }
-                            Err(rc) => {
+                            Err(err) => {
                                 // Can't leave the reactor half-nested
                                 // (threads whose fgrps aren't under
                                 // the sleeping reactor's fd_group
@@ -800,7 +794,7 @@ impl Reactor {
                                 // Capture the reason, stop trying to
                                 // nest further incomings, fall back
                                 // below after the drain completes.
-                                fallback_reason = Some((i.name().to_string(), rc));
+                                fallback_reason = Some((i.name().to_string(), err));
                             }
                         }
                     }
@@ -809,11 +803,11 @@ impl Reactor {
             self.threads.borrow_mut().push_back(i);
         }
 
-        if let Some((name, rc)) = fallback_reason {
+        if let Some((name, err)) = fallback_reason {
             error!(
-                "Core {}: add_incoming nest failed for thread '{}' (rc={}) — \
+                "Core {}: add_incoming nest failed for thread '{}' ({}) — \
                  falling back to poll mode",
-                self.lcore, name, rc,
+                self.lcore, name, err,
             );
             self.leave_interrupt_mode();
         }
--- a/io-engine/src/core/reactor.rs
+++ b/io-engine/src/core/reactor.rs
@@ -34,7 +34,7 @@ use std::{
     collections::VecDeque,
     fmt::{self, Debug, Display, Formatter},
     future::Future,
-    os::raw::c_void,
+    os::{fd::RawFd, raw::c_void},
     pin::Pin,
     slice::Iter,
     time::Duration,
@@ -69,6 +69,8 @@ pub enum ReactorState {
     Running,
     Shutdown,
     Delayed,
+    /// Epoll-based interrupt mode: reactor sleeps until I/O events arrive.
+    Interrupt,
 }
 
 impl Display for ReactorState {
@@ -78,6 +80,7 @@ impl Display for ReactorState {
             ReactorState::Running => "Running",
             ReactorState::Shutdown => "Shutdown",
             ReactorState::Delayed => "Delayed",
+            ReactorState::Interrupt => "Interrupt",
         };
         write!(f, "{s}")
     }
@@ -122,6 +125,14 @@ pub struct Reactor {
     /// through FFI
     sx: Sender<Pin<Box<dyn Future<Output = ()> + 'static>>>,
     rx: Receiver<Pin<Box<dyn Future<Output = ()> + 'static>>>,
+    /// Whether interrupt mode is enabled for this reactor.
+    interrupt_enabled: bool,
+    /// Reactor-level fd_group for interrupt mode. Thread fd_groups are
+    /// nested into this, and spdk_fd_group_wait() blocks until events.
+    fgrp: Option<spdk_rs::FdGroup>,
+    /// eventfd registered in the reactor's fd_group to wake from
+    /// fd_group_wait when Rust futures arrive via send_future().
+    wakeup_fd: RawFd,
 }
 
 thread_local! {
@@ -131,7 +142,7 @@ thread_local! {
 
 impl Reactors {
     /// initialize the reactor subsystem for each core assigned to us
-    pub fn init(developer_delay: bool) {
+    pub fn init(developer_delay: bool, interrupt_mode: bool) {
         REACTOR_LIST.get_or_init(|| {
             let mempool_sz = try_from_env(
                 "SPDK_DEFAULT_MSG_MEMPOOL_SIZE",
@@ -140,15 +151,39 @@ impl Reactors {
             if !(mempool_sz+1).is_power_of_two() {
                 tracing::warn!("The provided SPDK_DEFAULT_MSG_MEMPOOL_SIZE ({SPDK_DEFAULT_MSG_MEMPOOL_SIZE}) is not a power of 2 - 1. This is not optimal for memory consumption");
             }
+            // Enable SPDK interrupt mode globally if requested.
+            // Must be called BEFORE spdk_thread_lib_init_ext().
+            if interrupt_mode {
+                let rc = spdk_rs::Thread::interrupt_mode_enable();
+                if rc != 0 {
+                    error!(
+                        "Failed to enable SPDK interrupt mode (rc={}), \
+                         falling back to poll mode",
+                        rc
+                    );
+                } else {
+                    info!("SPDK interrupt mode enabled globally");
+                }
+            }
+
             let rc = unsafe {
                 spdk_thread_lib_init_ext(Some(Self::do_op), Some(Self::can_op), 0, mempool_sz)
             };
             assert_eq!(rc, 0);
 
+            let interrupt_available = interrupt_mode
+                && spdk_rs::Thread::interrupt_mode_is_enabled();
+
             Reactors(
                 Cores::count()
                     .into_iter()
-                    .map(|core| Reactor::new(core, developer_delay))
+                    .map(|core| {
+                        Reactor::new(
+                            core,
+                            developer_delay,
+                            interrupt_available,
+                        )
+                    })
                     .collect::<Vec<_>>(),
             )
         });
@@ -187,6 +222,14 @@ impl Reactors {
                     r.lcore,
                 );
                 r.incoming.push(mt);
+                // Wake the target reactor if it's another core sitting
+                // in fd_group_wait — it can't drain `incoming` until it
+                // returns from the blocking wait. Without this kick a
+                // slave reactor that entered interrupt mode with zero
+                // threads would never see threads scheduled to it later.
+                if r.lcore != Cores::current() {
+                    r.wake();
+                }
                 return true;
             }
             false
@@ -279,9 +322,54 @@ impl<'a> IntoIterator for &'a Reactors {
 
 impl Reactor {
     /// create a new ['Reactor'] instance
-    fn new(core: u32, developer_delay: bool) -> Self {
+    fn new(core: u32, developer_delay: bool, interrupt_enabled: bool) -> Self {
         // create a channel to receive futures on
         let (sx, rx) = unbounded::<Pin<Box<dyn Future<Output = ()> + 'static>>>();
+
+        let mut fgrp = None;
+        let mut wakeup_fd: RawFd = -1;
+
+        if interrupt_enabled {
+            match spdk_rs::FdGroup::create() {
+                Ok(fg) => {
+                    // Create an eventfd and register it in the reactor's
+                    // fd_group so send_future() can wake us from
+                    // fd_group_wait().
+                    let efd = unsafe { libc::eventfd(0, libc::EFD_NONBLOCK | libc::EFD_CLOEXEC) };
+                    if efd >= 0 {
+                        // Register as EVENTFD type so fd_group_wait()
+                        // auto-drains it. Without this, a single
+                        // send_future() write makes the fd permanently
+                        // readable (level-triggered), spinning the reactor.
+                        match fg.add_with_fd_type(
+                            efd,
+                            Self::wakeup_handler,
+                            std::ptr::null_mut(),
+                            spdk_rs::FD_TYPE_EVENTFD,
+                        ) {
+                            Ok(()) => {
+                                wakeup_fd = efd;
+                                info!("Reactor core {}: interrupt mode enabled", core);
+                                fgrp = Some(fg);
+                            }
+                            Err(rc) => {
+                                error!("Failed to add wakeup eventfd to fd_group (rc={})", rc,);
+                                unsafe { libc::close(efd) };
+                            }
+                        }
+                    } else {
+                        error!("Failed to create wakeup eventfd for core {}", core);
+                    }
+                }
+                Err(rc) => {
+                    error!(
+                        "Failed to create fd_group for core {} (rc={}), \
+                         interrupt mode disabled",
+                        core, rc,
+                    );
+                }
+            }
+        }
 
         Self {
             threads: RefCell::new(VecDeque::new()),
@@ -292,7 +380,18 @@ impl Reactor {
             tid: Cell::new(0),
             sx,
             rx,
+            interrupt_enabled: fgrp.is_some(),
+            fgrp,
+            wakeup_fd,
         }
+    }
+
+    /// Callback for the wakeup eventfd. The eventfd is auto-drained
+    /// by `fd_group_wait()` because it is registered as
+    /// `SPDK_FD_TYPE_EVENTFD`. The actual futures are processed by
+    /// `receive_futures()` in the poll loop.
+    extern "C" fn wakeup_handler(_ctx: *mut c_void) -> i32 {
+        0
     }
 
     /// this function gets called by DPDK
@@ -333,6 +432,17 @@ impl Reactor {
         F: Future<Output = ()> + 'static,
     {
         self.sx.send(Box::pin(future)).unwrap();
+        // Wake the reactor if it's sleeping in fd_group_wait.
+        if self.wakeup_fd >= 0 {
+            let val: u64 = 1;
+            unsafe {
+                libc::write(
+                    self.wakeup_fd,
+                    &val as *const u64 as *const libc::c_void,
+                    std::mem::size_of::<u64>(),
+                );
+            }
+        }
     }
 
     /// spawn a future locally on this core; note that you can *not* use the
@@ -404,7 +514,8 @@ impl Reactor {
             ReactorState::Init
             | ReactorState::Delayed
             | ReactorState::Shutdown
-            | ReactorState::Running => {
+            | ReactorState::Running
+            | ReactorState::Interrupt => {
                 self.flags.set(state);
             }
         }
@@ -458,14 +569,39 @@ impl Reactor {
         // Initialize TID for this reactor.
         self.tid.set(gettid());
 
+        // If interrupt mode is enabled, enter it immediately. The
+        // reactor will block in fd_group_wait() until events arrive,
+        // then poll, then block again.
+        if self.interrupt_enabled {
+            self.enter_interrupt_mode();
+        }
+
         loop {
             match self.get_state() {
-                // running is the default mode for all cores. All cores, except
-                // the master core spin within this specific loop
                 ReactorState::Running => {
                     self.poll_once();
                 }
+                ReactorState::Interrupt => {
+                    // Block until I/O events or wakeup from send_future.
+                    // fd_group_wait dispatches events to the registered
+                    // interrupt callbacks (which are the poller functions).
+                    self.wait_for_events();
+                    // Restore init_thread context before running Rust
+                    // futures. fd_group_wait wrappers save/restore the
+                    // context for each event, but we set it explicitly
+                    // so gRPC futures see the app thread (required by
+                    // spdk_bdev_register and other app-thread-only APIs).
+                    if let Some(init_t) = self.threads.borrow().front() {
+                        init_t.set_current();
+                    }
+                    self.receive_futures();
+                    self.run_futures();
+                    self.add_incoming();
+                }
                 ReactorState::Shutdown => {
+                    if self.interrupt_enabled {
+                        self.exit_interrupt_mode();
+                    }
                     info!("reactor {} shutdown requested", self.lcore);
                     break;
                 }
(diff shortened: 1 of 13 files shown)


Abhinandan-Purkait:
@jr42 Thanks for the contribution! There seems to be a conflict here.

Abhinandan-Purkait:
@jr42 Also, since you have the means to validate this, can you give us a rough idea on the performance drop between these two modes? Thanks!

abhilashshetty04:
bors try

jr42:
Rebased onto current develop — the conflict was just the spdk-rs submodule pointer diverging. Should be clean now.

jr42:
@Abhinandan-Purkait Here's a benchmark comparing poll mode vs interrupt mode across different `NVME_IOQ_POLL_PERIOD` values, following Longhorn v2's methodology.

### Test Environment

| Component | Detail |
|---|---|
| Host | MacBook Pro M4 Pro, 48GB RAM, Apple SSD AP1024Z (1TB NVMe) |
| VM | Colima (Apple Virtualization.framework), Ubuntu 24.04.2, kernel 6.8.0-64-generic |
| VM resources | 6 vCPU (aarch64), 12GB RAM, 100GB VirtIO disk on host NVMe |
| io-engine | `pr/interrupt-mode` branch, debug build, aarch64 native |
| SPDK | `v25.05-d1a7ada` (jr42/spdk fork with interrupt mode fixes) |
| fio | 3.36 |

### Topology

```
fio → kernel nvme-tcp → ms1 (nexus/initiator, core 2) → NVMe-oF TCP → ms0 (target, core 1) → aio bdev (2GiB file on VirtIO SSD)
```

Two io-engine instances on a docker bridge network exercising both the target path (event-driven via epoll) and the initiator path (periodic timerfd polling via `NVME_IOQ_POLL_PERIOD`).

### fio Parameters (matching Longhorn v2 wiki)

| Test | bs | iodepth | numjobs | rw |
|---|---|---|---|---|
| IOPS | 4K | 128 | 8 | randread / randwrite |
| Bandwidth | 128K | 16 | 4 | read / write |
| Latency | 4K | 1 | 1 | randread / randwrite |

All tests: `--direct=1 --ioengine=libaio --runtime=30 --ramp_time=5 --norandommap=1`

### Results

| Test | Poll (period=0) | Interrupt (100μs) | Interrupt (1000μs) |
|---|---|---|---|
| **Rand Read 4K** | 129,584 IOPS / 7.9ms | 92,190 (**-29%**) / 11.1ms | 15,472 (-88%) / 66ms |
| **Rand Write 4K** | 108,374 / 9.5ms | 73,958 (**-32%**) / 14ms | 12,580 (-88%) / 81ms |
| **Seq Read 128K** | 2,483 MiB/s / 3.2ms | 1,249 MiB/s (**-50%**) / 6.4ms | 125 MiB/s (-95%) / 64ms |
| **Seq Write 128K** | 2,345 MiB/s / 3.4ms | 2,043 MiB/s (**-13%**) / 3.9ms | 416 MiB/s (-82%) / 19ms |
| **Lat Read 4K** (qd=1) | 15,383 IOPS / 64μs | 4,962 (-68%) / 201μs | 499 (-97%) / 2ms |
| **Lat Write 4K** (qd=1) | 12,153 / 82μs | 4,647 (-62%) / 215μs | 491 (-96%) / 2ms |

### CPU Usage (no I/O in flight)

| Config | ms0 (target) | ms1 (initiator) |
|---|---|---|
| Poll | **100%** (1 full core) | **100%** (1 full core) |
| Interrupt (100μs) | **0.4%** | **7.4%** |
| Interrupt (1000μs) | **1.6%** | **3.9%** |

### Analysis

Results align with [Longhorn v2's findings](https://github.com/longhorn/longhorn/issues/9834) (~28% random read IOPS drop at `period=100μs` on AWS, we see 29%).

**`period=100μs` is probably the more sensible default for most workloads** — moderate IOPS impact at high queue depth, but CPU drops from a full core per reactor to near-zero when idle. `period=1000μs` makes sense for edge/homelab deployments where storage backends are slower and energy savings matter more than peak IOPS.

The impact scales with I/O path latency — the slower the storage or network, the less the poll period matters relative to overall latency. On fast NVMe paths it dominates; on SATA/HDD or 1GbE backends the 100μs overhead disappears into the noise.

Long term, SPDK's dynamic scheduler (automatically switching between poll and interrupt based on thread busyness) would remove the need for static `poll_period` tuning, but integrating it with the custom reactor would require substantially more changes.

Note: debug build in a VM — absolute throughput would be higher with a release build on bare metal, but the relative comparison between modes should hold.

io-engine/src/core/reactor.rs
```diff
@@ -140,15 +151,39 @@ impl Reactors {
...
             }
+            // Enable SPDK interrupt mode globally if requested.
+            // Must be called BEFORE spdk_thread_lib_init_ext().
+            if interrupt_mode {
+                let rc = spdk_rs::Thread::interrupt_mode_enable();
+                if rc != 0 {
+                    error!(
+                        "Failed to enable SPDK interrupt mode (rc={}), \
+                         falling back to poll mode",
+                        rc
```
tiagolobocastro, line 162: Could you make similar changes throughout the new code please?

```suggestion
                        "Failed to enable SPDK interrupt mode (rc={rc}), \
                         falling back to poll mode",
```

io-engine/src/core/reactor.rs
```diff
@@ -292,9 +380,20 @@ impl Reactor {
             tid: Cell::new(0),
             sx,
             rx,
+            interrupt_enabled: fgrp.is_some(),
```
tiagolobocastro, line 383: what happens if we fail to enable it on some reactors?

io-engine/src/core/reactor.rs
```diff
@@ -458,14 +569,39 @@ impl Reactor {
...
+                    // interrupt callbacks (which are the poller functions).
+                    self.wait_for_events();
+                    // Restore init_thread context before running Rust
+                    // futures. fd_group_wait wrappers save/restore the
+                    // context for each event, but we set it explicitly
+                    // so gRPC futures see the app thread (required by
+                    // spdk_bdev_register and other app-thread-only APIs).
+                    if let Some(init_t) = self.threads.borrow().front() {
+                        init_t.set_current();
+                    }
```
tiagolobocastro, line 596: After we wait for the events, is this then a "regular" poll? If so, could we just call `poll_once` ?
Also looks like we're not polling the spdk threads here?

io-engine/src/core/reactor.rs
```diff
@@ -502,6 +638,75 @@ impl Reactor {
         self.add_incoming();
     }
 
+    /// Switch all SPDK threads to interrupt mode and nest their
+    /// fd_groups into the reactor's fd_group. Called once at startup.
+    fn enter_interrupt_mode(&self) {
+        let fgrp = match &self.fgrp {
+            Some(fg) => fg,
+            None => return,
+        };
```
tiagolobocastro, line 647: nit: could probably do but up to you

```suggestion
        let Some(ref fgrp) = self.fgrp else {
            return;
        };
```

io-engine/src/core/reactor.rs
```diff
@@ -502,6 +638,75 @@ impl Reactor {
...
+
+        let threads = self.threads.borrow();
+        for t in threads.iter() {
+            let thread_fgrp = t.get_interrupt_fd_group();
+            if !thread_fgrp.is_null() {
+                if let Err(rc) = fgrp.nest(thread_fgrp) {
+                    warn!("Failed to nest thread '{}' fd_group (rc={})", t.name(), rc,);
```
tiagolobocastro, line 654: what happens in this case, would we fail to receive events from this thread?
If so, should we just resort back to polling?

io-engine/src/core/reactor.rs
```diff
@@ -502,6 +638,75 @@ impl Reactor {
...
+                }
+            }
+            t.set_current();
+            spdk_rs::Thread::set_interrupt_mode(true);
```
tiagolobocastro, line 658: I'm assuming it's like this instead of t.set_interrupt_mode(true) because the spdk api doesn't take the thread and relies on the current?

io-engine/src/core/reactor.rs
```diff
@@ -524,11 +729,56 @@ impl Reactor {
...
+                if !thread_fgrp.is_null() {
+                    if let Err(rc) = fgrp.nest(thread_fgrp) {
+                        warn!(
+                            "add_incoming: failed to nest thread '{}' fd_group on core {} (rc={})",
+                            i.name(),
+                            self.lcore,
+                            rc,
+                        );
+                    } else {
+                        info!(
```
tiagolobocastro, line 754: should this be a debug log?

io-engine/src/core/reactor.rs
```diff
@@ -670,6 +920,13 @@ impl Future for &'static Reactor {
                 cx.waker().wake_by_ref();
                 Poll::Pending
             }
+            ReactorState::Interrupt => {
+                // Interrupt state is handled by poll_reactor(), not here.
```
tiagolobocastro, line 924: hmm but isn't the master core interrupt handled here?

tiagolobocastro (commented):
Thanks @jr42, this is awesome.

I had previously wondered if we could change our reactor into just a poller for the rust futures, and then reuse the spdk reactor, though not sure how feasible that is.

Abhinandan-Purkait:
> @Abhinandan-Purkait Here's a benchmark comparing poll mode vs interrupt mode across different `NVME_IOQ_POLL_PERIOD` values, following Longhorn v2's methodology.
> 
> ### Test Environment
> Component	Detail
> Host	MacBook Pro M4 Pro, 48GB RAM, Apple SSD AP1024Z (1TB NVMe)
> VM	Colima (Apple Virtualization.framework), Ubuntu 24.04.2, kernel 6.8.0-64-generic
> VM resources	6 vCPU (aarch64), 12GB RAM, 100GB VirtIO disk on host NVMe
> io-engine	`pr/interrupt-mode` branch, debug build, aarch64 native
> SPDK	`v25.05-d1a7ada` (jr42/spdk fork with interrupt mode fixes)
> fio	3.36
> ### Topology
> ```
> fio → kernel nvme-tcp → ms1 (nexus/initiator, core 2) → NVMe-oF TCP → ms0 (target, core 1) → aio bdev (2GiB file on VirtIO SSD)
> ```
> 
> Two io-engine instances on a docker bridge network exercising both the target path (event-driven via epoll) and the initiator path (periodic timerfd polling via `NVME_IOQ_POLL_PERIOD`).
> 
> ### fio Parameters (matching Longhorn v2 wiki)
> Test	bs	iodepth	numjobs	rw
> IOPS	4K	128	8	randread / randwrite
> Bandwidth	128K	16	4	read / write
> Latency	4K	1	1	randread / randwrite
> All tests: `--direct=1 --ioengine=libaio --runtime=30 --ramp_time=5 --norandommap=1`
> 
> ### Results
> Test	Poll (period=0)	Interrupt (100μs)	Interrupt (1000μs)
> **Rand Read 4K**	129,584 IOPS / 7.9ms	92,190 (**-29%**) / 11.1ms	15,472 (-88%) / 66ms
> **Rand Write 4K**	108,374 / 9.5ms	73,958 (**-32%**) / 14ms	12,580 (-88%) / 81ms
> **Seq Read 128K**	2,483 MiB/s / 3.2ms	1,249 MiB/s (**-50%**) / 6.4ms	125 MiB/s (-95%) / 64ms
> **Seq Write 128K**	2,345 MiB/s / 3.4ms	2,043 MiB/s (**-13%**) / 3.9ms	416 MiB/s (-82%) / 19ms
> **Lat Read 4K** (qd=1)	15,383 IOPS / 64μs	4,962 (-68%) / 201μs	499 (-97%) / 2ms
> **Lat Write 4K** (qd=1)	12,153 / 82μs	4,647 (-62%) / 215μs	491 (-96%) / 2ms
> ### CPU Usage (no I/O in flight)
> Config	ms0 (target)	ms1 (initiator)
> Poll	**100%** (1 full core)	**100%** (1 full core)
> Interrupt (100μs)	**0.4%**	**7.4%**
> Interrupt (1000μs)	**1.6%**	**3.9%**
> ### Analysis
> Results align with [Longhorn v2's findings](https://github.com/longhorn/longhorn/issues/9834) (~28% random read IOPS drop at `period=100μs` on AWS, we see 29%).
> 
> **`period=100μs` is probably the more sensible default for most workloads** — moderate IOPS impact at high queue depth, but CPU drops from a full core per reactor to near-zero when idle. `period=1000μs` makes sense for edge/homelab deployments where storage backends are slower and energy savings matter more than peak IOPS.
> 
> The impact scales with I/O path latency — the slower the storage or network, the less the poll period matters relative to overall latency. On fast NVMe paths it dominates; on SATA/HDD or 1GbE backends the 100μs overhead disappears into the noise.
> 
> Long term, SPDK's dynamic scheduler (automatically switching between poll and interrupt based on thread busyness) would remove the need for static `poll_period` tuning, but integrating it with the custom reactor would require substantially more changes.
> 
> Note: debug build in a VM — absolute throughput would be higher with a release build on bare metal, but the relative comparison between modes should hold.

Thanks! @jr42 I see you tested on arm64, did you build arm64 images? or you ran it on emulation the arm64 host?

jr42:
**@Abhinandan-Purkait** on arm64 — native aarch64, no emulation. Colima on the M4 Pro runs an aarch64 Ubuntu guest via macOS Virtualization.framework; the io-engine ELF is EM_AARCH64. Debug build, so absolute numbers are conservative — relative deltas should be representative. Can redo on x86_64 release if useful.

**@tiagolobocastro** on the reactor-as-poller idea — I like it, and it feels like the right long-term shape (and would probably be the cleaner answer to #1745 via SPDK's scheduler). Full disclosure: I tried a naive first pass on using SPDK's  reactor and failed miserably, then settled for this more intermediate approach.

io-engine/src/core/reactor.rs
jr42: 👍 will do.

io-engine/src/core/reactor.rs
jr42: Fair point. Easiest fix is probably all-or-nothing: if any reactor fails to build its `fd_group`/eventfd, clear the mode globally and fall everything back to poll (with a loud error). Does that match what you'd want, or would you prefer a hard abort at startup?

io-engine/src/core/reactor.rs
jr42: My assumption here was that `fd_group_wait` dispatches the thread pollers itself via the nested fd_groups, so an extra `t.poll()` would be redundant (and spin the pollers that just fired). That's why I only drive the rust-future side after returning. Could be wrong on that — happy to re-check. `poll_once` also iterates `t.poll()` on every thread which is what I was trying to avoid here.

Will add a comment explaining the intent so the next person doesn't have to re-derive it.

io-engine/src/core/reactor.rs
jr42: 👍

io-engine/src/core/reactor.rs
jr42: 👍

io-engine/src/core/reactor.rs
jr42: Agreed, `warn!` and carry on is too quiet. My instinct is to fall the whole reactor back to poll on first nest failure — simpler than trying to keep the reactor half-in, half-out. OK with that?

io-engine/src/core/reactor.rs
jr42: Yeah, that was my read — the SPDK C function takes no thread arg and operates on `spdk_get_thread()`, which is why it ends up as an associated fn and needs the preceding `t.set_current()`. I'll leave a short comment there.

io-engine/src/core/reactor.rs
jr42: So this one I'm less sure about. My reading: master is driven by tokio via `impl Future for &Reactor` (from `try_join_all` in `env.rs`), slaves via `rte_eal_remote_launch → poll_reactor`. `enter_interrupt_mode()` only fires from `poll_reactor`, so master never actually transitions to `Interrupt` in the main binary — which is why the branch here is a fall-through rather than a real handler. The production CPU numbers (~100m+ left on master after the reduction) line up with that reading.

If that's right, I'd turn this into `unreachable!()` with a comment and call the limitation out in the PR description. If I'm misreading and master *does* hit that branch in some path I'm missing, I'd rather know before making it `unreachable!()` — does master ever get here that you're aware of?

(`casperf` does call `master().poll_reactor()` directly, so it's a real branch for that binary — but that's a separate entry point.)

io-engine/src/core/reactor.rs
tiagolobocastro: I don't have an answer, was more interested in understanding whether this is possible and if so should we handle it.

Is there any reason why this would ever fail (other than -ENOMEM)? If so then we should consider falling back to polling (or does running in "mixed" mode make sense here) ?
If no legitimate reason why this would fail, then perhaps we could abort to make it simpler?

io-engine/src/core/reactor.rs
tiagolobocastro: Yes looking at [`spdk_thread_poll`](https://github.com/spdk/spdk/blob/d711c4e421e5f51f87f80fdf8fa9ef761f76e89c/lib/event/reactor.c#L1011) I think you're correct.

io-engine/src/core/reactor.rs
tiagolobocastro: Yeah that seems safer; we've had an issue previously where some threads were not polled and caused all sorts of weird and hard to debug issues

io-engine/src/core/reactor.rs
tiagolobocastro: Yes I believe you are right, the master is polled from [here](https://github.com/openebs/mayastor/blob/881d4897be156e54c0c7abb6ba2baa0b5fa2ae23/io-engine/src/core/env.rs#L1262).

So I guess you'd have to add `master.enter_interrupt_mode` before adding it to the futures list?

Would be interesting to use Casperf with this and see what the numbers are.
Btw, to check release numbers you have to also use spdk release, like so:
```
nix-shell --run zsh --argstr spdk release
```

test/python/tests/rpc/docker-compose.yml
```diff
@@ -13,7 +13,10 @@ services:
         - NEXUS_NVMF_RESV_ENABLE=1
         - PATH=${LLVM_SYMBOLIZER_DIR:-}
         - ASAN_OPTIONS=detect_leaks=0
-    command: ${SRCDIR}/${IO_ENGINE_DIR}/io-engine -g 0.0.0.0 -l 3,4 -r /tmp/ms1.sock
+        - ENABLE_INTERRUPT_MODE=${ENABLE_INTERRUPT_MODE:-false}
+        - NVME_IOQ_POLL_PERIOD=${NVME_IOQ_POLL_PERIOD:-}
```
tiagolobocastro, line 17: Also we should have some tests with interrupt mode.

We can either have some tests in interrupt, or run a parallel test run with interrupt, any thoughts here?

CC @Abhinandan-Purkait @abhilashshetty04

jr42:
**casperf release-build benchmarks** (per @tiagolobocastro's request on reactor.rs:924)

Re-ran with the `--argstr spdk release` shell you suggested. Numbers are for the rebuilt stack with master-core interrupt mode wired in (`master.enter_interrupt_mode()` in `env.rs`) and the spdk-rs / spdk fork chain rebased onto current upstream.

### Environment

| | |
|---|---|
| Host / VM | MacBook M4 Pro, Colima aarch64 Linux, 6 vCPU, 12 GiB |
| Build | `cargo build --release`, SPDK release closure via `--argstr spdk release` |
| Backing | AIO bdev on 2 GiB file on Colima VirtIO disk |
| Topology | single reactor (core 1 = master), `casperf -b 4096 -q 64` |

### Results — 4K randread, qd=64

| Mode | IOPS | Bandwidth | vs poll |
|---|---|---|---|
| Poll | 1,914k | 7.47 GB/s | baseline |
| Interrupt, `NVME_IOQ_POLL_PERIOD=100μs` | 1,775k | 6.93 GB/s | **-7.3%** |
| Interrupt, `NVME_IOQ_POLL_PERIOD=1000μs` | 1,780k | 6.95 GB/s | -7.0% |

### Notes

- **Much smaller throughput cost than the fio path** (7% vs 29% on 4K randread). fio went through `kernel nvme-tcp → ms1 nexus → NVMe-oF TCP → ms0 → aio` — 5 hops, each with its own poll penalty. Casperf is direct `reactor → aio bdev` so we only pay the penalty once.
- **`100μs` and `1000μs` identical for AIO** as expected — `NVME_IOQ_POLL_PERIOD` only affects NVMe initiator timerfd polling, AIO completions are epoll-driven.
- **Master-core interrupt mode verified working**: casperf runs a single reactor pinned to core 1 (= master in single-core config). Non-trivial throughput confirms the master-core path is functional after the env.rs change.
- **Write numbers skipped** — Colima's VirtIO disk fsync tops out around 19k IOPS, so casperf writes are storage-bound, not reactor-bound; poll/interrupt comparison isn't meaningful there.

Absolute numbers are aarch64 VM; relative deltas should translate to bare metal x86_64.

io-engine/src/core/reactor.rs
jr42: Done in `84f27bdb`.

io-engine/src/core/reactor.rs
jr42: Done in `6ac1435d` — `interrupt_mode_enable()` and per-reactor fgrp/eventfd setup now panic on failure. No silent fallback; the mixed-mode case was the source of hard-to-debug deadlocks in prior incidents, and -ENOMEM at boot isn't recoverable anyway.

io-engine/src/core/reactor.rs
jr42: Done in `84f27bdb` — added a comment citing `spdk_thread_poll` in `lib/event/reactor.c`. No explicit `t.poll()` needed: thread pollers dispatch via the nested fd_groups during `fd_group_wait` itself, so `poll_once()` would just re-enter the same code path with nothing to do.

io-engine/src/core/reactor.rs
jr42: Done in `84f27bdb` — applied the same idiom to `exit_interrupt_mode` for consistency.

io-engine/src/core/reactor.rs
jr42: Done in `95d57971` — two-phase `enter_interrupt_mode` (nest all → flip all) so we cannot half-transition; first nest failure unnests what we did and returns without touching state. `add_incoming` uses the same fallback via a shared `leave_interrupt_mode` helper on any nest failure for runtime-added threads.

io-engine/src/core/reactor.rs
jr42: Done in `84f27bdb` — added a comment noting `spdk_thread_set_interrupt_mode` operates on `spdk_get_thread()` (no thread arg), hence the `t.set_current()` first.

io-engine/src/core/reactor.rs
jr42: Done in `84f27bdb`.

io-engine/src/core/reactor.rs
jr42: Done in `0be4c72e` — explicit `master.enter_interrupt_mode()` in `env.rs` before the futures list. The `Interrupt` arm comment now notes that master cannot block in `fd_group_wait` (would starve tokio); CPU savings on master come from `spdk_thread_poll`'s interrupt path, not from the reactor actually sleeping. Casperf release numbers posted [here](https://github.com/openebs/mayastor/pull/1966#issuecomment-4273526113).

test/python/tests/rpc/docker-compose.yml
jr42: Done in `e9f70d4c` — three smoke tests added to `test/python/tests/rpc/test_interrupt_mode.py`, reusing the existing rpc compose. They `@skipif` when `ENABLE_INTERRUPT_MODE` isn't set, so they cost nothing in poll-mode runs; they fire when any suite is run with the env var on.

- `test_reactor_state_is_interrupt` — log-grep assert that the env var was honoured and at least one reactor transitioned.
- `test_idle_cpu_is_low` — `docker stats` assert that the idle container stays under 50% CPU (busy-poll would be ~100% per core).
- `test_wakeup_from_sleep` — timed gRPC round-trip asserts the wakeup-eventfd path actually kicks the reactor out of `fd_group_wait`.

No new compose file — any existing suite can get interrupt coverage just by adding `ENABLE_INTERRUPT_MODE=true` as a CI matrix dimension. Happy to extend into a full parallel pytest job if you prefer that over the targeted smoke.

tiagolobocastro:
bors try

tiagolobocastro:
@jr42 could you please update to the tip of spdk-rs ?

jr42:
> @jr42 could you please update to the tip of spdk-rs ?

done

io-engine/src/core/reactor.rs
```diff
@@ -279,10 +313,50 @@ impl<'a> IntoIterator for &'a Reactors {
...
+        // Initialise the per-reactor fd_group + wakeup eventfd for
+        // interrupt mode. We fail the process on any setup error
+        // rather than silently falling back to poll mode: a mixed
+        // configuration where some reactors interrupt and others poll
+        // has been a source of hard-to-diagnose deadlocks in the past
+        // (see PR #1966 review). ENOMEM or init-order violations at
+        // boot are unrecoverable anyway.
+        let (fgrp, wakeup_fd) = if interrupt_enabled {
+            let fg = spdk_rs::FdGroup::create().unwrap_or_else(|err| {
+                panic!("Reactor core {}: failed to create fd_group ({})", core, err)
```
tiagolobocastro, line 329: Could you make similar changes below too?

```suggestion
                panic!("Reactor core {core}: failed to create fd_group ({err})")
```

tiagolobocastro (approved):
Could you fix this as well please:
```
thread 'main' panicked at io-engine/src/core/env.rs:649:13:
invalid reactor state during shutdown
```

tiagolobocastro:
@jr42 I'm looking into adding support for running the master reactor for app/thread only, and leave all I/O for the "remote" reactors.
The reasoning behind it is that on very large scale testing, even listing replicas can take a while, which may impact performance.

Could we setup things such that the "master" reactor is interrupt-based, and the "remotes" are in polling?

jr42:
Sorry @tiagolobocastro — total oversight on my end. Pushed a3ed0a28 with a regression test (`io-engine/tests/reactor_interrupt.rs`) that pins down the `Interrupt` arm so this can't creep back.

On master-interrupt / remote-poll: yes, doable. The per-reactor `interrupt_enabled` flag already exists and `poll_reactor` only enters interrupt when it's set (`reactor.rs:561`), so it's just a CLI knob — something like `--enable-interrupt-mode=off|master|all`. Happy to prototype as a follow-up PR if that works for you.

tiagolobocastro:
Sorry would you mind fixing the spdk-rs conflict? Another PR got merged with the updated spdk-rs, so that's probably why
Also would you mind please adding the signoff to your latest commit?

> On master-interrupt / remote-poll: yes, doable. The per-reactor `interrupt_enabled` flag already exists and `poll_reactor` only enters interrupt when it's set (`reactor.rs:561`), so it's just a CLI knob — something like `--enable-interrupt-mode=off|master|all`. Happy to prototype as a follow-up PR if that works for you.

That'd be awesome thank you!

Abhinandan-Purkait (approved):

tiagolobocastro:
bors try

tiagolobocastro:
@jr42 looks like the test timed out at `/home/ubuntu/_work/mayastor/mayastor/target/debug/deps/reactor_interrupt-ba193413923ba2ea`

jr42:
Sorry about the timeout @tiagolobocastro — the test was driving the full shutdown path and hit a separate hang in interrupt mode. Rewrote it to mirror the `poller.rs` pattern and isolate just the state check. Runs in ~0.2s; still reproduces the original panic when the fix is reverted. Pushed as 6ffa63dd.

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

tiagolobocastro:
merged, thanks for this awesome contribution @jr42 :rocket:  :fireworks: 
would you be adding a helm-chart variable to enable interrupt mode as well?

would be awesome if you could join the community meeting [next thursday, 30th](https://zoom-lfx.platform.linuxfoundation.org/meetings/openebs?view=list).

jr42:
> merged, thanks for this awesome contribution @jr42 :rocket:  :fireworks: 
> 
> would you be adding a helm-chart variable to enable interrupt mode as well?
> 
> 
> 
> would be awesome if you could join the community meeting [next thursday, 30th](https://zoom-lfx.platform.linuxfoundation.org/meetings/openebs?view=list).
> 
> 

Will do and also try to join, if work permits.

tiagolobocastro:
@jr42 I remember a crash on shutdown had been fixed but looks like still happening ?

>io-engine: fd_group.c:583: spdk_fd_group_destroy: Assertion `fgrp->parent == NULL' failed.

tiagolobocastro:
@jr42 raised #1997 for that if you could take a look.

PR #1967, opened 2026-04-16
refactor: use new impl of snapshot clone counting

tiagolobocastro:
The previous impl did not account for chaining. This was only found during control-plane tests which is showing that the snapshot testing in this repo is lacking!

I was not able to modify the existing impl to traverse the chain in an optimal way so instead I've added a hammer approach which iterates all blobs and finds clones.

It's still much faster than what we had before but there's still room for improvement, namely chaning the snapshot chain to rb tree.


niladrih (approved):

Abhinandan-Purkait (approved):

niladrih:
bors merge

tiagolobocastro:
setting up pre-requisites failed...
bors merge

PR #1968, opened 2026-04-21
refactor(pool/probe): implement probe for more devices and use import

tiagolobocastro:
fix: don't log debug of failed future
    
    This can be extremely large when listing large numbers of nexuses
    or replicas/snapshots.
    
---

    refactor(pool/probe): implement probe for more devices and use import
    
    Updates api to include probe option import.
    Also implement the probe for other devices, even though there's currently
    no simple way of probing (ie nvmf), as otherwise we'd be adding a breaking
    change when using the probe via the control-plane.

### DIFF
--- a/io-engine/src/bdev/aio.rs
+++ b/io-engine/src/bdev/aio.rs
@@ -186,7 +186,7 @@ impl CreateDestroy for Aio {
 }
 
 impl super::Probe for Aio {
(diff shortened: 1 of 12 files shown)


Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

PR #1969, opened 2026-04-21
refactor: add mayastor cli args to casperf

tiagolobocastro:
Allows us to tweak common mayastor args, such as the coming interrupt mode. Also whilst we're here, it's a good time to modernize using clap derive api.


tiagolobocastro:
Once #1966 lands, we'll be able to run it with interrupt mode, which is interesting to see if we can still peg it, and also to see if there's cycles to spared in case we have "fast cpu than storage".

Abhinandan-Purkait (approved):

niladrih (approved):

tiagolobocastro:
bors merge

PR #1976, opened 2026-04-30
Nexus - handle child hot removal

tiagolobocastro:
fix(nexus/nvmf): don't wait forever for the remove-channel
    
    The snapshot tests revealed a very weird race when the nexus child
    is retired, but re-created manually.
    In this case the device is not fully retired and the device handles
    remain in the nexus child.
    However, the listeners are gone! So when we try to close the nexus
    child the nexus does not receive the device removal event, and is
    unable to trigger the unplug.
    
    This kinda breaks the fix where we wait for proper device removal
    before completing the nexus child close...
    Whereas this will probably not happen outside of the tests, let's
    add a time limit whilst waiting for the unplug?

---

    test(lvs/hot-remove): test hot-removing and hot-attaching the disk
    
    When the pool disk is reattached, we have to ensure the initiator handles this
    correctly.
    Adding tests with the nexus present both local and remote to the lvol and also
    with or without I/O.
    
    This test revealed an issue with the current nvmf initiator where it does not
    handle namespace reattachment.
    
---

    fix(lvs/import): retry lvs sharing on import
    
    We now unshare all lvols prior to sharing them on lvs import.
    However, post calling unshare, the subsystem is not fully removed.
    I suspect this may have something to do with the fact that the subsystem
    now uses a queue of operations, though not entirely sure.
    
    Nevertheless, to address this we now retry this for up to 2 seconds, which
    seems to be enough for my testing, since the subsystem is not really doing
    anything as the namespace is actually removed.
    Even if this does not hold true, the control-plane can always re-share.
    
    I have tried to re-add the namespace, which was a simpler change, but this
    revelead an issue with our nvmf initiator when readding the namespace in
    namespace races with the child controller shutdown because both are trying
    to take each other controller's locks. There's probably also some other
    issue here.

---

    fix(nexus): retire child when bdev is hot-removed
    
    A bdev may be hot-removed underneath the nexus.
    The current behaviour was for the child to be unplugged.
    
    If there's no I/O then the nexus will no detect this to happen. This
    may have 2 adverse side effects:
    1. the nexus child is stuck as degraded, which the control-plane ignores,
    meaning there's no attempt to remove it
    2. the bdev is not fully removed until all handles are dropped, which
    can then cause other problems, such as lead to stuck pool unload as a result
    of a hot-remove.
    
    However there's currently no clear way of knowing hot-removal is being
    triggered in these cases so we have to inspect the child state to determine.
    In case we deem the bdev removal due to hot-remove, then we actually trigger
    the child to be retired which will update the pstor in addition to complete
    bdev removal.
    
---

    refactor: add options to the interface Share::unshare
    
    Add options to the unshare method part of the Share interface.
    This then allows us to add persistence flag to the unshare.
    
    With this flag we can then build on the previous commit and
    not update the persistence when unsharing the lvol as part of
    the re-import logic. This removes the potential race and also
    avoids the I/O updating the lvol metadata xattr!
    
---

    fix: clean up nvmf subsystems when pool is re-imported
    
    If the pool is hot-removed, the lvols are hot-removed which results
    in the nvmf namespaces getting hot-removed whilst leaving the nvmf
    target in place!
    
    We might need to consider if we need an SPDK change to tear down the
    target in such case, but for now the fix is to unshare the lvol first,
    allowing it to be re-shared again. We could look into simply attaching
    the namespace again, though this change seems safer.
    
    In order to do this we change the share detection to also probe the
    nvmf targets by nqn, otherwise it would fail to find the share since
    the lvol is not owned by the nvmf on re-import!

---

    fix(nexus/child): close child properly when hot-plug precedes close
    
    If a nexus child bdev is hot-removed, the unplug runs ahead of the close.
    When the unplug runs first time, it would not destroy the block-device
    because it's not closing yet...
    During close itself, it will destroy the device, which will again lead to
    another unplug... however the first unplug already signalled the removal
    channel, so the close does not wait for another unplug and would mark the
    child as not destroying... then the unplug runs again, but because the
    child is not destroying, then it would still not destroy the block-device!
    
    The fix proposed here is to make the close drain the remove device channel
    until the device is destroyed.
    I *think* this is safe, because the unplug should always happen, though if
    this is not true, it could lead to deadlocks... so we'll need both the CI
    and the E2E tests to prove this holds true.
    
    We also bump the remove_channel capacity to 2 just in case unplug runs
    before we drain the first entry...


tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors try

Abhinandan-Purkait (approved):

niladrih (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
oops typo found
bors cancel

tiagolobocastro:
bors merge

PR #1982, opened 2026-05-05
Pool stall fixes

tiagolobocastro:
fix(lvs/stall): i/o on non-init thread and concurrent stalls
    
    A collection of fixes and refactorings, namely:
    failures when the stalled i/o was sent from !init-thread
    multiple i/os failing and triggering the stall callback
    handling of concurrency when modifying the stall flag
    
---

    test(lvs/stall): make stall test more robust
    
    Test the stalls by issuing actual I/O against a replica bdev.
    Also send this I/O via master and non-master reactors, concurrently,
    as well as sending staggard I/O.
    
    All in all, this test needs to be made even more robust, perhaps
    running fio, though E2E should be sufficient as well.
    
---

    fix(handle/guard): store current thread in the handle guard
    
    When a device handle is dropped, it gets sent to the primary thread.
    However, this doesn't work when we open the device handle in another thread,
    which is how I found this whilst testing writing directly from another
    reactors thread.

---

    test: resume suspended lv's when cleaning up
    
    Some tests may suspend the lv's, so we should resume them, otherwise
    cleanup may get stuck..
    
---

    test: clean up stale lvm and loop devices properly
    
    The backing file for the loop may have been deleted, so rather than
    inspecting the back directory use losetup itself to find the leaked
    devices.

### DIFF
--- a/io-engine/src/bdev/device.rs
+++ b/io-engine/src/bdev/device.rs
@@ -1,16 +1,10 @@
 //!
 //! Trait implementation for native bdev
 
-use std::{
-    collections::HashMap,
-    convert::TryFrom,
-    os::raw::c_void,
-    sync::{Arc, Mutex},
-};
-
 use async_trait::async_trait;
 use nix::errno::Errno;
 use once_cell::sync::{Lazy, OnceCell};
+use std::{collections::HashMap, convert::TryFrom, os::raw::c_void, rc::Rc, sync::Mutex};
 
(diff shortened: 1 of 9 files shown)


tiagolobocastro:
bors try

Abhinandan-Purkait (approved):

niladrih (approved):

tiagolobocastro:
bors merge

PR #1985, opened 2026-05-08
fix(lvs/stall): properly handle bdevs which don't support reset

tiagolobocastro:
test(lvs/stall): concurrent test with aio and uring
    
    Refactor the lvs_stall test to support multiple bdev types, in
    this case aio and uring.
    Then run the aio and uring tests concurrently.
    
---

    fix(lvs/stall): properly handle bdevs which don't support reset
    
    After spdk fix, the lvs reset now fails with -ENOTSUP and doesn't
    increment the io error stat count.
    Then we simply proceed to the superblock read directly, and even
    though there's no reset, we expect the read to eventually complete.

### DIFF
--- a/.github/workflows/bdd.yml
+++ b/.github/workflows/bdd.yml
@@ -85,10 +85,10 @@ jobs:
             mv test/python/reports ci-report/python
           fi
           nix-shell --run "./scripts/ci-report.sh"
-        if: failure()
+        if: failure() || cancelled()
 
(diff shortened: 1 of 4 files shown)


Abhinandan-Purkait (approved):

niladrih (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1994, opened 2026-06-04
refactor: refactor lvol clusters and nvmf subsystem

tiagolobocastro:
refactor: prepare for subsystem stop-destroy

    A new spdk api stops the subsystem for destroy, preventing operations
    from further modifying the subsystem.
    This is useful to avoid a race where resume could race with unpublish.

    However we can't make use of this api yet because it'd break our current
    use of the queued subsystem operations.
    Once stop-for-destroy is issued, any operation will fail straight away
    with -ENODEV, rather than only once the subsystem is actually stopped.
    This means a pause would fail with -ENODEV straightaway, which is not
    what we want since we need all IOs to complete!!

    In addition, we currently have code which takes the subsystem and then
    loops inside with async code.
    This also does not seem safe as in case of a destroy we'd be holding a
    dangling pointer!

---

    refactor: replace cache with spdk num allocated cluster

    Revert the added blob cache and use the blob's num_allocated_clusters.
    The cache was just a short-term solution and it's not needed since we
    can get the clusters from the spdk blob.

### DIFF
--- a/io-engine/examples/lvs-eval/display.rs
+++ b/io-engine/examples/lvs-eval/display.rs
@@ -5,7 +5,7 @@ use io_engine::{
 use prettytable::{row, Table};
 use spdk_rs::libspdk::{
     spdk_bit_array, spdk_bit_array_capacity, spdk_bit_array_get, spdk_bit_pool,
-    spdk_bit_pool_capacity, spdk_bit_pool_is_allocated, spdk_blob_calc_used_clusters,
+    spdk_bit_pool_capacity, spdk_bit_pool_is_allocated, spdk_blob_get_num_allocated_clusters,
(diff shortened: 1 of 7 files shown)


tiagolobocastro:
bors try

abhilashshetty04 (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

PR #2001, opened 2026-06-11
Enable support for RDMA for both initiator and target

martino87r:
First attempt to enable RDMA for both initiator and target, removes the hard coded TCP transport and re-uses similar approach to existing spdk code.

### DIFF
--- a/io-engine/src/bdev/nvmx/controller.rs
+++ b/io-engine/src/bdev/nvmx/controller.rs
@@ -1114,7 +1114,10 @@ pub(crate) mod options {
 pub(crate) mod transport {
     use std::{ffi::CStr, fmt::Debug, str::FromStr};
 
-    use spdk_rs::{ffihelper::copy_str_with_null, libspdk::spdk_nvme_transport_id};
+    use spdk_rs::{
+        ffihelper::copy_str_with_null,
+        libspdk::{spdk_nvme_transport_id, SPDK_NVME_TRANSPORT_RDMA, SPDK_NVME_TRANSPORT_TCP},
+    };
 
     pub struct NvmeTransportId(spdk_nvme_transport_id);
 
@@ -1169,21 +1172,6 @@ pub(crate) mod transport {
     }
 
     #[derive(Debug, Default)]
-    #[allow(clippy::upper_case_acronyms)]
-    enum TransportId {
-        #[default]
-        TCP = 0x3,
-    }
-
-    impl From<TransportId> for String {
-        fn from(t: TransportId) -> Self {
-            match t {
-                TransportId::TCP => String::from("tcp"),
-            }
-        }
-    }
-
-    #[derive(Debug, Default)]
     #[allow(dead_code)]
     pub(crate) enum AdressFamily {
         #[default]
@@ -1194,10 +1182,34 @@ pub(crate) mod transport {
         NvmfAdrfamLoop = 0xfe,
     }
 
+    /// Transport selected for an outgoing nvmf connection.
+    #[derive(Copy, Clone, Debug, Default, PartialEq, Eq, strum_macros::EnumIter)]
+    #[allow(clippy::upper_case_acronyms)]
+    pub enum TrType {
+        #[default]
+        TCP,
+        RDMA,
+    }
+
+    impl TrType {
+        fn spdk_type(self) -> u32 {
+            match self {
+                Self::TCP => SPDK_NVME_TRANSPORT_TCP,
+                Self::RDMA => SPDK_NVME_TRANSPORT_RDMA,
+            }
+        }
+        fn trstring(self) -> &'static str {
+            match self {
+                Self::TCP => "TCP",
+                Self::RDMA => "RDMA",
+            }
+        }
+    }
+
     #[derive(Default, Debug)]
     #[allow(dead_code)]
     pub struct Builder {
-        trid: TransportId,
+        trtype: TrType,
         adrfam: AdressFamily,
         svcid: String,
         traddr: String,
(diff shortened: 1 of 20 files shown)


io-engine/src/bdev/nvmx/uri.rs
```diff
@@ -165,6 +168,16 @@ impl TryFrom<&Url> for NvmfDeviceTemplate {
             .trim_start_matches("[")
             .trim_end_matches("]")
             .to_string();
+
+        let trtype = match url.scheme() {
+            "nvmf" | "nvmf+tcp" => TrType::TCP,
+            "nvmf+rdma+tcp" => TrType::RDMA,
```
tiagolobocastro: probably also need to check if RDMA is enabled on this io-engine

io-engine/src/bdev/nvmx/controller.rs
```diff
@@ -1274,6 +1290,20 @@ pub(crate) mod transport {
...
             assert_eq!(transport.subnqn(), "nqn.2021-01-01:test.nqn");
             assert_eq!(transport.svcid(), "4420");
+            assert_eq!(transport.trtype(), "TCP");
+        }
+        #[test]
+        fn test_transport_rdma() {
+            let transport = transport::Builder::new()
+                .with_subnqn("nqn.2021-01-01.test.nqn")
+                .with_svcid("4420")
+                .with_traddr("127.0.0.1")
```
tiagolobocastro, line 1295: missing `.with_trtype` ?

io-engine/src/bdev/nvmx/controller.rs
```diff
@@ -1204,10 +1189,37 @@ pub(crate) mod transport {
...
+    #[derive(Copy, Clone, Debug, PartialEq, Eq)]
+    #[allow(clippy::upper_case_acronyms)]
+    pub enum TrType { TCP, RDMA }
+
+    impl Default for TrType {
+        fn default() -> Self { Self::TCP }
+    }
+
+    impl TrType {
+        #[inline]
```
tiagolobocastro: don't need these inlines here

io-engine/src/bdev/nvmx/controller.rs
```diff
@@ -1114,7 +1114,11 @@ pub(crate) mod options {
 pub(crate) mod transport {
     use std::{ffi::CStr, fmt::Debug, str::FromStr};
 
-    use spdk_rs::{ffihelper::copy_str_with_null, libspdk::spdk_nvme_transport_id};
+    use spdk_rs::{ffihelper::copy_str_with_null,
```
tiagolobocastro: can you run the linter: `./scripts/rust-style.sh`

tiagolobocastro (commented):
Haven't looked in detail here, though seems plausible
Can you try adding tests as well please? See existing `test_rdma_target` for reference

io-engine/src/bdev/nvmx/uri.rs
martino87r: I'm assuming the control plane passes rdma only when it's enabled with the env-var; though I would be glad for some inputs for where that validation should happen

io-engine/src/bdev/nvmx/controller.rs
```diff
@@ -1264,7 +1280,7 @@ pub(crate) mod transport {
         use crate::bdev::nvmx::controller::transport;
 
         #[test]
-        fn test_transport_id() {
+        fn test_transport_tcp() {
```
susobhandey: Do we really need separate function for tcp & rdma ? why can't we just have a single test function --> with argument trtype? fn test_transport(trtype) {

if trtype == TCP:
do this
else if trtype == RDMA
do this
else
NOT SUPPORTED

io-engine/src/bdev/nvmx/uri.rs
```diff
@@ -165,6 +168,16 @@ impl TryFrom<&Url> for NvmfDeviceTemplate {
...
+            other => return Err(BdevError::InvalidUri {
+                uri: url.to_string(),
+                message: format!("unsupported nvmf scheme '{other}'"),
+            }),
```
susobhandey: Here are few things need to be considered. Hope it is covered. Let's say if someone wants to run initiator side TCP & replica side RDMA or vice versa.

tiagolobocastro:
Thanks @martino87r, I had some some test changes but I can add those in following PR
Would you mind squashing all your commits into 1 and also add DCO signoff? https://github.com/openebs/community/blob/HEAD/CONTRIBUTING.md#sign-your-work
Once squashed to sign off: `git commit --amend -s`

tiagolobocastro:
Running the test I made a mistake and deleted the rdma-rxe device, which causes spdk assert.
This may not happen in prod, though worth noting:

```text
[ERROR mayastor::spdk:rdma_provider_verbs.c:130] rdma_disconnect failed, errno No such device or address (6)
io-engine: nvme_rdma.c:2327: nvme_rdma_ctrlr_delete_io_qpair: Assertion `rc == 0' failed.
```

susobhandey:
I have verified it with real rdma cluster and it is working, so what is the plan here? on top of it some more work is needed.

susobhandey (approved):
LGTM

io-engine/src/bdev/nvmx/controller.rs
tiagolobocastro: Done

io-engine/src/bdev/nvmx/uri.rs
tiagolobocastro: hmm this could be possible, though would require changing the existing plain `--enable-rdma=true/false` into something capable of specifying more options, ie:

```
enum RdmaModes {
  None,
  All,
  Nexus,
  Replica
}
```

Is there a use-case for wanting to run in a "half mode"?
Btw existing rdma targets already support TCP as well, so initiators without rdma can always connect in tcp mode.

Abhinandan-Purkait: Any specific reason why someone will want a mix? Also, do we have the fallback tcp mode for between nexus and replica if nexus and replica connect fail over rdma?

tiagolobocastro: If the initiator is TCP-only, it'll connect over TCP, no fallback in the sense, if RDMA fails to connect, try to connect over TCP
Do we do that in the csi-node?

abhilashshetty04: @Abhinandan-Purkait , csi-node does fallback to tcp if either of initiator or nexus node is not rdma capable, right?

Abhinandan-Purkait: Yes

Abhinandan-Purkait (approved):

io-engine/src/bdev/nvmx/uri.rs
tiagolobocastro: Yes does based on initial check for RDMA only
I've added the same here, PTAL @abhilashshetty04

abhilashshetty04 (approved):

PR #2002, opened 2026-06-11
fix: tackle various CI flaky tests

tiagolobocastro:
ci: add container logs to journalctl
    
    This will help diagnosing in case of failure.
    
---

    ci: check for coredumps even on failures

---

    fix: revert back to arc descriptor guard
    
    When a previous fix added the Thread it changed the descriptor guard
    to be protected by Rc.
    This turned out not to be sound because this guard is actually shared
    across the reactors which may run on different cores!
    
    For now we add the Arc back and we mark the guard as Sync, since the
    thread is only used via Drop and should not be freed by SPDK whilst
    we still have the guard anyway.
    
    todo: we should get the thread from the descriptor itself, rather than
    keep another copy of it.
    Also to consider creating separate guards in the various cores, rather
    than reusing the same one?
    
---

    test: wait until stall clears before IO
    
    vbdev_lvs_bs_bdev_reset() must drain before new I/Os succeed, so there's
    a race between dmsetup resume and trying to send new I/Os.
    Instead we now wait until the stall is cleared.

### DIFF
--- a/.github/workflows/bdd.yml
+++ b/.github/workflows/bdd.yml
@@ -62,6 +62,11 @@ jobs:
           # Check if nvme config is valid, otherwise apply it
           sudo ./scripts/nvme-conf.sh --check --apply --overwrite
 
+      - name: Configure Docker logging driver
+        run: |
+          echo '{"log-driver": "journald"}' | sudo tee /etc/docker/daemon.json
+          sudo systemctl restart docker
+
       - name: Setup VENV
         run: nix-shell --run "./test/python/setup.sh"
 
--- a/.github/workflows/bdd.yml
+++ b/.github/workflows/bdd.yml
@@ -94,6 +94,7 @@ jobs:
           path: ./ci-report/ci-report.tar.gz
 
       - name: Check Coredumps
+        if: always()
         run: sudo ./scripts/check-coredumps.sh --since "${TEST_START_DATE}"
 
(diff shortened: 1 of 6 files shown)


niladrih (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
Resolves https://github.com/openebs/mayastor/issues/2004

tiagolobocastro:
There's still one outstanding issue, null pointer dereference:

```
0  nvmf_check_subsystem_active (req=0x55b3987ae100) at ctrlr.c:4879
#1  0x000055b379018c54 in spdk_nvmf_request_exec (req=0x55b3987ae100) at ctrlr.c:4995
#2  0x000055b37904b6b9 in nvmf_tcp_req_process (ttransport=0x55b3986393a0, tcp_req=0x55b3987ae100) at tcp.c:3268
```

It's only on develop, there's already a PR fix, so I'll merge this for now and will update spdk-rs and this repo on another PR so we can back port this to the release.

bors merge

tiagolobocastro:
@mergify backport release/2.11

PR #2003, opened 2026-06-12

susobhandey:
Kubernetes CPU Manager running with cpuManagerPolicy=static
may rewrite container cpusets when Guaranteed pods are
admitted or removed.

When cpuset.cpus.effective is updated, the kernel resets
task affinity masks for threads in the cgroup. This causes
Mayastor reactor threads to lose their CPU pinning and
Tokio worker threads to lose their unaffinitized CPU mask.

Add a cpuset monitor that watches for cpuset updates and
automatically restores reactor affinity. Tokio worker TIDs
are tracked so worker unaffinity can also be re-applied,
preserving separation between reactor and runtime threads.

- cpuManagerPolicy=static
- isolcpus
- KubeVirt dedicated CPUs

admission/removal events.


io-engine/src/core/affinity.rs
```diff
@@ -0,0 +1,64 @@
+use std::path::Path;
+
+use inotify::{Inotify, WatchMask};
+
+use crate::core::{runtime, Mthread, Reactors};
+
+const CPUSET_FILE: &str = "/sys/fs/cgroup/cpuset.cpus.effective";
+
+pub fn start_monitor() {
+    std::thread::Builder::new()
```
tiagolobocastro: This thread itself also needs to be reaffinitized right?
Instead of manually creating here, use the existing runtime which you've already modified for this anyhow, so we'll get the re-affinity for free

io-engine/src/core/affinity.rs
```diff
@@ -0,0 +1,64 @@
...
+        .watches()
+        .add(CPUSET_FILE, WatchMask::MODIFY | WatchMask::CLOSE_WRITE)
+        .unwrap();
+
+    tracing::info!("watching {} for cpuset changes", CPUSET_FILE);
+
+    let mut buffer = [0u8; 4096];
+
+    loop {
+        match inotify.read_events_blocking(&mut buffer) {
```
tiagolobocastro: @Abhinandan-Purkait IIRC when you had used this in another project there was an issue causing cpu hog, could you please check here?

io-engine/src/core/affinity.rs
```diff
@@ -0,0 +1,64 @@
...
+            Ok(events) => {
+                let mut changed = false;
+
+                for _ in events {
```
tiagolobocastro: not sure what's happening here, we just seem to be looping events and not checking for changes?

io-engine/src/core/runtime.rs
```diff
@@ -3,13 +3,14 @@
...
 use crate::core::Reactor;
 use futures::{channel::oneshot, Future};
 use once_cell::sync::Lazy;
+use parking_lot::Mutex;
+use std::collections::HashSet;
 use tokio::task::JoinHandle;
-
-use super::Mthread;
-
+static TOKIO_WORKERS: Lazy<Mutex<HashSet<libc::pid_t>>> = Lazy::new(|| Mutex::new(HashSet::new()));
```
tiagolobocastro: Rename this, since it's not just the tokio workers, it's the device monitor, reactor monitor, etc

io-engine/src/core/affinity.rs
```diff
@@ -0,0 +1,64 @@
+use std::path::Path;
+
+use inotify::{Inotify, WatchMask};
+
+use crate::core::{runtime, Mthread, Reactors};
+
+const CPUSET_FILE: &str = "/sys/fs/cgroup/cpuset.cpus.effective";
```
tiagolobocastro: Is this the only thing we need to listen to?

tiagolobocastro (commented):
Looks reasonable! Would be good if you could test this with a K8s cluster and verify the change and then the fix with the patch

io-engine/src/core/runtime.rs
susobhandey: fixed!

io-engine/src/core/affinity.rs
susobhandey: Likely fixed.

PR #2007, opened 2026-06-13
feat(nexus): persist clean_shutdown on share/unshare transitions

yugchaudhari:
This is the data-plane change for offline volume rebuild iteration 4 (OEP openebs/openebs#4208).

## The core idea

A nexus is "clean" (no possible dirty data) iff it is currently unshared. While unshared, no initiator can connect and no I/O can dirty the children, so a crash from that state is equivalent to a graceful shutdown.

Today the only thing that flips `NexusInfo.clean_shutdown` to `true` is `PersistOp::Shutdown`, called during destroy/shutdown. An unshared nexus that crashes therefore looks "dirty" to the control plane on recovery, and `scheduling.rs::healthy_children` falls into the `HealthyChildItems::One` branch which forces every non-elected child into a full rebuild. For offline-rebuild nexuses (which by design never carry frontend I/O) that's a lot of needless copying.

## What changes

- New `PersistOp::SetCleanShutdown { clean }` variant. Wires through `nexus_info.clean_shutdown` like the other persist ops.
- `register_instance` calls `SetCleanShutdown { clean: true }` right after `PersistOp::Create`. A freshly created nexus is unshared by construction.
- `share_ext` calls `SetCleanShutdown { clean: false }` before the share actually goes live. A crash mid-share now leaves the nexus marked dirty, which is the safe direction, the existing share path already starts the nexus from a known-clean state, so over-marking-dirty just means a full rebuild on recovery instead of incorrectly trusting children.
- `unshare_nexus` calls `SetCleanShutdown { clean: true }` after the target is cleared. Best-effort: a persistence failure here is logged and swallowed, because the worst case is reverting to today's behaviour (full rebuild on next start), not data loss.
- The save-failure path that self-shutdowns now treats `SetCleanShutdown { clean: true }` the same as `PersistOp::Shutdown`, both are correctness-neutral, so a transient etcd hiccup shouldn't take the nexus down. `SetCleanShutdown { clean: false }` still self-shutdowns on failure, because that path is load-bearing for the dirty-window guarantee.

## Crash semantics

| Scenario | `clean_shutdown` after crash | Control-plane recovery |
|---|---|---|
| Crash before `share_ext` runs (offline rebuild) | `true` | Treat children as clean |
| Crash during `share_ext` (after the persist, before the share completes) | `false` | Existing full-rebuild behaviour |
| Crash on a shared nexus | `false` | Existing full-rebuild behaviour |
| Crash after `unshare_nexus` | `true` | Treat children as clean |
| Graceful shutdown | `true` | Treat children as clean |

No control-plane changes needed for this PR. `scheduling.rs::healthy_children` already does the right thing the moment it sees `clean_shutdown = true`.

## Testing

Local build + lint sweep on a Linux nix-shell environment:

```
cargo fmt --check
./scripts/rust-linter.sh          # cargo clippy --all --all-targets --features=io-engine-testing -- -D warnings
cargo build --bins --features=io-engine-testing
./scripts/cargo-test.sh
./scripts/pytest-tests.sh         # BDD
```

(All pass locally)

## Follow-up

There's a small follow-up on the control-plane side: today the offline-rebuild temp nexus is constructed from healthy replicas only, so when the offline replica returns it's added as a fresh child and the existing partial-rebuild / I/O-log machinery in `nexus_child.rs::start_io_log` never gets a chance to give it a cheap return path. Including the faulted child at temp-nexus-creation time fixes that, and lines up with `handle_faulted_child` / `faulted_child_wait` that already exist for shared volumes. Will raise a separate PR against mayastor-control-plane once this is merged.

cc @tiagolobocastro @Abhinandan-Purkait

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev.rs
@@ -802,6 +802,9 @@ impl<'n> Nexus<'n> {
         };
 
         // Persist the fact that the nexus is now successfully open.
+        // `PersistOp::Create` records the initial clean-while-unshared state,
+        // which then gets flipped on share/unshare to maintain the invariant
+        // `clean_shutdown <=> currently unshared`.
         // We have to do this before setting the nexus to open so that
         // nexus list does not return this nexus until it is persisted.
         if let Err(e) = nex.persist(PersistOp::Create).await {
@@ -831,7 +834,9 @@ impl<'n> Nexus<'n> {
     pub async fn destroy_ext(mut self: Pin<&mut Self>, sigterm: bool) -> Result<(), Error> {
         info!("{:?}: destroying nexus...", self);
 
-        self.as_mut().unshare_nexus().await?;
+        // Destroy already issues `PersistOp::Shutdown` below (which sets
+        // `clean_shutdown = true`), so skip the redundant persist here.
+        self.as_mut().unshare_nexus_internal(false).await?;
 
         // wait for all rebuild jobs to be cancelled before proceeding with the
         // destruction of the nexus
--- a/io-engine/src/bdev/nexus/nexus_persistence.rs
+++ b/io-engine/src/bdev/nexus/nexus_persistence.rs
@@ -82,6 +82,13 @@ pub(crate) enum PersistOp<'a> {
     },
     /// Save the clean shutdown variable.
     Shutdown,
+    /// Explicitly set the clean shutdown flag while the nexus is alive.
+    /// This is used to track the invariant `clean_shutdown <=> currently
+    /// unshared`: while no I/O can possibly flow (unshared), the children
+    /// cannot get dirty, so a crash from that state is equivalent to a clean
+    /// shutdown. Set to `true` on nexus create and on unshare, set to `false`
+    /// when the nexus transitions to shared and I/O becomes possible.
+    SetCleanShutdown { clean: bool },
 }
 
 impl std::fmt::Debug for PersistOp<'_> {
@@ -97,6 +104,9 @@ impl std::fmt::Debug for PersistOp<'_> {
                 child_uri, healthy, ..
             } => write!(f, "UpdateChildCond: {child_uri}: {healthy}"),
             PersistOp::Shutdown => write!(f, "Shutdown"),
+            PersistOp::SetCleanShutdown { clean } => {
+                write!(f, "SetCleanShutdown: {clean}")
+            }
         }
     }
 }
@@ -146,6 +156,11 @@ impl<'n> Nexus<'n> {
                         reason: "only child is unhealthy".to_string(),
                     });
                 }
+                // A freshly created nexus is intrinsically unshared: no
+                // initiator can connect and no I/O can dirty the children
+                // until `share_ext` runs. Record the clean state up front so
+                // that a crash from here is not treated as a dirty shutdown.
+                nexus_info.clean_shutdown = true;
             }
             PersistOp::AddChild { child_uri, healthy } => {
                 // Add the state of a new child. This should only be called
@@ -238,6 +253,12 @@ impl<'n> Nexus<'n> {
                 // This should only be called when destroying a nexus.
                 nexus_info.clean_shutdown = true;
             }
+            PersistOp::SetCleanShutdown { clean } => {
+                // Only update the clean shutdown variable. This is used by
+                // share/unshare/create transitions to track whether the nexus
+                // is currently in a state where I/O can dirty the children.
+                nexus_info.clean_shutdown = *clean;
+            }
         }
 
         match self.save(&persistent_nexus_info).await {
@@ -246,8 +267,10 @@ impl<'n> Nexus<'n> {
                 Ok(())
             }
             Err(e) => {
-                // If the operation was an update for shutdown, no need to
-                // shutdown in the case of an error.
+                // `PersistOp::Shutdown` is on the destroy path already, so
+                // self-shutting-down on top of that is pointless. Every other
+                // op falls through to the safe-by-default self-shutdown
+                // behaviour, since the save layer already retries internally.
                 if matches!(op, PersistOp::Shutdown) {
                     error!("{self:?}: failed to update persistent store: {e}");
                 } else {
--- a/io-engine/src/bdev/nexus/nexus_share.rs
+++ b/io-engine/src/bdev/nexus/nexus_share.rs
@@ -3,9 +3,12 @@ use async_trait::async_trait;
 use snafu::ResultExt;
 use std::pin::Pin;
 
-use super::{nexus_err, Error, NbdDisk, Nexus, NexusTarget};
+use super::{nexus_err, Error, NbdDisk, Nexus, NexusTarget, PersistOp};
 
-use crate::core::{NvmfShareProps, Protocol, PtplProps, Share, UpdateProps};
+use crate::{
+    core::{NvmfShareProps, Protocol, PtplProps, Share, UpdateProps},
+    subsys::NvmfSubsystem,
+};
 
 ///
 /// The sharing of the nexus is different compared to regular bdevs
@@ -38,6 +41,16 @@ impl Share for Nexus<'_> {
         let uri = match self.shared() {
             Some(Protocol::Off) | None => {
                 info!("{:?}: sharing NVMF target...", self);
+
+                // Persist `clean_shutdown = false` before the share goes
+                // live, so that a crash from here cannot leave the nexus
+                // marked clean while an initiator could be connecting and
+                // dirtying children. Lives here (rather than in `share_ext`)
+                // so the same guarantee holds for any direct caller of the
+                // trait method.
+                self.as_mut()
+                    .persist(PersistOp::SetCleanShutdown { clean: false })
+                    .await?;
 
                 let name = self.name.clone();
                 self.as_mut()
@@ -166,6 +179,14 @@ impl<'n> Nexus<'n> {
             // right now Off is mapped to Nbd, will clean up the Nbd related
             // code once we refactor the rust tests that use nbd.
             Protocol::Off => {
+                // Persist `clean_shutdown = false` before the NBD share goes
+                // live. The NVMe-oF path persists the same flag inside
+                // `Share::share_nvmf`, so each protocol owns its own
+                // transition guarantee.
+                self.as_mut()
+                    .persist(PersistOp::SetCleanShutdown { clean: false })
+                    .await?;
+
                 let disk = NbdDisk::create(&self.name)
                     .await
                     .context(nexus_err::ShareNbdNexus {
@@ -198,8 +219,20 @@ impl<'n> Nexus<'n> {
         }
     }
 
-    /// TODO
+    /// Unshare the nexus target and persist the now-clean state so a crash
+    /// from here behaves like a graceful shutdown.
     pub async fn unshare_nexus(mut self: Pin<&mut Self>) -> Result<(), Error> {
+        self.as_mut().unshare_nexus_internal(true).await
+    }
+
+    /// Inner unshare path shared by the public `unshare_nexus` and the
+    /// destroy path. `persist_clean` controls whether the now-unshared state
+    /// is recorded; destroy passes `false` because `PersistOp::Shutdown`
+    /// issued shortly afterwards already flips `clean_shutdown` to `true`.
+    pub(super) async fn unshare_nexus_internal(
+        mut self: Pin<&mut Self>,
+        persist_clean: bool,
+    ) -> Result<(), Error> {
         match unsafe { self.as_mut().get_unchecked_mut().nexus_target.take() } {
             Some(NexusTarget::NbdDisk(disk)) => {
                 info!("{:?}: destroying NBD device target...", self);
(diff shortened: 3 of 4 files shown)


io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -812,6 +812,25 @@ impl<'n> Nexus<'n> {
...
 
+        // The nexus is freshly created and intrinsically unshared: no
+        // initiator is connected and no I/O is possible until `share_ext`
+        // runs. Mark it as cleanly shutdown so that a crash from this state
+        // is not treated as a dirty shutdown by the control plane.
+        // The flag is flipped back to `false` on share, and back to `true` on
+        // unshare, so the invariant `clean_shutdown <=> currently unshared`
+        // holds for the rest of the nexus lifetime.
+        if let Err(e) = nex
+            .persist(PersistOp::SetCleanShutdown { clean: true })
```
tiagolobocastro: Can this not be wired up via the Create above?

io-engine/src/bdev/nexus/nexus_share.rs
```diff
@@ -162,6 +162,14 @@ impl<'n> Nexus<'n> {
             });
         }
 
+        // Persist `clean_shutdown = false` before the share goes live, so that
+        // a crash between this point and the share completing does not leave
+        // the nexus marked clean while an initiator could be connecting and
+        // dirtying children. Safer to over-mark dirty than to under-mark it.
+        self.as_mut()
+            .persist(PersistOp::SetCleanShutdown { clean: false })
```
tiagolobocastro: hmm should this be wired up in the `Share` implementation for the Nexus bdev instead?

Also to consider, what if the bdev is shared via SPDK directly, do we need to cater for that as well? Perhaps adding rdonly semantics would be a way to tackle this?

io-engine/src/bdev/nexus/nexus_share.rs
```diff
@@ -214,7 +222,20 @@ impl<'n> Nexus<'n> {
...
-        self.as_mut().unshare(None).await
+        self.as_mut().unshare(None).await?;
+
+        // No initiator can dirty the children any longer. Record the clean
+        // state so a crash from here behaves the same as a graceful shutdown.
+        // Best-effort: a persistence failure does not compromise data
+        // correctness here (the worst case is the existing crash-recovery
+        // behaviour kicking in, i.e. a full rebuild on next startup), and is
+        // already logged inside `persist`.
+        let _ = self
```
tiagolobocastro: May need some flag here on when to set this because if the nexus is being destroyed, we already set this over there, so would be good to avoid setting it twice

tiagolobocastro (commented):
hmm I think this approach of the clean flag modification has a limitation
Let's say one of the children has a temporary blip, then we'd be marking it a faulted still, even though there's no I/O ongoing
Conversely we can't just ignore it because if the nexus is shared, we have to ensure the health is still accurate. 

An alternative would be to use a different mode of operation altogether, though that' be more involved...
Maybe we can take up this approach first and then improve in another iteration?

io-engine/src/bdev/nexus/nexus_bdev.rs
yugchaudhari: Done, folded into `PersistOp::Create`. The Create handler now sets `nexus_info.clean_shutdown = true` after the children are populated, and I have removed the separate persist call in `register_instance`.

io-engine/src/bdev/nexus/nexus_share.rs
yugchaudhari: `unshare_nexus` now takes a `persist_clean: bool`:

- `destroy_ext` calls `unshare_nexus(false)`, `PersistOp::Shutdown` a few lines below already sets `clean_shutdown = true`, so the unshare-side persist would be redundant.
- gRPC handlers call `unshare_nexus(true)`.
- In-tree test callers (`nexus_io`, `mount_fs`, `ftl_mount_fs`, `add_child`, `replica_timeout`) updated to pass `true`, matching the gRPC behaviour they're standing in for.

io-engine/src/bdev/nexus/nexus_share.rs
yugchaudhari: Good points. On placement: I left it in `share_ext` because that's the single funnel for both the NVMe-oF and the NBD branches. The `Share` implementation on `Nexus` only sees `share_nvmf`, moving the persist there would miss the NBD path in `share_ext`. Happy to refactor if you'd prefer the persist to live in the bdev `Share` impl and drop NBD entirely as part of this.

On direct SPDK share / rdonly semantics: agreed that's the more principled redesign, flipping a nexus to read-only-while-unshared (and treating the persisted "clean" flag as a derivation of that) covers both this case and the "temporary blip with no I/O" case you mentioned. That feels like the iter5+ improvement. Want me to file a separate tracking issue or keep it on the OEP thread (openebs/openebs#4208)?

yugchaudhari:
> hmm I think this approach of the clean flag modification has a limitation Let's say one of the children has a temporary blip, then we'd be marking it a faulted still, even though there's no I/O ongoing Conversely we can't just ignore it because if the nexus is shared, we have to ensure the health is still accurate.
> 
> An alternative would be to use a different mode of operation altogether, though that' be more involved... Maybe we can take up this approach first and then improve in another iteration?

Yes, the limitation is real. A shared nexus that's just quiet still gets marked dirty on a child blip, this PR doesn't help there. The case it helps are offline rebuild and other explicitly-unshared nexuses, where we know no initiator can be connected so the flag is safe to flip.

The rdonly / different-mode direction you mentioned is the cleaner fix for the shared-but-quiet case, same mechanism would cover both. Happy to pick that up as iter5 after this lands. Will track on openebs/openebs#4208 unless you want a separate issue.

io-engine/src/bdev/nexus/nexus_persistence.rs
```diff
@@ -246,9 +267,27 @@ impl<'n> Nexus<'n> {
...
+                //
+                // `PersistOp::Shutdown` is on the destroy path already, no
+                // point self-shutting-down.
+                //
+                // `PersistOp::SetCleanShutdown { clean: true }` is the
+                // create-time / post-unshare marker: the data plane is fine,
+                // we just couldn't tell etcd that the children are clean.
+                // Falling back to the existing crash-recovery behaviour
+                // (treat as dirty, full rebuild next time) is wasteful but
+                // safe. Self-shutdown would be destructive overkill.
```
tiagolobocastro: hmm I'm not sure if there any pitfalls in not updating etcd, but I'd rather keep things consistent for sanity, if we fail to write to etcd (note that the save does retries, it's not a single try), self shutdown

io-engine/src/bdev/nexus/nexus_share.rs
tiagolobocastro: hmm yeah that makes it tricky, maybe for nvmf keep it in the `Share` impl and for nbd in the `share_ext`? that should cover both.

Though not covering the sharing of the bdev directly. I'd say keep the tracking in the same oep issue

yugchaudhari:
@tiagolobocastro just checking in if you got time to go through the PR again :)

io-engine/src/bdev/nexus/nexus_share.rs
```diff
@@ -214,7 +227,22 @@ impl<'n> Nexus<'n> {
             }
         }
 
-        self.as_mut().unshare(None).await
+        self.as_mut().unshare(None).await?;
+
+        if persist_clean {
```
tiagolobocastro, line 252: there's currently a race where a resume could start a stopped subsystem
So I think here you should also double check if the `NvmfSubsystem` is really gone

io-engine/src/bdev/nexus/nexus_bdev.rs
```diff
@@ -830,7 +833,9 @@ impl<'n> Nexus<'n> {
     pub async fn destroy_ext(mut self: Pin<&mut Self>, sigterm: bool) -> Result<(), Error> {
         info!("{:?}: destroying nexus...", self);
 
-        self.as_mut().unshare_nexus().await?;
+        // Destroy already issues `PersistOp::Shutdown` below (which sets
+        // `clean_shutdown = true`), so skip the redundant persist here.
+        self.as_mut().unshare_nexus(false).await?;
```
tiagolobocastro: Maybe add separate private method to avoid modifying all unshare and exposing it

tiagolobocastro (commented):
Sorry about the delay

tiagolobocastro:
> @tiagolobocastro just checking in if you got time to go through the PR again :)

ah good timing, was actually reviewing it now :)

io-engine/src/bdev/nexus/nexus_share.rs
yugchaudhari: Done. The persist for NVMe-oF lives inside `Share::share_nvmf` now (in the `Off | None` arm, before the actual share goes live), and the NBD branch keeps its own persist inside `share_ext`. Same effective behaviour for the `share_ext` callers, and a direct caller of `Share::share_nvmf` now gets the same guarantee for free.

The direct-bdev-share gap is tracked separately in openebs/openebs#4208.

io-engine/src/bdev/nexus/nexus_persistence.rs
yugchaudhari: Fair point on the retries. Reverted this, the save-failure path is now uniform again, `PersistOp::Shutdown` is the only op that skips self-shutdown (since it's already on the destroy path). Every other op, including the `SetCleanShutdown` ones, falls through to the existing self-shutdown behaviour.

io-engine/src/bdev/nexus/nexus_share.rs
yugchaudhari: Added a post-unshare `NvmfSubsystem::nqn_lookup(&self.name)`. If the subsystem is still around, we log a warn and skip the `SetCleanShutdown { clean: true }` persist, so `clean_shutdown` stays false and the existing crash-recovery path handles it on next start.

io-engine/src/bdev/nexus/nexus_bdev.rs
yugchaudhari: Done. Pulled the unshare body into a private `unshare_nexus_internal(persist_clean: bool)` helper, kept the public `unshare_nexus()` as a no-arg wrapper that calls it with `true`. `destroy_ext` is the only caller that hits the internal directly (with `false`). All the gRPC and test callers go back to plain `unshare_nexus()`.

io-engine/src/bdev/nexus/nexus_share.rs
```diff
@@ -3,9 +3,10 @@ use async_trait::async_trait;
 use snafu::ResultExt;
 use std::pin::Pin;
 
-use super::{nexus_err, Error, NbdDisk, Nexus, NexusTarget};
+use super::{nexus_err, Error, NbdDisk, Nexus, NexusTarget, PersistOp};
 
 use crate::core::{NvmfShareProps, Protocol, PtplProps, Share, UpdateProps};
+use crate::subsys::NvmfSubsystem;
```
tiagolobocastro: can you run our linter script, it should combine this with line above

tiagolobocastro (approved):

tiagolobocastro:
bors try

io-engine/src/bdev/nexus/nexus_share.rs
yugchaudhari: Done.

yugchaudhari:
> bors try

@tiagolobocastro 

- Linter fixed: the new `use crate::subsys::NvmfSubsystem` is now merged into the existing `use crate::{...}` block by `imports_granularity=Crate`.
- Test fix for the two `bors try` failures in `io-engine/tests/persistence.rs`. `persist_clean_shutdown` and `persist_unexpected_restart` both asserted `!clean_shutdown` right after `create_nexus`, but with the iter4 invariant (`clean_shutdown <=> currently unshared`) a freshly created, never-shared nexus is clean by construction. Flipped the assertions and updated the comment on `persist_unexpected_restart` to describe the new behaviour.
- Rebased on latest `develop`.

io-engine/src/bdev/nexus/nexus_share.rs
```diff
@@ -214,7 +247,29 @@ impl<'n> Nexus<'n> {
...
+            if NvmfSubsystem::nqn_lookup(&self.name).is_some() {
+                warn!(
+                    "{self:?}: NvmfSubsystem still present after unshare, \
+                    skipping clean-shutdown persist (subsystem resume race)"
+                );
+            } else {
+                let _ = self
+                    .as_mut()
+                    .persist(PersistOp::SetCleanShutdown { clean: true })
+                    .await;
```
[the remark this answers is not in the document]
yugchaudhari: Replaced the `let _ = ...` with `if let Err(e) = ... { warn!(...) }` so a failure is logged at the call site with the "best-effort, falls back to existing crash-recovery behaviour" context. `persist` still does its own error-level logging internally; this just makes the unshare-side intent obvious from a single read of the file.

io-engine/tests/persistence.rs
```diff
@@ -111,7 +113,7 @@ async fn persist_clean_shutdown() {
 
     // Check the persisted nexus info is correct.
 
-    assert!(!nexus_info.clean_shutdown);
+    assert!(nexus_info.clean_shutdown);
```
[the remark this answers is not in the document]
yugchaudhari: Added a `publish_nexus` step between create and destroy so the test now exercises the full `true → false → true` transition:

- Assert clean after create (unshared)
- Share → assert dirty after share
- Destroy → assert clean again

The destroy path's flip is meaningfully covered now.

yugchaudhari:
@tiagolobocastro @Abhinandan-Purkait should we merge this if this looks good? so that I can start on the control plane side of this iteration!

Abhinandan-Purkait (approved):
LGTM

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

PR #2009, opened 2026-06-17
refactor(gpt): add gpt ops interface

tiagolobocastro:
Add a generic ops interface, allowing us to reuse the gpt code even outside of the io-engine core.
Adds implementation with the existing BlockDeviceHandle. Note that we could make the BlockDeviceHandle the generic interface, but at the moment that is too coupled to spdk itself.


Abhinandan-Purkait (approved):

niladrih (approved):
Seems ok. Lack detailed context.

PR #2010, opened 2026-06-17
feat(gpt): add simply binary to write/read

tiagolobocastro:
Adds simply binary to read and write the nexus gpt. It makes use of the added gpt ops interface.
This allows is to reuse the gpt code whilst not running in spdk at all!


abhilashshetty04 (approved):

niladrih (approved):
LGTM

PR #2012, opened 2026-06-17
fix: nexus bdev size off by 1

tiagolobocastro:
The size of the nexus was off by 1.
Because we're increasing rather than decreasing there's no issue with the client side.

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev.rs
@@ -726,15 +726,16 @@ impl<'n> Nexus<'n> {
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


Abhinandan-Purkait (approved):

Abhinandan-Purkait:
As I understand this is on a nexus level, so a republish should ideally take care of this right for existing volumes?

tiagolobocastro:
> As I understand this is on a nexus level, so a republish should ideally take care of this right for existing volumes?

yes, the bdev will be larger by 1 sector

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

PR #2017, opened 2026-06-25

susobhandey:
Kubernetes CPU Manager running with cpuManagerPolicy=static may rewrite container cpusets when Guaranteed pods are admitted or removed.

When cpuset.cpus.effective is updated, the kernel resets task affinity masks for threads in the cgroup. This causes Mayastor reactor threads to lose their CPU pinning and Tokio worker threads to lose their unaffinitized CPU mask.

Add a cpuset monitor that watches for cpuset updates and automatically restores reactor affinity. Tokio worker TIDs are tracked so worker unaffinity can also be re-applied, preserving separation between reactor and runtime threads.

- cpuManagerPolicy=static
- isolcpus
- KubeVirt dedicated CPUs

admission/removal events.

fix(io-engine): update reactor repinning handling

refactor(affinity): run cpuset monitor on runtime blocking pool

Run the cpuset affinity monitor via runtime::spawn_blocking() instead of creating a dedicated std::thread.

This allows the monitor thread to participate in the existing thread unaffinity restoration mechanism and avoids maintaining a separate unmanaged thread.

fix(io-engine): re-pin reactors on CPUManager cpuset updates

Monitor kubelet cpu_manager_state and re-pin reactor threads when CPUManager updates the node cpuset allocation.

Handle cpu_manager_state inode replacement by re-establishing the inotify watch after DELETE_SELF and IGNORED events to ensure subsequent CPUManager updates are detected.

### DIFF
--- a/io-engine/src/core/affinity.rs
+++ b/io-engine/src/core/affinity.rs
@@ -1,0 +1,143 @@
+use std::{
+    path::Path,
+    time::{Duration, Instant},
+};
+
+use inotify::{EventMask, Inotify, WatchMask};
+
+use crate::core::{runtime, Mthread, Reactors};
+
+const CPUSET_FILE: &str = "/var/lib/kubelet/cpu_manager_state";
+
+/// Starts the CPUManager cpuset monitoring thread.
+///
+/// The monitor watches the kubelet `cpu_manager_state` file for
+/// cpuset updates and restores reactor and runtime thread affinity
+/// when Kubernetes CPUManager changes the effective cpuset.
+pub fn start_monitor() {
+    if std::env::var("KUBERNETES_SERVICE_HOST").is_err() {
+        return;
+    }
+    runtime::spawn_blocking(run);
+}
+
+fn run() {
+    let mut inotify = match Inotify::init() {
+        Ok(inotify) => inotify,
+        Err(error) => {
+            tracing::error!(
+                ?error,
+                "failed to initialize CPUManager affinity monitor; affinity restoration disabled"
+            );
+            return;
+        }
+    };
+
+    loop {
+        if !Path::new(CPUSET_FILE).exists() {
+            tracing::warn!("{CPUSET_FILE} does not exist, waiting...");
+            std::thread::sleep(std::time::Duration::from_secs(1));
+            continue;
+        }
+
+        if let Err(error) = inotify.watches().add(
+            CPUSET_FILE,
+            WatchMask::ATTRIB | WatchMask::MODIFY | WatchMask::CLOSE_WRITE | WatchMask::DELETE_SELF,
+        ) {
+            tracing::warn!(?error, "failed to watch {CPUSET_FILE}, retrying");
+            std::thread::sleep(std::time::Duration::from_secs(1));
+            continue;
+        }
+
+        tracing::trace!("watching {CPUSET_FILE}");
+
+        let mut buffer = [0u8; 4096];
+        let mut recreate_watch = false;
+
+        while !recreate_watch {
+            match inotify.read_events_blocking(&mut buffer) {
+                Ok(events) => {
+                    let mut changed = false;
+
+                    for event in events {
+                        tracing::trace!("event {:?}", event.mask);
+
+                        if event.mask.contains(EventMask::ATTRIB)
+                            || event.mask.contains(EventMask::MODIFY)
+                            || event.mask.contains(EventMask::CLOSE_WRITE)
+                        {
+                            changed = true;
+                        }
+
+                        if event.mask.contains(EventMask::DELETE_SELF)
+                            || event.mask.contains(EventMask::IGNORED)
+                        {
+                            recreate_watch = true;
+                        }
+                    }
+
+                    if changed {
+                        reconcile();
+                    }
+                }
+                Err(error) => {
+                    tracing::warn!(?error, "cpu_manager_state monitor failed, recreating watch");
+                    recreate_watch = true;
+                }
+            }
+        }
+
+        tracing::debug!(
+            "inotify watch lost (DELETE_SELF/IGNORED), recreating watch on new cpuset file"
+        );
+
+        std::thread::sleep(std::time::Duration::from_millis(100));
+    }
+}
+
+/// Upper bound on how long [`reconcile`] waits for kubelet to apply the cgroup
+/// cpuset after a `cpu_manager_state` event.
+///
+/// kubelet writes `cpu_manager_state` (which triggers this reconcile) as soon
+/// as a Guaranteed pod is admitted/removed, but it applies the container's
+/// cgroup cpuset asynchronously, on its CPUManager reconcile period (default
+/// 5s). Measured apply latencies on the target cluster were ~5s (grow) and
+/// ~7-9s (shrink), so we must keep watching well past that; otherwise we miss
+/// the apply and leave the baseline stale, which then mis-triggers the next
+/// event. This is only a ceiling -- reconcile breaks the moment the change is
+/// observed, so it normally returns in ~5-9s and only approaches this bound if
+/// the apply never arrives (in which case we fall back to the current
+/// baseline).
+const RECONCILE_SETTLE_WINDOW: Duration = Duration::from_secs(20);
+/// Interval between cpuset re-checks within [`RECONCILE_SETTLE_WINDOW`].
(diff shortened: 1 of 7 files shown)


io-engine/src/core/affinity.rs
```diff
@@ -0,0 +1,89 @@
+use std::path::Path;
+
+use inotify::{EventMask, Inotify, WatchMask};
+
+use crate::core::{runtime, Mthread, Reactors};
+
+const CPUSET_FILE: &str = "/var/lib/kubelet/cpu_manager_state";
```
tiagolobocastro, line 10: this should probably be pass via the cli args since `/var/lib/kubelet` is not universal

io-engine/src/core/affinity.rs
```diff
@@ -0,0 +1,89 @@
...
+
+pub fn start_monitor() {
```
tiagolobocastro, line 17: Add doc comments for all public functions
Also to consider if they should even be public

io-engine/src/core/affinity.rs
```diff
@@ -0,0 +1,89 @@
...
+    runtime::spawn_blocking(run);
```
tiagolobocastro, line 21: might want to consider bumping the max blocking threads since we're adding another

io-engine/src/core/affinity.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+fn run() {
+    Mthread::unaffinitize();
```
tiagolobocastro: This should not be needed as it's already added by the spawner?

io-engine/src/core/affinity.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    loop {
+        if !Path::new(CPUSET_FILE).exists() {
+            tracing::warn!("{} does not exist, waiting...", CPUSET_FILE);
```
tiagolobocastro: nit: whenever possible use this new format

```suggestion
            tracing::warn!("{CPUSET_FILE} does not exist, waiting...");
```

io-engine/src/core/affinity.rs
```diff
@@ -0,0 +1,89 @@
...
+            std::thread::sleep(std::time::Duration::from_secs(1));
+            continue;
+        }
+
+        let mut inotify = Inotify::init().expect("failed to initialise inotify");
```
tiagolobocastro: Should we handle the error rather than panic?

io-engine/src/core/affinity.rs
```diff
@@ -0,0 +1,89 @@
...
+        inotify
+            .watches()
+            .add(
+                CPUSET_FILE,
+                WatchMask::ATTRIB
+                    | WatchMask::MODIFY
+                    | WatchMask::CLOSE_WRITE
+                    | WatchMask::DELETE_SELF,
+            )
+            .unwrap();
```
tiagolobocastro: Same here

io-engine/src/core/affinity.rs
```diff
@@ -0,0 +1,89 @@
...
+
+        tracing::info!("watching {}", CPUSET_FILE);
```
tiagolobocastro: If we loose the notifications frequently we should not log this message as info, perhaps trace

io-engine/src/core/affinity.rs
tiagolobocastro: Also can we reuse the inotify by removing the old watch and adding a new one?

susobhandey: you mean to say just init inotify once and keep adding new watches?

io-engine/src/core/affinity.rs
susobhandey: Is it ok in the first go let the change gets in and maybe another pr for config/cli driven. otherwise changes are becoming big

tiagolobocastro: Sure, can do that

io-engine/src/core/affinity.rs
```diff
@@ -0,0 +1,98 @@
...
+                            || event.mask.contains(EventMask::MODIFY)
+                            || event.mask.contains(EventMask::CLOSE_WRITE)
+                        {
+                            changed = true;
+                        }
+
+                        if event.mask.contains(EventMask::DELETE_SELF)
+                            || event.mask.contains(EventMask::IGNORED)
+                        {
+                            recreate_watch = true;
```
tiagolobocastro, line 75: So what is the behaviour with the latest changes, do we fall here or into the error below on 77?
And how often does that happen, is frequency based on new guaranteed pods being created/deleted on fixed interval?

io-engine/src/core/affinity.rs
tiagolobocastro: Yes, if possible ofc

susobhandey: yes it is working and changes are already here. only one time init

io-engine/src/core/affinity.rs
susobhandey: I never see we failed at inotify.read_events_blocking() as we make sure always look for fresh file

io-engine/src/core/reactor.rs
```diff
@@ -934,6 +934,30 @@ impl Reactor {
         }
     }
 
+    pub fn reapply_affinity(&self) {
+        let tid = self.tid();
+
+        if tid == 0 {
+            return;
```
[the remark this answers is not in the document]
tiagolobocastro: This can also be fixed by scheduling reapply_affinity in the correct reactor core i suppose

io-engine/src/core/affinity.rs
```diff
@@ -0,0 +1,98 @@
...
+            return;
+        }
+    };
+
+    loop {
+        if !Path::new(CPUSET_FILE).exists() {
+            tracing::warn!("{CPUSET_FILE} does not exist, waiting...");
+            std::thread::sleep(std::time::Duration::from_secs(1));
+            continue;
+        }
```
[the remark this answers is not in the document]
tiagolobocastro: Good point, if not running in K8s this should not log, a simple check for K8s env would suffice, ie:
```rust
pub fn start_monitor() {
    if std::env::var("KUBERNETES_SERVICE_HOST").is_err() {
        return;
    }
...
}
```

io-engine/src/core/affinity.rs
```diff
@@ -0,0 +1,98 @@
...
+
+        tracing::debug!("watch lost, recreating on new inode");
+
+        std::thread::sleep(std::time::Duration::from_millis(100));
+    }
+}
+
+fn reconcile() {
+    tracing::warn!("Detected cpuset update, restoring affinity");
+
```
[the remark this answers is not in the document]
tiagolobocastro: Should use debug, don't want to get logs everytime new pod is cheduled

io-engine/src/core/affinity.rs
```diff
@@ -0,0 +1,98 @@
...
+                    }
+                }
+                Err(error) => {
+                    tracing::warn!(?error, "cpu_manager_state monitor failed, recreating watch");
+                    recreate_watch = true;
+                }
+            }
+        }
+
+        tracing::debug!("watch lost, recreating on new inode");
```
tiagolobocastro: This is a big generic, what watch, maybe add more context

tiagolobocastro (commented):
Please check the copilot reviews, otherwise lgtm

io-engine/src/core/reactor.rs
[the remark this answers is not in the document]
susobhandey: In that case runtime::reapply_workers_unaffinity() may execute before the reactors are repinned ??

io-engine/src/core/runtime.rs
```diff
@@ -54,8 +64,14 @@ static RUNTIME: Lazy<Runtime> = Lazy::new(|| {
...
-        .max_blocking_threads(6)
-        .on_thread_start(Mthread::unaffinitize)
+        .max_blocking_threads(7)
+        .on_thread_start(|| {
+            let tid = unsafe { libc::syscall(libc::SYS_gettid) as libc::pid_t };
+
+            UNAFFINITIZED_WORKERS.lock().insert(tid);
+
+            Mthread::unaffinitize();
+        })
```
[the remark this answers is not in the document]
susobhandey: not required

tiagolobocastro:
bors try

Abhinandan-Purkait (approved):

PR #2018, opened 2026-06-25
Implementation of the V2 nexus layout

tiagolobocastro:
feat(gpt): add v2 layout implementation
    
    We reduce the size of the metadata partition to 3MiB.
    By doing this we are aligning the data partition to a 4MiB
    cluster boundary.
    Also we adjust the size of the data partition to multiples of 4MiB
    
    Also add tests for both V1 and V2
    
---

    feat(gpt): stitch up the label version from the nexus gRPC
    
    Process the label version from the create nexus handler, and use it
    to determine the nexus data partition placement.
    Note that the metadata is not currently in-use.
    
---

    feat(gpt): add version and block-size to the gpt binary
    
    Allows us to create GPT with either version as well as
    specifying different block sizes for testing.
    
---

    refactor(gpt): add maya label and version support
    
    This is scaffolding ready for the upcoming changes, with a V2
    version which uses different offset and end alignment.


niladrih (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #2029, opened 2026-07-03
feat(pool): expose disk SMART/health via GetPoolHealth

susobhandey:
Implement a generic device_health() on the BlockDevice trait and wire up the GetPoolHealth gRPC handler so a pool's backing disk health can be queried at runtime.

Health is read by attachment, not device type:
- kernel-backed disks (aio://, uring:// — SATA/SAS/NVMe without VFIO, which expose a /dev node) are read via `smartctl --json --all` and parsed for NVMe, ATA and SCSI SMART shapes;
- VFIO-bound NVMe (pcie://, no /dev node) is read through the existing bdev NVMe admin passthru via BdevHandle::nvme_get_smart(), issuing Get Log Page 0x02, then decoded from the 512-byte log.

Adds core/device_health.rs (DeviceHealth type, smartctl reader/parser, NVMe log decoder, unit tests), the trait method and SpdkBlockDevice implementation, BdevHandle::nvme_get_smart(), and the GetPoolHealth handler with proto<->type conversions. The device is opened read-only so it coexists with the pool's exclusive-write claim.

### DIFF
--- a/io-engine/src/bdev/device.rs
+++ b/io-engine/src/bdev/device.rs
@@ -24,9 +24,9 @@ use spdk_rs::{
 use crate::core::{
     mempool::MemoryPool, Bdev, BdevHandle, BlockDevice, BlockDeviceDescriptor, BlockDeviceHandle,
     BlockDeviceIoStats, CoreError, DeviceEventDispatcher, DeviceEventSink, DeviceEventType,
-    DeviceIoController, IoCompletionCallback, IoCompletionCallbackArg, IoCompletionStatus,
-    NvmeStatus, ReadOptions, SnapshotParams, ToErrno, UntypedBdev, UntypedBdevHandle,
-    UntypedDescriptorGuard,
+    DeviceHealth, DeviceIoController, IoCompletionCallback, IoCompletionCallbackArg,
+    IoCompletionStatus, NvmeStatus, ReadOptions, SnapshotParams, ToErrno, UntypedBdev,
+    UntypedBdevHandle, UntypedDescriptorGuard,
 };
 
 #[cfg(feature = "fault-injection")]
@@ -131,6 +131,106 @@ impl BlockDevice for SpdkBlockDevice {
         let disp = map.entry(self.device_name()).or_default();
         disp.add_listener(listener);
         Ok(())
+    }
+
+    async fn device_health(&self) -> Result<DeviceHealth, CoreError> {
+        let driver = self.driver_name();
+        let name = self.device_name();
+
+        // Kernel-backed disks (aio/uring) — SAS, SATA and NVMe used without
+        // VFIO — have a /dev node, read generically with smartctl. This
+        // excludes file-backed aio/uring bdevs (e.g.
+        // `aio:///var/local/openebs/blob?size=100GiB`), whose name is a
+        // regular file path, not a `/dev` node -- there's no physical disk to
+        // query, so they fall through to the generic NotSupported below.
+        if (driver == "aio" || driver == "uring") && name.starts_with("/dev/") {
+            return crate::core::device_health::read_device_health(&name).await;
+        }
+
+        // VFIO-bound NVMe (pcie:// -> SPDK nvme bdev) has no /dev node; read the
+        // SMART/Health log page via the existing bdev NVMe admin passthru path.
+        if driver != "nvme" {
+            return Err(CoreError::NotSupported {
+                source: Errno::ENXIO,
+            });
+        }
+
+        let hdl = UntypedBdevHandle::open_with_bdev(&self.0, false)?;
+        let mut buf = DmaBuf::new(512, self.alignment())
+            .map_err(|_| CoreError::DmaAllocationFailed { size: 512 })?;
+        hdl.nvme_get_smart(&mut buf).await?;
+        // A short/garbled log page is a decode failure, not a missing
+        // device (we already know the device exists -- we just read from
+        // it) -- report EIO rather than ENXIO.
+        let mut health = DeviceHealth::from_nvme_smart(buf.as_slice())
+            .ok_or(CoreError::NotSupported { source: Errno::EIO })?;
+
+        // The SMART log only reports how many error-log entries exist;
+        // fetch the entries themselves (Log Page 01h) when there are any.
+        if let Some(count) = health.num_error_log_entries {
+            if count > 0 {
+                const MAX_ENTRIES: u32 = 16;
+                let max_entries = count.min(MAX_ENTRIES as u128) as u32;
+                let entry_size = crate::core::device_health::NVME_ERROR_LOG_ENTRY_SIZE as u64;
+                let mut err_buf = DmaBuf::new((max_entries as u64) * entry_size, self.alignment())
+                    .map_err(|_| CoreError::DmaAllocationFailed {
+                        size: (max_entries as u64) * entry_size,
+                    })?;
+                match hdl.nvme_get_error_log(&mut err_buf, max_entries).await {
+                    Ok(()) => {
+                        health.error_log_entries =
+                            crate::core::device_health::parse_nvme_error_log(err_buf.as_slice());
+                    }
+                    Err(error) => {
+                        warn!("failed to read NVMe error log for '{name}': {error}");
+                    }
+                }
+            }
+        }
+
+        // Identity is static for the device's lifetime, unlike the health
+        // counters above -- fetch the Identify Controller admin command
+        // only once per device and reuse it on every subsequent poll.
+        health.identity = match crate::core::device_health::cached_nvme_identity(&name) {
+            Some(mut identity) => {
+                // Capacity/sector size can change if the namespace is
+                // resized after identity was cached; refresh them from the
+                // bdev (cheap, no admin command needed) on every poll
+                // rather than serving a stale snapshot.
+                identity.capacity_bytes = Some(self.size_in_bytes());
+                identity.logical_sector_size = Some(self.block_len() as u32);
+                Some(identity)
+            }
+            None => {
+                let mut ident_buf = DmaBuf::new(4096, self.alignment())
+                    .map_err(|_| CoreError::DmaAllocationFailed { size: 4096 })?;
+                let identity = match hdl.nvme_identify_ctrlr(&mut ident_buf).await {
+                    Ok(()) => crate::core::device_health::identity_from_nvme_identify(
+                        ident_buf.as_slice(),
+                    ),
+                    Err(error) => {
+                        // Health itself already succeeded; don't fail the
+                        // whole call just because identity didn't, but
+                        // don't swallow it silently either.
+                        warn!("failed to identify NVMe controller for '{name}': {error}");
+                        None
+                    }
+                }
+                .map(|mut identity| {
+                    // Capacity/sector size are already known generically
+                    // for any bdev -- no need for a namespace Identify.
+                    identity.capacity_bytes = Some(self.size_in_bytes());
+                    identity.logical_sector_size = Some(self.block_len() as u32);
+                    identity
+                });
+                if let Some(identity) = identity.clone() {
+                    crate::core::device_health::cache_nvme_identity(&name, identity);
+                }
+                identity
+            }
+        };
+
+        Ok(health)
     }
 }
 
@@ -694,6 +794,10 @@ pub fn bdev_event_callback<T: BdevOps>(event: spdk_rs::BdevEvent, bdev: spdk_rs:
     let event = match event {
         spdk_rs::BdevEvent::Remove => {
             info!("Received SPDK remove event for bdev '{}'", dev.name());
+            // Drop any cached VFIO NVMe identity for this device name -- it
+            // must not be served to a different physical device that later
+            // reuses the same bdev name (e.g. the pool recreated).
+            crate::core::device_health::evict_nvme_identity(dev.name());
             DeviceEventType::DeviceRemoved
         }
         spdk_rs::BdevEvent::Resize => {
--- a/io-engine/src/bin/io-engine-client/v1/pool_cli.rs
+++ b/io-engine/src/bin/io-engine-client/v1/pool_cli.rs
@@ -40,6 +40,8 @@ enum PoolCommands {
     ClearErrors(ClearErrorsArgs),
     /// Probes storage pool
     Probe(ProbeArgs),
+    /// Get SMART / health of the pool's disk(s)
+    GetHealth(GetHealthArgs),
 }
 
 #[derive(Debug, Args)]
@@ -92,6 +94,14 @@ struct DestroyArgs {
     uuid: Option<Uuid>,
     #[arg(short = 't', long = "type", default_value = "lvs")]
     pool_type: PoolType,
+}
+
+#[derive(Debug, Args)]
+struct GetHealthArgs {
+    /// Storage pool name
+    pool: String,
+    #[arg(short = 'u', long)]
+    uuid: Option<Uuid>,
 }
 
 #[derive(Debug, Args)]
@@ -169,6 +179,7 @@ pub async fn handler(ctx: Context, args: PoolArgs) -> crate::Result<()> {
         PoolCommands::List(args) => list(ctx, args).await,
         PoolCommands::ClearErrors(args) => clear_errors(ctx, args).await,
         PoolCommands::Probe(args) => probe(ctx, args).await,
+        PoolCommands::GetHealth(args) => get_health(ctx, args).await,
     }
 }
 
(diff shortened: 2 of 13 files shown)


io-engine/src/bin/io-engine-client/v1/pool_cli.rs
```diff
@@ -40,6 +40,9 @@ enum PoolCommands {
     ClearErrors(ClearErrorsArgs),
     /// Probes storage pool
     Probe(ProbeArgs),
+    /// Get SMART / health of the pool's disk(s)
+    #[command(name = "get-health")]
```
tiagolobocastro: Others don't have the command, can we omit it if not required?

io-engine/src/bdev/device.rs
```diff
@@ -132,6 +132,67 @@ impl BlockDevice for SpdkBlockDevice {
...
         Ok(())
     }
+
+    async fn device_health(&self) -> Result<DeviceHealth, CoreError> {
+        let driver = self.driver_name();
+        let name = self.device_name();
+
+        // Kernel-backed disks (aio/uring) — SAS, SATA and NVMe used without
+        // VFIO — have a /dev node, read generically with smartctl.
+        if (driver == "aio" || driver == "uring") && name.starts_with("/dev/") {
```
tiagolobocastro, line 146: Are we not supporting file based? ie `aio:///var/local/openebs/blob?size=100GiB` ?

io-engine/src/bdev/device.rs
```diff
@@ -132,6 +132,67 @@ impl BlockDevice for SpdkBlockDevice {
...
+
+        // VFIO-bound NVMe (pcie:// -> SPDK nvme bdev) has no /dev node; read the
+        // SMART/Health log page via the existing bdev NVMe admin passthru path.
+        if driver == "nvme" {
+            let hdl = UntypedBdevHandle::open_with_bdev(&self.0, false)?;
+            let mut buf = DmaBuf::new(512, self.alignment())
+                .map_err(|_| CoreError::DmaAllocationFailed { size: 512 })?;
+            hdl.nvme_get_smart(&mut buf).await?;
+            let mut health =
+                DeviceHealth::from_nvme_smart(buf.as_slice()).ok_or(CoreError::NotSupported {
```
tiagolobocastro: Is this error due to parsing the received data? If so is this not supported with enxio right?

io-engine/src/bdev/device.rs
```diff
@@ -132,6 +132,67 @@ impl BlockDevice for SpdkBlockDevice {
...
+            // only once per device and reuse it on every subsequent poll.
+            health.identity = match crate::core::device_health::cached_nvme_identity(&name) {
+                Some(identity) => Some(identity),
+                None => {
+                    let mut ident_buf = DmaBuf::new(4096, self.alignment())
+                        .map_err(|_| CoreError::DmaAllocationFailed { size: 4096 })?;
+                    let identity = hdl
+                        .nvme_identify_ctrlr(&mut ident_buf)
+                        .await
+                        .ok()
```
tiagolobocastro: Should this be returned as an error?

io-engine/src/bdev/device.rs
```diff
@@ -132,6 +132,67 @@ impl BlockDevice for SpdkBlockDevice {
...
+            hdl.nvme_get_smart(&mut buf).await?;
+            let mut health =
+                DeviceHealth::from_nvme_smart(buf.as_slice()).ok_or(CoreError::NotSupported {
+                    source: Errno::ENXIO,
+                })?;
+
+            // Identity is static for the device's lifetime, unlike the health
+            // counters above -- fetch the Identify Controller admin command
+            // only once per device and reuse it on every subsequent poll.
+            health.identity = match crate::core::device_health::cached_nvme_identity(&name) {
```
tiagolobocastro: Doesn't the capacity need updating here in case the device is extended?

io-engine/src/core/handle.rs
```diff
@@ -243,6 +243,19 @@ impl<T: BdevOps> BdevHandle<T> {
         self.nvme_admin(&cmd, Some(buffer)).await
     }
 
+    /// read the SMART / Health Information log page (Log Page Identifier 02h)
```
tiagolobocastro, line 246: Btw don't we need the error and event log as well?

io-engine/src/grpc/v1/pool.rs
```diff
@@ -448,15 +516,8 @@ impl PoolGrpc {
...
-        match clear {
-            ClearErrors::ClearAll => {
-                self.pool.reset_errors().await?;
-                self.pool.reset_stall_transitions().await?;
-            }
-            ClearErrors::ClearIoErrors => self.pool.reset_errors().await?,
-            ClearErrors::ClearIoStallTransitions => self.pool.reset_stall_transitions().await?,
-        }
+    async fn clear_errors(&self) -> Result<(), tonic::Status> {
+        self.pool.reset_errors().await?;
```
tiagolobocastro: rebase conflict?

io-engine/src/grpc/v1/pool.rs
```diff
@@ -820,6 +881,47 @@ impl PoolRpc for PoolService {
...
+    ) -> GrpcResult<GetPoolHealthResponse> {
+        self.locked(
+            GrpcClientContext::new(&request, function_name!()),
+            async move {
+                crate::spdk_submit!(async move {
+                    info!("{:?}", request.get_ref());
+
+                    let pool = GrpcPoolFactory::finder(request.into_inner()).await?;
+
+                    let mut disks = Vec::new();
```
tiagolobocastro: Making sure disks is a reference

```suggestion
                    let mut disks = Vec::with_capacity(pool.disks().len());
```

io-engine/src/grpc/v1/pool.rs
```diff
@@ -820,6 +881,47 @@ impl PoolRpc for PoolService {
...
+                    for uri in pool.disks() {
+                        // Resolve the disk uri to its registered bdev and ask
+                        // the device for its health. Dispatch — smartctl for
+                        // aio/uring (incl. kernel NVMe), SMART log page for
+                        // VFIO NVMe — is handled by the BlockDevice impl.
+                        let health = match crate::bdev::dev::device_name(&uri) {
+                            Ok(name) => match crate::bdev::device_lookup(&name) {
+                                Some(dev) => dev.device_health().await.ok(),
```
tiagolobocastro: we're flattening all errors here?
I see you're using supported in the health, maybe could we also keep an error message field?

io-engine/src/grpc/v1/pool.rs
```diff
@@ -820,6 +881,47 @@ impl PoolRpc for PoolService {
         .await
     }
 
+    #[named]
+    async fn get_pool_health(
```
tiagolobocastro, line 922: Have you thought having supporting a get_pools_health or something like that? ie all health for all pools

io-engine/src/core/device_health.rs
```diff
@@ -0,0 +1,593 @@
...
+    pub transport: Option<String>,
+    pub link_speed: Option<String>,
+}
+
+/// A single SMART attribute table entry (ATA devices only -- SAS/NVMe report
+/// an empty list here since their equivalent thresholds are already
+/// first-class fields on [`DeviceHealth`]).
+#[derive(Debug, Clone)]
+pub struct SmartAttribute {
+    pub id: u8,
```
tiagolobocastro, line 84: add doc comments on all pubs

io-engine/src/core/device_health.rs
```diff
@@ -0,0 +1,593 @@
...
+        let u128le = |o: usize| {
+            let mut b = [0u8; 16];
+            b.copy_from_slice(&page[o..o + 16]);
+            u128::from_le_bytes(b)
+        };
+        let temp_k = u16le(1);
+        let temperature_celsius = if temp_k == 0 {
+            None
+        } else {
+            Some(temp_k as i16 - 273)
```
tiagolobocastro: might want to do safe subtract here to avoid panic in case of bogus data

io-engine/src/core/device_health.rs
```diff
@@ -0,0 +1,593 @@
...
+}
+
+/// Read SMART / health for a kernel block device by its `/dev` path, via
+/// `smartctl --json --all`.
+///
+/// smartctl uses a bitmask exit status where non-zero frequently just signals
+/// SMART warnings while still emitting complete JSON, so the exit code is
+/// ignored and the JSON on stdout is parsed regardless.
+pub fn read_device_health(path: &str) -> Result<DeviceHealth, CoreError> {
+    let output = Command::new("smartctl")
```
tiagolobocastro, line 388: We should run this in trampolin with async command
You can check the lvm command run IIRC

io-engine/src/core/device_health.rs
```diff
@@ -0,0 +1,593 @@
...
+    pub threshold: u8,
+    pub raw_value: u64,
+}
+
+/// Cache of [`DeviceIdentity`] for VFIO-bound NVMe devices, keyed by bdev
+/// name. Identity requires a separate NVMe Identify Controller admin command
+/// from the health log page, and — unlike health — never changes, so it's
+/// fetched once per device and reused rather than re-issued on every
+/// `GetPoolHealth` call.
+static NVME_IDENTITY_CACHE: Lazy<Mutex<HashMap<String, DeviceIdentity>>> =
```
tiagolobocastro, line 125: We should also remove entry from here when pool is deleted

io-engine/src/bin/io-engine-client/v1/pool_cli.rs
susobhandey: Fixed

io-engine/src/bdev/device.rs
susobhandey: Is it required?

io-engine/src/bdev/device.rs
susobhandey: Changed to EIO

io-engine/src/bdev/device.rs
susobhandey: Kept it non-fatal — health itself already succeeded by this point, so I didn't want an Identify failure to fail the whole GetPoolHealth call for that disk

io-engine/src/bdev/device.rs
susobhandey: Fixed

io-engine/src/core/handle.rs
susobhandey: Added

io-engine/src/grpc/v1/pool.rs
susobhandey: yes :)

io-engine/src/core/device_health.rs
susobhandey: done

io-engine/src/grpc/v1/pool.rs
susobhandey: Good idea, but I'd rather keep it as a follow-up PR than fold it into this one

io-engine/src/core/device_health.rs
susobhandey: Fixed

io-engine/src/grpc/v1/pool.rs
susobhandey: fixed

io-engine/src/grpc/v1/pool.rs
abhilashshetty04: Could be useful to have listing capability. Control plane could also expose getting all pools health in one api.

https://github.com/openebs/mayastor-dependencies/pull/171/changes#r3850068137

io-engine/src/bdev/device.rs
```diff
@@ -132,6 +132,107 @@ impl BlockDevice for SpdkBlockDevice {
...
+            let mut health = DeviceHealth::from_nvme_smart(buf.as_slice())
+                .ok_or(CoreError::NotSupported { source: Errno::EIO })?;
+
+            // The SMART log only reports how many error-log entries exist;
+            // fetch the entries themselves (Log Page 01h) when there are any.
+            if let Some(count) = health.num_error_log_entries {
+                if count > 0 {
+                    const MAX_ENTRIES: u32 = 16;
+                    let max_entries = count.min(MAX_ENTRIES as u128) as u32;
+                    let mut err_buf = DmaBuf::new((max_entries as u64) * 64, self.alignment())
```
tiagolobocastro: what is 64? Maybe place that with an appropriate name in a const

io-engine/src/bdev/device.rs
```diff
@@ -132,6 +132,107 @@ impl BlockDevice for SpdkBlockDevice {
...
+        // `aio:///var/local/openebs/blob?size=100GiB`), whose name is a
+        // regular file path, not a `/dev` node -- there's no physical disk to
+        // query, so they fall through to the generic NotSupported below.
+        if (driver == "aio" || driver == "uring") && name.starts_with("/dev/") {
+            return crate::core::device_health::read_device_health(&name).await;
+        }
+
+        // VFIO-bound NVMe (pcie:// -> SPDK nvme bdev) has no /dev node; read the
+        // SMART/Health log page via the existing bdev NVMe admin passthru path.
+        if driver == "nvme" {
```
tiagolobocastro: Maybe reverse for early return to reduce identation

```suggestion
        if driver != "nvme" {
            return;
```

io-engine/src/grpc/v1/pool.rs
```diff
@@ -820,6 +894,69 @@ impl PoolRpc for PoolService {
...
+            GrpcClientContext::new(&request, function_name!()),
+            async move {
+                crate::spdk_submit!(async move {
+                    info!("{:?}", request.get_ref());
+
+                    let pool = GrpcPoolFactory::finder(request.into_inner()).await?;
+
+                    // disks() allocates a fresh Vec<String> per call (it's
+                    // owned, not a reference) -- call it once rather than
+                    // once for the length and again for the loop.
```
tiagolobocastro: this seems clear from code itself

```suggestion
```

io-engine/src/grpc/v1/pool.rs
```diff
@@ -820,6 +894,69 @@ impl PoolRpc for PoolService {
...
+                    let uris = pool.disks();
+                    let mut disks = Vec::with_capacity(uris.len());
+                    for uri in uris {
+                        // LVM pools report their PVs as plain kernel device
```
tiagolobocastro: IIRC we use 100 or 100 something for commends, so can extend each line

io-engine/src/grpc/v1/pool.rs
```diff
@@ -820,6 +894,69 @@ impl PoolRpc for PoolService {
...
+                        // paths (e.g. `/dev/nvme1n1`, from LVM's `pv_name`),
+                        // not bdev URIs -- they're never registered as SPDK
+                        // bdevs, so parsing them as a URI always fails and
```
tiagolobocastro: then should we implement `read_device_health` on the pool interface?

io-engine/src/core/device_health.rs
```diff
@@ -0,0 +1,870 @@
...
+        serial_number: ascii_field(4, 20),
+        firmware_revision: ascii_field(64, 8),
+        wwn: None,
+        capacity_bytes: None,
+        logical_sector_size: None,
+        physical_sector_size: None,
+        rotation_rate: Some(0), // NVMe is always non-rotating media.
+        form_factor: None,
+        transport: Some("PCIe".to_string()),
+        link_speed: None,
```
tiagolobocastro, line 340: Is this not available? This would be useful if this is the PCIe slot speed

io-engine/src/core/device_health.rs
```diff
@@ -0,0 +1,870 @@
...
+
+/// smartctl uses a bitmask exit status where non-zero frequently just signals
+/// SMART warnings while still emitting complete JSON, so the exit code is
+/// ignored and the JSON on stdout is parsed regardless.
+async fn run_smartctl(path: &str) -> Result<DeviceHealth, CoreError> {
+    let output = Command::new("smartctl")
+        .args(["--json", "--all", path])
+        .output()
+        .await
+        .map_err(|_| CoreError::NotSupported {
```
tiagolobocastro: We should capture the error string into the error for visibility

io-engine/src/core/device_health.rs
```diff
@@ -0,0 +1,870 @@
...
+            // smartctl binary not found / not executable.
+            source: Errno::ENOENT,
+        })?;
+
+    let json: Value = serde_json::from_slice(&output.stdout)
+        .map_err(|_| CoreError::NotSupported { source: Errno::EIO })?;
```
tiagolobocastro: We should capture the error string into the error for visibility

tiagolobocastro (commented):
There's no integration tests (ie `io-engine/tests`)
Can we use loop device to at least check it runs without panic?

io-engine/src/bdev/device.rs
susobhandey: Fixed

io-engine/src/bdev/device.rs
susobhandey: Fixed

io-engine/src/core/device_health.rs
susobhandey: Fixed

io-engine/src/grpc/v1/pool.rs
susobhandey: removed

tiagolobocastro (approved):

PR #2030, opened 2026-07-06
refactor: don't reuse clean shutdown for nexus share/unshare

tiagolobocastro:
In previous change we had modified the clean shutdown, keeping it as true in case the nexus was not shared.
This is a breaking change which affects existing control-plane and e2e tests.

Instead here we add a new shared flag which we can use without modifying the existing flags.

The control-plane will then also have to be changed to take this flag into account when determining the replica health.

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev.rs
@@ -802,9 +802,8 @@ impl<'n> Nexus<'n> {
         };
 
         // Persist the fact that the nexus is now successfully open.
-        // `PersistOp::Create` records the initial clean-while-unshared state,
-        // which then gets flipped on share/unshare to maintain the invariant
-        // `clean_shutdown <=> currently unshared`.
+        // `PersistOp::Create` records the initial unshared state, which then
+        // gets transitioned via `PersistOp::SetShared` on share/unshare.
         // We have to do this before setting the nexus to open so that
         // nexus list does not return this nexus until it is persisted.
         if let Err(e) = nex.persist(PersistOp::Create).await {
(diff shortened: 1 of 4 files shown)


tiagolobocastro:
@yugchaudhari please take a look at this change, keeping existing flag as is and tracking shared with new flag

yugchaudhari (approved):
Yes, this looks cleaner. A dedicated `shared` marker is the right approach. The test additions look solid too, the `test.kill()` mid-shared case is exactly the crash scenario iteration 4 wanted to solve.

yugchaudhari:
@tiagolobocastro 

Would be happy to pick up the CP side once this lands. As far as I can tell, the only place that needs updating is `scheduling.rs::healthy_children`, which today is:

```rust
if !info_inner.clean_shutdown {
    return Ok(HealthyChildItems::One(info, ...));
}
```

With the new `shared` flag, that becomes something like `!info_inner.clean_shutdown && info_inner.shared`, so an unshared nexus that crashed still returns `HealthyChildItems::All` and we avoid the wasteful full rebuild. 

Anything else needed on CP-side, or is that all?

Abhinandan-Purkait (approved):

tiagolobocastro:
> @tiagolobocastro
> 
> Would be happy to pick up the CP side once this lands. As far as I can tell, the only place that needs updating is `scheduling.rs::healthy_children`, which today is:
> 
> ```rust
> if !info_inner.clean_shutdown {
>     return Ok(HealthyChildItems::One(info, ...));
> }
> ```
> 
> With the new `shared` flag, that becomes something like `!info_inner.clean_shutdown && info_inner.shared`, so an unshared nexus that crashed still returns `HealthyChildItems::All` and we avoid the wasteful full rebuild.
> 
> Anything else needed on CP-side, or is that all?

Yeah that would be great thank you!
I think that's all but remember to handle that `shared` field may not be present at all if the entry was entry by an old version and even at runtime when you may have a mix of engines during the upgrade.

niladrih (approved):

yugchaudhari:
> Yeah that would be great thank you! I think that's all but remember to handle that `shared` field may not be present at all if the entry was entry by an old version and even at runtime when you may have a mix of engines during the upgrade.

Got it, thanks. Will make the CP-side `shared` field `Option<bool>` and fall back to conservative default (treat as shared) when the value is missing, so old entries and mid-upgrade mixed-engine writes still return `HealthyChildItems::One`. Something like `shared.unwrap_or(true)` in the check.

abhilashshetty04 (approved):

abhilashshetty04:
bors try

tiagolobocastro:
bors merge

PR #2032, opened 2026-07-07
fix: handle nexus info transaction

tiagolobocastro:
On transaction updates, we actually load the existing value and compare it to the in-memory.

This is not fully covered in the data-plane tests today, and was only caught on control-plane tests.

Since we load it, we now have to ensure we can load older values where shared did not exist.

todo: add transaction tests here


Abhinandan-Purkait (approved):

niladrih (approved):

tiagolobocastro:
bors merge

PR #2031, opened 2026-07-07
feat: add pbjson workspace dependencies for updated events-api

rohan2794:
Add pbjson workspace dependencies for updated events-api.

PR #2033, opened 2026-07-08
fix(nexus/grpc): add read_only field to v1 struct literals

yugchaudhari:
Bumps utils/dependencies to include openebs/mayastor-dependencies#162 which adds an `optional bool read_only` field to both the v1 `Nexus` proto message and `PublishNexusRequest`. Two sites in the io-engine construct these as struct literals without `..Default::default()`, so bumping the submodule alone breaks the build with `E0063: missing field read_only`.

Add the field explicitly as `None` at both sites:

  - `io-engine/src/grpc/v1/nexus.rs:403`: the `Nexus { ... }` response literal in `into_grpc()`
  - `io-engine/src/bin/io-engine-client/v1/nexus_cli.rs:480`: the `PublishNexusRequest { ... }` in the CLI publish path

The io-engine doesn't know how to track ROX state yet, so `None` is the honest default until the consumer PR lands and populates them from actual state.

Also adds `pbjson`/`pbjson-build` to the workspace dependency table because the newer `events-api` inside the bumped submodule needs them. This includes the same submodule bump and `pbjson` workspace deps as openebs/mayastor#2031, plus the missing struct literal fixes that were blocking #2031's CI.

Refs: openebs/mayastor-dependencies#162
Refs: openebs/mayastor#2031

cc @abhilashshetty04 @rohan2794 @tiagolobocastro

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1341,6 +1341,8 @@ dependencies = [
  "bytes",
  "chrono",
  "futures",
+ "pbjson",
+ "pbjson-build",
  "prost",
  "prost-extend",
  "serde",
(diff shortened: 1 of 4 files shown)


abhilashshetty04 (approved):
Thanks :)

tiagolobocastro (approved):

tiagolobocastro:
bors merge

abhilashshetty04:
Changes on Nexus breaks `mayastor-control-plane` build also. Can you please take a look @yugchaudhari

PR #2035, opened 2026-07-10
feat(nexus): support publish-time read_only mode for ROX

yugchaudhari:
Iteration 1 of the ROX-fs OEP ([openebs/openebs#4242](https://github.com/openebs/openebs/issues/4242)) on the io-engine side, consuming the `optional bool read_only` field added to `PublishNexusRequest` in openebs/mayastor-dependencies#162.

A nexus published with `read_only=true` rejects write-family I/O at submit time so multiple initiators can safely share the target for read-only workloads. Re-sharing with a different value flips the mode (unshare + share carry the state transition cleanly).

## Where the state lives

`read_only: AtomicCell<bool>` on `Nexus<'n>`, alongside the existing `shutdown_requested` field. Interior mutability lets the share/unshare paths flip the flag without requiring a pinned mutable borrow, and reads from the I/O hot path are a single relaxed load.

Two public accessors, `Nexus::is_read_only` and `Nexus::set_read_only`, expose the flag for external tests. Internal callers use the field directly.

## Plumbing

- `share_ext` gets a new `read_only: bool` argument. On the Nvmf branch the flag is recorded before the target goes live (so any I/O from an initiator connecting during the share window is subject to the gate), and rolled back if `share_nvmf` itself fails.
- `unshare_nexus_internal` clears the flag: without an active target no front-end I/O can flow, so the default RWO shape is the honest state.
- `nexus_io::submit_request` returns NVMe status `NAMESPACE_IS_PROTECTED` for `Write` / `WriteZeros` / `Unmap` / `Reset` when the flag is set. Reads pass through unchanged.
- The v1 `publish_nexus` gRPC handler unwraps `args.read_only.unwrap_or(false)` and threads it into `share_ext`; the v0 handler hardcodes `false` since the v0 proto has no `read_only` field.
- `Nexus::into_grpc` reports the effective flag only while the nexus is currently shared, and `None` while unshared.

## What's deliberately not touched

NVMe reservation acquisition. `reservation_acquire` in `nexus_bdev_children.rs` is:

1. Off by default (gated on the `NEXUS_NVMF_RESV_ENABLE` env var, which is not set in production), and
2. When enabled, defaults to `NvmeReservation::WriteExclusiveAllRegs`, which already supports multiple registered initiators writing per the NVMe spec.

Neither branch conflicts with the ROX submit-time gate, so no reservation refactor is needed for this iteration. If a future scenario needs to gate reservation acquisition on `read_only` (e.g. strict `WriteExclusive` type + ROX combo), that can layer in as iter2 without touching this PR's shape.

## Tests

`io-engine/tests/nexus_read_only.rs` creates a nexus in-process with two aio children, then:

1. Verifies default `read_only = false` and baseline write + read succeed.
2. Flips the flag to `true` via the public accessor, asserts writes fail and reads still succeed.
3. Flips back to `false`, asserts writes succeed and the written data reads back correctly.

Runs green locally: `test result: ok. 1 passed; 0 failed; finished in 0.39s`.

## Refs

- OEP: openebs/openebs#4242
- Proto change: openebs/mayastor-dependencies#162
- Companion PR that added the field to develop: openebs/mayastor#2033

cc @tiagolobocastro @Abhinandan-Purkait

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_io.rs
+++ b/io-engine/src/bdev/nexus/nexus_io.rs
@@ -11,7 +11,8 @@ use spdk_rs::{
     libspdk::{
         spdk_bdev_io, spdk_bdev_io_complete_nvme_status, spdk_io_channel,
         SPDK_NVME_SC_ABORTED_SQ_DELETION, SPDK_NVME_SC_CAPACITY_EXCEEDED,
-        SPDK_NVME_SC_INVALID_OPCODE, SPDK_NVME_SC_RESERVATION_CONFLICT,
+        SPDK_NVME_SC_COMMAND_NAMESPACE_IS_PROTECTED, SPDK_NVME_SC_INVALID_OPCODE,
+        SPDK_NVME_SC_RESERVATION_CONFLICT,
     },
     BdevIo,
 };
@@ -165,6 +166,26 @@ impl<'n> NexusBio<'n> {
     /// TODO
     pub(super) fn submit_request(mut self) {
         if !self.accept_request() {
+            return;
+        }
+
+        // ROX gate: reject state-mutating I/O at submit time when the nexus is
+        // currently published read-only. Reads pass through untouched. The
+        // status maps to NVMe "Namespace is Write Protected" so initiators
+        // report a clean, well-known error rather than a generic failure.
+        // The flag is read from the per-core `NexusChannel` snapshot rather
+        // than the atomic on `Nexus` — the write side pushes updates via
+        // `set_nexus_read_only`, keeping this branch a plain bool load on
+        // the hot path.
+        if self.channel().is_read_only()
+            && matches!(
+                self.io_type(),
+                IoType::Write | IoType::WriteZeros | IoType::Unmap | IoType::Reset
+            )
+        {
+            self.fail_nvme_status(NvmeStatus::Generic(
+                SPDK_NVME_SC_COMMAND_NAMESPACE_IS_PROTECTED,
+            ));
             return;
         }
 
--- a/io-engine/src/bdev/nexus/nexus_share.rs
+++ b/io-engine/src/bdev/nexus/nexus_share.rs
@@ -149,22 +149,39 @@ impl<'n> Nexus<'n> {
         protocol: Protocol,
         key: Option<String>,
     ) -> Result<String, Error> {
-        self.share_ext(protocol, key, vec![]).await
+        self.share_ext(protocol, key, vec![], false).await
     }
 
-    /// TODO
+    /// Share the nexus over the given protocol. When `read_only` is true the
+    /// nexus is published in ROX mode: `nexus_io::submit_request` rejects
+    /// write I/O at submit time so multiple initiators can safely share the
+    /// target for read-only workloads. The flag is scoped to the current
+    /// publish and is cleared on unshare, so re-sharing with a different
+    /// value flips the mode.
     pub async fn share_ext(
         mut self: Pin<&mut Self>,
         protocol: Protocol,
         _key: Option<String>,
         allowed_hosts: Vec<String>,
+        read_only: bool,
     ) -> Result<String, Error> {
         // This function should be idempotent as it's possible that
         // we get called more than once for some odd reason.
         if let Some(target) = &self.nexus_target {
             // We're already shared ...
             if Protocol::from(target) == protocol {
-                // Same protocol as that requested, simply return Ok()
+                // Same protocol as requested. `read_only` is negotiated with
+                // the initiator at connect time (NVMe identify data is cached
+                // per session), so a mid-life flip wouldn't propagate to
+                // already-connected clients. Reject a mismatched value here
+                // and require unshare + re-share to change the mode.
+                if self.is_read_only() != read_only {
+                    return Err(Error::ReadOnlyChangeNotAllowed {
+                        name: self.name.clone(),
+                        current: self.is_read_only(),
+                    });
+                }
+
                 warn!("{} is already shared", self.name);
 
                 self.as_mut()
@@ -204,6 +221,14 @@ impl<'n> Nexus<'n> {
                 Ok(uri)
             }
             Protocol::Nvmf => {
+                // Record the effective read-only state before the target goes
+                // live, so any I/O that arrives from an initiator connecting
+                // during the share window is subject to the ROX gate in
+                // `nexus_io::submit_request`. `set_nexus_read_only` updates
+                // the source-of-truth atomic and pushes the flag out to every
+                // per-core channel so the submit path can read a plain bool.
+                self.as_mut().set_nexus_read_only(read_only).await;
+
                 let props = NvmfShareProps::new()
                     .with_range(Some((
                         self.nvme_params.min_cntlid,
@@ -212,7 +237,16 @@ impl<'n> Nexus<'n> {
                     .with_ana(true)
                     .with_allowed_hosts(allowed_hosts)
                     .with_ptpl(self.create_ptpl()?);
-                let uri = self.as_mut().share_nvmf(Some(props)).await?;
+                let uri = match self.as_mut().share_nvmf(Some(props)).await {
+                    Ok(uri) => uri,
+                    Err(e) => {
+                        // Roll back the ROX flag if the share itself failed:
+                        // the target never went live, so leaving `read_only`
+                        // set would misreport state on a subsequent query.
+                        self.as_mut().set_nexus_read_only(false).await;
+                        return Err(e);
+                    }
+                };
 
                 unsafe {
                     self.as_mut().get_unchecked_mut().nexus_target =
@@ -251,6 +285,12 @@ impl<'n> Nexus<'n> {
         }
 
         self.as_mut().unshare(None).await?;
+
+        // Clear the ROX flag: without an active target no front-end I/O can
+        // flow, so read-only doesn't apply here. Reset to `false` (the default
+        // RWO shape) rather than remembering the last-published value. Uses
+        // `set_nexus_read_only` so per-core channel snapshots come along.
+        self.as_mut().set_nexus_read_only(false).await;
 
         if persist_clean {
             // Double-check the `NvmfSubsystem` is really gone before marking
--- a/io-engine/tests/nexus_read_only.rs
+++ b/io-engine/tests/nexus_read_only.rs
@@ -1,0 +1,226 @@
+//! Nexus ROX (ReadOnlyMany) submit-time rejection tests.
+//!
+//! Verifies that when the nexus is marked read-only, write-family I/O
+//! submitted to the nexus bdev is rejected at submit time and reads
+//! continue to succeed. The read-only flag is set via
+//! `Nexus::set_nexus_read_only`, which stores on the atomic on `Nexus` and
+//! fans the new value out to every per-core `NexusChannel` snapshot so the
+//! submit gate reads a plain bool on the hot path. The share/unshare paths
+//! go through the same setter; this test drives it directly to isolate the
+//! submit gate from the rest of the share pipeline.
+
+use common::{bdev_io, MayastorTest};
+use io_engine::{
+    bdev::nexus::{nexus_create, nexus_lookup, nexus_lookup_mut, Error},
+    core::{MayastorCliArgs, Protocol},
+};
+use once_cell::sync::OnceCell;
+
+pub mod common;
+
+static TESTDIR: &str = "/tmp/io-engine-tests";
+
+static DISKNAME1: &str = "/tmp/io-engine-tests/nexus_read_only_1.img";
+static DISKNAME2: &str = "/tmp/io-engine-tests/nexus_read_only_2.img";
+static BDEVNAME1: &str = "aio:///tmp/io-engine-tests/nexus_read_only_1.img?blk_size=512";
+static BDEVNAME2: &str = "aio:///tmp/io-engine-tests/nexus_read_only_2.img?blk_size=512";
+
+static NEXUS_NAME: &str = "nexus_read_only";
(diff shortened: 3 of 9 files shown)


io-engine/src/bdev/nexus/nexus_share.rs
```diff
@@ -204,7 +216,16 @@ impl<'n> Nexus<'n> {
...
                     .with_ptpl(self.create_ptpl()?);
-                let uri = self.as_mut().share_nvmf(Some(props)).await?;
+                let uri = match self.as_mut().share_nvmf(Some(props)).await {
+                    Ok(uri) => uri,
+                    Err(e) => {
+                        // Roll back the ROX flag if the share itself failed:
+                        // the target never went live, so leaving `read_only`
+                        // set would misreport state on a subsequent query.
+                        self.read_only.store(false);
+                        return Err(e);
```
Abhinandan-Purkait, line 247: Hmm, I quite did not get this part. Can you explain a bit more.

io-engine/src/bdev/nexus/nexus_share.rs
```diff
@@ -244,6 +265,11 @@ impl<'n> Nexus<'n> {
 
         self.as_mut().unshare(None).await?;
 
+        // Clear the ROX flag: without an active target no front-end I/O can
+        // flow, so read-only doesn't apply here. Reset to `false` (the default
+        // RWO shape) rather than remembering the last-published value.
+        self.read_only.store(false);
```
Abhinandan-Purkait: While I get the intent here to flip it back to RWO, but I am wondering how will this execute in the entire workflow, if a volume is asked to be ROX then under what case we would like it to be RWO again from the top level k8s flow? Also how does the user configure this?

io-engine/src/bdev/nexus/nexus_share.rs
yugchaudhari: The flag (`self.read_only.store(read_only)`) is set before `share_nvmf` is called, so if an initiator connects mid-share, it's already subject to the ROX gate in `nexus_io::submit_request`.

If `share_nvmf` then fails, the target never went live, but the flag is still set to whatever the caller requested. Rolling it back to `false` keeps internal state consistent with reality: no active target, no ROX property. Otherwise a follow-up query would report ROX on a nexus that isn't actually sharing anything.

io-engine/src/bdev/nexus/nexus_share.rs
yugchaudhari: `read_only` is a per-session property, not a persistent policy. Once we unshare and `nexus_target` clears, there is no live share to gate anymore. The next publish calls `share_ext(read_only=?)` from the CP and sets the flag fresh, so resetting to false here is just cleanup of the previous session, not a policy call.

For the K8s flow: io-engine doesn't decide the mode, it just responds to what the CP passes on each publish. The K8s side composes RWO → ROX via `reclaimPolicy: Retain` + sequential PVC re-binds (writer PVC → delete → strip claimRef → ROX PVC). The reverse is symmetric. User opts in via a StorageClass parameter (`rox_fs: "true"`), and the PVC access mode (`ReadOnlyMany`) tells what the CP requests on publish.

yugchaudhari:
@tiagolobocastro @Abhinandan-Purkait Please take a look whenever you have some time.

io-engine/src/bdev/nexus/nexus_io.rs
```diff
@@ -170,6 +171,22 @@ impl<'n> NexusBio<'n> {
             return;
         }
 
+        // ROX gate: reject state-mutating I/O at submit time when the nexus is
+        // currently published read-only. Reads pass through untouched. The
+        // status maps to NVMe "Namespace is Write Protected" so initiators
+        // report a clean, well-known error rather than a generic failure.
+        if self.nexus().read_only.load()
```
tiagolobocastro: this is not a good idea, an atomic in the happy path here will adversely affect the performance

yugchaudhari: One way I am thinking is, moving the flag off the atomic on the hot path by caching it on `NexusChannel`, so `submit_request` reads a plain bool from the per-core snapshot it already dereferences (right next to `self.channel().is_frozen()`).

- Add `read_only: bool` on `NexusChannel<'n>`, seeded in `NexusChannel::new` from the nexus.
- `Nexus::set_nexus_read_only(&self, ro: bool).await`, mirrors the existing `set_nexus_io_mode` shape: stores on the atomic, then `traverse_io_channels_async(ro, |ch, ro| ch.set_read_only(*ro))` to extend to every per-core channel.
- `share_ext` / `unshare_nexus_internal` call `set_nexus_read_only` instead of `self.read_only.store(..)` directly.
- Atomic on Nexus stays as the source of truth for late-created channels (constructor seed) and for the gRPC query path.
- `submit_request` becomes `if self.channel().read_only && matches!(...)`, plain load from a struct we already have a reference to.

Does this sound right, or should I move to per-core state entirely and drop the atomic?

tiagolobocastro: Yes I think the `NexusChannel` probably is the right place for it.
Also to consider, should we claim reads are unsupported when we enter this mode? Would need to check if this is checked on every io or only when there are changes (probably don't want to pay extra cost if it's checked per io)
Also can we set the nvme namespace as readonly?

yugchaudhari: `read_only` now lives on `NexusChannel<'n>` as a plain bool, seeded from the nexus at channel creation and flipped via `Nexus::set_nexus_read_only` (mirrors `set_nexus_io_mode`, fans out through `traverse_io_channels_async`). Submit path is now a plain bool load next to the existing `is_frozen()` check, no atomic on the hot path. Atomic on `Nexus` stays as the source of truth for late-created channels and the gRPC query. `share_ext` / `unshare_nexus_internal` call through the async setter, and the existing unit test drives it directly to isolate the submit gate.

yugchaudhari: On the two other suggestions:

Claiming writes unsupported via `io_type_supported`: SPDK queries this at `bdev-open` / `capability-negotiation` time, not per-IO, so no hot-path cost. Catch is SPDK treats the answer as effectively static: flipping it after an initiator has connected doesn't propagate to that initiator, and the identify data it already cached says writes are supported. Useful as a static hint for new connections but doesn't replace the runtime gate under our current `share_ext` / `unshare` cycle, and asymmetric behavior (some connected clients see writable, some don't) is worse than a uniform runtime reject. We can leave it currently out of iteration1, but if you want, I can think something for iteration 2 onwards if it's possible and how.

NVMe namespace write-protect: this seems right, and it's the cleanest signal (initiator kernel sees the namespace as read-only from identify data, returns `EROFS` locally, writes never hit the wire). But `spdk_nvmf_ns_get_write_protect` / the corresponding setter isn't bound in `spdk-rs` right now, so it needs a `spdk-rs` binding PR first plus the plumbing here. Happy to do it as a follow-up iteration. If we go with this approach, `io_type_supported` becomes redundant.

io-engine/src/bdev/nexus/nexus_share.rs
```diff
@@ -141,15 +141,21 @@ impl<'n> Nexus<'n> {
...
+    /// write I/O at submit time so multiple initiators can safely share the
+    /// target for read-only workloads. The flag is scoped to the current
+    /// publish and is cleared on unshare, so re-sharing with a different
+    /// value flips the mode.
     pub async fn share_ext(
         mut self: Pin<&mut Self>,
         protocol: Protocol,
         _key: Option<String>,
         allowed_hosts: Vec<String>,
+        read_only: bool,
```
tiagolobocastro, line 166: what if the nexus_target already exists?

io-engine/src/bdev/nexus/nexus_io.rs
tiagolobocastro: Sorry for delay, pls feel free to ping next time

Would be good to add support for setting the I think returning the `spdk_nvmf_ns_get_write_protect` in spdk as a next task; returning the `SPDK_NVME_SC_COMMAND_NAMESPACE_IS_PROTECTED` seems ok for now.

tiagolobocastro: I've got a question on the "capability-negotiation", is this done when the target is created?
If so then we could add a restriction that the rd_only flag is never changed, and target must be destroyed/re-created in order to change this

io-engine/src/bdev/nexus/nexus_share.rs
yugchaudhari: Good catch, currently the same-protocol re-share branch drops the `read_only` arg silently. Fixed: `share_ext` now returns a new `Error::ReadOnlyChangeNotAllowed` when a live target's `read_only` differs from the requested value. Same-value re-share still no-ops. Also addresses the `capability-negotiation`: unshare + re-share is now the only way to flip the mode. Added a unit test (`nexus_read_only_immutable_on_shared_target`) that walks RWO share → same-value no-op → mismatched-value reject → unshare → ROX share → mirror reject.

io-engine/src/bdev/nexus/nexus_io.rs
yugchaudhari: > I've got a question on the "capability-negotiation", is this done when the target is created? If so then we could add a restriction that the rd_only flag is never changed, and target must be destroyed/re-created in order to change this

Yes, NVMe identify data is fetched at connect time and cached per session, so a mid-life flip wouldn't propagate to already-connected initiators. Agreed on the restriction: `read_only` is now immutable for the target's lifetime. `share_ext` rejects a mismatched value with `Error::ReadOnlyChangeNotAllowed` when a target already exists. The CP will have the mirror guard on `PublishVolume` (`VolumeAccessModeConflict` on RWO↔ROX without unpublish), so io-engine and CP will be aligned.

yugchaudhari: > Would be good to add support for setting the I think returning the `spdk_nvmf_ns_get_write_protect` in spdk as a next task; returning the `SPDK_NVME_SC_COMMAND_NAMESPACE_IS_PROTECTED` seems ok for now.

Agreed, keeping `NAMESPACE_IS_PROTECTED` for iter 1. I'll raise the spdk-rs binding PR for `spdk_nvmf_ns_get`/`set_write_protect` after this lands and thread it into the nexus share path as a follow-up iteration so future connects also see the initiator-side `EROFS`. Tracking it as a work item on the OEP so it doesn't get lost.

io-engine/tests/nexus_read_only.rs
```diff
@@ -0,0 +1,223 @@
...
+use io_engine::{
+    bdev::nexus::{nexus_create, nexus_lookup, nexus_lookup_mut, Error},
+    core::{MayastorCliArgs, Protocol},
+};
+use once_cell::sync::OnceCell;
+use std::process::Command;
+
+pub mod common;
+
+static DISKNAME1: &str = "/tmp/nexus_read_only_1.img";
```
tiagolobocastro: Try not to use custom names otherwise it's more files we have to know to delete if test crashes.
Or otherwise use the new tmp location I started using on `lvs_pool_test` as the cleanup script clears that path.

```suggestion
static DISKNAME1: &str = "/tmp/disk1.img";
```

io-engine/tests/nexus_read_only.rs
```diff
@@ -0,0 +1,223 @@
...
+        MayastorTest::new(MayastorCliArgs {
+            enable_io_all_thrd_nexus_channels: true,
+            ..Default::default()
+        })
+    })
+}
+
+fn create_backing_files(paths: &[&str]) {
+    for path in paths {
+        let _ = Command::new("rm").args(["-f", path]).output();
```
tiagolobocastro: we already have common::* functions for this, please reuse them

io-engine/tests/nexus_read_only.rs
```diff
@@ -0,0 +1,223 @@
...
+        bdev_io::write_some(NEXUS_NAME, 0, 2, 0xbb)
+            .await
+            .expect("write should succeed after flipping back to RWO");
+        bdev_io::read_some(NEXUS_NAME, 0, 2, 0xbb)
+            .await
+            .expect("read after RWO write should reflect the written data");
+    })
+    .await;
+
+    remove_backing_files(&[DISKNAME1, DISKNAME2]);
```
tiagolobocastro: We already have functions for this, something like so:


```suggestion
    common::delete_file(&[DISKNAME1, DISKNAME2]);
```

tiagolobocastro (commented):
Thanks @yugchaudhari, just a few testing nits

yugchaudhari:
> Thanks @yugchaudhari, just a few testing nits

@tiagolobocastro 

Applied all three:
- Moved backing files to /tmp/io-engine-tests/ 
- Dropped the custom `create_backing_files` / `remove_backing_files` helpers and now call `common::truncate_file` + `common::delete_file` directly.
- Kept distinct filenames for the two tests in this binary so they can't race on setup/teardown

@Abhinandan-Purkait please take a look whenever you have some time.

tiagolobocastro (approved):

abhilashshetty04 (approved):

Abhinandan-Purkait (approved):

Abhinandan-Purkait:
@yugchaudhari Can you please join tomorrow's community meeting?

yugchaudhari:
> @yugchaudhari Can you please join tomorrow's community meeting?

Sure thing!

tiagolobocastro:
bors merge

yugchaudhari:
> bors merge

@tiagolobocastro can we try merging again, last time the runner died mid-test so bdd tests could not finish.

tiagolobocastro:
bors merge

PR #2036, opened 2026-07-14
Nexus Label V2 - implement requested size constraint

tiagolobocastro:
refactor(cli): support specifying label version and size required
    
    Allows for us to use the new api whilst keeping the existing one
    in-place for back-compat.
    
---

    feat: add request size constraint enforcement
    
    Ensure the nexus bdev size matches the exact request size when
    the caller specifies this to be the case.
    
---

    refactor(label/v2): ensure size aligned to 1MiB with 4MiB void
    
    Goes back to the original scope, as otherwise rounding to 4MiB multiple
    to the size again leads to size inconsistency.


abhilashshetty04 (approved):

niladrih (approved):

tiagolobocastro:
bors merge

PR #2040, opened 2026-07-21
fix: don't detach devices on submission error

tiagolobocastro:
The control-plane test republished_nexus_io_engine_txn_fail revealed a
race in the nexus.
The test pauses the io-engine container, and then thaws the container.
Upon thaw the container resumes BUT the io channel for nvmf children
is in a broken state and starts failing the submission of IO itself.

The code was currently handling this scenario but detaching the handle
and disconnecting it altogether.
I think this is actually wrong becauses subsequent IOs on the channel
would skip this handle and complete back to the host!

I'm not entirely sure why it was done this way, but I suspect we wanted
to avoid resubmission loops?

A potentially good fix for this would be to freeze the channel, which
is what the retire already does anyway, but it runs some-time after
only.
We could freeze the channel ahead of time, though we have to take care
and ensure we don't leave it frozen.

So for this patch the safer way of at least fixing the issue for now
is to freeze these IOs without changing the state of the channel and
then defrostign them sometime later.

Since we don't want to keep doing this forever, we resubmit only for
a certain number of times and then we let them fail.
To avoid writing to healthy children continuously we change the order
so that this handle is first in the list.

Also, I don't know why we stop at first failure, this should be
investigated further :/

### DIFF
--- a/.github/workflows/bdd.yml
+++ b/.github/workflows/bdd.yml
@@ -105,7 +105,7 @@ jobs:
       - name: Cleanup
         if: always()
         run: nix-shell --run "./scripts/pytest-tests.sh --clean-all-exit"
-# debugging
+      # debugging
       # - name: Setup tmate session
       #   if: ${{ failure() }}
       #   timeout-minutes: 240
--- a/.github/workflows/unit-int.yml
+++ b/.github/workflows/unit-int.yml
@@ -84,10 +84,11 @@ jobs:
       - name: Run JS Grpc Tests
         run: |
           echo "TEST_START_DATE=$(date +"%Y-%m-%d %H:%M:%S")" >> $GITHUB_ENV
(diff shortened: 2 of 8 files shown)


tiagolobocastro:
bors try

Abhinandan-Purkait (approved):

niladrih (approved):

tiagolobocastro:
bors try

tiagolobocastro:
seems like some GitHub runner failure: `The self-hosted runner lost communication with the server. Verify the machine is running and has a healthy network connection. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.` ?

bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

PR #2042, opened 2026-07-29
fix: abort frozen ios from shutdown nexus on unshare

tiagolobocastro:
When a nexus goes into shutdown its frozen IOs remain, and must be
eventually failed otherwise unshare will get stuck..

For now we tackle this by failing them on unshare itself, though
there may be some cases where we may want to handle this a bit
differently.

I'll tackle a solution for shutdown-after-unshare starts on a follow up.

Another interesting topic here is that after shutdown newer IOs may
be received and will again be frozen!
This IIRC is to ensure IOs don't fail back to initiator, and give some
time for a new nexus to be setup.

We should look into this, perhaps kicking initiators out of the nexus
and ensuring no new connection may be established would be sufficient?
This would require some research and testing so not tacking this here.

However, during unshare when nexus is shutdown we can already handle this
by at least ensuring new IOs are rejected.

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_bdev.rs
+++ b/io-engine/src/bdev/nexus/nexus_bdev.rs
@@ -1075,6 +1075,34 @@ impl<'n> Nexus<'n> {
         Ok(())
     }
 
+    /// Aborts all frozen IOs of a shutdown Nexus.
+    /// # Warning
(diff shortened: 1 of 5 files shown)


abhilashshetty04 (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #2045, opened 2026-07-31
Pool hot remove

tiagolobocastro:
test: use new rescan to trigger hot-remove
    
    Changes the hot-remove tests to use rescan instead of grow.

---

    feat: add pool hot-remove rescanner
    
    Device removal is not always handled automatically by the pool device layer
    
    PCIe processes hot-removal events and will trigger hot-removal of the
    underlying device and the pool.
    
    AIO/URING devices need IO to be submitted to detect hot-removal, so if the
     device is not being used, it will not be detected as removed.
    Rescan on these devices inspects the file handles and will trigger a
     hot-removal event if the file handle is no longer valid.
    
    A cli argument is added to enable the rescan and also configure how often
    it should run (rescan period).

### DIFF
--- a/io-engine/src/pool_backend.rs
+++ b/io-engine/src/pool_backend.rs
@@ -1,8 +1,9 @@
 use crate::{
     bdev::crypto::EncryptionKey,
-    core::{BdevStater, BdevStats, ToErrno},
+    core::{BdevStater, BdevStats, Reactors, ToErrno},
     replica_backend::ReplicaOps,
 };
+use futures::channel::oneshot;
 use nix::errno::Errno;
 use std::ops::Deref;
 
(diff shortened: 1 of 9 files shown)


io-engine/src/pool_backend.rs
```diff
@@ -169,6 +170,9 @@ pub trait PoolOps: IPoolProps + BdevStater<Stats = BdevStats> + std::fmt::Debug
     /// The pool will no longer be listable until it is imported again.
     async fn export(self: Box<Self>) -> Result<(), Error>;
 
+    /// Rescan the pool to detect any hot-removals.
```
abhilashshetty04: Maybe we don't say rescan to detect hot-removals?

io-engine/src/pool_backend.rs
```diff
@@ -322,3 +326,68 @@ impl PoolFactory {
...
+/// Device removal is not always handled automatically by the pool device layer:
+///
+/// 1. PCIe processes hot-removal events and will trigger hot-removal of the underlying device and the pool
+///
+/// 2. AIO/URING devices need IO to be submitted to detect hot-removal, so if the device is not being used, it will not be detected as removed.
+///    Rescan on these devices inspects the file handles and will trigger a hot-removal event if the file handle is no longer valid.
+///
+/// # Note
+///
+/// Pools are not currently grown automatically, but this would expose a new size of the underlying device to the pool.
```
abhilashshetty04:
```suggestion
/// Pools are not currently grown automatically, but this would expose a new size of the underlying device to the pool if its extended.
```

abhilashshetty04 (approved):

io-engine/src/pool_backend.rs
tiagolobocastro: Sorry didn't understand, you're saying we shouldn't mention hot-removals here?

Abhinandan-Purkait (approved):

PR #2047, opened 2026-08-03
fix: await close children correctly

tiagolobocastro:
The NexusChild::close tried to be somewhat idempotent by returning early if the child is already in the closing process.

But this is actually unsafe because an in-progress close may still be using the remove-channel and therefore may see a corrupted value there leading to a potential deadlock for example.

This is fixed by ensuring the early close awaits proper close completion by use of a close channel, which lets callers wait on it.

### DIFF
--- a/io-engine/src/bdev/nexus/nexus_child.rs
+++ b/io-engine/src/bdev/nexus/nexus_child.rs
@@ -281,6 +281,8 @@ pub(crate) enum ChildDestroyState {
     None,
     /// The child device is being destroyed.
     Destroying,
+    /// The child device is destroyed.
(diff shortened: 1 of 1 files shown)


Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
```
The self-hosted runner lost communication with the server. Verify the machine is running and has a healthy network connection. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.
```
bors merge

Issue #1458, opened 2023-07-19
IO Engine Cannot set affinity


sfxworks:
**Describe the bug**
First install of mayastor, I'm getting a "Cannot set affinity" error

**To Reproduce**
1. Install mayastor with values below
```yaml
USER-SUPPLIED VALUES:
agents:
  core:
    capacity:
      thin:
        poolCommitment: 250%
        snapshotCommitment: 40%
        volumeCommitment: 40%
        volumeCommitmentInitial: 40%
    logLevel: info
    partialRebuildWaitPeriod: ""
    priorityClassName: ""
    resources:
      limits:
        cpu: 1000m
        memory: 128Mi
      requests:
        cpu: 500m
        memory: 32Mi
    tolerations: []
  ha:
    cluster:
      logLevel: info
      resources:
        limits:
          cpu: 100m
          memory: 64Mi
        requests:
          cpu: 100m
          memory: 16Mi
    enabled: true
    node:
      logLevel: info
      priorityClassName: ""
      resources:
        limits:
          cpu: 100m
          memory: 64Mi
        requests:
          cpu: 100m
          memory: 64Mi
      tolerations: []
apis:
  rest:
    logLevel: info
    priorityClassName: ""
    replicaCount: 1
    resources:
      limits:
        cpu: 100m
        memory: 64Mi
      requests:
        cpu: 50m
        memory: 32Mi
    service:
      nodePorts:
        http: 30011
        https: 30010
      type: ClusterIP
    tolerations: []
base:
  cache_poll_period: 30s
  default_req_timeout: 5s
  imagePullSecrets:
    enabled: false
    secrets:
    - name: login
  initContainers:
    containers:
    - command:
      - sh
      - -c
      - trap "exit 1" TERM; until nc -vzw 5 {{ .Release.Name }}-agent-core 50051;
        do date; echo "Waiting for agent-core-grpc services..."; sleep 1; done;
      image: busybox:latest
      name: agent-core-grpc-probe
    - command:
      - sh
      - -c
      - trap "exit 1" TERM; until nc -vzw 5 {{ .Release.Name }}-etcd {{.Values.etcd.service.port}};
        do date; echo "Waiting for etcd..."; sleep 1; done;
      image: busybox:latest
      name: etcd-probe
    enabled: true
  initCoreContainers:
    containers:
    - command:
      - sh
      - -c
[...]

PR #2050, opened 2026-08-04
feat(io-engine): derive core list from kubelet-assigned cgroup cpuset

therayquaza:
## Summary
- Adds a `--cores-from-cgroup` CLI flag (env `CORES_FROM_CGROUP`) to io-engine that derives the reactor core list from the CPU set actually assigned to the pod's cgroup (`cpuset.cpus.effective` on cgroup v2, `cpuset.effective_cpus` on v1), instead of a statically configured `-l`/`-m`.
- Falls back to `-l`/`-m` if the cgroup cpuset can't be determined.
- Addresses #1458: with the Kubernetes CPU Manager static policy, the exact cores assigned to a pod aren't known ahead of time, so a statically configured core list/mask can mismatch the cores actually granted.

## Test plan
- [x] Unit tests for cgroup parsing and candidate-path resolution (`io-engine/src/core/cgroup_cpuset.rs`)
- [x] Manual test on a kubelet node with CPU Manager static policy enabled

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/io-engine/src/core/cgroup_cpuset.rs
+++ b/io-engine/src/core/cgroup_cpuset.rs
@@ -1,0 +1,105 @@
+//! Reads the CPU set assigned to this process' cgroup by the container
+
+use std::{fs, path::PathBuf};
+
+/// Returns the effective cgroup CPU set as a DPDK `-l`-compatible core-list
+/// string (e.g. `"0-2,7"`), or `None` if it can't be determined.
+pub fn cpuset_from_cgroup() -> Option<String> {
+    let cgroups = read_self_cgroup(&fs::read_to_string("/proc/self/cgroup").ok()?);
+
+    candidate_cpuset_files(&cgroups).into_iter().find_map(|path| {
+        let cpus = fs::read_to_string(&path).ok()?;
+        let cpus = cpus.trim();
+        if cpus.is_empty() {
+            return None;
+        }
+        debug!("Read cgroup cpuset '{cpus}' from {}", path.display());
+        Some(cpus.to_string())
+    })
+}
+
+/// Parses `/proc/self/cgroup` into `(controllers, path)` pairs.
+fn read_self_cgroup(contents: &str) -> Vec<(String, String)> {
+    contents
+        .lines()
+        .filter_map(|line| {
+            let mut parts = line.splitn(3, ':');
+            let _hierarchy_id = parts.next()?;
+            let controllers = parts.next()?.to_string();
+            let path = parts.next()?.to_string();
+            Some((controllers, path))
+        })
+        .collect()
+}
+
+/// Candidate paths to the cpuset "effective" CPUs file, v2 then v1.
+fn candidate_cpuset_files(cgroups: &[(String, String)]) -> Vec<PathBuf> {
+    let mut candidates = Vec::new();
+
+    if let Some((_, path)) = cgroups.iter().find(|(controllers, _)| controllers.is_empty()) {
+        candidates.push(PathBuf::from(format!(
+            "/sys/fs/cgroup{path}/cpuset.cpus.effective"
+        )));
+    }
+
+    if let Some((_, path)) = cgroups
+        .iter()
+        .find(|(controllers, _)| controllers.split(',').any(|c| c == "cpuset"))
+    {
+        candidates.push(PathBuf::from(format!(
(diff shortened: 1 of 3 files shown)


io-engine/src/core/env.rs
```diff
@@ -712,7 +720,14 @@ impl MayastorEnvironment {
             rpc_addr: args.rpc_address,
             hugedir: args.hugedir,
             env_context: args.env_context,
-            core_list: args.core_list,
+            core_list: if args.cores_from_cgroup {
+                cgroup_cpuset::cpuset_from_cgroup().or_else(|| {
+                    warn!("--cores-from-cgroup was set but the cgroup cpuset could not be determined; falling back to -l/-m");
```
tiagolobocastro, line 725: I think this should be a failure rather than reverting back to core list, or is there some case where this can happen?

io-engine/src/core/cgroup_cpuset.rs
```diff
@@ -0,0 +1,105 @@
+//! Reads the CPU set assigned to this process' cgroup by the container
+
+use std::{fs, path::PathBuf};
+
+/// Returns the effective cgroup CPU set as a DPDK `-l`-compatible core-list
+/// string (e.g. `"0-2,7"`), or `None` if it can't be determined.
+pub fn cpuset_from_cgroup() -> Option<String> {
+    let cgroups = read_self_cgroup(&fs::read_to_string("/proc/self/cgroup").ok()?);
```
tiagolobocastro, line 8: Do we need to read from cgroup or can we do

```rust
libc::sched_getaffinity(0, std::mem::size_of::<libc::cpu_set_t>(), &mut set)
```

Then we don't need to worry about group v1 or v2?

tiagolobocastro: Or is there some advantage in using the sysfs?

therayquaza: No, I dont think there is an advantage, I did not know about that syscall.
Will take a look :pray:

io-engine/src/core/env.rs
therayquaza: As you wish, I guess a failure would makes more sense but I felt like warning and fall back was safer :man_shrugging:

tiagolobocastro: Also I think we may need some tweaking to allow for this to work if K8s changes the core allocation
Perhaps we can build upon the work being done [here](https://github.com/openebs/mayastor/pull/2017). 
This can be another PR once both these merge

Scorplionn:
Re: reading the cpuset — `sched_getaffinity` is the more robust choice here, and
not only because it avoids the cgroup v1/v2 branch.

The io-engine container runs privileged with the host cgroup hierarchy
bind-mounted, so reading `/sys/fs/cgroup/cpuset.cpus.effective` from inside it
returns the *host* view rather than the pod's. On our nodes:

    $ kubectl exec -n openebs <io-engine-pod> -- cat /sys/fs/cgroup/cpuset.cpus.effective
    0-31

while the CPU manager had actually granted cores 2-3:

    $ sudo cat /var/lib/kubelet/cpu_manager_state | jq .
    {
      "policyName": "static",
      "defaultCpuSet": "4-31",
      "entries": { "3883fe77-...": { "io-engine": "2-3" } }
    }

Process affinity reports the correct set:

    $ grep Cpus_allowed_list /proc/self/status
    Cpus_allowed_list:  2-3

We confirmed this by wrapping the entrypoint in a shell that derived `-l` from
`Cpus_allowed_list` (equivalent to `sched_getaffinity`) — reactors bound to
exactly the granted cores. Reading sysfs would have picked up all 32 and spawned
a reactor on every core.

Environment: Kubernetes v1.35, cgroup v2, Rocky Linux 10 (kernel 6.12),
`cpuManagerPolicy: static` with `strict-cpu-reservation: true`.

Scorplionn:
@therayquaza are you still planning to address the review feedback here
(`sched_getaffinity` instead of the cgroup read, failing instead of falling back
to `-l`)? openebs/mayastor-extensions#962 already has two approvals and is only
waiting on this.

@tiagolobocastro if the author doesn't have time, I'm happy to open a follow-up
building on this work with those changes, and test it on a cluster running
`cpuManagerPolicy: static` with `strict-cpu-reservation: true`.

Issue #2055, opened 2026-08-13
io-engine panics in Lvol::committed during replica destruction: assertion failed: !blob.is_null()


deathalt:
**Describe the bug**

The Mayastor `io-engine` process panics while destroying a replica.

The panic occurs in `Lvol::committed()` when the underlying SPDK blob pointer is already null:

```text
thread 'main' panicked at io-engine/src/lvs/lvs_lvol.rs:1056:9:
assertion failed: !blob.is_null()
```

This terminates the entire `io-engine` process with exit code `101`. Kubernetes restarts the container automatically, but all pools and single-replica volumes hosted by that io-engine become temporarily unavailable while the pools are being imported.

In our incident, the panic caused three active volumes to report `No online replicas`. The volumes recovered after their Nexus resources were recreated.

There were no disk I/O errors or stalled pools. After recovery, all affected DiskPools were `Online` and `Healthy`, with:

```text
ioErrorCount: 0
ioStalled: false
```

This appears to be an io-engine lifecycle/concurrency bug rather than a physical disk failure.

Immediately before the panic, a Nexus was destroyed and its replica was being removed:

```text
[2026-08-12T15:57:49.522332740+00:00 INFO]
Nexus '68a0b686-3790-4d57-a608-00029de8f2a5': nexus destroyed ok

[2026-08-12T15:57:51.090336783+00:00 INFO]
DestroyReplicaRequest {
  uuid: "2997a85e-eda8-4b00-ba01-552aee3101b5",
  pool: Some(PoolName("<affected-pool>"))
}

[2026-08-12T15:57:51.090422648+00:00 INFO]
Lvol '<affected-pool>/<lvol-uuid>/2997a85e-eda8-4b00-ba01-552aee3101b5'
[thin 150.00 GiB]: unshared

thread 'main' panicked at io-engine/src/lvs/lvs_lvol.rs:1056:9:
assertion failed: !blob.is_null()
```

Stack trace:

```text
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: <io_engine::lvs::lvs_lvol::Lvol as io_engine::core::logical_volume::LogicalVolume>::committed
   4: io_engine::lvs::<impl io_engine::pool_backend::IPoolProps for io_engine::lvs::lvs_store::Lvs>::committed
[...]

PR #2059, opened 2026-08-18
Lvol race: listing pools when destroying replicas

tiagolobocastro:
fix: don't list lvs lvols which are deleting
    
    Root cause of the race (all on the single-threaded reactor, inside SPDK):
    1. `DestroyReplica` calls `vbdev_lvol_destroy`, which first
        `spdk_bdev_unregister`s the lvol bdev. The bdev destruct closes the
        lvol and `lvol_close_blob_cb` sets `lvol->blob = NULL`.
    2. The lvol is NOT unlinked from the `lvs->lvols` list until later, in
        `lvol_delete_blob_cb`, after `spdk_bs_delete_blob` completes.
    3. Between those two SPDK callbacks the lvol is still enumerable via
        `Lvs::lvols()` but its blob is already null.
    
    When listing lvols from an lvolstore, we now skip the ones which are deleting.
    This means a pool list call will not compute the committed or any other
    information from these deleting lvols.
    
---

    test: concurrent pool list and replica destroy
    
    Adds a test to reproduce the reported issue with the metrics export doing list
    call whilst a replica is being destroyed.

### DIFF
--- a/io-engine/src/lvs/lvs_lvol.rs
+++ b/io-engine/src/lvs/lvs_lvol.rs
@@ -303,7 +303,11 @@ impl Lvol {
         if lvol.is_null() {
             return None;
         }
-        Some(Self::from_inner_ptr(lvol))
+        let lvol = Self::from_inner_ptr(lvol);
+        if lvol.is_deleting() {
+            return None;
+        }
+        Some(lvol)
     }
 
     /// Wipe the first 8MB if unmap is not supported on failure the operation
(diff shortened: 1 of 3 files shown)


tiagolobocastro:
Resolves #2055

susobhan87 (approved):
Is there any other places where we need to skip is_deleting() apart from list?

abhilashshetty04 (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
> Is there any other places where we need to skip is_deleting() apart from list?

Yes it's possible, though if it's through the replica service likely won't happen; still I've added it to `lookup_by_uuid_str` just to be safe.

PR #2063, opened 2026-08-27
feat(nexus): add read-child selection policy (round-robin default, sm…

susobhandey:
Nexus reads have always been round-robined across every healthy child regardless of locality, so with N replicas only 1/N of reads land on a local (same-node) replica when one is available - the rest needlessly cross the network. NexusChild::is_local() has existed since 2020 and was wired into rebuild-source selection in 2024 to cut network traffic, but never into the read path. And also 49% performance improve

Add a NexusReadPolicy enum:
  - RoundRobin (default): unchanged behaviour, rotate across all healthy children.
  - Smart: prefer healthy local reader(s); fall back to round-robin over remote readers only when no local reader is healthy.

Configurable via --policy / NEXUS_READ_POLICY (default round-robin, so existing deployments see no behaviour change unless opted in).

Live-verified on a 3-node cluster with a hand-built nexus (1 local + 2 remote NVMe-oF children): round-robin reproduced the reported 2/3-of-reads-cross-network behaviour exactly (293088/293088/293088 reads across the 3 replicas); --policy smart eliminated all network reads (1311459/0/0) and increased read IOPS ~49% in the same run.

### DIFF
--- a/io-engine/src/bdev/nexus/mod.rs
+++ b/io-engine/src/bdev/nexus/mod.rs
@@ -183,3 +183,36 @@ pub static ENABLE_NEXUS_CHANNEL_DEBUG: AtomicBool = AtomicBool::new(false);
 /// Whether the nexus channel should have readers/writers configured.
 /// This must be set true ONLY from tests.
 pub static ENABLE_IO_ALL_THRD_NX_CHAN: AtomicBool = AtomicBool::new(false);
+
+/// Policy governing how a nexus channel picks which child to read from.
+/// Configured once at startup via `--policy` / `NEXUS_READ_POLICY` (see
+/// `MayastorCliArgs::nexus_read_policy`) and picked up by each
+/// `NexusChannel` from `MayastorEnvironment` when it is created (see
+/// `NexusChannel::new()`).
+#[derive(clap::ValueEnum, Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub enum NexusReadPolicy {
+    /// Rotate reads across all healthy children, local or remote alike.
+    /// This is the default, preserving Mayastor's historical behaviour.
+    #[default]
+    RoundRobin,
+    /// Prefer a healthy local (same-node) replica for reads, keeping read
+    /// I/O off the network whenever a local copy of the data is available.
+    /// Falls back to rotating across remote readers only when no local
+    /// reader is currently healthy.
+    LocalPreferred,
+}
+
+impl std::fmt::Display for NexusReadPolicy {
+    /// Renders via the same kebab-case name clap uses for `--policy`
+    /// (`ValueEnum`), so the two never drift apart. Needed for
+    /// `#[clap(default_value_t = ..)]` on `nexus_read_policy`.
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{}",
+            clap::ValueEnum::to_possible_value(self)
+                .expect("no skipped variants")
+                .get_name()
+        )
+    }
+}
--- a/io-engine/src/bdev/nexus/nexus_channel.rs
+++ b/io-engine/src/bdev/nexus/nexus_channel.rs
@@ -7,16 +7,28 @@ use std::{
     sync::atomic::Ordering,
 };
 
-use super::{FaultReason, IOLogChannel, Nexus, NexusBio};
+use super::{FaultReason, IOLogChannel, Nexus, NexusBio, NexusReadPolicy};
 
-use crate::core::{BlockDeviceHandle, CoreError, Cores};
+use crate::core::{BlockDeviceHandle, CoreError, Cores, MayastorEnvironment};
 use spdk_rs::Thread;
 
 /// I/O channel, per core.
 #[repr(C)]
 pub struct NexusChannel<'n> {
     writers: Vec<Box<dyn BlockDeviceHandle>>,
+    /// Reader handles, with any local (same-node) readers kept at the front
+    /// of the vector and remote (e.g. NVMe-oF) readers after them. See
+    /// `local_reader_count`.
     readers: Vec<Box<dyn BlockDeviceHandle>>,
+    /// Number of entries at the front of `readers` that are local. When
+    /// non-zero, `select_reader()` rotates only over that local prefix so
+    /// reads stay off the network while a local copy is healthy.
+    local_reader_count: usize,
+    /// Nexus read policy in effect for this channel, snapshotted from
+    /// `MayastorEnvironment` when the channel was created. Kept on the
+    /// channel (rather than read from a global on every I/O) so
+    /// `select_reader()` stays off the shared-state path entirely.
+    read_policy: NexusReadPolicy,
     detached: Vec<Box<dyn BlockDeviceHandle>>,
     io_logs: Vec<IOLogChannel>,
     previous_reader: UnsafeCell<usize>,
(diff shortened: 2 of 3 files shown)


io-engine/src/bdev/nexus/mod.rs
```diff
@@ -183,3 +185,48 @@ pub static ENABLE_NEXUS_CHANNEL_DEBUG: AtomicBool = AtomicBool::new(false);
...
+pub enum NexusReadPolicy {
+    /// Rotate reads across all healthy children, local or remote alike.
+    /// This is the default, preserving Mayastor's historical behaviour.
+    #[default]
+    RoundRobin,
+    /// Prefer a healthy local (same-node) replica for reads, keeping read
+    /// I/O off the network whenever a local copy of the data is available.
+    /// Falls back to rotating across remote readers only when no local
+    /// reader is currently healthy.
+    Smart,
```
tiagolobocastro: Maybe be clearer:

```suggestion
    LocalPreferred,
```

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -198,26 +198,53 @@ impl<'n> NexusChannel<'n> {
...
-            Some(self.readers[idx].as_ref())
+            return None;
         }
+
+        // The local readers, if any, are always kept at the front of
+        // `readers` (see `connect_children()`). Under the `Smart` policy,
+        // bounding the rotation to `local_reader_count` restricts it to
+        // just those; under `RoundRobin` (the default) we rotate over
+        // every reader, exactly as before this policy existed.
+        let pool_len = if NexusReadPolicy::current() == NexusReadPolicy::Smart
```
tiagolobocastro: This is placing atomic in the IO path, we should rather ensure the policy is in the channel, which is `&self`

io-engine/src/bdev/nexus/mod.rs
```diff
@@ -183,3 +185,48 @@ pub static ENABLE_NEXUS_CHANNEL_DEBUG: AtomicBool = AtomicBool::new(false);
...
+}
+
+impl NexusReadPolicy {
+    const ROUND_ROBIN: u8 = 0;
```
tiagolobocastro: This seems like C in rust :)
Let's use the enum values instead

io-engine/src/core/env.rs
```diff
@@ -1180,6 +1181,8 @@ impl MayastorEnvironment {
             nexus::ENABLE_IO_ALL_THRD_NX_CHAN.store(true, SeqCst);
         }
 
+        self.nexus_read_policy.set_current();
```
tiagolobocastro: I think this is not needed, you can read from `MayastorEnvironment` when creating the channel
See change for the ROX, IIRC was similar to this

io-engine/src/bdev/nexus/mod.rs
```diff
@@ -183,3 +185,48 @@ pub static ENABLE_NEXUS_CHANNEL_DEBUG: AtomicBool = AtomicBool::new(false);
 /// Whether the nexus channel should have readers/writers configured.
 /// This must be set true ONLY from tests.
 pub static ENABLE_IO_ALL_THRD_NX_CHAN: AtomicBool = AtomicBool::new(false);
+
+/// Policy governing how a nexus channel picks which child to read from.
+/// Configured once at startup via `--policy` / `NEXUS_READ_POLICY` (see
+/// `MayastorCliArgs::nexus_read_policy`) and read on every read I/O by
+/// `NexusChannel::select_reader()`.
+#[derive(clap::ValueEnum, Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub enum NexusReadPolicy {
```
tiagolobocastro, line 193: Given we're still doing round robin for the remotes, and prefer single only if local I wonder if we could make it the default, at least in the helm chart?

io-engine/src/bdev/nexus/mod.rs
susobhandey: Fixed

io-engine/src/bdev/nexus/mod.rs
susobhandey: Fixed

io-engine/src/core/env.rs
susobhandey: Fixed

io-engine/src/bdev/nexus/mod.rs
susobhandey: Need Discussion currently by default round-robin

Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

io-engine/src/bdev/nexus/nexus_channel.rs
```diff
@@ -122,6 +135,8 @@ impl<'n> NexusChannel<'n> {
         let mut res = Self {
             writers: Vec::new(),
             readers: Vec::new(),
+            local_reader_count: 0,
+            read_policy: MayastorEnvironment::global_or_default().nexus_read_policy,
```
tiagolobocastro: nit: can probably just use `global` here

```suggestion
            read_policy: MayastorEnvironment::global().nexus_read_policy,
```

io-engine/src/core/env.rs
```diff
@@ -294,6 +294,19 @@ pub struct MayastorCliArgs {
...
     pub enable_coredump: Option<Option<u64>>,
+    /// Policy governing how a nexus selects which child to read from. {n}
+    /// `round-robin` (default) rotates reads across all healthy children,
+    /// local or remote alike. {n}
+    /// `local-preferred` prefers a healthy local (same-node) replica for
+    /// reads, keeping read I/O off the network, and falls back to
+    /// round-robin over remote readers only when no local reader is
+    /// healthy.
+    #[clap(
+        long = "policy",
```
tiagolobocastro, line 305: I think cli arg should be `nexus-read-policy` just like the variable name, `policy` is too broad
You can write up or just remove and it will default to that

```suggestion
```

io-engine/src/core/env.rs
```diff
@@ -294,6 +294,19 @@ pub struct MayastorCliArgs {
...
+        env = "NEXUS_READ_POLICY",
+        default_value = "round-robin"
```
```suggestion
        default_value_t = nexus::NexusReadPolicy::RoundRobin
```

io-engine/src/bdev/nexus/mod.rs
tiagolobocastro: @Abhinandan-Purkait @abhilashshetty04 what do you think?

tiagolobocastro (commented):
lgtm, please squash the commits for merge

io-engine/src/bdev/nexus/mod.rs
Abhinandan-Purkait: Not sure, maybe better to keep the existing behavior as default?

tiagolobocastro: I suppose we could leave as default here and modify in the helm chart

susobhandey: I think only this is the open topic rest review I addressed I believe

io-engine/src/bdev/nexus/nexus_channel.rs
susobhandey: Fixed

io-engine/src/core/env.rs
susobhandey: Fixed

io-engine/src/bdev/nexus/mod.rs
tiagolobocastro: Default can be modified later

PR #2071, opened 2026-09-12
Add fips enable switch to the io-engine

tiagolobocastro:
feat: back our tls stacks with FIPS validated crypto
    
    --enable-fips only covered diskpool encryption, while the TLS we do - to the
    kube api for secrets, and to nats for events - ran on whatever crypto rustls
    picked. Depend on the new fips crate, which pins rustls and aws-lc-rs to the
    FIPS validated AWS-LC for every rustls user in the graph, and call
    fips::init_if() first thing in main, before the logger can open a nats
    connection. It installs the FIPS provider as the process-wide rustls default,
    which is the only lever we have over the configs kube-client and async-nats
    build for us, and fails startup if the linked module is not validated. The
    --enable-fips help text now says so too.
    
    The aws-lc-rs and rustls entries in workspace.dependencies are not used
    directly by anything here: they are what the fips crate's `workspace = true`
    resolves against when it is built as part of this workspace rather than its
    own, so they have to match the ones in utils/dependencies.
    
    aws-lc-fips-sys builds the validated module from source, so the dev shell and
    the nix build gain cmake, go and perl.
    
---

    feat: add option to enable fips mode
    
    If fips is requested we cannot provide encryption as it stands currently, as
    we use isa-l-crypto without fips enabled.


io-engine/src/core/env.rs
```diff
@@ -479,12 +489,17 @@ impl MayastorFeatures {
         let lvm = env::var("ENABLE_LVM").as_deref() == Ok("true");
         let snapshot_rebuild = env::var("ENABLE_SNAPSHOT_REBUILD").as_deref() == Ok("true");
         let rdma_capable_io_engine = env::var("ENABLE_RDMA").as_deref() == Ok("true");
-        let diskpool_encryption = env::var("ENABLE_DISKPOOL_ENCRYPTION").as_deref() == Ok("true");
+        let fips = env::var("ENABLE_FIPS").as_deref() == Ok("true");
+        // The crypto module which backs diskpool encryption is not FIPS
+        // validated, so encryption must not be offered in FIPS mode.
+        let diskpool_encryption =
+            !fips && env::var("ENABLE_DISKPOOL_ENCRYPTION").as_deref() == Ok("true");
```
Abhinandan-Purkait, line 496: Ohh, then for fips variant we would have encryption? Any alternatives that would work here, not for this PR, a question in general.

tiagolobocastro: No there's no alternative at the moment, if you want to be able to pass a fips certification you can't use our current encryption (we'll have to modify it), so have to turn it off if fips is enabled

Abhinandan-Purkait (approved):

io-engine/src/core/env.rs
```diff
@@ -313,6 +313,16 @@ pub struct MayastorCliArgs {
...
     pub nexus_read_policy: nexus::NexusReadPolicy,
+    /// Run in FIPS mode. {n}
+    /// The SPDK crypto module which backs diskpool encryption is not FIPS
+    /// validated, so encryption is not supported in this mode: encrypted
+    /// pools can neither be created nor imported, and the diskpool
+    /// encryption capability is reported as unavailable to the
+    /// control-plane. {n}
+    /// Our TLS is put on a FIPS validated crypto module, and startup fails if
+    /// the module this was built against turns out not to be one.
+    #[clap(long = "enable-fips", env = "ENABLE_FIPS", value_parser = delay_compat)]
```
niladrih: nit: Don't mean to break convention or anything... but... with `delay_compat`, your flag usage turns out to be like this: `io-engine --enable-fips` means --enable-fips=false, which is a confusing option in my book. Maybe `--fips` with delay_compat, because likely the delay_compat is required for the env

niladrih (approved):

PR #2072, opened 2026-09-16

Abhinandan-Purkait:
Kubernetes CPU Manager running with cpuManagerPolicy=static may rewrite container cpusets when Guaranteed pods are admitted or removed.

When cpuset.cpus.effective is updated, the kernel resets task affinity masks for threads in the cgroup. This causes Mayastor reactor threads to lose their CPU pinning and Tokio worker threads to lose their unaffinitized CPU mask.

Add a cpuset monitor that watches for cpuset updates and automatically restores reactor affinity. Tokio worker TIDs are tracked so worker unaffinity can also be re-applied, preserving separation between reactor and runtime threads.

- cpuManagerPolicy=static
- isolcpus
- KubeVirt dedicated CPUs

admission/removal events.

fix(io-engine): update reactor repinning handling

refactor(affinity): run cpuset monitor on runtime blocking pool

Run the cpuset affinity monitor via runtime::spawn_blocking() instead of creating a dedicated std::thread.

This allows the monitor thread to participate in the existing thread unaffinity restoration mechanism and avoids maintaining a separate unmanaged thread.

fix(io-engine): re-pin reactors on CPUManager cpuset updates

Monitor kubelet cpu_manager_state and re-pin reactor threads when CPUManager updates the node cpuset allocation.

Handle cpu_manager_state inode replacement by re-establishing the inotify watch after DELETE_SELF and IGNORED events to ensure subsequent CPUManager updates are detected.


Abhinandan-Purkait:
Depends on openebs/spdk-rs#131 — the submodule is pinned to that PR's head commit (`71cd990`) for now and must be re-pinned to the merge commit before this lands.

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

